PR #14, opened 2021-09-08
Add zigbee contact driver

greens:


drivers/SmartThings/zigbee-contact/src/init.lua
```diff
@@ -0,0 +1,32 @@
...
+local constants = require "st.zigbee.constants"
+local defaults = require "st.zigbee.defaults"
+
+local zigbee_contact_driver_template = {
+  supported_capabilities = {
+    capabilities.contactSensor,
+    capabilities.temperatureMeasurement,
+    capabilities.battery,
+  },
+  sub_drivers = { require("battery-overrides") },
```
FreeMasen, line 26: Does it not work to use a bare string here over the wrapped parens?

drivers/SmartThings/zigbee-contact/src/battery-overrides/init.lua
```diff
@@ -0,0 +1,27 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local battery_defaults = require "st.zigbee.defaults.battery_defaults"
+
+local battery_range_2_1v_handler = {
+  NAME = "Cell battery 2.1v",
+  lifecycle_handlers = {
+    init = battery_defaults.build_linear_voltage_init(2.1, 3.0)
```
FreeMasen, line 20: Will this be taken care of automatically somehow on `added` as well?

FreeMasen (approved):

drivers/SmartThings/zigbee-contact/src/battery-overrides/init.lua
greens: has to be in `init` (or any time before `added`) because that's when the zigbee driver base builds its cluster configurations.

drivers/SmartThings/zigbee-contact/src/init.lua
greens: Not certain what you're asking. This is the standard way of including sub-drivers

posborne: I believe @FreeMasen is asking if it could just be written as `require "battery-overrides"` here, which I believe it could.  I don't think we have a strict policy for omission of parens for function calls or `require`s specifically so I would say we leave this as-is apart from any clear standard.

posborne (approved):
I've been using this driver regularly for the past couple of weeks with the Sonoff SNZB-04 (eWeLink/DS01) without any issues to report.

prjct92eh2:
@posborne does it work correctly for the Sonoff SNZB-04 even though it doesn't support temperature measurement? Doesn't look like there is a sub-driver without temperature capability.

greens:
> @posborne does it work correctly for the Sonoff SNZB-04 even though it doesn't support temperature measurement? Doesn't look like there is a sub-driver without temperature capability.

The nice thing about drivers is that the driver can support a superset of all the devices it supports' capabilities. You just need to point to a different profile that doesn't include the capability and it won't be displayed.

prjct92eh2:
ah, ok. I see the different `deviceProfileName` defined for each fingerprint now. Nifty! Thanks!

drivers/SmartThings/zigbee-contact/src/init.lua
FreeMasen: Since the other `require` calls in this file do not use the parenthesis I was curious why this would be different. Regardless of style, it is typical to try and keep usage consistent internally for a file.

drivers/SmartThings/zigbee-contact/src/battery-overrides/init.lua
FreeMasen: `init` is fired for every device know tho the driver at start up, is this the behavior that you are looking for?

It just strikes me as odd that you would want it to fire for a device init but not for a device add

PR #32, opened 2021-12-02
Add github action for automatic channel creation

greens:
When a new PR is opened that contains a modified driver, a new channel should be created with that driver assigned and an invite link should be posted on the PR

### DIFF
--- a/.github/workflows/create-driver-channel.yml
+++ b/.github/workflows/create-driver-channel.yml
@@ -1,0 +1,182 @@
+name: Create Driver Channel
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - 'drivers/**'
+jobs:
+  find-changed-drivers:
+    runs-on: ubuntu-latest
+    outputs:
+      changed-drivers: ${{ steps.set-outputs.outputs.changed-drivers }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: shallow fetch base revision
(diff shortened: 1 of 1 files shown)


.github/workflows/create-driver-channel.yml
```diff
@@ -0,0 +1,101 @@
+name: Create Driver Channel
+on:
+  pull_request:
+    types: [opened, synchronize]
```
tpmanley, line 4: What will happen if additional changes are added to the PR after it's opened? Will those be pushed to the channel?

greens: `synchronize` = updates to PR like new commits

tpmanley: So a new channel would be created each time, right? Do we have a way to "deactivate" the previous invites that only the latest one can be used?

greens: The first functioning version would delete the previous comment before adding a new one so only one invite _link_ will be visible at the same time, but the invites will all point to the same channel anyway that will continue to get its driver updated, so which invite link you use should not matter.

greens:
This is now in what I consider a good state. Only one invite/channel _should_ be created per PR.

Let me know if you can think of a corner case I might have missed.

Future enhancements I could add now, depending on how folks feel:

- deleting the channel after the PR is closed/merged

Current limitations I'm aware of:

- if one commit includes changes to a driver, and then those changes are reverted in a later commit, the channel will still have that version of the driver. I can't think of any way to avoid this.

PR #63, opened 2022-05-06
CHAD-8368 Emit all component button events from main

greens:


KKlimczukS (approved):
lgtm

drivers/SmartThings/zigbee-button/src/init.lua
```diff
@@ -35,6 +35,9 @@ local generate_event_from_zone_status = function(driver, device, zone_status, zb
       zb_rx.address_header.src_endpoint.value,
       event)
   end
+  if device:get_component_id_for_endpoint(zb_rx.address_header.src_endpoint.value) ~= "main" then
```
varzac, line 38: This check is outside the `event ~= nil` check.  You can also use the fact that the `emit_event_for_endpoint` function returns the event which would include the component as a part of it https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/master/lua_libs/st/device.lua#L131 and just check that component instead of using this function.  Although using this function might be a bit clearer.

varzac:
Not sure why a bunch of the test files are classified as binary. Otherwise looks  good.  I didn't notice any buttons where this would be a problem, but we'll just have to make sure in the future that if we do this that `main` isn't a separate button, but always just a combined thing.

tpmanley (approved):
I tested these changes with an Ecosmart 4-button device

inasail (approved):
I tested with SiHAS(4 button)/ikea(two button), it works well.

varzac (approved):

PR #71, opened 2022-05-11
[CHAD-8243] Added metadata info for some siren profiles

dsobierajsk:
@KKlimczukS @greens @varzac 

### DIFF
--- a/drivers/SmartThings/zwave-siren/profiles/aeon-siren.yml
+++ b/drivers/SmartThings/zwave-siren/profiles/aeon-siren.yml
@@ -12,7 +12,7 @@ components:
   - name: Siren
 metadata:
   mnmn: SmartThingsEdge
-  vid: Aeon_Siren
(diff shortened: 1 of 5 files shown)


greens (commented):
Since it's not encapsulated in the names can you leave a comment in the profiles that indicates what's different about these vids?

drivers/SmartThings/zwave-siren/profiles/aeon-siren.yml
```diff
@@ -10,6 +10,9 @@ components:
     version: 1
   categories:
   - name: Siren
+metadata:
+  mnmn: SmartThingsEdge
```
KKlimczukS, line 14: Please use "SmartThings" instead of "SmartThingsEdge" - applies to all files in this PR

KKlimczukS (changes requested):

KKlimczukS (approved):

PR #72, opened 2022-05-11
added fingerprint Third Reality 3RWS18BZ to zigbee-water-leak-sensor driver for CHAD-8543

sjl2178:


greens (commented):
This is another fingerprint that has some manufacturer-specific behavior, so a fp wouldn't be sufficient for integration.

PR #75, opened 2022-05-13
CHAD-8544 - Third Reality 3RMS16BZ

KKlimczukS:
@greens 
Please, could You take a look at the code ?
cc: @SmartThingsCommunity/srpol-pe-team

### DIFF
--- a/drivers/SmartThings/zigbee-motion-sensor/src/thirdreality/init.lua
+++ b/drivers/SmartThings/zigbee-motion-sensor/src/thirdreality/init.lua
@@ -14,8 +14,11 @@
 
 local capabilities = require "st.capabilities"
 local zcl_clusters = require "st.zigbee.zcl.clusters"
+local Basic = zcl_clusters.Basic
 local PowerConfiguration = zcl_clusters.PowerConfiguration
 local utils = require "st.utils"
+
+local SOFTWARE_VERSION = "software_version"
 
 local ZIGBEE_MOTION_SENSOR_FINGERPRINTS = {
   { mfr = "Third Reality, Inc", model = "3RMS16BZ"},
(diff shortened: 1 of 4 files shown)


drivers/SmartThings/zigbee-motion-sensor/src/thirdreality/init.lua
```diff
@@ -0,0 +1,52 @@
...
+  for _, fingerprint in ipairs(ZIGBEE_MOTION_SENSOR_FINGERPRINTS) do
+      if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
+          return true
+      end
+  end
+  return false
+end
+
+local function battery_percentage_handler(driver, device, raw_value, zb_rx)
+  -- if ((manufacturer == "Third Reality, Inc" || manufacturer == "THIRDREALITY") && application.toInteger() <= 17) {
```
tpmanley, line 35: I think the `application.toInteger() <= 17` part of the DTH code is important to preserve. That sounds like older versions of the firmware had a bug in the battery reporting that was fixed in later versions. @varzac Is the application version available to drivers so we can check it here?

varzac: I don't believe so, although we can just check to make certain.  The driver can request and store the value itself if it is functionally necessary.

greens: The Axis handlers do this, I believe: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/tree/main/drivers/SmartThings/zigbee-window-treatment/src/axis

drivers/SmartThings/zigbee-motion-sensor/src/thirdreality/init.lua
```diff
@@ -0,0 +1,52 @@
...
+  { mfr = "Third Reality, Inc", model = "3RMS16BZ"},
+  { mfr = "THIRDREALITY", model = "3RMS16BZ"}
+}
+
+local is_third_reality_motion_sensor = function(opts, driver, device)
+  for _, fingerprint in ipairs(ZIGBEE_MOTION_SENSOR_FINGERPRINTS) do
+      if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
+          return true
+      end
+  end
```
greens, line 30: your tab widths are inconsistent here

KKlimczukS: fixed

drivers/SmartThings/zigbee-motion-sensor/src/thirdreality/init.lua
KKlimczukS: @tpmanley yup, I already started a discussion about that in the task.  I'll check how it's done in the "axis" sub-driver, thank You @greens

KKlimczukS:
@greens @tpmanley @varzac 
The way the FV version is handled in the axis sub-driver works here too, but since we don't have any device here - I'm not sure how does the version string look like. For now, I assumed that the last 2 digits are significant, but it may be changed after integration tests.

tpmanley:
[app_version.txt](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/files/8741759/app_version.txt)

I can't push directly to your branch but here's a patch. I switched from SWBuild to ApplicationVersion to match the DTH. I confirmed this is working with my device that has version 1.0.23 (23 is 0x17) so it has the battery reporting bug. With the DTH my device reported about 60% and it was still the same after I tested with the updated version of the driver.

KKlimczukS:
> 

Thank You @tpmanley! It's super helpful!

tpmanley (approved):

greens (approved):

PR #80, opened 2022-05-16
Chad 8645 zigbee-button: Make button status visible in dashboard

sjl2178:
Default presentation uses the first capability listed for the dashboard status and a couple profiles in this driver listed battery first, thus showing battery percentage rather than button status. This fixes the problem by listing button first in the capabilities and adding an emit_event to the added_handler.

https://smartthings.atlassian.net/browse/CHAD-8645

### DIFF
--- a/drivers/SmartThings/zigbee-button/src/init.lua
+++ b/drivers/SmartThings/zigbee-button/src/init.lua
@@ -65,6 +65,7 @@ end
 local function added_handler(self, device)
   device:emit_event(capabilities.button.supportedButtonValues({"pushed","held","double"}))
   device:emit_event(capabilities.button.numberOfButtons({value = 1}))
+  device:emit_event(capabilities.button.button.pushed({state_change = false}))
 end
 
 local zigbee_button_driver_template = {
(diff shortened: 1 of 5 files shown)


drivers/SmartThings/zigbee-button/src/init.lua
```diff
@@ -65,6 +65,7 @@ end
 local function added_handler(self, device)
   device:emit_event(capabilities.button.supportedButtonValues({"pushed","held","double"}))
   device:emit_event(capabilities.button.numberOfButtons({value = 1}))
+  device:emit_event(capabilities.button.button({"pushed"}))
```
tpmanley, line 68: Can you update the unit tests to check for this too?

tpmanley: Also I got this from @greens 's PR, but this is slightly better because it doesn't generate a state change event: `device:emit_event(capabilities.button.button.pushed({state_change = false}))`. I already tested it and it worked fine with this change.

tpmanley (approved):
Looks good!

greens (approved):
Looks good.

PR #85, opened 2022-05-19
Have git stop treating lua files as binary

greens:


varzac:
This does not appear to have worked, at least on github.  I'm still seeing no diff and just a BIN designation.  Are there any non-ascii characters in these files or something?

greens:
I'm guessing it won't take effect until after this merge but I am not positive.

The issue was the various `\x00` bytes that, when they got copied over, caused git to think the whole file was not UTF-8. 

Open to other ideas.

varzac (approved):

PR #102, opened 2022-05-31
zigbee-button: Change category (and icon) for SmartThings button

tpmanley:
The default icon for the RemoteController category is different than the icon the SmartThings button uses when using a DTH. The new Button category will be mapped to the same icon that device uses when backed by a DTH.

https://smartthings.atlassian.net/browse/CHAD-8642


greens (approved):
There are definitely other profiles that are only a single button. Do we want to pre-emptively switch those?

sjl2178 (approved):

tpmanley:
@greens We could consider this in the future, but I think we should just keep it to this one device for now.

PR #111, opened 2022-06-03
CHAD-8725 - adds a new device profile for Aeotec aerQ sensors

KKlimczukS:
CHAD-8725 - new device profile for Aeotec aerQ sensors
@SmartThingsCommunity/srpol-pe-team @greens


dsobierajsk (approved):

greens (commented):
both the devices that use this profile use the same ocf device type

you should not have created a new profile here, and now the original profile also needs to be fixed because, upon inspection, the Popp multisensor also uses a thermsotat device type.

KKlimczukS:
> both the devices that use this profile use the same ocf device type
> 
> you should not have created a new profile here, and now the original profile also needs to be fixed because, upon inspection, the Popp multisensor also uses a thermsotat device type.

@greens You're right, that was in the `definition()` of the DTH. New PR is coming.

PR #121, opened 2022-06-13
HCS-2753: Added '000B' as fingerprint to zwave-thermostat

Bhavik-T:
Bugfix HCS-2753: Added '000B' as fingerprint to zwave-thermostat

### DIFF
--- a/drivers/SmartThings/zwave-thermostat/fingerprints.yml
+++ b/drivers/SmartThings/zwave-thermostat/fingerprints.yml
@@ -47,6 +47,12 @@ zwaveManufacturer:
     productType: 0x6501
     productId: 0x000C
     deviceProfileName: base-thermostat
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/zwave-thermostat/fingerprints.yml
```diff
@@ -41,7 +41,7 @@ zwaveManufacturer:
     productType: 0x6401
     productId: 0x0107
     deviceProfileName: base-thermostat
-  - id: "0098/6501/000C"
+  - id: "0098/6501/000C/000B"
```
varzac, line 44: This is just changing the ID.  You will need to duplicate the entire entry (everything between this `-` entry and  the next one), but change the `productID` for the copy to use `000B` instead of C

varzac (approved):

PR #129, opened 2022-06-14
CHAD-8864 - Fixed an issue where energy consumption was not updated in the Energy…

inasail:
CHAD-8864
Added vid for Dawon outlet.
Changed the unit of energy and deltaEnergy values (kWh -> Wh) like DTH.


inasail:
@greens Would you please review this commit? After this commit is merged, this modification should be updated to the 'production' branch & default channel for public user ASAP.

inasail:
@greens Thanks for the review. BTW, Can you teach me how can we get the merge permission? If it is impossible, please merge this commit instead of me.
and this modification should be updated to the 'production' branch & default channel for public user ASAP. Korean customers are already complaining about this issue.

nravi-n (approved):
LGTM.

PR #140, opened 2022-06-22
CHAD-8872 - Add missing fingerprints in power-meter and vid for EZEX power meter

inasail:
[CHAD-8872](https://smartthings.atlassian.net/browse/CHAD-8872)
- Add specific vid for EZEX power meter to display in Energy plugin.
- Fix bug

[CHAD-8897](https://smartthings.atlassian.net/browse/CHAD-8897)
- Add missing fingerprints in [DTH](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy) and add sub-drivers for them.

### DIFF
--- a/drivers/SmartThings/zigbee-power-meter/fingerprints.yml
+++ b/drivers/SmartThings/zigbee-power-meter/fingerprints.yml
@@ -2,6 +2,21 @@ zigbeeManufacturer:
   - id: "E240-KR080Z0-HA"
     deviceLabel: Energy Monitor
     model: "E240-KR080Z0-HA"
+    deviceProfileName: power-meter-consumption-report
+  - id: "Develco/ZHEMI101"
+    deviceLabel: frient Energy Monitor
+    manufacturer: Develco
+    model: "ZHEMI101"
+    deviceProfileName: power-meter
+  - id: "Develco/EMIZB-132"
+    deviceLabel: frient Energy Monitor
+    manufacturer: Develco Products A/S
+    model: "EMIZB-132"
+    deviceProfileName: power-meter
+  - id: "ShinaSystem/PMM-300Z1"
+    deviceLabel: SiHAS Energy Monitor
+    manufacturer: ShinaSystem
+    model: "PMM-300Z1"
     deviceProfileName: power-meter
 zigbeeGeneric:
   - id: "genericMeter"
--- a/drivers/SmartThings/zigbee-power-meter/src/ezex/init.lua
+++ b/drivers/SmartThings/zigbee-power-meter/src/ezex/init.lua
@@ -1,0 +1,79 @@
+-- Copyright 2022 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local capabilities = require "st.capabilities"
+local ZigbeeDriver = require "st.zigbee"
+local defaults = require "st.zigbee.defaults"
+local constants = require "st.zigbee.constants"
+local clusters = require "st.zigbee.zcl.clusters"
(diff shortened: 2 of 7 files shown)


inasail:
@greens Would you please review and merge this commit? Thank you.

drivers/SmartThings/zigbee-power-meter/fingerprints.yml
```diff
@@ -2,6 +2,21 @@ zigbeeManufacturer:
...
+  - id: "Develco/EMIZB-132"
+    deviceLabel: frient Energy Monitor
+    manufacturer: Develco Products A/S
+    model: "EMIZB-132"
+    deviceProfileName: power-meter
+  - id: "ShinaSystem/PMM-300Z1"
+    deviceLabel: SiHAS Energy Monitor
+    manufacturer: ShinaSystem
+    model: "PMM-300Z1"
     deviceProfileName: power-meter
```
greens, line 20: shouldn't these three also support `powerConsumptionReport`?

drivers/SmartThings/zigbee-power-meter/src/shinasystems/init.lua
```diff
@@ -0,0 +1,56 @@
...
+local shinasystems_power_meter_handler = {
+  NAME = "shinasystems power meter handler",
+  lifecycle_handlers = {
+    init = device_init,
+    doConfigure = do_configure,
+  },
+  can_handle = is_shinasystems_power_meter
+}
+
+return shinasystems_power_meter_handler
```
greens, line 56: this file is identical to the frient handler above, why wouldn't you just include the fingerprints for all of the devices?

drivers/SmartThings/zigbee-power-meter/src/ezex/init.lua
```diff
@@ -0,0 +1,79 @@
...
+  if current_power_consumption ~= nil then
+    delta_energy = math.max(raw_value_watts - current_power_consumption.energy, 0.0)
+  end
+  device:emit_event(capabilities.powerConsumptionReport.powerConsumption({energy = raw_value_watts, deltaEnergy = delta_energy })) -- the unit of these values should be 'Wh'
+
+  local multiplier = device:get_field(constants.SIMPLE_METERING_MULTIPLIER_KEY) or 1
+  local divisor = device:get_field(constants.SIMPLE_METERING_DIVISOR_KEY) or 1000000
+  local converted_value = raw_value_miliwatts * multiplier/divisor -- unit: kWh
+  device:emit_event(capabilities.energyMeter.energy({value = converted_value, unit = "kWh"}))
+end
```
greens, line 61: Please refactor/re-nest the sub-drivers that support this capability so that this code is not duplicated.

A useful thing to do might be to make a powerConsumptionReport sub-handler that, for it's `can_handle` function checks the device profile to see if it supports that capability, rather than using mfr fingerprints.

drivers/SmartThings/zigbee-power-meter/fingerprints.yml
inasail: No, they don't. Only EZEX sensor supports `powerConsumptionReport`.
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy#L111

drivers/SmartThings/zigbee-power-meter/src/shinasystems/init.lua
inasail: `frient` and `shinasystems` have different ELECTRICAL_MEASUREMENT_DIVISOR_KEY value.
I considered to merge each sub drivers into one. but It was hard to naming the sub driver.I n fact, I personally prefer to distinguish between drivers by manufacturer, unless there are many devices in fingerprint and therefore code reuse needs to be considered a lot.

drivers/SmartThings/zigbee-power-meter/src/ezex/init.lua
inasail: This handler is only for EZEX and It is already overrided from energyMeter_defaut because of powerConsumptionReport capability. Each sensors have different input value unit(W,mW) and different divisor and so on. so I think it is okay to separate their code for maintenance.

drivers/SmartThings/zigbee-power-meter/src/shinasystems/init.lua
greens: I missed that difference.

greens (approved):

PR #141, opened 2022-06-22
[CHAD-8882] Added holdTime preference to dimming-remote devices

dsobierajsk:


KKlimczukS (approved):
lgtm
@greens

drivers/SmartThings/zigbee-button/src/dimming-remote/init.lua
```diff
@@ -37,14 +37,32 @@ local function can_handle_zigbee_dimming_remote(opts, driver, device, ...)
...
+    local comp = device.profile.components[string.format("button%d", button_number)]
+
+    device:emit_event(event)
+    device:emit_component_event(comp, event)
+  end
+end
+
+local function button_hold_handler(button_number)
+  return function(driver, device, value, zb_rx)
+    device:set_field(string.format(BUTTON_X_HOLD, button_number), true, { persist = true })
```
greens, line 51: this shouldn't need to use persistent storage, considering the timeframe

greens (commented):
I'm in favor of keeping the button_utils version because of the timeout threshold. If we miss a released event I don't want us sending a false held event later.

drivers/SmartThings/zigbee-button/src/dimming-remote/init.lua
```diff
@@ -37,14 +38,44 @@ local function can_handle_zigbee_dimming_remote(opts, driver, device, ...)
...
+local function send_held_event_if_applicable(device, button_number)
+  local event = capabilities.button.button.held({ state_change = true })
+  local comp = device.profile.components[string.format("button%d", button_number)]
+  local button_hold_time = get_button_hold_time(device, button_number)
+
+  device:set_field(string.format(BUTTON_X_HOLD, button_number), nil)
+
+  if button_hold_time and button_hold_time < TIMEOUT_THRESHOLD then
+    device:emit_event(event)
+    device:emit_component_event(comp, event)
```
greens, line 72: Shouldn't we send a `press` event if the hold time is less than some period of time?

dsobierajsk: Imo pushed event should be sent only when OnOff message is received. Device recognizes itself if button was hold or pushed only

greens: Looks like the DTH sends a pushed if it's not a hold, unless I'm reading wrong?

We want to be sending whatever the DTH would have sent.

dsobierajsk: > Looks like the DTH sends a pushed if it's not a hold, unless I'm reading wrong?

Yes, it is

greens: so you'll be making a change here to send a pushed if the hold is shorter than the held threshold?

dsobierajsk:
Hi @varzac, could you look at this too? Thanks

drivers/SmartThings/zigbee-button/src/dimming-remote/init.lua
greens: I think right now we want to copy DTH functionality.

dsobierajsk: Idk why, but button_utils started working properly. I've noticed that device should have also holdTime preference so I added it and rollbacked previous changes.

KKlimczukS (approved):

drivers/SmartThings/zigbee-button/fingerprints.yml
```diff
@@ -108,12 +108,12 @@ zigbeeManufacturer:
...
     manufacturer: OSRAM
     model: LIGHTIFY Dimming Switch
-    deviceProfileName: two-buttons-battery
+    deviceProfileName: osram-two-buttons-battery
   - id: CentraLite/3130
     deviceLabel: Centralite Button
     manufacturer: CentraLite
     model: 3130
-    deviceProfileName: two-buttons-battery
+    deviceProfileName: osram-two-buttons-battery
```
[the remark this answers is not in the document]
dsobierajsk: It doesn't matter. I have a CentraLite device and it is branded as OSRAM, but you might be right and this name should be better.

greens (approved):

varzac (approved):

PR #145, opened 2022-06-23
Add fpr to zigbee-button

sjl2178:
Added fingerprints to zigbee-button for CHAD-9538

https://smartthings.atlassian.net/browse/CHAD-8538


greens (commented):
the DTH has some manufacturer-specific behavior associated with these devices that would need to be added

PR #159, opened 2022-07-03
zigbee-switch: Fix color handling for Ikea RGB bulbs

tpmanley:
The original code wasn't handling the setHue and setSaturation commands
correctly. The light would flash but not change to the desired color. To
fix the problem, a short timer is started when receiving either the setHue
or setSaturation commands instead of sending a command to the device right
away. The hue (or saturation) value is cached temporarily, and if the other
command is received before the timer expires then both the cached value and
the value from the second command are used to calculate the x/y and are sent
to the device. If only one of the two commands is sent before the timer
expires then the value from that command is combined with the latest known
color state to calculate the x/y.

https://smartthings.atlassian.net/browse/CHAD-8948

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
@@ -23,6 +23,9 @@ local ColorControl = clusters.ColorControl
 local CURRENT_X = "current_x_value" -- y value from xyY color space
 local CURRENT_Y = "current_y_value" -- x value from xyY color space
 local Y_TRISTIMULUS_VALUE = "y_tristimulus_value" -- Y tristimulus value which is used to convert color xyY -> RGB -> HSV
+local HUESAT_TIMER = "huesat_timer"
+local TARGET_HUE = "target_hue"
+local TARGET_SAT = "target_sat"
 
 local IKEA_XY_COLOR_BULB_FINGERPRINTS = {
   ["IKEA of Sweden"] = {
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
```diff
@@ -66,36 +69,79 @@ end
...
+          hue = hue,
+          saturation = sat
+        }
+      }
+      set_color_handler(driver, device, cmd)
+      return
+    end
+  end
+  device:set_field(TARGET_HUE, hue)
+  huesat_timer = device.thread:call_with_delay(0.2, huesat_timer_callback(driver, device, cmd))
```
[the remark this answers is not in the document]
tpmanley: In my testing the `setHue` and `setSaturation` commands received by the driver less than 100ms apart. I'm pretty sure they were sent as a batch to the hub although I don't have logs in front of me to confirm that. So the timer doesn't need to be exactly 0.2.

drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
```diff
@@ -66,36 +69,79 @@ end
...
-local function set_saturation_handler(driver, device, cmd)
-  local hue = device:get_latest_state("main", capabilities.colorControl.ID, capabilities.colorControl.hue.NAME)
-  local x, y, Y = utils.safe_hsv_to_xy(hue, cmd.args.saturation)
-  store_xyY_values(device, x, y, Y)
-  switch_defaults.on(driver,device,cmd)
+local huesat_timer_callback = function(driver, device, cmd)
+  return function()
+    local hue = device:get_field(TARGET_HUE)
+    local sat = device:get_field(TARGET_SAT)
+    hue = hue ~= nil and hue or device.state_cache.main.colorControl.hue.value
```
greens, line 89: On initial join it's possible to have both `TARGET_HUE` and the state cache not have a value yet for the hue. (or vice versa with saturation)

Also I'm concerned that adding in the delay here as you've done hasn't caused any test cases to require adjustment.

greens: I'd expect they would need them, since you're adding at least a .2s delay that wasn't there before, no?

tpmanley: That's a good point. I think there is something about the mock time handling that isn't quite right that was allowing the tests to pass before. I updated the tests with what I think should have been required and confirmed they are still passing. I think we need to talk more about the issues I was seeing but I don't think it needs to hold up the PR.

drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
```diff
@@ -64,38 +67,87 @@ local query_device = function(device)
...
+    local sat = device:get_field(TARGET_SAT)
+    if sat ~= nil then
+      cmd.args = {
+        color = {
+          hue = hue,
+          saturation = sat
+        }
+      }
+      set_color_handler(driver, device, cmd)
+      return
```
greens, line 122: multiple exit points tsk tsk

greens (approved):

drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
tpmanley: That's a fair nit, plus there is duplicate code. Take a look at the fixup commit and see if you think this is better.

drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
```diff
@@ -104,50 +104,41 @@ local huesat_timer_callback = function(driver, device, cmd)
...
+  else
+    return false
+  end
+end
+
+local function set_hue_handler(driver, device, cmd)
+  local hue = cmd.args.hue
+  if not set_hue_sat_helper(driver, device, cmd, hue, device:get_field(TARGET_SAT)) then
+    device:set_field(TARGET_HUE, hue)
+    device:set_field(HUESAT_TIMER, device.thread:call_with_delay(0.2, huesat_timer_callback(driver, device, cmd)))
```
tpmanley, line 132: I could also move this call into the helper function

greens (approved):
it's fine as is

drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
```diff
@@ -64,38 +67,73 @@ local query_device = function(device)
...
-  local hue = cmd.args.hue > 99 and 99 or cmd.args.hue
-  local x, y, Y = utils.safe_hsv_to_xy(hue, sat)
-  store_xyY_values(device, x, y, Y)
-  switch_defaults.on(driver,device,cmd)
+local huesat_timer_callback = function(driver, device, cmd)
+  return function()
+    local hue = device:get_field(TARGET_HUE)
+    local sat = device:get_field(TARGET_SAT)
+    hue = hue ~= nil and hue or device:get_latest_state("main", capabilities.colorControl.ID, capabilities.colorControl.hue.NAME)
+    sat = sat ~= nil and sat or device:get_latest_state("main", capabilities.colorControl.ID, capabilities.colorControl.saturation.NAME)
```
varzac, line 96: So in the case that the device has not ever reported it's hue/sat it is possible for these values to still end up `nil`.  It's quite unlikely, but it's probably worth protecting against.

drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
```diff
@@ -64,38 +67,73 @@ local query_device = function(device)
...
+    elseif sat ~= nil then
+      device:set_field(TARGET_SAT, sat)
+    end
+    device:set_field(HUESAT_TIMER, device.thread:call_with_delay(0.2, huesat_timer_callback(driver, device, cmd)))
+  end
+end
 
-  device.thread:call_with_delay(2, query_device(device))
+local function set_hue_handler(driver, device, cmd)
+  set_hue_sat_helper(driver, device, cmd, cmd.args.hue, device:get_field(TARGET_SAT))
```
varzac, line 132: So it looks like this same cmd argument table is passed through the whole time.  It seems a bit questionable to be passing around a `setHue` command with non-standard args, and passing it into a `setColor` command handler.  In this case everything _works_ but it's a bit fragile.

drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
tpmanley: `set_color_handler` uses `utils.safe_hsv_to_xy` which will treat either hue or sat as 0 if `nil` is passed in so I think we're covered. There are also tests that test when either hue or sat is nil.

drivers/SmartThings/zigbee-switch/src/zll-dimmer-bulb/ikea-xy-color-bulb/init.lua
varzac: I guess since this is a specific sub-driver I'm OK with it.  I was originally worried about accidental usage of this not maintaining the assumptions, but given it's for a single driver it seems fine.

PR #160, opened 2022-07-05
[CHAD-8941] Changed the range of Z-Wave dimmers to [1,100]

Bhavik-T:


greens:
@bflorian does this look kosher to you?

@Bhavik-T I think you're on the right track. This particular profile is for a zigbee dimming remote, but I do think we'll want to apply the [1,100] range to *all* profiles which include the Switch Level preference.

greens (approved):
Looks good but we're waiting on upstream changes.

lelandblue:
Hey @bflorian Can you please check this?

iplantemn:
@greens @bflorian updated for the model changes. Not sure why the unit tests are failing, I just changed the profile YMLs

greens:
@ctowns when you get time could you update this PR to align to the new format

greens (approved):
Looks good but we'll probably have to do this again with the profiles that have been added since this was created.

greens:
@ctowns could you resolve the conflict with the `inovelli-dimmer` profile?

PR #166, opened 2022-07-06
Bugfix/CHAD-8838

Bhavik-T:
[CHAD-8838] Created a new profile with colorTemperature of range 2700k to 6500k. Updated OSRAM Light devices in fingerprint file to reflect changes to profile.


drivers/SmartThings/zigbee-dimmer-remote/profiles/switch-level-button.yml
```diff
@@ -6,6 +6,10 @@ components:
       version: 1
     - id: switchLevel
       version: 1
+      configuration:
+      - values:
+          key: "level.value"
+          range: [1, 100]
```
greens, line 12: I don't think you meant to include the changes to this file in this PR

PR #181, opened 2022-07-14
[CHAD-8788] zwave-sensor: Set correct range on temperature measurement values

Bhavik-T:
Updated zwave sensor's temperaturemeasurement to have a range of -20C to 100C.


greens:
@bflorian how does this work for a capability that has units?

bflorian:
@greens temperature ranges in Canvas should alway be in Celsius. There's special handling in the plugin to scale properly depending on the current units specified in the attribute value.

greens (approved):

iplantemn:
@greens @bflorian updated capability configurations

lelandblue:
Hello @ctowns Please change   "configuration:" to "config:" thank you!

PR #204, opened 2022-07-29
Add zigbee-window-treatment for Vimar xx594 and 03982

maxelweb-vimar:
Add support for: 
- Vimar Smart Roller Shutter (xx594)
- Vimar Smart Roller Shutter Module (03982)


greens (commented):
removed

jodyalbritton:
> It seems (from a quick glance) that this device would work fine if it were added to the IKEA blinds sub-driver.
> 
> Also there's no reason to remove the firmwareUpdate capability. That's managed by the hub for you.

The device does  supports firmware updates, but does them via bluetooth.

greens (approved):

PR #208, opened 2022-08-03
Aqara zigbee motion sensor

sbkimaqara:


drivers/SmartThings/zigbee-motion-sensor/src/aqara/init.lua
```diff
@@ -0,0 +1,154 @@
...
+
+local function attr_handler(driver, device, value, zb_rx)
+  -- The low 16 bits for Illuminance
+  -- The high 16 bits for Motion Detection
+  if value.value > 65536 then
+    -- active
+    local lux = value.value - 65536
+    device:emit_event(capabilities.illuminanceMeasurement.illuminance(lux))
+
+    motion_active_handler(driver, device, value, zb_rx)
```
inasail, line 95: Is this necessary? there is separate handler(occupancy_attr_handler) for motion detection. If attr_handler/occupancy_attr_handler called separately, motion_active_handler() will be called twice.

drivers/SmartThings/zigbee-motion-sensor/src/test/test_aqara_motion.lua
```diff
@@ -0,0 +1,141 @@
...
+
+-- Mock out globals
+local test = require "integration_test"
+local clusters = require "st.zigbee.zcl.clusters"
+local cluster_base = require "st.zigbee.cluster_base"
+local data_types = require "st.zigbee.data_types"
+local capabilities = require "st.capabilities"
+local zigbee_test_utils = require "integration_test.zigbee_test_utils"
+local t_utils = require "integration_test.utils"
+local test = require "integration_test"
```
inasail, line 23: duplicated

drivers/SmartThings/zigbee-motion-sensor/src/aqara/init.lua
sbkimaqara: 1) Motion Sensor T1 sends the detected value  and the brightness value when motion is detected. This uses the Private cluster attribute 0x0112.

2) High Precision Motion Sensor sends the detected value only when motion is detected. This uses the OccupancySensing cluster attribute.

3) Therefore, each device is mutually exclusive and the attribute handler is not executed twice.

drivers/SmartThings/zigbee-motion-sensor/src/test/test_aqara_motion.lua
sbkimaqara: I removed it. Thanks.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/init.lua
inasail: okay, got it.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/init.lua
```diff
@@ -0,0 +1,158 @@
...
+end
+
+local write_motion_pref_attribute = function(device, cluster, attr, value)
+  device:send(cluster_base.write_manufacturer_specific_attribute(device, cluster, attr, MFG_CODE,
+    data_types.Uint8, value))
+end
+
+local function added_handler(self, device)
+  device:emit_event(capabilities.motionSensor.motion.inactive())
+  device:emit_event(sensitivityAdjustment.sensitivityAdjustment.Medium())
```
inasail, line 53: This should be only for 'agl04', need to check the model name before emit_event. also need to change the TC.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/init.lua
```diff
@@ -0,0 +1,158 @@
...
+  device:emit_event(detectionFrequency.detectionFrequency(FREQUENCY_DEFAULT_VALUE))
+  device:emit_event(capabilities.battery.battery(100))
+end
+
+local do_configure = function(self, device)
+  device:configure()
```
inasail, line 59: It isn't maybe necessary. Please test after remove this with real device. (also need to change the TC)

drivers/SmartThings/zigbee-motion-sensor/src/test/test_aqara_motion.lua
```diff
@@ -0,0 +1,140 @@
...
+local sensitivityAdjustment = capabilities["stse.sensitivityAdjustment"]
+local detectionFrequency = capabilities["stse.detectionFrequency"]
+
+local PowerConfiguration = clusters.PowerConfiguration
+local OccupancySensing = clusters.OccupancySensing
+local FREQUENCY_DEFAULT_VALUE = 120
+
+local mock_device = test.mock_device.build_test_zigbee_device(
+  {
+    profile = t_utils.get_profile_definition("motion-battery-aqara.yml"),
```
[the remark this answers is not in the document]
sbkimaqara: I wrote the test code for each model.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/init.lua
sbkimaqara: I wrote the test code for each model.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/init.lua
sbkimaqara: Thanks for the suggestion.

veonua:
any progress with this PR? @inasail

inasail:
@veonua 
Aqara will request review officially to SmartThings soon.
After that, SmartThings guys will review this PR.

greens (changes requested):
These custom capabilities seem like they'd be better suited to be preferences instead.

Kwang-Hui:
> These custom capabilities seem like they'd be better suited to be preferences instead.
Hi @greens ,
As described at REQ-8173, Aqara device need to fallback mechanism in case of their configuration has failed if the user doesn't press the physical button of device. but the preference can't get status value from edge driver. that's why custom capability is used instead of preference.

Kwang-Hui:
> @veonua Aqara will request review officially to SmartThings soon. After that, SmartThings guys will review this PR. @sbkimaqara , What is the current progress? Please discuss with '조광희' for the next step.

I requested PR review with jira ticket to Alissa. Thanks!

greens:
@Kwang-Hui you could build a mechanism to re-send the preference on every refresh, for example, until it receives the read attribute response from the device.

Kwang-Hui:
> @Kwang-Hui you could build a mechanism to re-send the preference on every refresh, for example, until it receives the read attribute response from the device.

Hi @greens , I understand your intend to re-enforce device's status with using preference value as SSOT. But that could cause inconsistency between UI and real device status. I wrote slack message about this to you :)

veonua:
> could build a mechanism to re-send the preference on every refresh, for example, until it receives the read attribute response from the device.

@greens  it's the ST platform responsibility.
In other way you force *Every* driver developer fix this flow.

Kwang-Hui:
Hi @sbkimaqara , could you explain how the configuration (sensitivity, detection frequency) update works with physical button press?
@greens is wondering how does it work without additional packet handling except immediate zigbee write.

sbkimaqara:
> Hi @sbkimaqara , could you explain how the configuration (sensitivity, detection frequency) update works with physical button press? @greens is wondering how does it work without additional packet handling except immediate zigbee write.

1. Send a write attribute value to the device in the driver code.
2. The device receive a write attribute event.
3. The device update a value if the user press a device's button within 30 seconds.
4. The device send a write attribute success response to the hub.
5. Receive a write attribute success response from the device in the driver code.
6. Update the UI in the driver code.

greens (commented):
at the very least, please `git rm` the `.DS_Store` file

On top of that, there is significant code overlap between the two drivers. Since you already have one parent driver for these two new drivers, it would make sense to move this common behavior into the parent driver. Specifically, the detection frequency handler seems the same between both drivers and the `write_attr_res_handler` that handles both custom attributes, and the handling of the frequency_attribute_id can also be consolidated.

The code in each handler should only be the code that is different between them. If you're duplicating functions or constants, consider creating a separate file that can be imported by your drivers that keeps them all in once place.

sbkimaqara:
> at the very least, please `git rm` the `.DS_Store` file
> 
> On top of that, there is significant code overlap between the two drivers. Since you already have one parent driver for these two new drivers, it would make sense to move this common behavior into the parent driver. Specifically, the detection frequency handler seems the same between both drivers and the `write_attr_res_handler` that handles both custom attributes, and the handling of the frequency_attribute_id can also be consolidated.
> 
> The code in each handler should only be the code that is different between them. If you're duplicating functions or constants, consider creating a separate file that can be imported by your drivers that keeps them all in once place.

I refactored the code according to your advice. Please review it.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/high-precision/init.lua
```diff
@@ -0,0 +1,119 @@
...
+    init = device_init,
+    added = added_handler
+  },
+  capability_handlers = {
+    [aqara_utils.sensitivityAdjustmentId] = {
+      [aqara_utils.sensitivityAdjustmentCommand] = aqara_utils.sensitivity_adjustment_handler,
+    },
+    [aqara_utils.detectionFrequencyId] = {
+      [aqara_utils.detectionFrequencyCommand] = aqara_utils.detection_frequency_handler,
+    }
```
greens, line 96: What I meant was that this should be moved to the parent aqara driver rather than having to set it twice.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/high-precision/init.lua
```diff
@@ -0,0 +1,119 @@
...
+  },
+  zigbee_handlers = {
+    global = {
+      [aqara_utils.PRIVATE_CLUSTER_ID] = {
+        [zcl_commands.WriteAttributeResponse.ID] = aqara_utils.write_attr_res_handler
```
greens, line 101: same here

drivers/SmartThings/zigbee-motion-sensor/src/aqara/high-precision/init.lua
```diff
@@ -0,0 +1,119 @@
...
+    },
+    attr = {
+      -- High Precision Motion Sensor
+      [OccupancySensing.ID] = {
+        [OccupancySensing.attributes.Occupancy.ID] = aqara_utils.occupancy_attr_handler
+      },
+      [aqara_utils.PRIVATE_CLUSTER_ID] = {
+        -- Prefs
+        [aqara_utils.FREQUENCY_ATTRIBUTE_ID] = aqara_utils.frequency_attr_handler,
+        [aqara_utils.SENSITIVITY_ATTRIBUTE_ID] = aqara_utils.sensitivity_attr_handler
```
greens, line 112: same here

drivers/SmartThings/zigbee-motion-sensor/src/aqara/high-precision/init.lua
```diff
@@ -0,0 +1,119 @@
...
+      }
+    },
+    attr = {
+      -- High Precision Motion Sensor
+      [OccupancySensing.ID] = {
+        [OccupancySensing.attributes.Occupancy.ID] = aqara_utils.occupancy_attr_handler
+      },
+      [aqara_utils.PRIVATE_CLUSTER_ID] = {
+        -- Prefs
+        [aqara_utils.FREQUENCY_ATTRIBUTE_ID] = aqara_utils.frequency_attr_handler,
```
greens, line 111: same here

drivers/SmartThings/zigbee-motion-sensor/src/aqara/high-precision/init.lua
```diff
@@ -0,0 +1,119 @@
...
+  device:send(aqara_utils.read_custom_attribute(device, aqara_utils.PRIVATE_CLUSTER_ID,
+    aqara_utils.FREQUENCY_ATTRIBUTE_ID))
+  device:send(aqara_utils.read_custom_attribute(device, aqara_utils.PRIVATE_CLUSTER_ID,
+    aqara_utils.SENSITIVITY_ATTRIBUTE_ID))
+end
+
+local aqara_motion_handler = {
+  NAME = "Aqara Motion Handler",
+  lifecycle_handlers = {
+    init = device_init,
```
greens, line 87: your init function is also identical between the sub-drivers, so it can be put in the parent and the sub-drivers will "inherit" it

drivers/SmartThings/zigbee-motion-sensor/src/aqara/aqara_utils.lua
```diff
@@ -0,0 +1,186 @@
...
+
+local MOTION_DETECTED_UINT32 = 65536
+aqara_utils.MOTION_DETECTED_UINT32 = MOTION_DETECTED_UINT32
+local MOTION_DETECTED_NUMBER = 1
+aqara_utils.MOTION_DETECTED_NUMBER = MOTION_DETECTED_NUMBER
+
+local UNOCCUPIED_TIMER = "unoccupiedTimer"
+aqara_utils.UNOCCUPIED_TIMER = UNOCCUPIED_TIMER
+
+local PREF_CHANGED_KEY = "prefChangedKey"
```
greens, line 42: anything you're only using in this utils file doesn't need to be put into the map that gets loaded by the two sub-drivers

drivers/SmartThings/zigbee-motion-sensor/src/aqara/high-precision/init.lua
sbkimaqara: Delete duplicate codes.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/high-precision/init.lua
sbkimaqara: Delete duplicate codes.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/high-precision/init.lua
sbkimaqara: Delete duplicate codes.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/high-precision/init.lua
sbkimaqara: Delete duplicate codes.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/aqara_utils.lua
sbkimaqara: Delete duplicate codes.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/init.lua
```diff
@@ -0,0 +1,44 @@
...
+local is_aqara_products = function(opts, driver, device)
+  for _, fingerprint in ipairs(FINGERPRINTS) do
+    if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
+      return true
+    end
+  end
+  return false
+end
+
+local aqara_motion_handler = {
```
Kwang-Hui, line 18: it seems same name `aqara_motion_handler` for all sub drivers also.  @sbkimaqara  Could you check this is working as expected?

sbkimaqara: The device worked correctly. But I will modify the code according to your advice.

Kwang-Hui: thanks. but if it works. I'm just fine.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/motion-illuminance/init.lua
```diff
@@ -0,0 +1,68 @@
...
+local aqara_motion_illuminance_handler = {
+  NAME = "Aqara Motion Illuminance Handler",
+  lifecycle_handlers = {
+    init = device_init,
+    added = added_handler
+  },
+  zigbee_handlers = {
+    attr = {
+      [aqara_utils.PRIVATE_CLUSTER_ID] = {
+        [aqara_utils.MOTION_ILLUMINANCE_ATTRIBUTE_ID] = aqara_utils.motion_illuminance_attr_handler,
```
greens, line 59: Since this function is only used in this file, it makes sense for it to be defined here rather than in the shared utils file.

Same for the other handlers unique to the other driver/device. 

Your utils file should be the things that need to be shared by both sub-drivers that can't be included in the parent driver.

Now that you've refactored your code like this it looks like all you might need in the utils file is just some constants.

sbkimaqara: I updated thanks :)

drivers/SmartThings/zigbee-motion-sensor/capabilities/sensitivityAdjustment.yaml
```diff
@@ -0,0 +1,55 @@
...
+    setter: setSensitivityAdjustment
+    enumCommands:
+      - command: High
+        value: High
+      - command: Medium
+        value: Medium
+      - command: Low
+        value: Low
+      - command: Desc
+        value: Desc
```
greens, line 30: for my own education, what does `Desc` mean in this context?

greens (approved):
Looks fine. Waiting on confirmation about the custom capabilities.

drivers/SmartThings/zigbee-motion-sensor/capabilities/sensitivityAdjustment.yaml
sbkimaqara: The sensitivityAdjustment capability has been extended to display not only the title but also the description to the user.

veonua:
Good job guys! Thank you

PR #209, opened 2022-08-03
[CHAD-8803] "thermostatOperatingState” and “temperatureAlarm” updated

Bhavik-T:
thermostatOperatingState” and “temperatureAlarmt” updated to  only contain capabilities that are supported by the device.

### DIFF
--- a/drivers/SmartThings/zigbee-thermostat/profiles/base-thermostat-fanless.yml
+++ b/drivers/SmartThings/zigbee-thermostat/profiles/base-thermostat-fanless.yml
@@ -16,6 +16,13 @@ components:
     version: 1
   - id: thermostatOperatingState
     version: 1
+    config:
+      values:
+        - key: "temperatureAlarm.value"
+          enabledValues:
+            - cleared
+            - freeze
+            - heat
   - id: firmwareUpdate
     version: 1
   - id: refresh
--- a/drivers/SmartThings/zigbee-thermostat/profiles/thermostat-temperature-temperaturealarm.yml
+++ b/drivers/SmartThings/zigbee-thermostat/profiles/thermostat-temperature-temperaturealarm.yml
@@ -18,8 +18,8 @@ components:
       values:
         - key: "thermostatOperatingState.value"
           enabledValues:
-            - Heating
-            - Idle
+            - heating
+            - idle
   - id: temperatureAlarm
     version: 1
     config:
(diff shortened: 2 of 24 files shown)


greens (commented):
so not every thermostat only supports heating or idle

most would also support cooling

we have a couple of devices that are heat only, these would be the profiles that only have the thermostat heating setpoint capability without the cooling setpoint capability. Those should have heating and idle, but all other profiles should have Cooling as well

greens (approved):

iplantemn:
@bflorian @greens updated capability configurations

drivers/SmartThings/zigbee-thermostat/profiles/thermostat-temperature-temperaturealarm.yml
```diff
@@ -10,8 +10,21 @@ components:
...
+            - Idle
   - id: temperatureAlarm
     version: 1
+    configuration:
+      values:
+        - key: "thermostatOperatingState.value"
+          enabledValues:
+            - Normal
+            - Overheating
+            - Freezing
```
tpmanley, line 27:
```suggestion
        - key: "temperatureAlarm.value"
          enabledValues:
            - cleared
            - freeze
            - heat
```

tpmanley: Looks like there are a number of similar problems with `temperatureAlarm` in other profiles. @greens it may make more sense for you or someone from your team to fix those up rather than asking @iplantemn to figure out what is correct.

drivers/SmartThings/zigbee-thermostat/profiles/base-thermostat-fanless.yml
```diff
@@ -12,6 +12,13 @@ components:
     version: 1
   - id: thermostatOperatingState
     version: 1
+    configuration:
```
tpmanley, line 15:
```suggestion
    config:
```

@ctowns a similar change will need to be made to all the profiles in this PR

ctowns: done

greens:
@ctowns can you update this one too?

greens (approved):

ctowns (approved):

PR #200, opened 2022-07-28
[CHAD-8901] Updated all thermostatHeatingSetpoint to correct range

Bhavik-T:

PR #210, opened 2022-08-03
[CHAD-8812] Updated qubino-flush-thermostat's thermostatHeatingSetpoint

Bhavik-T:
Updated qubino-flush-thermostat's thermostatHeatingSetpoint to -13F to 176F


greens:
@bflorian about the question of units again

lelandblue:
Hey @ctowns Can you please take a look at this its in F and not in C and I believe @bflorian said this needs to be in C. Thankyou.

ctowns:
Converted from F to C

greens (approved):

iplantemn:
@bflorian @greens updated capability configuration

drivers/SmartThings/zwave-thermostat/profiles/qubino-flush-thermostat.yml
```diff
@@ -6,6 +6,10 @@ components:
     version: 1
   - id: thermostatHeatingSetpoint
     version: 1
+    configuration:
```
tpmanley, line 9:
```suggestion
    config:
```

drivers/SmartThings/zwave-thermostat/profiles/qubino-flush-thermostat.yml
```diff
@@ -6,6 +6,10 @@ components:
     version: 1
   - id: thermostatHeatingSetpoint
     version: 1
+    config:
+      values:
+        - key: "temperature.value"
```
greens, line 11: yeah I think this is the wrong attribute name

ctowns:
It looks like the thermostatHeatingSetpoint for this device was already set as a part of this PR: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/200

In that PR. the range was set to [4, 28] with correct attribute name:
`  - id: thermostatHeatingSetpoint
    version: 1
    config:
      values:
        - key: "heatingSetpoint.value"
          range: [ 4, 28 ]`
          
In this patch however, it is being set to [-25, 80]. Which of these numbers should be considered "correct"? [CHAD-8812](https://smartthings.atlassian.net/browse/CHAD-8812) calls out that the temperature range should be [-25, 80], but I am wondering if that is a misunderstanding. According to the [qubino thermostat technical specification](https://qubino.com/products/flush-onoff-thermostat/flush-on-off-thermostat-technical-specifications2/), the temperature sensor range is [-25, 80] C but the "operational range" is only [-10, 40] C. I am not sure where the [4, 28] value from the other PR came from, but which of these values is correct?

greens:
@ctowns I'm guessing the sensor can detect -25,80, but the setpoints are restricted to -10,40, so -10,40 is probably what we want.

My guess is that Bhavik went with a general range from the DTH side of things and applied it to all thermostats, and some of them have specific values and that's what this PR is about.

greens (approved):

PR #233, opened 2022-08-23
Zigbee lock fixes

varzac:

PR #238, opened 2022-08-24
Bugfix newly added tests that are failing due to integration test bug

cjswedes:
The 44.x release artifact fixed an integration test bug that impacts
tests where messages were expected to be sent but they were not sent.

### DIFF
--- a/drivers/SmartThings/zigbee-contact/src/test/test_aqara_contact_sensor.lua
+++ b/drivers/SmartThings/zigbee-contact/src/test/test_aqara_contact_sensor.lua
@@ -90,7 +90,7 @@ test.register_coroutine_test(
   "closed contact events",
   function()
     local attr_report_data = {
-      { IASZone.ID, data_types.Int16.ID, 0x0020}
+      { IASZone.attributes.ZoneStatus.ID, data_types.Bitmap16.ID, 0x0020 }
     }
     test.socket.zigbee:__queue_receive({
       mock_device.id,
@@ -104,7 +104,7 @@ test.register_coroutine_test(
   "open contact events",
   function()
     local attr_report_data = {
-      { IASZone.ID, data_types.Int16.ID, 0x0021}
+      { IASZone.attributes.ZoneStatus.ID, data_types.Bitmap16.ID, 0x0021}
     }
     test.socket.zigbee:__queue_receive({
       mock_device.id,
--- a/drivers/SmartThings/zigbee-power-meter/src/test/test_zigbee_power_meter_consumption_report.lua
+++ b/drivers/SmartThings/zigbee-power-meter/src/test/test_zigbee_power_meter_consumption_report.lua
@@ -28,7 +28,7 @@ local mock_device = test.mock_device.build_test_zigbee_device(
         [1] = {
           id = 1,
           model = "E240-KR080Z0-HA",
-          server_clusters = {}
+          server_clusters = {SimpleMetering.ID, ElectricalMeasurement.ID}
         }
       }
     }
@@ -144,6 +144,5 @@ test.register_coroutine_test(
       mock_device:expect_metadata_update({ provisioning_state = "PROVISIONED" })
     end
 )
-
 
 test.run_registered_tests()
--- a/drivers/SmartThings/zwave-switch/src/test/test_zooz_zen_30_dimmer_relay.lua
+++ b/drivers/SmartThings/zwave-switch/src/test/test_zooz_zen_30_dimmer_relay.lua
@@ -67,17 +67,6 @@ test.register_coroutine_test(
     test.socket.zwave:__expect_send(
       zw_test_utils.zwave_test_build_send_command(
         mock_zooz_zen_30_dimmer_relay,
-        SwitchBinary:Get({},
-          {
-            encap = zw.ENCAP.AUTO,
-            src_channel = 0,
-            dst_channels={}
-          })
-      )
-    )
-    test.socket.zwave:__expect_send(
-      zw_test_utils.zwave_test_build_send_command(
-        mock_zooz_zen_30_dimmer_relay,
         SwitchMultilevel:Get({},
           {
             encap = zw.ENCAP.AUTO,
--- a/drivers/SmartThings/zwave-thermostat/src/test/test_qubino_flush_thermostat.lua
+++ b/drivers/SmartThings/zwave-thermostat/src/test/test_qubino_flush_thermostat.lua
@@ -46,8 +46,20 @@ local mock_device = test.mock_device.build_test_zwave_device(
     zwave_product_id = 0x0054,
   }
 )
+
+local mock_device_cooling = test.mock_device.build_test_zwave_device(
+  {
+    profile = t_utils.get_profile_definition("qubino-flush-thermostat-cooling.yml"),
+    zwave_endpoints = thermostat_endpoints,
+    zwave_manufacturer_id = 0x0159,
+    zwave_product_type = 0x0005,
+    zwave_product_id = 0x0054,
+  }
+)
+
 local function test_init()
   test.mock_device.add_test_device(mock_device)
+  test.mock_device.add_test_device(mock_device_cooling)
 end
 test.set_test_init_function(test_init)
 
@@ -218,7 +230,7 @@ test.register_message_test(
       {
         channel = "zwave",
         direction = "receive",
-        message = { mock_device.id,
+        message = { mock_device_cooling.id,
                     zw_test_utilities.zwave_test_build_receive_command(ThermostatSetpoint:Report({
                                                                                                    setpoint_type = ThermostatSetpoint.setpoint_type.COOLING_1,
                                                                                                    scale = 0,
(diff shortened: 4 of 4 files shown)


greens (commented):
I don't think we should just be removing all of these. Some of them seem like functionality we'd expect to work...

cjswedes:
@greens is this in line with your understanding of why the aqara tests fail? The subdriver is saying it can handle the aqara IASZone updates, but the subdriver is not specifying handlers for that attribute, so they are just dropped because no handler is found. When I run tests locally I can confirm that the defaults are not being called either. When I add tests for other subdrivers I see them fail as well; only those that use the base driver impl pass the contact sensor tests.

I thought that the base drivers handler would be called if there isn't a handler in the subdriver, but that doesn't appear to be the case.

drivers/SmartThings/zwave-thermostat/src/test/test_qubino_flush_thermostat.lua
```diff
@@ -212,26 +212,6 @@ test.register_message_test(
     }
 )
 
-test.register_message_test(
```
cjswedes, line 215: This can be removed because the device [profile](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/494974b8c37e1ab3274001b8d43ebe2e7083ca80/drivers/SmartThings/zwave-thermostat/profiles/qubino-flush-thermostat.yml#L1) does not support the coolingSetpointCapability. I would assume that the profile is correct, but I am not sure how this device works given this [preference](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/494974b8c37e1ab3274001b8d43ebe2e7083ca80/drivers/SmartThings/zwave-thermostat/profiles/qubino-flush-thermostat.yml#L24-L34) that tells the user to re-add the device after changing it. Any ideas on if the test or the profile should change?

drivers/SmartThings/zwave-switch/src/test/test_zooz_zen_30_dimmer_relay.lua
```diff
@@ -64,17 +64,6 @@ test.register_coroutine_test(
       mock_zooz_zen_30_dimmer_relay.id,
       { capability = "refresh", component = "main", command = "refresh", args = {} }
     })
-    test.socket.zwave:__expect_send(
```
cjswedes, line 67: This seems to be a copy paste error, and can be removed safely. This test is testing that this command is sent to the device, but it was erroneously expecting that it be sent twice.

drivers/SmartThings/zigbee-power-meter/src/test/test_zigbee_power_meter_consumption_report.lua
```diff
@@ -100,50 +100,66 @@ test.register_message_test(
   }
 )
 
+-- TODO should this test be removed, or should the driver be fixed so it passes?
```
cjswedes, line 103: The [device](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/494974b8c37e1ab3274001b8d43ebe2e7083ca80/drivers/SmartThings/zigbee-power-meter/fingerprints.yml#L2-L5) being tested in this test uses the [ezex](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/494974b8c37e1ab3274001b8d43ebe2e7083ca80/drivers/SmartThings/zigbee-power-meter/src/ezex/init.lua#L24) subdriver. It seems like this is another case where the subdriver writer expected default behavior to exist, in this case the default configured attributes, but it doesn't since the subdriver doesn't specify them. It's not clear to me whether the subdriver should be overriding the configure handler, or even what the configure functionality for this device should be; no attributes are being configured on this device.

drivers/SmartThings/zigbee-contact/src/test/test_aqara_contact_sensor.lua
```diff
@@ -86,32 +86,34 @@ test.register_coroutine_test(
...
-      { IASZone.ID, data_types.Int16.ID, 0x0020}
-    }
-    test.socket.zigbee:__queue_receive({
-      mock_device.id,
-      zigbee_test_utils.build_attribute_report(mock_device, IASZone.ID, attr_report_data, 0x115F)
-    })
-    test.socket.capability:__expect_send( mock_device:generate_test_message("main", capabilities.contactSensor.contact.closed()))
-  end
-)
+--TODO There are no handlers defined for these attr reports
```
cjswedes, line 89: As alluded to in another comment, the aqara subdriver appears to be expecting some default handlers for the contact sensor events, but no default handlers exist. Similar issues exist for all other subdrivers of the zigbee-contact driver, its just that they dont have tests for contact events. It seems like the subdrivers will all have to specify their own contact event handlers.

drivers/SmartThings/zwave-switch/src/test/test_zooz_zen_30_dimmer_relay.lua
greens: this is sending to the unencapsulated device though

drivers/SmartThings/zwave-thermostat/src/test/test_qubino_flush_thermostat.lua
greens: it uses profile switching I think, but looks like it's using default handling for these attributes anyway

drivers/SmartThings/zigbee-power-meter/src/test/test_zigbee_power_meter_consumption_report.lua
cjswedes: Actually, it looks like the mock device doesn't specify any server clusters which is why no attributes are configured.

drivers/SmartThings/zwave-switch/src/test/test_zooz_zen_30_dimmer_relay.lua
cjswedes: Im not sure I understand. The [handler](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/494974b8c37e1ab3274001b8d43ebe2e7083ca80/drivers/SmartThings/zwave-switch/src/zooz-zen-30-dimmer-relay/init.lua#L110-L114) only sends 3 commands. Why would we expect a fourth in the test?

drivers/SmartThings/zigbee-contact/src/test/test_aqara_contact_sensor.lua
varzac: I see defaults for zigbee contactSensor.  I would expect these messages to be handled.

drivers/SmartThings/zigbee-contact/src/test/test_aqara_contact_sensor.lua
```diff
@@ -86,32 +86,34 @@ test.register_coroutine_test(
...
+-- This and other subdrivers are not handling these reports
+-- test.register_coroutine_test(
+--   "closed contact events",
+--   function()
+--     local attr_report_data = {
+--       { IASZone.ID, data_types.Int16.ID, 0x0020}
```
varzac, line 95: This line is wrong, it needs to be:

```
{ IASZone.attributes.ZoneStatus.ID, data_types.Bitmap16.ID, 0x0020}
```
and the same is true in the other test.

cjswedes: > The subdriver is saying it can handle the aqara IASZone updates, but the subdriver is not specifying handlers for that attribute, so they are just dropped because no handler is found. When I run tests locally I can confirm that the defaults are not being called either. When I add tests for other subdrivers I see them fail as well; only those that use the base driver impl pass the contact sensor tests.

^^ this is not true, which is what I was almost sure of, but I just couldn't figure out why the defaults weren't called. Ill make the test changes then.

cjswedes:
@varzac https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/233 broke a couple of tests:
```
Unit test failures in /home/parallels/SmartThingsEdgeDrivers/drivers/SmartThings/zwave-lock/src/test/test_samsung_lock.lua:
    All user codes should be reported as deleted upon changing Master Code
Unit test failures in /home/parallels/SmartThingsEdgeDrivers/drivers/SmartThings/zwave-lock/src/test/test_zwave_lock.lua:
    Setting a user code name should be handled
```

greens:
@cjswedes those are broken because of upstream changes.

they'll be fixed with the next library release

drivers/SmartThings/zwave-switch/src/test/test_zooz_zen_30_dimmer_relay.lua
greens: oh I see it was a duplicate now

greens:
@cjswedes as far as I can tell the Aqara contact driver isn't saying anything about handling those attributes beyond what the defaults do

greens (approved):

cjswedes:
> they'll be fixed with the next library release

Should I comment these out or can I merge?

greens:
merge

PR #241, opened 2022-08-28
Added switchLevel capability support.   Parameter support for Leviton 4 speed controller.

csstup:
- Added support for switchLevel capability.   Previous default "zwave fan switch" DTH supported it, and its needed for better Alexa support.   For instance requesting the fan turn on to "25%" instead of having to know the level names.   And the speed is only controllable from the Alexa app via switchLevel. 
- Added preferences/settings for leviton 4 speed fan.

Tested on Leviton 4 speed fan controller model ZW4SF.

Please let me know if I can add anything else!

### DIFF
--- a/drivers/SmartThings/zwave-fan/fingerprints.yml
+++ b/drivers/SmartThings/zwave-fan/fingerprints.yml
@@ -28,4 +28,4 @@ zwaveManufacturer:
     manufacturerId: 0x001D
     productType: 0x0038
     productId: 0x0002
-    deviceProfileName: fan-4speed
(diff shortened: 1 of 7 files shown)


greens:
We made a conscious decision to drop the switchLevel capability as it was duplicating behavior. As you mentioned, it was originally included to make integration with SmartLighting possible, but we're trying to be a bit stricter about our capability mappings these days.

greens:
@csstup Definitely agree with you. I'll see if I can bring it up with our voice team.

PR #242, opened 2022-08-30
PHX-183 Added missing state_change = true to events

dsobierajsk:


greens (approved):

greens:
Are there any other buttons events that need this? We probably just want this true in all cases for button events.

dsobierajsk:
As I see, that's all

PR #243, opened 2022-08-30
Add lock support for lock codes after migration from DTHs

varzac:
This allows us to use the lock codes synced as a part of the DTH->Edge migration to pre-populate our lock state to avoid extra events, and give us necessary context.

### DIFF
--- a/drivers/SmartThings/zigbee-lock/src/lock_utils.lua
+++ b/drivers/SmartThings/zigbee-lock/src/lock_utils.lua
@@ -21,7 +21,8 @@ local lock_utils =  {
   -- Constants
   LOCK_CODES      = "lockCodes",
   CHECKING_CODE   = "checkingCode",
-  CODE_STATE      = "codeState"
+  CODE_STATE      = "codeState",
+  MIGRATION_COMPLETE = "migrationComplete",
 }
 
 lock_utils.get_lock_codes = function(device)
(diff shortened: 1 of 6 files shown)


drivers/SmartThings/zigbee-lock/src/lock_utils.lua
```diff
@@ -73,4 +74,23 @@ function lock_utils.code_deleted(device, code_slot)
...
+  if device.data.lockCodes ~= nil and device:get_field(lock_utils.MIGRATION_COMPLETE) ~= true then
+    -- build the lockCodes table
+    local lockCodes = {}
+    for k, v in pairs(device.data.lockCodes) do
+      lockCodes[k] = v
+    end
+    -- Populate the devices `lockCodes` field
+    device:set_field(lock_utils.LOCK_CODES, utils.deep_copy(lockCodes), { persist = true })
+    -- Populate the devices state history cache
+    -- TODO: "main" usage
```
varzac, line 87: I can remove this.  It might make sense to create a `st.constants` or something similar for something like this, but this is pretty well defined so I'm not too concerned about the raw string usage.

drivers/SmartThings/zwave-lock/src/init.lua
```diff
@@ -68,6 +92,10 @@ local function added_handler(self, device)
   end
 end
 
+local init_handler = function(driver, device, event)
+  populate_state_from_data(device)
```
varzac, line 96: We can discuss if it makes sense to even have this handling in `init` vs just `added` or including `infoChanged`.  I believe that `added` alone should be sufficient, this was just intended as a fallback, so I would be OK with removing it.  I didn't really want to use `infoChanged` because 1) we _shouldn't_ ever receive the data via that path and 2) `infoChanged` happens more frequent than the other lifecycle events and so adding an additional lookup/check seemed like extra work.

drivers/SmartThings/zigbee-lock/src/samsungsds/init.lua
```diff
@@ -49,10 +50,18 @@ local function unlock_cmd_handler(driver, device, command)
 end
 
 local device_added = function(self, device)
+  lock_utils.populate_state_from_data(device)
   device:emit_event(capabilities.lock.lock.unlocked())
   device:emit_event(capabilities.battery.battery(100))
 end
 
+local battery_init = battery_defaults.build_linear_voltage_init(4.0, 6.0)
```
greens, line 58: you could just inline this in `init`

unless it's used elsewhere in the driver I can't see in the web diff?

greens (approved):

drivers/SmartThings/zigbee-lock/src/samsungsds/init.lua
varzac: That would end up with a duplicate function per device where this is just a single function for the driver.  So just a minor optimization 🤷

greens (approved):

tpmanley (approved):
Looks good but just double check the unit test failures to make sure there is nothing unexpected

varzac:
The lock migration tests are failing because of this change: https://github.ecodesamsung.com/iot-hub/scripting-engine/commit/93a4b76fa8acfd9a3ea27e282192f457d4a66c20 that is not yet released that corrects the mock datastore comparison to use stringify_table when comparing tables instead of just comparing the memory address.  All tests pass locally.

rboy1:
@tpmanley @varzac Would this also work if drivers are switched from the stock to a custom driver? The same issue would exist with that scenario.

varzac:
> @tpmanley @varzac Would this also work if drivers are switched from the stock to a custom driver? The same issue would exist with that scenario.

It will not.  As of right now that is not something we support.  This is specifically for the transition from DTH to driver.

PR #244, opened 2022-08-31
HCS-2732: Always start a presence timeout on init

varzac:
This changes the logic to always start a presence timeout timer on device init.
This means that even if the device was in the offline state it will trigger a
not present event.

### DIFF
--- a/drivers/SmartThings/zigbee-presence-sensor/src/arrival-sensor-v1/init.lua
+++ b/drivers/SmartThings/zigbee-presence-sensor/src/arrival-sensor-v1/init.lua
@@ -122,6 +122,7 @@ local function init_handler(self, device, event, args)
       )
     end
   )
+  presence_utils.create_presence_timeout(device)
 end
 
 local arrival_sensor_v1 = {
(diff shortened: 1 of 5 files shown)


drivers/SmartThings/zigbee-presence-sensor/src/arrival-sensor-v1/init.lua
```diff
@@ -111,19 +111,18 @@ end
...
   device:set_field(
-    timer_const.PRESENCE_CALLBACK_CREATE_FN,
+    presence_utils.PRESENCE_CALLBACK_CREATE_FN,
     function(device)
       return device.thread:call_with_delay(
               3 * CHECKIN_INTERVAL + 1,
               function()
                 device:emit_event(PresenceSensor.presence("not present"))
-                device:emit_event(SignalStrength.lqi(0))
-                device:emit_event(SignalStrength.rssi({value = -100, unit = 'dBm'}))
```
varzac, line 121: these signal strength events should have been previously removed when the capabilities were removed from the profile, but this case was missed.

varzac:
This will be easier to review by commit as there was a bit of noise added by a rename, and cleanup of an older change.

greens (approved):

PR #255, opened 2022-09-08
Change visibility and state_change of zigbee lock codes events to mat…

greens:
…ch DTHs

### DIFF
--- a/drivers/SmartThings/zigbee-lock/src/samsungsds/init.lua
+++ b/drivers/SmartThings/zigbee-lock/src/samsungsds/init.lua
@@ -46,7 +46,7 @@ local function unlock_cmd_handler(driver, device, command)
           DoorLock.ID,
           SAMSUNG_SDS_MFR_SPECIFIC_UNLOCK_COMMAND,
           SAMSUNG_SDS_MFR_CODE,
-          "1235"))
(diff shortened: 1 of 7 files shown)


greens:
sister commit to: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/254

drivers/SmartThings/zigbee-lock/src/samsungsds/init.lua
```diff
@@ -46,7 +46,7 @@ local function unlock_cmd_handler(driver, device, command)
           DoorLock.ID,
           SAMSUNG_SDS_MFR_SPECIFIC_UNLOCK_COMMAND,
           SAMSUNG_SDS_MFR_CODE,
-          "1235"))
+          "\x10\x04\x31\x32\x33\x35"))
```
varzac, line 49: This change seems unrelated?  Was it due to a failing test?

varzac (approved):

drivers/SmartThings/zigbee-lock/src/samsungsds/init.lua
greens: some of these explicit hex values weren't copied over correctly to the public repo

PR #258, opened 2022-09-09
Remove security CC from generic GDO fingerprint

greens:
The security command class is not included for the purposes of fingerprinting a device


greens:
not the issue

PR #274, opened 2022-09-21
Don't issue a code reload after migration

varzac:


greens (approved):
The migration of the data wasn't happening first?

varzac:
> The migration of the data wasn't happening first?

It is but the pin response handling that happens after we query generates a `codeChanged` event with state change regardless: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-lock/src/init.lua#L101

PR #275, opened 2022-09-21
Call default state for zwave-siren capabilities

ctowns:
https://smartthings.atlassian.net/browse/CHAD-9325

### DIFF
--- a/drivers/SmartThings/zwave-siren/src/init.lua
+++ b/drivers/SmartThings/zwave-siren/src/init.lua
@@ -13,6 +13,7 @@
 -- limitations under the License.
 
 local capabilities = require "st.capabilities"
+local cap_defaults = require "st.capabilities.defaults"
 --- @type st.zwave.defaults
 local defaults = require "st.zwave.defaults"
 --- @type st.zwave.Driver
@@ -68,6 +69,14 @@ local function do_configure(driver, device)
   end
 end
 
+--- Handle device added
+---
+--- @param self st.zwave.Driver
+--- @param device st.zwave.Device
+local function added_handler(self, device)
+  cap_defaults.emit_default_events(device, self.supported_capabilities)
+end
+
 --------------------------------------------------------------------------------------------
 -- Register message handlers and run driver
 --------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ctowns:
Any suggestions on how I should go about testing this? The unit tests are still passing but I am unsure how to check that this addresses the missing state behavior in described in the ticket.

drivers/SmartThings/zwave-siren/src/init.lua
```diff
@@ -68,6 +68,14 @@ local function do_configure(driver, device)
   end
 end
 
+--- Handle device added
+---
+--- @param self st.zwave.Driver
+--- @param device st.zwave.Device
+local function added_handler(self, device)
+  capabilities.defaults.emit_default_events(device, self.supported_capabilities)
```
varzac, line 76: My only concern here is if we want the default to be to do this for _all_ the capabilities or if we want to be more specific to the problem events?  Also if any of the sub-drivers have their own `added` handlers this won't happen for them so we should review them to determine if this needs to be added to any of their `added` handlers as well.

greens: I feel pretty confident we don't lose anything by emitting it for everything

greens (approved):

greens (approved):

greens (changes requested):
oh wait, can we add a unit test?

ctowns:
I added a unit test, but it required some rework to the 'capabilities.defaults.emit_default_events()' function. I pushed up those changes here: https://github.ecodesamsung.com/iot-hub/scripting-engine/pull/1407

With the above changes, the 'emit_default_events()' function can complete successfully.

greens (approved):

ctowns:
@lelandblue This change required a fix in the scripting engine in order to call the `emit_default_events()` function. Without the fix, the function call will cause an error. The scripting engine fix is in 45.8, but this has not been released to the general population yet. Since this is a breaking change for anyone not on the latest 45, we are waiting to merge until everyone gets the scripting engine change in 45.8.

Essentially because this causes problems for anyone without 45.8 who uses this device, we are waiting until 45.8 goes out to the general population in order to prevent issues. @greens do you know when we can expect 45.8 to be out to the general population?

greens:
I am okay to merge this to alpha, since most folks using alpha will have 45. We can discuss on Tuesday next week whether it gets merged to 45 or not.

greens (approved):

PR #277, opened 2022-09-26
Namespaced preferences must be prefixed in code

greens:


varzac (approved):
Seems OK, I would prefer this wasn't necessary, but the changes look acceptable.

PR #278, opened 2022-09-27
Fix error in min/max

greens:


greens:
btw @chancehiblinds my bad for not seeing this

iplantemn (approved):

PR #292, opened 2022-10-06
IQ-80 Fixes updateCodes method

greens:
Both zigbee and z-wave had the same errors in their updateCodes method around improper detection of when deletion was needed and not including enough delays between commands


greens:
The plan is to merge this change so that the changes to the default methods being in 46 do not block anything. We're overwriting the default method in the base driver.

cjswedes (approved):

sjl2178 (approved):

PR #293, opened 2022-10-06
[CHAD-9168] Re-Port to Parent Child: Qubino Flush 2 Relay

dsobierajsk:


drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
```diff
@@ -11,43 +11,180 @@
...
-    if component_id == "main" then
-      return { 1 }
-    elseif component_id == "extraTemperatureSensor" then
-      return { 3 }
-    else
-      local ep_num = math.floor(component_id:match("switch(%d)"))
-      return { ep_num and tonumber(ep_num)}
-    end
+  if component_id == "main" then
+    return { 0, 1, 2 }
```
greens, line 38: wouldn't this multicast to all these endpoints? I imagine you probably just want to remove the multi-component functions

dsobierajsk: The component to endpoint function is set and called only by the parent device and this is used to multicast commands to all these endpoints from parent device. (During porting I've noticed that ep 0 represents "status" of the whole device, you can send only Get zwave commands to this endpoint if you want to switch on/off the device you need to send set commands to ep 1 and 2 also; Ep 1, 2 are the children devices you can send both Gets and Sets to these endpoints)

drivers/SmartThings/zwave-switch/src/qubino-switches/init.lua
```diff
@@ -54,7 +54,7 @@ local function getDeviceProfile(device, isTemperatureSensorOnboard)
 end
 
 local function can_handle_qubino_flush_relay(opts, driver, device, cmd, ...)
-  return device.zwave_manufacturer_id == constants.QUBINO_MFR
+  return device:id_match(constants.QUBINO_MFR)
```
greens, line 57: did this method get fixed to handle nil arguments? I think guess it did, right?

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
```diff
@@ -11,43 +11,163 @@
...
+  end
 end
 
-local function endpoint_to_component(device, ep)
-    if ep == 2 then
-      return string.format("switch%d", ep)
-    elseif ep == 3 then
-      return "extraTemperatureSensor"
+local function do_refresh(driver, device)
+  for i = 1, 2, 1 do
```
greens, line 95: 1 2 1? why twice to ep 1?

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
```diff
@@ -11,43 +11,163 @@
...
 
-local function map_components(self, device)
-    device:set_endpoint_to_component_fn(endpoint_to_component)
-    device:set_component_to_endpoint_fn(component_to_endpoint)
+local function set_switch(value)
+  return function(driver, device, cmd)
+    local endpoint = device:component_to_endpoint(cmd.component)
+    local delay = constants.DEFAULT_GET_STATUS_DELAY
+    local query_device = function()
+      for _, ep in pairs(endpoint) do
```
greens, line 123: this is written as though `component_to_endpoint` will ever return anything other than `{1}`

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
```diff
@@ -11,43 +11,163 @@
...
-      return { 3 }
-    else
-      local ep_num = math.floor(component_id:match("switch(%d)"))
-      return { ep_num and tonumber(ep_num)}
+  return { 1 }
+end
+
+local function find_child(parent, src_channel)
+  if src_channel == 1 then
+    return parent
```
varzac, line 41: you should just return `nil` if there isn't a child.

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
```diff
@@ -11,43 +11,163 @@
...
+  else
+    return parent:get_child_by_parent_assigned_key(string.format("%02X", src_channel))
+  end
+end
+
+local function device_init(driver, device)
+  if device.network_type == st_device.NETWORK_TYPE_ZWAVE then
+    device:set_find_child(find_child)
+    device:set_component_to_endpoint_fn(component_to_endpoint)
```
varzac, line 50: Is this actually needed?  We should not need both the component->endpoint mapping and the child stuff.

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
varzac: This is a loop starting from 1, going to 2 and stepping by 1.  In this case I think this would just be clearer to duplicate the call to `send_refersh_to_endpoint` twice.

drivers/SmartThings/zwave-switch/src/test/test_qubino_flush_2_relay.lua
```diff
@@ -17,124 +17,487 @@ local t_utils = require "integration_test.utils"
...
+  profile = parent_profile,
+  zwave_endpoints = qubino_flush_2_relay_endpoints,
+  zwave_manufacturer_id = 0x0159,
+  zwave_product_type = 0x0002,
+  zwave_product_id = 0x0051
+})
+
+local mock_child_2_device = test.mock_device.build_test_child_device({
+  profile = child_relay_profile,
+  device_network_id = string.format("%s:%02X", mock_parent_device.device_network_id, 2),
```
varzac, line 65: This is not what the `DNI` will be within edge drivers.  the `device_network_id` will just be a UUID and you shouldn't include it in these templates.

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
```diff
@@ -11,43 +11,163 @@
...
+local function device_added(driver, device)
+  if device.network_type ~= st_device.NETWORK_TYPE_CHILD then
+    driver:try_create_device(get_child_metadata(device, 2))
+    if device:is_cc_supported(cc.SENSOR_MULTILEVEL) then
+      driver:try_create_device(get_child_metadata(device, 3))
     end
+  end
+end
+
+local function send_refresh_to_endpoint(ep, driver, device)
```
varzac, line 84: Instead of having this function take an endpoint and depend on that to determine what refresh commands should be sent, can we not just use the device profile/capabilities supported to determine it?

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
varzac: This also currently means that a `refresh` on the child will refresh both enpoint 1 and 2 which doesn't seem correct

drivers/SmartThings/zwave-switch/src/qubino-switches/init.lua
dsobierajsk: It did.

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
dsobierajsk:
```
function ZwaveChildDevice:component_to_endpoint(component_id)
  local com_to_ep_fn = self:get_field(ZwaveDevice.COMPONENT_TO_ENDPOINT_FUNC)
  if com_to_ep_fn ~= nil then
    return ZwaveDevice.component_to_endpoint(self, component_id)
  else
    return self:get_dst_channel()
  end
end
```
If child device call this function it will return dst_channel `{2}`

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
dsobierajsk: if I return nil in this place reports for ep `{1}` are not handled

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
dsobierajsk: Without it for parent device which has ep `{1}` the reports are not handled properly (commands are sent to ep 0)

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
```diff
@@ -11,43 +11,163 @@
...
+    parent_assigned_child_key = string.format("%02X", endpoint),
+    vendor_provided_label = name
+  }
+end
+
+local function device_added(driver, device)
+  if device.network_type ~= st_device.NETWORK_TYPE_CHILD then
+    driver:try_create_device(get_child_metadata(device, 2))
+    if device:is_cc_supported(cc.SENSOR_MULTILEVEL) then
+      driver:try_create_device(get_child_metadata(device, 3))
```
KKlimczukS, line 79: I would add constants, for example CHILD_SWITCH_ENDPOINT and TEMPERATURE_SENSOR_ENDPOINT and use them to call `get_child_metadata` instead of passing 2 or 3.

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
dsobierajsk: @varzac fixed

drivers/SmartThings/zwave-switch/src/test/test_qubino_flush_2_relay.lua
dsobierajsk: done

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
dsobierajsk: done

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
varzac: This can be done by setting the `component_to_endpoint` function, but I'm curious if we've had to do this for other Z-Wave devices, looking at the code it looks like we have always defaulted everything to ep 0, but it looks like this device doesn't use 0 at all?

greens: ep 0 is the same as unencapsulated, I'm pretty sure

dsobierajsk: I mentioned in one of comments that ep 0 in the device is to report aggregated information about it's state for example in one of switches is on then report will return ON_ENABLE value, also power/energy meter reports for is sum of values in ep 1 and 2. Ealier I did the driver which includes parent (ep 0) child_switch_1 (ep 1) child_switch_2 (ep 2) optional_child_temp_sensor (ep 3), but I've changed it to parent (ep 1) child_switch_2 (ep 2) optional_child_temp_sensor (ep 3) to provide identical behavior to DTH to make it pass migration tests.

dsobierajsk: And not exactly it can be done by conponent_to_endpoint
```
function ZwaveDevice:emit_event_for_endpoint(endpoint, event)
  local find_child_fn = self:get_field(base_device.FIND_CHILD_KEY)
  if find_child_fn ~= nil then
    local child = find_child_fn(self, endpoint)
    if child ~= nil then
      child:emit_event(event)
      return
    end
  else
    local component_id = self:endpoint_to_component(endpoint)
    self:emit_component_event(self.profile.components[component_id], event)
  end
end
```
I need to use emit_event_for_endpoint which does not use the component_to_endpoint function and as you can see here parent device has find_child function so it will not use any of comp -> ep or ep -> comp functions

varzac: Yeah, this is a bug, so you are right for now you need to return the parent from the find_child_fn.  I can create a PR to update this in the future, that won't break returning the parent, but right now the way it works is different between how zigbee and zwave works

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
varzac: Here is an example of what I'm suggesting: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/0f1e5807ecde7a82285405ba496e3719bbba8946

dsobierajsk: Done

drivers/SmartThings/zwave-switch/src/qubino-switches/qubino-relays/qubino-flush-2-relay/init.lua
```diff
@@ -11,43 +11,159 @@
...
+    local query_device = function()
+      do_refresh(driver, device, cmd)
+    end
+
+    device:send_to_component(SwitchBinary:Set({ switch_value = value }), cmd.component)
+    device.thread:call_with_delay(delay, query_device)
+  end
+end
+
+local function sensor_multilevel_report(driver, device, cmd)
```
greens, line 129: I think the DTH also will create the child temperature sensor if it receives a temp event after it's been added.

dsobierajsk: done

greens (approved):

PR #297, opened 2022-10-10
CHAD-9391 - changes double to pressed_2x

KKlimczukS:
CHAD-9391 - changes double to pressed_2x


greens (commented):
I think you probably also want to add a `supportedButtonValues` event for this device. It looks to be missing one?

dsobierajsk (approved):

KKlimczukS:
> I think you probably also want to add a `supportedButtonValues` event for this device. It looks to be missing one?

Sure, done - I had to create a sub-sub driver to handle this since not every Inovelli device in this sub-driver supports button capability.

NikitaKuchta (approved):

PR #300, opened 2022-10-11
[CHAD-9170] Re-Port parent/child zigbee switch power

NikitaKuchta:
Re-Port parent/child zigbee switch power

### DIFF
--- a/drivers/SmartThings/zigbee-switch/profiles/switch-power-2.yml
+++ b/drivers/SmartThings/zigbee-switch/profiles/switch-power-2.yml
@@ -12,11 +12,3 @@ components:
         version: 1
     categories:
       - name: SmartPlug
-  - id: switch1
-    capabilities:
-      - id: switch
-        version: 1
-      - id: powerMeter
-        version: 1
-    categories:
-      - name: SmartPlug
--- a/drivers/SmartThings/zigbee-switch/profiles/switch-power-2.yml
+++ b/drivers/SmartThings/zigbee-switch/profiles/switch-power-2.yml
@@ -12,3 +12,11 @@ components:
         version: 1
     categories:
       - name: SmartPlug
+  - id: switch1
+    capabilities:
+      - id: switch
+        version: 1
+      - id: powerMeter
+        version: 1
+    categories:
+      - name: SmartPlug
--- a/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
@@ -17,6 +17,11 @@ local clusters = require "st.zigbee.zcl.clusters"
 local OnOff = clusters.OnOff
 local SimpleMetering = clusters.SimpleMetering
 
+local ENDPOINTS = {
+  parent = 1,
+  child = 2
+}
+
 local ZIGBEE_DUAL_METERING_SWITCH_FINGERPRINT = {
   {mfr = "Aurora", model = "DoubleSocket50AU"}
 }
@@ -38,7 +43,7 @@ local function device_added(driver, device, event)
       label = name,
       profile = "switch-power-2",
       parent_device_id = device.id,
-      parent_assigned_child_key = string.format("%02X", 2),
+      parent_assigned_child_key = string.format("%02X", ENDPOINTS.child),
       vendor_provided_label = name,
     }
     driver:try_create_device(metadata)
@@ -54,7 +59,7 @@ local function find_child(parent, ep_id)
 end
 
 local function component_to_endpoint(device, component)
-  return 1
+  return ENDPOINTS.parent -- in zigbee drivers we are mapping components to numbers, not to array of numbers as in zwave cases
 end
 
 local function device_init(driver, device)
--- a/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
@@ -38,10 +38,7 @@ end
 
 local function do_refresh(self, device)
   device:send(OnOff.attributes.OnOff:read(device))
-  device:send(SimpleMetering.attributes.Divisor:read(device))
-  device:send(SimpleMetering.attributes.Multiplier:read(device))
-  device:send(ElectricalMeasurement.attributes.ACPowerDivisor:read(device))
-  device:send(ElectricalMeasurement.attributes.ACPowerMultiplier:read(device))
+  device:send(ElectricalMeasurement.attributes.ActivePower:read(device))
 end
 
 local function device_added(driver, device, event)
--- a/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
@@ -17,8 +17,6 @@ local clusters = require "st.zigbee.zcl.clusters"
 local OnOff = clusters.OnOff
 local SimpleMetering = clusters.SimpleMetering
 
-local CURRENT_STATUS = "current_status"
-
 local ZIGBEE_DUAL_METERING_SWITCH_FINGERPRINT = {
   {mfr = "Aurora", model = "DoubleSocket50AU"}
 }
@@ -71,10 +69,8 @@ local function on_off_command_handler(driver, device, value, zb_rx)
   
   if value == OnOff.server.commands.On.ID then
     event = capabilities.switch.switch.on()
-    device:set_field(CURRENT_STATUS, "on")
   elseif value == OnOff.server.commands.Off.ID then
     event = capabilities.switch.switch.off()
-    device:set_field(CURRENT_STATUS, "off")
   end
   
   if event ~= nil then
--- a/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
@@ -15,13 +15,9 @@ local capabilities = require "st.capabilities"
 local st_device = require "st.device"
 local clusters = require "st.zigbee.zcl.clusters"
 local OnOff = clusters.OnOff
-local SimpleMetering = clusters.SimpleMetering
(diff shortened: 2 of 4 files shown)


drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
```diff
@@ -25,31 +32,91 @@ local function can_handle_zigbee_dual_metering_switch(opts, driver, device, ...)
...
 end
 
-local function endpoint_to_component(device, endpoint)
-  if endpoint == 2 then
-    return "switch1"
-  else
-    return "main"
+local function device_added(driver, device, event)
+  if device.network_type == st_device.NETWORK_TYPE_ZIGBEE then
+    for i = 1,1 do
```
greens, line 37: why loop?

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
```diff
@@ -25,31 +32,91 @@ local function can_handle_zigbee_dual_metering_switch(opts, driver, device, ...)
...
 
-local function map_components(self, device)
-  device:set_endpoint_to_component_fn(endpoint_to_component)
-  device:set_component_to_endpoint_fn(component_to_endpoint)
+local function switch_on_command_handler(driver, device, command)
+  device:send(OnOff.server.commands.On(device))
+  
+end
+
+local switch_off_command_handler = function(driver, device, command)
```
greens, line 67: nit: inconsistent function initialization here

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
```diff
@@ -25,31 +32,91 @@ local function can_handle_zigbee_dual_metering_switch(opts, driver, device, ...)
...
+  },
+  capability_handlers = {
+    [capabilities.refresh.ID] = {
+      [capabilities.refresh.commands.refresh.NAME] = do_refresh
+    },
+    [capabilities.switch.ID] = {
+      [capabilities.switch.commands.on.NAME] = switch_on_command_handler,
+      [capabilities.switch.commands.off.NAME] = switch_off_command_handler
+    }
+  },
```
greens, line 116: it looks like you're overwriting default functions with copies of the default functions here. what am I missing?

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
NikitaKuchta: > why loop?

no need, fixed

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
NikitaKuchta: > it looks like you're overwriting default functions with copies of the default functions here. what am I missing?

no need to declare default functions at the beginning (I deleted it, thnx), but implementation in switch_on_command handler and switch_off_command handler are a bit different (in defaults we are sending commands to componets, here we don't have components, but devices)

greens: won't `command.component` just be "main", though?

`device:send` should be the same as `device:send_to_component("main"...`

NikitaKuchta: don't know why, but it works only if I write this method with this implementation:
`local function switch_on_command_handler(driver, device, command)
  device:send(OnOff.server.commands.On(device))
end`

greens: Did you get an error when you ran the tests locally? I'm trying to figure out why it wouldn't work.

NikitaKuchta: Unit test are ok when I run them loacally, but when I test the driver with device, it doesn't work correctly. Commands for parent device are not handled

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
```diff
@@ -25,31 +32,91 @@ local function can_handle_zigbee_dual_metering_switch(opts, driver, device, ...)
...
+local switch_off_command_handler = function(driver, device, command)
+  device:send(OnOff.server.commands.Off(device))
+end
+
+local function on_off_command_handler(driver, device, value, zb_rx)
+  local event
+  
+  if value == OnOff.server.commands.On.ID then
+    event = capabilities.switch.switch.on()
+    device:set_field(CURRENT_STATUS, "on")
```
varzac: where are these used?

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
varzac: We should not override the defaults.  If there are issues, provide us logs so we can investigate why it's not functioning as expected.

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
NikitaKuchta: that is outdated piece of code, was not needed :)

NikitaKuchta:
@varzac @greens I found out what was wrong with not working default methods. The problem was in mapping parent's endpoint. Now it is working correct and tests are passing

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
```diff
@@ -25,31 +35,56 @@ local function can_handle_zigbee_dual_metering_switch(opts, driver, device, ...)
...
   end
 end
 
-local function map_components(self, device)
-  device:set_endpoint_to_component_fn(endpoint_to_component)
-  device:set_component_to_endpoint_fn(component_to_endpoint)
+local function do_refresh(self, device)
+  device:send(OnOff.attributes.OnOff:read(device))
+  device:send(SimpleMetering.attributes.Divisor:read(device))
+  device:send(SimpleMetering.attributes.Multiplier:read(device))
```
KKlimczukS, line 75: @NikitaKuchta don't You need to read power meter attributes too ? (from all endpoints)

NikitaKuchta: done

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
```diff
@@ -25,31 +36,61 @@ local function can_handle_zigbee_dual_metering_switch(opts, driver, device, ...)
...
-    return "switch1"
-  else
-    return "main"
+local function do_refresh(self, device)
+  for endpoint = ENDPOINTS.parent, ENDPOINTS.child do
+    device:send(OnOff.attributes.OnOff:read(device):to_endpoint(endpoint))
+    device:send(SimpleMetering.attributes.Divisor:read(device):to_endpoint(endpoint))
+    device:send(SimpleMetering.attributes.Multiplier:read(device):to_endpoint(endpoint))
+    device:send(ElectricalMeasurement.attributes.ACPowerDivisor:read(device):to_endpoint(endpoint))
+    device:send(ElectricalMeasurement.attributes.ACPowerMultiplier:read(device):to_endpoint(endpoint))
```
KKlimczukS, line 45: @NikitaKuchta I think we may refresh only one device at a time. 
Could You also write unit tests for that ?

KKlimczukS (changes requested):

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
```diff
@@ -25,31 +36,59 @@ local function can_handle_zigbee_dual_metering_switch(opts, driver, device, ...)
...
-  if endpoint == 2 then
-    return "switch1"
-  else
-    return "main"
+local function do_refresh(self, device)
+  device:send(OnOff.attributes.OnOff:read(device))
+  device:send(SimpleMetering.attributes.Divisor:read(device))
+  device:send(SimpleMetering.attributes.Multiplier:read(device))
+  device:send(ElectricalMeasurement.attributes.ACPowerDivisor:read(device))
+  device:send(ElectricalMeasurement.attributes.ACPowerMultiplier:read(device))
```
KKlimczukS, line 44: - Reading divisor and multiplier might be done once after inclusion (but please double check if defaults do it already at some point)
- DTH reads from ELECTRICAL_MEASUREMENT_CLUSTER, attribute 0x050B
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/78747fa5a3b42a0c801f1095f8f7521da968ba42/devicetypes/smartthings/zigbee-multi-switch-power.src/zigbee-multi-switch-power.groovy#L128

KKlimczukS (changes requested):

drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
NikitaKuchta: done

KKlimczukS (approved):
This one might be good to go.
@greens @varzac @dsobierajsk

drivers/SmartThings/zigbee-switch/profiles/switch-power-2.yml
```diff
@@ -10,13 +10,5 @@ components:
...
       - id: refresh
         version: 1
-    categories:
-      - name: SmartPlug
-  - id: switch1
-    capabilities:
-      - id: switch
-        version: 1
-      - id: powerMeter
-        version: 1
```
greens, line 20: It might be more useful to change the profile the fingerprint points to, since deleting this component and still having it called "switch power 2" doesn't make too much sense. We can't move or delete it because Sprocket will throw an error, but this profile is probably useful for some other device later.

NikitaKuchta: @greens done

greens (approved):

PR #301, opened 2022-10-11
CHAD-9609 IKEA button zigbee group binding fix

greens:
Properly fall back to group 0 only if no other binding table entries are found.


varzac (approved):
This looks fine.  Was there also an issue related to adding the device to a group in the fallback case @tpmanley ?

tpmanley (approved):

PR #302, opened 2022-10-11
Release matter drivers

greens:


cjswedes (approved):
Note: Tests will succeed after we release the next library version, and these drivers will remain non functional until Matter is enabled across the platform.

PR #305, opened 2022-10-12
Add Neo Coolcam motion sensors 0083, 1083

jds13:
Add fingerprints for the 83 variant of this sensor (0083 US, 1083 Europe)


jds13:
Typo error

PR #308, opened 2022-10-14
Add metadata info to matter-lock base-lock profile

yhoyoon:
This PR adds UI metadata of matter-lock device. Since Matter Clusters and SmartThings Capabilities is different, it should be displayed as a different UI.

### DIFF
--- a/drivers/SmartThings/matter-lock/profiles/base-lock.yml
+++ b/drivers/SmartThings/matter-lock/profiles/base-lock.yml
@@ -14,3 +14,6 @@ components:
     version: 1
   categories:
   - name: SmartLock
+metadata:
+  mnmn: SmartThingsEdge
+  vid: matter-lock


cjswedes:
What does adding this custom metadata fix? Usually we try to avoid this wherever possible.

yhoyoon:
### Problem

The problem is that the lock state condition that cannot be executed in the automation UI of the matter-lock device is displayed. Five lock states are displayed in the automated UI, but the matter-lock device actually only supports three lock states. The user can even create automation by selecting conditions that the matter-lock device does not support.

### Cause

Since base-lock profile of matter-lock driver has not declared that it uses custom UI metadata, it displays the UI using a generic lock device UI metadata. This is made based on SmartThings Capabilities. The lock attribute of the lock capabilty has the following values:

- locked - The device is locked
- unknown - The state of the device is unknown
- unlocked - The device is unlocked
- unlocked with timeout - The device is unlocked with a timeout

In the automation UI's lock state conditions of the generic lock device UI metadata, based on above values, the following menus are displayed:

- Locked or Unlocked
- Locked
- Unlocked
- Unknown
- Unlocked with timeout

Unfortunately, LockState Attribute has only three possible values in Matter Clusters:

- Not Fully Locked
- Locked
- Unlocked

Only Locked and Unlocked between Matter Clusters and SmartThings Capabilities can be mapped to each other in automation UI.	If the user selects "Unknown" or "Unlocked with timeout" from the automation UI, it can never run.

### Solution

First, I added matter-lock( this is presentation id, not driver name ), a custom UI metadata that only displays the lock/unlock related menus in the automation UI. This PR uses this in the base-lock profile of matter-lock driver.

### References

- https://developer-preview.smartthings.com/docs/devices/capabilities/capabilities-reference/#lock

cjswedes:
Once the feature is released, I think we will prefer #314 over this to avoid having to maintain and understand ui metadata for devices.

yhoyoon:
@cjswedes , @greens I just installed and tested #314 and it works well. Thank you, I will close this PR.

PR #298, opened 2022-10-10
Remove unsupported alarm options for Philio Sound Siren

ctowns:
CHAD-9563: https://smartthings.atlassian.net/browse/CHAD-9563

PR #290, opened 2022-10-04
zigbee-switch: Update white color temp ranges

ctowns:
Updates white color temp ranges with embedded profile config based on vid from former DTH fingerprint.
CHAD-9469: https://smartthings.atlassian.net/browse/CHAD-9469

PR #314, opened 2022-10-18
Matter-lock lock state configuration to exclude unsupported states

cjswedes:
The Lock Capability supports locked, unknown, unlocked, and unlocked with timeout. Matter cluster only ever maps to locked, unknown, and unlocked, so we should only have those states available in the devices capability to not mislead the user when setting automations based on state. 

This is an alternative to #308 that doesn't require managing metadata separately. 

@greens @bflorian I am currently unable to package this driver in prod, and am curious if the capability configuration feature has been released yet or not?

### DIFF
--- a/drivers/SmartThings/matter-lock/profiles/base-lock.yml
+++ b/drivers/SmartThings/matter-lock/profiles/base-lock.yml
@@ -4,6 +4,13 @@ components:
   capabilities:
   - id: lock
     version: 1
+    config:
+      values:
+      - key: "lock.value"
+        enabledValues:
+        - locked
+        - unlocked
+        - unknown
   - id: lockCodes
     version: 1
   - id: tamperAlert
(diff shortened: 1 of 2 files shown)


greens (commented):
It's not released yet.

greens (approved):
Looks good for whenever it _is_ released.

drivers/SmartThings/matter-lock/profiles/base-lock.yml
```diff
@@ -4,6 +4,13 @@ components:
   capabilities:
   - id: lock
     version: 1
+    configuration:
```
tpmanley, line 7:
```suggestion
    config:
```

drivers/SmartThings/matter-lock/profiles/lock-without-codes.yml
```diff
@@ -4,6 +4,13 @@ components:
   capabilities:
   - id: lock
     version: 1
+    configuration:
```
tpmanley, line 7:
```suggestion
    config:
```

drivers/SmartThings/matter-lock/profiles/base-lock.yml
cjswedes: Since you seem to have inside knowledge of this feature, any update on when it will be available in prod? I saw this comment on #298 and #290 as well.

PR #321, opened 2022-10-20
CHAD-9480 Add fallback to group 0

greens:
The ikea motion sensor and the somfy buttons both also had this fallback behavior in their DTHs, even though it is not on-spec.

### DIFF
--- a/drivers/SmartThings/zigbee-button/src/zigbee-multi-button/somfy/init.lua
+++ b/drivers/SmartThings/zigbee-button/src/zigbee-multi-button/somfy/init.lua
@@ -12,6 +12,8 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
+local clusters = require "st.zigbee.zcl.clusters"
+local Groups = clusters.Groups
 local mgmt_bind_resp = require "st.zigbee.zdo.mgmt_bind_response"
 
 local function zdo_binding_table_handler(driver, device, zb_rx)
@@ -22,6 +24,8 @@ local function zdo_binding_table_handler(driver, device, zb_rx)
       return
     end
   end
+  driver:add_hub_to_zigbee_group(0x0000) -- fallback if no binding table entries found
+  device:send(Groups.commands.AddGroup(device, 0x0000))
 end
 
 local somfy = {
--- a/drivers/SmartThings/zigbee-motion-sensor/src/ikea/init.lua
+++ b/drivers/SmartThings/zigbee-motion-sensor/src/ikea/init.lua
@@ -24,6 +24,7 @@ local battery_defaults = require "st.zigbee.defaults.battery_defaults"
 
 local OnOff = clusters.OnOff
 local PowerConfiguration = clusters.PowerConfiguration
+local Groups = clusters.Groups
 
 local IKEA_MOTION_SENSOR_FINGERPRINTS = {
     { mfr = "IKEA of Sweden", model = "TRADFRI motion sensor" }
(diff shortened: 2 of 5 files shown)


drivers/SmartThings/zigbee-motion-sensor/src/ikea/init.lua
```diff
@@ -63,6 +63,7 @@ local function zdo_binding_table_handler(driver, device, zb_rx)
       return
     end
   end
+  driver:add_hub_to_zigbee_group(0x0000) -- fallback if no binding table entries found
```
tpmanley, line 66: The DTH also added the device to group 0x0000 so I think we also need to do that here.

greens: you mean this line:

`cmds += zigbee.command(CLUSTER_GROUPS, 0x00, "${zigbee.swapEndianHex(zigbee.convertToHexString(groupAddr, 4))} 00")`

tpmanley: Yep that's the one

drivers/SmartThings/zigbee-motion-sensor/src/ikea/init.lua
```diff
@@ -63,6 +64,8 @@ local function zdo_binding_table_handler(driver, device, zb_rx)
...
+  device:send(Groups.commands.AddGroup(device, 0x0000, "00"))
```
greens, line 68: @tpmanley this is how I parsed that command from the DTH, but would love some confirmation.

drivers/SmartThings/zigbee-motion-sensor/src/test/test_ikea_motion.lua
```diff
@@ -68,62 +70,83 @@ test.register_message_test(
...
-          }
-      )
-      test.socket.capability:__expect_send(mock_device:generate_test_message("main", capabilities.motionSensor.motion.active()))
-      test.wait_for_events()
-      test.mock_time.advance_time(180)
-      test.socket.capability:__expect_send(mock_device:generate_test_message("main", capabilities.motionSensor.motion.inactive()))
-    end
+  "ZDO Message handler and adding hub to group",
+  function()
+    local binding_table = mgmt_bind_response.BindingTableListRecord("\x6A\x9D\xC0\xFE\xFF\x5E\xCF\xD0", 0x01, 0x0006, 0x01, 0xB9F2)
```
greens, line 134: wanted to add a confirming test here, but not sure how to construct the packet that prompts the fallback

tpmanley: Could set `binding_table_list_count` to 0x00 and not provide any entries

drivers/SmartThings/zigbee-motion-sensor/src/ikea/init.lua
tpmanley: @cjswedes @varzac , double check me here, but I think for a `CharString` the length is handled automatically so we can just pass an empty string here.

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/somfy/init.lua
```diff
@@ -22,6 +22,7 @@ local function zdo_binding_table_handler(driver, device, zb_rx)
       return
     end
   end
+  driver:add_hub_to_zigbee_group(0x0000) -- fallback if no binding table entries found
```
tpmanley, line 25: Looking at the DTH Somfy also needs to add the device to group 0x0000 in this case.

drivers/SmartThings/zigbee-motion-sensor/src/ikea/init.lua
varzac: Correct, just the value needs to be included

greens: oh, that `00` in the DTH was just encoding the length of the (empty) group name? Got it.

tpmanley (approved):

PR #322, opened 2022-10-20
CHAD-9658 Fibaro Smoke: TemperatureAlarm

greens:
Adds cleared temperatureAlarm event on add

corrects profile used for tests, add simple test of temperature alarm defaults


greens:
ideally we'd just use the capability default thing here, but in the interest of expedience, and the added handler already emitting some events I just added a new synthetic event

ctowns (approved):

PR #323, opened 2022-10-20
Fix battery voltage reporting on Sinope valve

ctowns:


greens (approved):
Was 1-indexing the culprit here?

ctowns:
Yeah 1-indexing is the culprit, I keep forgetting Lua does that for some reason

PR #325, opened 2022-10-21
[CHAD-5883] Port 'Z-Wave Device Multichannel' DTH to Lua driver

dsobierajsk:
Original DTH: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zwave-device-multichannel.src/zwave-device-multichannel.groovy


dsobierajsk:
@greens @varzac  I'm wondering if it is possible that non-MCD device which has multiple endpoints can support cc MultiChannel ? If so, I don't have any other ideas how to port this DTH to lua, because in this form if non-MCD device supports MultiChannel cc it going to be impacted by those changes.

greens:
So here's what I would do:

Add only one fingerprint that checks whether the device supports the multichannel cc. Assign it to a copy of the `switch-multilevel` profile with a unique name. Use that profile name in the `can_handle` function of a sub-driver that includes all the code you need to do the interrogation of the different endpoints using `device:is_cc_supported(cc, endpoint)`. You'll probably need a couple of new profiles for the sensor-only endpoints, but the behavior in the defaults from the base driver should be sufficient.

dsobierajsk:
@greens I've tried to make it following your instructions, but I can't dig into device profile name. Do you know how to get it?

greens:
@dsobierajsk Hmm. You are absolutely correct... 

What you could do is add the "zwMultichannel" capability to the profile and check for its existence as your `can_handle`

dsobierajsk:
Hi @greens @varzac what do you thing about creating new main driver "zwave-multichannel" to cover those devices. I'm asking because I've noticed that for every `device.zwave_endpoints` there is also `device_class` which could help determine child type for each endpoint. Without it, there will be duplicated profiles and lua code in `zwave-switch` and `zwave-sensor` drivers.

greens:
@dsobierajsk I think you want to only be working in `zwave-switch`. Other than sensor profiles it should already have everything you need.

dsobierajsk:
2 TC's for sensors are failing now because some changes in battery defaults are needed

greens:
If you look at the multichannel DTH, the parent device is always a multilevel switch. The children can be switches, dimmers, or sensors, based on an interrogation of the supported command classes of each endpoint. What you have now has the parent device being either a sensor or a switch, and the children of a switch can only be binary switches and the children of a sensor can only be sensors.

The behavior of the DTH is why I said you should base your work in the zwave-switch driver, only creating a few child profiles for the sensor possibilities.

If you want to get really clever with the child devices, you can interrogate an endpoint that supports multilevel sensor with a SUPPORTED_GET and parse the result to determine the type of sensor it is, which is (as you've probably noticed) more in-depth than we did with the DTH. That's extra, though, and a generic sensor profile is sufficient.

lelandblue:
I thought this integration was going to be its own driver? Did something change or is that needed? 

cc @varzac

greens:
@lelandblue I don't think it's necessary. I have a bit of a hack in mind to avoid it.

dsobierajsk:
Hi @greens, this is what you were thinking about, weren't you? Ofc, changes in battery defaults are still needed. I will prepare PR with fix asap.

KKlimczukS (commented):
@dsobierajsk 
few unit tests are failing:
```
Unit test failures in /home/runner/work/SmartThingsEdgeDrivers/SmartThingsEdgeDrivers/drivers/SmartThings/zwave-switch/src/test/test_multichannel_device.lua:
    Notification reports (child 4) for POWER_MANAGEMENT events (battery) should be handled
    Notification reports (child 4) for POWER_MANAGEMENT events (battery) should be handled
```

dsobierajsk:
I know, PR in scripting-engine that fixes it is waiting for review

drivers/SmartThings/zwave-switch/profiles/child-generic-sensor.yml
```diff
@@ -0,0 +1,40 @@
...
+    version: 1
+  - id: temperatureMeasurement
+    version: 1
+  - id: voltageMeasurement
+    version: 1
+  # Meter CC based capabilities
+  - id: energyMeter
+    version: 1
+  - id: powerMeter
+    version: 1
```
greens, line 33: for the time being I'd stick to the capabilities handled by the DTH implementation, namely: smoke, motion, water, and contact

drivers/SmartThings/zwave-switch/src/init.lua
```diff
@@ -79,12 +100,49 @@ local function do_configure(driver, device)
...
+    if command_class.value == cc.SENSOR_MULTILEVEL or
+        command_class.value == cc.SENSOR_BINARY or
+        command_class.value == cc.SENSOR_ALARM or
+        command_class.value == cc.NOTIFICATION then
+      return "child-generic-sensor"
+    elseif command_class.value == cc.SWITCH_MULTILEVEL then
+      return "switch-level"
+    end
+  end
+  return "child-switch"
```
greens, line 119: DTH implementation for both of these includes metering

drivers/SmartThings/zwave-switch/src/init.lua
```diff
@@ -79,12 +100,49 @@ local function do_configure(driver, device)
...
     end
   end
 end
 
+local function get_profile(command_classes)
+  for _, command_class in pairs(command_classes) do
+    if command_class.value == cc.SENSOR_MULTILEVEL or
+        command_class.value == cc.SENSOR_BINARY or
+        command_class.value == cc.SENSOR_ALARM or
+        command_class.value == cc.NOTIFICATION then
```
greens, line 113: DTH uses device class here, which should be available in the objects stored in `zwave_endpoints`

greens: (on real devices, you'll have to hardcode it for tests)

greens: and you'll only want the first two bytes, the generic device class

greens: @cjswedes says this might not actually be supported

greens: You can get the configuration of each endpoint by sending a multichannel capability get to each endpoint and having a handler for the capability report, since this info isn't accurate on the device object yet

drivers/SmartThings/zwave-switch/profiles/child-generic-sensor.yml
dsobierajsk: Ok, there are any additional capabilities? I don't know how the default DTH implementation handles Sensor reports.

tools/run_driver_tests.py
```diff
@@ -53,7 +53,7 @@ def run_tests(verbosity_level, filter, junit, coverage_files):
         last_line = ""
         in_progress_test_name = ""
         test_cases = []
-        test_suite = junit_xml.TestSuite(str(test_file)[str(test_file).rindex('/')+1:])
+        test_suite = junit_xml.TestSuite(str(test_file)[str(test_file).find('/')+1:])
```
greens: was this breaking something?

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
```diff
@@ -0,0 +1,93 @@
...
+    parent_device_id = device.id,
+    parent_assigned_child_key = string.format("%02X", endpoint),
+    vendor_provided_label = name
+  }
+end
+
+local function device_added(driver, device)
+  if device.network_type ~= st_device.NETWORK_TYPE_CHILD  then
+    for index, endpoint in pairs(device.zwave_endpoints) do
+      device:send(MultiChannel:CapabilityGet({ end_point = index - 1 }))
```
greens, line 67: I think, since lua is 1-indexed, you can just use `index` here, without subtracting. Endpoint 0 is "special"

NikitaKuchta (approved):

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
dsobierajsk: done

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
```diff
@@ -0,0 +1,93 @@
...
+  if src_channel == 0 then
+    return device
+  else
+    return device:get_child_by_parent_assigned_key(string.format("%02X", src_channel))
+  end
+end
+
+local function device_init(driver, device)
+  if device.network_type == st_device.NETWORK_TYPE_ZWAVE then
+      device:set_find_child(find_child)
```
greens, line 43: nit: spacing

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
```diff
@@ -0,0 +1,93 @@
...
+
+local multichannel_device = {
+  NAME = "Z-Wave Device Multichannel",
+  lifecycle_handlers = {
+    init = device_init,
+    added = device_added
+  },
+  zwave_handlers = {
+    [cc.MULTI_CHANNEL] = {
+      [MultiChannel.CAPABILITY_REPORT] = capability_get_report_handler
```
greens, line 87: I'd like to see a test of this handler.

greens (commented):
This looks like it's almost there. I have a device here that I'm going to try to test this with.

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
dsobierajsk: I've created 2 unit tests (one for the parent device which has the child, and one for parent without children) and for the second one the unit tests fails with following stacktrace: 
```
INFO  || <ZwaveDevice: 00000000-1111-2222-3333-000000000001 [01] (Z-Wave Switch Multichannel)> received Z-Wave command: {args={command_class="", dynamic=false, end_point=1, generic_device_class=16, specific_device_class=0}, cmd_class="MULTI_CHANNEL", cmd_id="CAPABILITY_REPORT", dst_channels={}, encap="AUTO", payload="\x01\x10\x00", src_channel=0, version=2}
TRACE || Found ZwaveDispatcher handler in zwave_switch -> Z-Wave Device Multichannel
Received unexpected error: ...ojects\SmartThingsEdgeDrivers\lua_libs-api_v1\cosock.lua:296: ...SmartThingsEdgeDrivers\lua_libs-api_v1\st\dispatcher.lua:233: Error encountered while processing event for <ZwaveDevice: 00000000-1111-2222-3333-000000000001 [01] (Z-Wave Switch Multichannel)>:
    arg1: {args={command_class="", dynamic=false, end_point=1, generic_device_class=16, specific_device_class=0}, cmd_class="MULTI_CHANNEL", cmd_id="CAPABILITY_REPORT", dst_channels={}, encap="AUTO", payload="\x01\x10\x00", src_channel=0, version=2}
...SmartThingsEdgeDrivers\lua_libs-api_v1\st\dispatcher.lua:233: Error encountered while processing event for <ZwaveDevice: 00000000-1111-2222-3333-000000000001 [01] (Z-Wave Switch Multichannel)>:
    arg1: {args={command_class="", dynamic=false, end_point=1, generic_device_class=16, specific_device_class=0}, cmd_class="MULTI_CHANNEL", cmd_id="CAPABILITY_REPORT", dst_channels={}, encap="AUTO", payload="\x01\x10\x00", src_channel=0, version=2}
...cts\SmartThingsEdgeDrivers\lua_libs-api_v1\st\driver.lua:424: attempt to call a nil value (field 'create_device')
stack traceback:
	[C]: in function 'error'
	...cts\SmartThingsEdgeDrivers\lua_libs-api_v1\st\thread.lua:76: in function <...cts\SmartThingsEdgeDrivers\lua_libs-api_v1\st\thread.lua:54>
stack traceback:
	[C]: in function 'error'
	...ojects\SmartThingsEdgeDrivers\lua_libs-api_v1\cosock.lua:296: in function 'cosock.run'
	...cts\SmartThingsEdgeDrivers\lua_libs-api_v1\st\driver.lua:758: in function 'st.driver.run'
	./init.lua:147: in main chunk
	[C]: in function 'require'
	[C]: in function 'xpcall'
	...ngsEdgeDrivers\lua_libs-api_v1\integration_test\init.lua:220: in upvalue 'run_configured_test'
	...ngsEdgeDrivers\lua_libs-api_v1\integration_test\init.lua:336: in function 'integration_test.run_registered_tests'
	...hings\zwave-switch\src\test\test_multichannel_device.lua:2309: in main chunk
	[C]: in ?
FAILED
```

It is unable to call `create_device'. You can see it in next commit I've pushed.

cjswedes: In this case, perhaps you should comment out the test and leave it with a TODO to address after the 046x release of the lua libraries (hopefully such a thing can get in that release, but no guarantees). We will have to test it manually if possible.

greens: That sounds good. Thanks @cjswedes

greens (commented):
Worked pretty well for me locally!

dsobierajsk:
I've checked it also with Qubino Flush 2 Relay (I've commented the fingerprints) and it also works well.

KKlimczukS (approved):

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
```diff
@@ -0,0 +1,93 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+local cc = require "st.zwave.CommandClass"
+local capabilities = require "st.capabilities"
+local st_device = require "st.device"
+local MultiChannel = (require "st.zwave.CommandClass.MultiChannel")({ version = 3 })
+
+local map_device_class_to_profile = {
+  [0x10] = "metering-switch",
```
cjswedes, line 20: Switch binary device type maps to metering-switch? Im guessing this is just an assumption to better support all devices because they dont report functionality well via Z-Wave protocol?

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
```diff
@@ -0,0 +1,93 @@
...
+  [0x31] = "metering-switch",
+  [0x11] = "metering-dimmer",
+  [0x08] = "metering-sensor",
```
cjswedes, line 23: Thermostat generic type maps to Metering sensor?

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
```diff
@@ -0,0 +1,93 @@
...
+
+local function device_init(driver, device)
+  if device.network_type == st_device.NETWORK_TYPE_ZWAVE then
+    device:set_find_child(find_child)
+  end
+end
+
+local function get_profile(generic_device_class)
+  local device_class = generic_device_class or 0x10
+  return map_device_class_to_profile[device_class] or "metering_switch"
```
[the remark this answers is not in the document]
dsobierajsk: You are right

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
dsobierajsk: It is mapped to metering switch because [Switch Endpoint](https://github.com/SmartThingsCommunity/SmartThingsPrivate/blob/master/devicetypes/smartthings/switch-endpoint.src/switch-endpoint.groovy) also contains power and energy capabilities

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
dsobierajsk: Ok, I looked into the DTH again an it maps to Multi Sensor Endpoint, which contains temperature, luminance and humidity and Sensor Endpoints contains water, motion, smoke, water

dsobierajsk:
Thanks @cjswedes for comments. Now I'm 90% sure that current version will work in the same way as DTH. (The missing 10% are device labels 😄)

drivers/SmartThings/zwave-switch/src/multichannel-device/init.lua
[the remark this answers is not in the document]
dsobierajsk: I mean about this fragment of code https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/78747fa5a3b42a0c801f1095f8f7521da968ba42/devicetypes/smartthings/zwave-device-multichannel.src/zwave-device-multichannel.groovy#L98

greens (approved):

PR #328, opened 2022-10-24
[CHAD-9167] Re-port multi-metering switch to parent/child

NikitaKuchta:


drivers/SmartThings/zwave-switch/profiles/switch-binary.yml
```diff
@@ -1,10 +1,10 @@
...
-  categories:
-  - name: Switch
+  - id: main
+    capabilities:
+    - id: switch
+      version: 1
+    - id: refresh
+      version: 1
+    categories:
+    - name: Switch
```
greens, line 10: nit:  whitespace changes

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,186 @@
...
+  else
+    return "metering-switch"
+  end
+end
+
+local function device_added(driver, device, event)
+  if device.network_type == st_device.NETWORK_TYPE_ZWAVE then
+    local children_amount = get_children_amount(device)
+    for i = 2, children_amount+1 do
+      local name = string.format("%s %d", device.label, i)
```
greens, line 68: won't this result in "Zooz Power Strip 1 2", for example?

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,186 @@
...
+  if ep_id == 1 then
+    return parent
+  else
+    return parent:get_child_by_parent_assigned_key(string.format("%02X", ep_id))
+  end
+end
+
+local function endpoint_to_component(device, endpoint)
+  return "main"
+end
```
greens, line 92: I think this can be omitted.

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
NikitaKuchta: > won't this result in "Zooz Power Strip 1 2", for example?

correct, my bad, fixed now

drivers/SmartThings/zwave-switch/profiles/switch-binary.yml
NikitaKuchta: > nit: whitespace changes

reverted

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
NikitaKuchta: > I think this can be omitted.

done

drivers/SmartThings/zwave-switch/fingerprints.yml
```diff
@@ -253,25 +272,19 @@ zwaveManufacturer:
...
     manufacturerId: 0x010F
     productType: 0x1C01
     productId: 0x1000
     deviceProfileName: fibaro-walli-dimmer
-  - id: "ZoozPowerStrip"
-    deviceLabel: Zooz Power Strip
-    manufacturerId: 0x027A
-    productType: 0xA000
-    productId: 0xA004
-    deviceProfileName: switch-multicomponent-generic-5
```
KKlimczukS, line 274: @NikitaKuchta why was this fingerprint removed ?

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,182 @@
...
+  if device:get_manufacturer() == 0x015F then
+    return "switch-binary"
+  else
+    return "metering-switch"
+  end
+end
+
+local function device_added(driver, device, event)
+  if device.network_type == st_device.NETWORK_TYPE_ZWAVE then
+    local children_amount = get_children_amount(device)
```
KKlimczukS, line 66: @NikitaKuchta You may utilize similar mechanism to this one:  https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zwave-switch/src/zwave-dual-switch/dual_switch_configurations.lua
to obtain all needed config information (like: profile and number of child devices) and avoid 'if manufacturer' statements in the code.

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,182 @@
...
+  if device:supports_capability_by_id(capabilities.powerMeter.ID) or device:supports_capability_by_id(capabilities.energyMeter.ID) then
+    device:send_to_component(Meter:Get({ scale = Meter.scale.electric_meter.WATTS }), command.component)
+    device:send_to_component(Meter:Get({ scale = Meter.scale.electric_meter.KILOWATT_HOURS }), command.component)
+  end
+end
+
+local function switch_binary_report_handler(driver, device, cmd)
+  local event
+  local newValue
+  if(cmd.args.target_value ~= nil) then
```
KKlimczukS, line 117: nit: space after `if`

KKlimczukS (changes requested):

KKlimczukS (approved):
lgtm
@greens @varzac @dsobierajsk

greens (commented):
We can't delete already-uploaded and packaged profiles. Sprocket will reject them because it could leave "orphaned" devices.

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,168 @@
...
+      unit = supportedUnit
+    }
+    
+    local capabilityAttribute = map_scale_to_capability[cmd.args.scale]
+    device:emit_event_for_endpoint(
+      cmd.src_channel,
+      capabilityAttribute(event_arguments)
+    )
+  end
+end
```
greens, line 144: The defaults don't work here because you have to specifically filter out events from channel `0`, right?

I'd still prefer if you called the default handlers from within this method, though. After you filter by channel, you should be able to pass to the Power and Energy defaults like we do here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zwave-lock/src/keywe-lock/init.lua#L57

Since they both have built-in unit checking, you can just pass the `cmd` without checking the unit. That should also allow you to delete those maps.

NikitaKuchta: Ok, I deleted these methods, so now it is handled by defaults

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,117 @@
...
+      return true
+    end
+  end
+  return false
+end
+
+local function device_added(driver, device, event)
+  if device.network_type == st_device.NETWORK_TYPE_ZWAVE then
+    local children_amount = MULTI_METERING_SWITCH_CONFIGURATION_MAP.get_child_amount(device)
+    for i = 2, children_amount+1, 1 do
```
ctowns, line 54: The zwave-multi-metering-switch [DTH](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/78747fa5a3b42a0c801f1095f8f7521da968ba42/devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy#L374) creates child switches for values [2, numberOfSwitches]. In a case where number of switches = 5, this would then create switches on endpoint 2, 3, 4, 5 (4 total switches).

In this LUA driver, it creates child switches for values [2, childrenAmount + 1] and therefore results in 5 child switches. Is this change in behavior expected?

ctowns: If this behavior is expected, then would we need to decrease the number of children? For example in the DTH the Zooz Zen Strip has 4 child switches created (as explained above): https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/78747fa5a3b42a0c801f1095f8f7521da968ba42/devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy#L107

Here in this driver, the amount of children for the Zooz Zen Strip is set to 5, which results in 5 switches, so should the number of children be decreased?

NikitaKuchta: Yes, you're correct. 
I decreased the number of children in configuration file, so now the driver creates child switches according to DTH with correct amount.

greens (commented):
Looking at the DTH, the way the number of child endpoints is calculated is:

```
		if (isZoozZenStripV2()) {
			addChildSwitches(5)
		} else if (isZoozDoublePlug()) {
			addChildSwitches(2)
		} else {
			addChildSwitches(cmd.endPoints)
		}
	}
```

I'd prefer to keep that implementation rather than adding additional hard-coding.

KKlimczukS:
> Looking at the DTH, the way the number of child endpoints is calculated is:
> 
> ```
> 		if (isZoozZenStripV2()) {
> 			addChildSwitches(5)
> 		} else if (isZoozDoublePlug()) {
> 			addChildSwitches(2)
> 		} else {
> 			addChildSwitches(cmd.endPoints)
> 		}
> 	}
> ```
> 
> I'd prefer to keep that implementation rather than adding additional hard-coding.

@greens Do You men to avoid using [multi_metering_switch_configurations.lua](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/328/files#diff-7c641dc5cce08a302ee472cd601432bd45967c8e611e6918c20a49f4424afb88) ?
This file keeps not only the number of child switches but also device profile names and this sub-driver calls:
- MULTI_METERING_SWITCH_CONFIGURATION_MAP.get_child_amount(device)
- MULTI_METERING_SWITCH_CONFIGURATION_MAP.get_child_switch_device_profile(device)

greens:
@KKlimczukS I mean to use that field for the devices that have a defined value in the DTH, but use the number of endpoints for the devices where the DTH does not. So if it's not a ZoozZenStripV2 or a ZoozDoublePlug, then use the number of endpoints to determine the number of child devices, like the DTH does.

KKlimczukS:
> @KKlimczukS I mean to use that field for the devices that have a defined value in the DTH, but use the number of endpoints for the devices where the DTH does not. So if it's not a ZoozZenStripV2 or a ZoozDoublePlug, then use the number of endpoints to determine the number of child devices, like the DTH does.

@greens 
and to use this table ? `device.zwave_endpoints[]`

greens:
@KKlimczukS right, the number of entries should be the same

drivers/SmartThings/zwave-switch/src/multi-metering-switch/multi_metering_switch_configurations.lua
```diff
@@ -17,110 +17,110 @@ local devices = {
...
       mfrs = 0x0086,
       product_types = 0x0003,
-      product_ids = 0x0084,
-      children = 1
+      product_ids = 0x0084
     },
     CONFIGURATION = {
-      child_switch_device_profile = "metering-switch"
+      child_switch_device_profile = "metering-switch",
+      children = 1
```
greens, line 24: this should be unset for any device that's not one of the two zooz devices in the DTH

NikitaKuchta: @greens I deleted this file due to needless use. Now the logic is as in DTH. The only difference is the lack of one fingerprint (Zooz Double Plug) beacuse of moving it to another driver (zwave-dual-switch).

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,140 @@
...
+  end
+end
+
+local function device_added(driver, device, event)
+  if device.network_type == st_device.NETWORK_TYPE_ZWAVE then
+    local children_amount
+    if is_zooz_power_strip(device) then
+      children_amount = 4;
+    else
+      children_amount = device.zwave_endpoints
```
KKlimczukS, line 86: Since `device.zwave_endpoints` is a table, shouldn't You get the length of it here ? (excluding main endpoint)

NikitaKuchta: fixed

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,140 @@
...
+  else
+    return "metering-switch"
+  end
+end
+
+local function is_zooz_power_strip(device)
+  if device:get_manufacturer() == 0x027A and device:get_model() == 0xA004 then
+    return true
+  end
+end
```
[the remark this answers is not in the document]
KKlimczukS: > @KKlimczukS I have similar thoughts about it. Maybe sub-sub-drivers should be created instead of current checks.

@dsobierajsk this PR previously contained a configuration file that returned number of child devices and a child device profile. I'm wondering here if we should use it again and rely on device.zwave_endpoints[] if the configuration file doesn't return anything.

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,140 @@
...
+  local component = command and command.component and command.component or "main"
+  if device:is_cc_supported(cc.SWITCH_BINARY) then
+    device:send_to_component(SwitchBinary:Get({}), component)
+  elseif device:is_cc_supported(cc.BASIC) then
+    device:send_to_component(Basic:Get({}), component)
+  end
+  if device:supports_capability_by_id(capabilities.powerMeter.ID) or device:supports_capability_by_id(capabilities.energyMeter.ID) then
+    device:send_to_component(Meter:Get({ scale = Meter.scale.electric_meter.WATTS }), component)
+    device:send_to_component(Meter:Get({ scale = Meter.scale.electric_meter.KILOWATT_HOURS }), component)
+  end
```
greens, line 123: is this not handled by the default?

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,126 @@
...
+
+local function create_child_device(driver, device, children_amount)
+  if children_amount == 0 then return end
+  for i = 2, children_amount+1, 1 do
+    local device_name_without_number = string.sub(driver.label, 0,-2)
+    local name = string.format("%s%d", device_name_without_number, i)
+    local metadata = {
+      type = "EDGE_CHILD",
+      label = name,
+      profile = MULTI_METERING_SWITCH_CONFIGURATION_MAP.get_child_switch_device_profile(device),
```
KKlimczukS, line 59: Could You move it to `device_added` and call `create_child_device` with `device_profile` parameter ?

NikitaKuchta: done

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,129 @@
...
+      vendor_provided_label = name,
+    }
+    driver:try_create_device(metadata)
+  end
+end
+
+local function device_added(driver, device, event)
+  if device.network_type == st_device.NETWORK_TYPE_ZWAVE then
+    local children_amount = MULTI_METERING_SWITCH_CONFIGURATION_MAP.get_child_amount(device)
+    local profile = MULTI_METERING_SWITCH_CONFIGURATION_MAP.get_child_switch_device_profile(device)
```
KKlimczukS, line 70: nit: what about renaming it to `device_profile`?

NikitaKuchta: done

KKlimczukS (approved):
lgtm

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
greens: I would have expected that setting the `component_to_endpoint` function to handle that, no?

drivers/SmartThings/zwave-switch/src/multi-metering-switch/multi_metering_switch_configurations.lua
```diff
@@ -0,0 +1,143 @@
...
+  },
+  WYFY_TOUCH_4_1 = {
+    MATCHING_MATRIX = {
+      mfrs = 0x015F,
+      product_types = 0x3141,
+      product_ids = 0x5102,
+    },
+    CONFIGURATION = {
+      child_switch_device_profile = "switch-binary"
+    }
```
greens, line 115: you could probably leave this blank for the other devices and just default to `metering-switch` if this field is not present

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
```diff
@@ -0,0 +1,129 @@
...
+  end
+end
+
+local function device_added(driver, device, event)
+  if device.network_type == st_device.NETWORK_TYPE_ZWAVE then
+    local children_amount = MULTI_METERING_SWITCH_CONFIGURATION_MAP.get_child_amount(device)
+    local device_profile = MULTI_METERING_SWITCH_CONFIGURATION_MAP.get_child_switch_device_profile(device)
+    if children_amount == nil then
+      local endpoints = 0
+      for _, _ in pairs(device.zwave_endpoints) do
```
greens, line 73: we have a `utils.table_size`

NikitaKuchta: done

drivers/SmartThings/zwave-switch/src/multi-metering-switch/multi_metering_switch_configurations.lua
NikitaKuchta: done

drivers/SmartThings/zwave-switch/src/multi-metering-switch/init.lua
NikitaKuchta: I assume that something is wrong with default refresh, because as you see I set `component_to_endpoint` function on 88 line. The "Refresh on parent device sends commands to 1 endpoint" test doesn't pass if I delete this overriden refresh function.

greens: @cjswedes Can you help me out here? It looks like we might have an error in the z-wave default refresh builder if the parent/main device uses an endpoint rather than no encapsulation

greens: Oh. Looks like it's fixed in v46.

greens (approved):
This looks good, but can you put a note next to the refresh overwrite reminding us we should delete it when v46 is released?

PR #342, opened 2022-10-31
Remove samsung fingerprints for matter devices

yhoyoon:
The fingerprinting logic can find the driver of the device by using the generic fingerprint instead of manufacturer specific fingerprint. For this reason, this change removes samsung fingerprints for matter devices.


cjswedes (approved):
Looks like all failing unit tests are ones that are unrelated to this PR

greens (approved):

PR #352, opened 2022-11-02
removed handle refresh from matter window covering

sjl2178:
Removed handle refresh from matter window covering driver as it is not longer needed and to take care of a previous nil value.`if read_request = nil` in handle refresh function.


sjl2178:
@cjswedes I forget if I need to make a call to the default refresh function cooper wrote?

tpmanley (approved):

cjswedes (approved):

PR #355, opened 2022-11-03
[Aqara] Motion detection updated 

sbkimaqara:


Kwang-Hui (commented):
Major changes are.
- change default `frequency` value to firmware's value
- fix missing send event in case of 0 lux

drivers/SmartThings/zigbee-motion-sensor/src/aqara/motion-illuminance/init.lua
```diff
@@ -56,20 +62,18 @@ local function motion_illuminance_attr_handler(driver, device, value, zb_rx)
   -- The low 16 bits for Illuminance
   -- The high 16 bits for Motion Detection
 
-  if value.value > MOTION_DETECTED_UINT32 then
```
greens, line 59: I think you probably should just change this to `>=` to avoid the possibility that you emit a negative value for lux.

sbkimaqara: You are right. 
I also changed to the default value of detection frequency interval variable to the device's default value (1 min).

Kwang-Hui: Hi @sbkimaqara, could you update like below?
```diff
-  if value.value >= MOTION_DETECTED_UINT32 then
+  if value.value >= MOTION_DETECTED_UINT32 then
```

sbkimaqara: The IF conditional statement is not needed. The handler is called when motion is detected, and values include motion and illuminance values. So the motion UI is updated without conditional statements and the illuminance UI updated using low bit of received value.

sbkimaqara: > You are right. I also changed to the default value of detection frequency interval variable to the device's default value (1 min).

Sorry. I think my answer makes misunderstood you. Please refer to the comments below.

sbkimaqara: I'll push again after cleaning the code.

sbkimaqara:
The variable value returns nil after reboot the hub. Is there any other way to store a variable forever?
```
device:set_field(PREF_FREQUENCY_KEY, value, { persist = true })
```

Kwang-Hui:
> The variable value returns nil after reboot the hub. Is there any other way to store a variable permanently?
> 
> store
> 
> ```
> device:set_field(PREF_FREQUENCY_KEY, value, { persist = true })
> ```
> 
> retrive
> 
> ```
> local detect_duration = device:get_field(PREF_FREQUENCY_KEY) or PREF_FREQUENCY_VALUE_DEFAULT
> ```

@greens Could you answer for this question?

greens:
I believe Zach answered you in Slack, but to repeat what he said, filesystem writes are not instant.

greens (approved):
Looks okay. Let me know if you're ready to merge.

sbkimaqara:
> Looks okay. Let me know if you're ready to merge.

I ready! Thank you 👍

PR #359, opened 2022-11-04
CHAD-9560: fixed on function in zwave virtual momentary contact

sjl2178:
The on function in zwave virtual momentary contact immediately sent the off command after the on command without the three second delay.  The `momentary_switch_on` function was edited to include the 3 second delay. The issue was that the value of the `switch_off` function was being passed into the delay command not the actual function, so `switch_off` was called and then its return value was passed into the delay function, which was a `nil` value. Tests were also edited to test for 3 second delay.

https://smartthings.atlassian.net/browse/CHAD-9560


cjswedes (approved):

greens (approved):
Nice catch.

PR #366, opened 2022-11-08
added read request for battery percent for ikea remote control

sjl2178:
Added read request in added handler for Ikea remote control in zigbee button driver. Will send a read `BatteryPercentageRemaining` upon the device being handled to fix initial nil battery status, also fixed tests. 

https://smartthings.atlassian.net/browse/BUG-6272

### DIFF
--- a/drivers/SmartThings/zigbee-button/src/zigbee-multi-button/ikea/TRADFRI_remote_control.lua
+++ b/drivers/SmartThings/zigbee-button/src/zigbee-multi-button/ikea/TRADFRI_remote_control.lua
@@ -57,6 +57,7 @@ local function added_handler(self, device)
       device:emit_component_event(comp, capabilities.button.numberOfButtons({value = 1}, {visibility = { displayed = false }}))
     end
   end
+  device:send(PowerConfiguration.attributes.BatteryPercentageRemaining:read(device))
   device:send(PowerConfiguration.attributes.BatteryVoltage:read(device))
   device:emit_event(capabilities.button.button.pushed({state_change = false}))
 end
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/zigbee-button/src/zigbee-multi-button/ikea/TRADFRI_remote_control.lua
```diff
@@ -57,6 +57,7 @@ local function added_handler(self, device)
       device:emit_component_event(comp, capabilities.button.numberOfButtons({value = 1}, {visibility = { displayed = false }}))
     end
   end
+  device:send(PowerConfiguration.attributes.BatteryPercentageRemaining:read(device))
   device:send(PowerConfiguration.attributes.BatteryVoltage:read(device))
```
greens, line 61: I think you probably just want to delete the `added` method here and use the parent's 

Looks like this device uses percentage, not voltage, but I missed it when fixing a long time ago: https://github.com/SmartThingsCommunity/SmartThingsEdgeDriversBeta/pull/589/files

sjl2178: @greens Looking at this `added_handler` it seems to be emitting different values for button5 than the other buttons, which is slightly more complicated than what the parent's `added_handler` does. I could probably still use the parent handler but it would need to be edited a bit, should I jut edit the parent handler or just keep this `added` handler?

greens: ah, good eye. in that case you just want to switch the voltage read with a percentage read

greens (approved):

cjswedes (approved):
LGTM. I also tested with my ikea button, and it looks good.

PR #370, opened 2022-11-09
Feature/matter remove cluster element list usage

cjswedes:
This is needed since the next library release (0.46.x) will completely deprecate and remove some of the opts functionality of the `MatterDevice:get_endpoints(cluster_id, opts)`. 

These changes also discovered a unit test bug that will be fixed in 0.46.x where `require "version"` would fail when run with the unit test framework.


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -15,10 +15,19 @@
 local capabilities = require "st.capabilities"
 local log = require "log"
 local clusters = require "st.matter.clusters"
-
+local version = require "version"
 local MatterDriver = require "st.matter.driver"
 local utils = require "st.utils"
 
+local ThermostatClusterAttributeList
+if version.api < 3 then
```
cjswedes, line 23: Note that explicitly branching on `version.api` is not completely necessary; we could just require the attribute class packaged with the driver for all cases, but I did it this way to document with code the reasoning for including the cluster library file in the driver package.

cjswedes:
> I would push to switch over to using the latest dev artifact, but I worry that would lead to us accidentally pushing changes to prod that will break on production hubs.

I would worry this too, and think that we cannot do this. We could do a beta artifact release that comes before users actually have the artifact on their hubs, and we could run tests against the beta artifact and the most recent production release. We could even do it without doing any extra release of artifacts, but would need to clearly define what test failures are a blocker for release.

drivers/SmartThings/matter-thermostat/src/init.lua
cjswedes: This is done.

cjswedes:
@tpmanley @greens @ctowns This is ready for a review. It is essentially adding the support for thermostats that have ThermostatOperatingState capability, which is based off the optional attribute ThermostatRunningState. I believe all hubs have FW 0.46 or greater, so various workarounds have been removed and it seems to work with the more recent profile changes for nobattery.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -212,6 +213,17 @@ local function system_mode_handler(driver, device, ib, response)
   end
 end
 
+local function attr_list_handler(driver, device, ib, response)
+  for _, attr_id in ipairs (ib.data.elements or {}) do
+    if attr_id.value == clusters.Thermostat.attributes.ThermostatRunningState.ID then
+      local new_profile = string.gsub(device:get_field("profile_name"), "-nostate", "")
+      device.log.info(string.format("Updating device profile to %s.", new_profile))
+      device:try_update_metadata({ profile = new_profile })
```
greens, line 221: do you need to `set_field` here?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -415,6 +428,7 @@ local matter_driver_template = {
         [clusters.Thermostat.attributes.AbsMinCoolSetpointLimit.ID] = setpoint_limit_handler(setpoint_limit_device_field.MIN_COOL),
         [clusters.Thermostat.attributes.AbsMaxCoolSetpointLimit.ID] = setpoint_limit_handler(setpoint_limit_device_field.MAX_COOL),
         [clusters.Thermostat.attributes.MinSetpointDeadBand.ID] = min_deadband_limit_handler,
+        [clusters.Thermostat.attributes.AttributeList.ID] = attr_list_handler
```
greens, line 431: what triggers a read here? the subscription?

hcarter-775:
I think this can be closed- it appears the added functionality from this PR has been introduced separately since then

PR #371, opened 2022-11-09
Add automated check for duplicate profiles on new PRs

ctowns:


greens:
funny enough, this causes the channel creation CI to fail because the name (in the file itself) of the profile you added is a duplicate: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/actions/runs/3441859101

.github/workflows/duplicate-profiles.yml
```diff
@@ -0,0 +1,42 @@
+name: Check for duplicate profiles
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - 'drivers/**'
```
greens, line 6: you could make this `drivers/**/profiles/**` _I think_

greens: maybe even `drivers/**/profiles/*.yml`

.github/scripts/test_python.py
```diff
@@ -0,0 +1,76 @@
...
+                                if len(new_component["capabilities"]) == len(current_component["capabilities"]):
+                                    #check if top capability is the same
+                                    if new_component["capabilities"][0] == current_component["capabilities"][0]:
+                                        # check if capabilities are a direct match with ordering
+                                        if new_component["capabilities"] == current_component["capabilities"]:
+                                            print("capabilities are the exact same with same ordering")
+                                            is_duplicate = True
+                                        # check if capabilities are the same with same top capability but different subsequent ordering
+                                        else:
+                                            new_cap_set = set()
```
greens, line 53: `new_cap_set = set([capability["id"] for capability in new_component["capabilities"])`

I think that should do what you want in a nice pythonic one-liner

.github/scripts/test_python.py
```diff
@@ -0,0 +1,76 @@
...
+                if profile != new_profile and Path(profile).suffix == ".yml":
+                    print("\n---------COMPARISON-----------\ncomparing %s vs %s" % (new_profile, profile))
+                    is_duplicate = False
+                    with open(new_profile) as new_data, open(profile) as current_data:
+                        new_profile_map = yaml.safe_load(new_data)
+                        current_profile_map = yaml.safe_load(current_data)
+
+                        # compare each component
+                        if len(new_profile_map["components"]) == len(current_profile_map["components"]):
+                            for y in range(len(new_profile_map["components"])):
```
greens, line 38: you can save yourself a line by doing:

`for y, new_component in enumnerate(new_profile_map["components"])`

.github/scripts/test_python.py
ctowns: Condensing 5 lines into 1 is peak python, thanks!

.github/scripts/test_python.py
ctowns: Nice! I did not know about that nifty python trick, thank you for the heads up!

.github/workflows/duplicate-profiles.yml
```diff
@@ -0,0 +1,47 @@
...
+      #   uses: peter-evans/create-or-update-comment@v2
+      #   with:
+      #     body-file: 'comment-body.md'
+
+      - name: Post comment
+        uses: peter-evans/create-or-update-comment@v2
+        with:
+          body-file: 'profile-comment-body.md'
+          edit-mode: replace
+          issue-number: ${{ github.event.number }}
```
greens, line 43: I'd recommend updating this to just update one comment, rather than posting a new one every time.

the last two steps in this workflow: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/.github/workflows/delete-channel.yml#L35 now do this, if you want an example

.github/scripts/check_duplicates.py
```diff
@@ -0,0 +1,100 @@
...
+        f.write("%s == %s\n" % (duplicate[0], duplicate [1]))
+    f.write("\nPlease see [documentation] for more information on duplicate profiles.")
+    f.close()
+    f = open("profile-comment-body.md", "r")
+    print(f.read())
+else:
+    f.write("No duplicate profiles detected.")
+
+
+f.close()
```
greens, line 99: I'd recommend using the `with` syntax for `f` like you did for `csvfile`

ctowns:
I think this should be pretty close to done 🤞 . It should determine whether two profiles are the same based on the criteria outlined [here](https://smartthings.atlassian.net/wiki/spaces/~62de3b63b6b0b70770d7905d/pages/2602013567/Edge+Drivers+Repo+Automated+Testing#Definition-of-duplicate-profile%3A).

Right now I have a bunch of made up profiles to use for testing various test cases (embedded configs present, capabilities in different order, different categories, etc.). I will delete those before submitting, but for now they are present to show that the script is actually working!


Future ideas:
* Check for mismatch between top capability and recommended category
* Check that profile name matches file name - would this be useful?

greens (approved):
👍

ctowns:
Quick fixes before submitting: just removed all the test profiles and made some minor changes to comments. No functional changes.

PR #372, opened 2022-11-10
Add metadata to window-covering-profile

HunsupJung:
There are some clusters which is supported exclusive on Routines. users can make a wrong routine. so this PR is needed.


lelandblue:
Hello, we are trying not to add ui meta data when ever possible, we will review this pr.

HunsupJung:
@lelandblue 
Hello, I saw your mention in Slack channel. (https://samsung.slack.com/archives/C018FE9TDK7/p1668192592312229)
But there's no way to configure exclusion. So I'd like to know how to configure exclusion option by using Embedded Device Config.

PR #373, opened 2022-11-11
CHAD-9856: added Aeon Multisensor 5 fingerprint

sjl2178:
added Aeon Multisensor 5 fingerprint to zwave-sensor driver

https://smartthings.atlassian.net/browse/CHAD-9856

### DIFF
--- a/drivers/SmartThings/zwave-sensor/fingerprints.yml
+++ b/drivers/SmartThings/zwave-sensor/fingerprints.yml
@@ -456,6 +456,11 @@ zwaveManufacturer:
     manufacturerId: 0x0086
     productId: 0x0005
     deviceProfileName: motion-battery-temperature-illuminance-humidity
+  - id: "aeon/multisensor/gen5"
+    deviceLabel: Aeon Multipurpose Sensor
+    manufactureId: 0x0086
+    productId: 0x004A
+    deviceProfileName: motion-battery-temperature-illuminance-humidity
   - id: 021F/0003/0085
     deviceLabel: Dome Water Leak Sensor
     manufacturerId: 0x021F


greens:
Also needs some additions here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zwave-sensor/src/configurations.lua#L25

sjl2178:
@greens It looks like its already been added to the matching_matrix? `fingerprint mfr:"0086", prod:"0102", model:"004A"` is the device info which is already under `AOTEC_MULTISENSOR_GEN5`.

greens:
@sjl2178 Looks like the FP you've added already exists in the file

I think you need to add this one:
`mfr: 0086, model: 0005`

and, based on installed devices, this one:
`mfr: 0371, model: 0005`

greens:
I think 0x0371, based on our DTH codebase is another mfr code for Aeotec.

sjl2178:
@greens It seems like both those fps are already present?
`mfr: 0086, model: 0005` is included in `AEON_MULTISENSOR` in the configuration file
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/845724c8d7720e78b1c23d6518601db5725c09e9/drivers/SmartThings/zwave-sensor/fingerprints.yml#L454

`mfr: 0371, model: 0005` is in `AOTEC_TRISENSOR`
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/845724c8d7720e78b1c23d6518601db5725c09e9/drivers/SmartThings/zwave-sensor/fingerprints.yml#L73

greens:
you're absolutely correct, so I guess we don't need to do anything here and just need to confirm to Chris that he's testing the right device

PR #374, opened 2022-11-15
Sengled Windows/Door Sensor WWST

willyan2021:
1. add the init.lua file to the path ~\SmartThingsEdgeDrivers\drivers\SmartThings\zigbee-contact\src\sengled;
2. Modified ~\SmartThingsEdgeDrivers\drivers\SmartThings\zigbee-contact\fingerprints.yml file, a new device type E2D-G73 is added.

### DIFF
--- a/drivers/SmartThings/zigbee-contact/src/sengled/init.lua
+++ b/drivers/SmartThings/zigbee-contact/src/sengled/init.lua
@@ -1,0 +1,56 @@
+local clusters = require "st.zigbee.zcl.clusters"
+local battery_defaults = require "st.zigbee.defaults.battery_defaults"
+
+local IASZone = clusters.IASZone
+local PowerConfiguration = clusters.PowerConfiguration
+
+local FINGERPRINTS = {
+  { mfr = "sengled", model = "E2D-G73" }
+}
+
+local CONFIGURATIONS = {
+  {
+    cluster = IASZone.ID,
+    attribute = IASZone.attributes.ZoneStatus.ID,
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zigbee-contact/src/sengled/init.lua
```diff
@@ -0,0 +1,56 @@
...
+  for _, fingerprint in ipairs(FINGERPRINTS) do
+    if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
+      return true
+    end
+  end
+  return false
+end
+
+local function device_init(driver, device)
+  battery_defaults.build_linear_voltage_init(2.0, 3.0)(driver, device)
```
greens, line 40: If you use this method, the battery voltage attribute will be configured for you using the defaults. The only difference between your config and the defaults is the max reporting interval, which is 21600 in the defaults.

Similarly, your IAS Zone configuration is not that far removed from the defaults (a lower max interval of 300) and a `nil` reportable_change.

I think, if you wanted, all you'd have to have is the `linear_voltage_init`. You could use tests to confirm that the results of the configure lifecycle event are what you expect.

greens:
@willyan2002 Could you rebase your fork? There have been some CI updates since your fork was created that are integral in the certification process.

lelandblue:
Hello - I am closing this PR because it is covered in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/716/files

PR #375, opened 2022-11-15
Default capability update for Samjin multipurpose

dwmaxlee:
Default capability (threeAxis) update for Samjin multipurpose


greens:
The acceleration attribute should get read on add of the device. Is the device not responding?

dwmaxlee:
> 
The device is not reporting the acceleration attribute, even after binding success. Reonboarding the device can solve the issue. Anyway, what QA team said is that UI looks ugly and this patch is our best what we can do.

greens:
This solution may trigger automations when migration happens from DTH-backed devices. We're currently having discussions about alternative approaches.

greens:
@lelandblue correct. @dwmaxlee is that okay?

PR #380, opened 2022-11-17
Update supportedPlaybackCommands and supportedTrackControlCommands fo…

HunsupJung:
Update supportedPlaybackCommands and supportedTrackControlCommands for matter-media

### DIFF
--- a/drivers/SmartThings/matter-media/src/init.lua
+++ b/drivers/SmartThings/matter-media/src/init.lua
@@ -28,6 +28,13 @@ local configure_handler = function(self, device)
     capabilities.mediaPlayback.commands.play.NAME,
     capabilities.mediaPlayback.commands.pause.NAME,
     capabilities.mediaPlayback.commands.stop.NAME,
+    capabilities.mediaPlayback.commands.rewind.NAME,
+    capabilities.mediaPlayback.commands.fastForward.NAME,
+  }))
+
+  device:emit_event(capabilities.mediaTrackControl.supportedTrackControlCommands({
+    capabilities.mediaTrackControl.commands.previousTrack.NAME,
+    capabilities.mediaTrackControl.commands.nextTrack.NAME,
   }))
 
   device:emit_event(capabilities.keypadInput.supportedKeyCodes({
--- a/drivers/SmartThings/matter-media/src/init.lua
+++ b/drivers/SmartThings/matter-media/src/init.lua
@@ -24,13 +24,24 @@ local function device_init(driver, device)
 end
 
 local configure_handler = function(self, device)
-  device:emit_event(capabilities.mediaPlayback.supportedPlaybackCommands({
-    capabilities.mediaPlayback.commands.play.NAME,
-    capabilities.mediaPlayback.commands.pause.NAME,
-    capabilities.mediaPlayback.commands.stop.NAME,
-    capabilities.mediaPlayback.commands.rewind.NAME,
-    capabilities.mediaPlayback.commands.fastForward.NAME,
-  }))
+  local variable_speed_eps = device:get_endpoints(clusters.MediaPlayback.ID, {feature_bitmap = clusters.MediaPlayback.types.MediaPlaybackFeature.VARIABLE_SPEED})
+
+  if #variable_speed_eps > 0 then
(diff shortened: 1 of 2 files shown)


cjswedes (dismissed):
Thanks for finding this!

drivers/SmartThings/matter-media/src/init.lua
```diff
@@ -28,6 +28,13 @@ local configure_handler = function(self, device)
     capabilities.mediaPlayback.commands.play.NAME,
     capabilities.mediaPlayback.commands.pause.NAME,
     capabilities.mediaPlayback.commands.stop.NAME,
+    capabilities.mediaPlayback.commands.rewind.NAME,
```
tpmanley, line 31: Fastforward and rewind are only available if the device supports the `VS` or "Variable Speed" feature. At the time the driver was written the ZAP file did not define this feature so we couldn't use it in the driver, but thanks to @cjswedes it is now available https://github.com/project-chip/connectedhomeip/blob/76d834e409f5568434837fcad42d23dd944c1d6b/src/app/zap-templates/zcl/data-model/chip/media-playback-cluster.xml#L120. 

@HunsupJung can you check for the `VS` feature flag support and only include fastforward/rewind are supported if the feature flag is supported?

cjswedes: Here is an example of checking a device for feature support: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/5a78971034bf6f6e6e02a0436300b91b91311a5f/drivers/SmartThings/matter-thermostat/src/init.lua#L43

drivers/SmartThings/matter-media/src/init.lua
```diff
@@ -28,6 +28,13 @@ local configure_handler = function(self, device)
...
+    capabilities.mediaPlayback.commands.fastForward.NAME,
+  }))
+
+  device:emit_event(capabilities.mediaTrackControl.supportedTrackControlCommands({
+    capabilities.mediaTrackControl.commands.previousTrack.NAME,
```
tpmanley, line 36: Next and previous track are optional commands so we can't assume they are supported. I think we need to remove `mediaTrackControl` from the default profile and then have the read the `Accepted Command List` attribute and if those commands are supported then switch the profile to one that includes `mediaTrackControl`. 

@cjswedes @ctowns can you check my thinking?

cjswedes: Yeah, this is the way. I'll note that cluster library support for `AcceptedCommandList` will not be available until the next Lua library release which is preventing us from releasing such a change (similar to the situation that #370 is in).

ctowns:
This is ready for review. I will address the other comment in a separate PR since it is waiting for the next lua library release. However, this feature map functionality is ready to go.

drivers/SmartThings/matter-media/src/test/test_matter_media_video_player.lua
```diff
@@ -63,7 +91,19 @@ local function test_init()
   end
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
   test.mock_device.add_test_device(mock_device)
+
+  local subscribe_request = cluster_subscribe_list[1]:subscribe(mock_device_variable_speed)
+  for i, cluster in ipairs(cluster_subscribe_list) do
+    print(i)
```
tpmanley, line 97: Remove?

tpmanley (approved):
Approved but let's get the support for detecting whether the `Next` and `Previous` commands are supported in as soon as possible

drivers/SmartThings/matter-media/src/init.lua
cjswedes: Just to confirm, the plan is to support these by default despite the cluster commands backing them being optional, correct? I am fine with this, just want to make sure we are on the same page.

cjswedes (approved):

drivers/SmartThings/matter-media/src/init.lua
ctowns: Hey Carter, I am going to separate this work to optionally support the next/prevTrack commands in a separate PR since it can not go in yet

PR #383, opened 2022-11-17
Add checks for child existence on device added events

cjswedes:
This mitigates a situation where when a device is switched from one driver to another, there is an Added event that creates duplicate children. Note this situation should be prevented by the ST platform, and will be fixed in a future release.

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
@@ -37,8 +37,14 @@ local function do_refresh(self, device)
   device:send(ElectricalMeasurement.attributes.ActivePower:read(device))
 end
 
+local function find_child(parent, ep_id)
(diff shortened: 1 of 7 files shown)


drivers/SmartThings/zigbee-switch/src/zigbee-dual-metering-switch/init.lua
```diff
@@ -37,8 +37,13 @@ local function do_refresh(self, device)
...
+  return parent:get_child_by_parent_assigned_key(string.format("%02X", ep_id))
+end
+
 local function device_added(driver, device, event)
-  if device.network_type == st_device.NETWORK_TYPE_ZIGBEE then
+  if device.network_type == st_device.NETWORK_TYPE_ZIGBEE and
+     find_child(device, CHILD_ENDPOINT) == nil then
     local name = "AURORA Outlet 2"
     local metadata = {
       type = "EDGE_CHILD",
```
greens, line 49: nit: spacing

dsobierajsk (approved):

NikitaKuchta (approved):

PR #392, opened 2022-11-22
Set switchLevel range to [1,100] for zwave/zigbee dimmers

ctowns:
Some zwave/zigbee profiles have been added since https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/160, so this essentially extends that work to these new profiles. 

https://smartthings.atlassian.net/browse/CHAD-8941


ctowns:
Also note this only covers zwave/zigbee profiles. There are some matter profiles with the switchLevel capability as well. Should I get those also? I can put up another PR for those if needed.

greens (approved):
We can change matter if someone makes a fuss, but I'd say leave it for now. Enough on our plate as is.

greens (approved):

PR #394, opened 2022-11-22
Create new profile for fibaro radiator thermostat with [10, 30C] heat…

ctowns:
…ing range

Based on feedback here: https://smartthings.atlassian.net/browse/CHAD-9316

I created a new profile to set the heating range to [10, 30C] for the fibaro radiator thermostat when the extra sensor is not connected.


greens (approved):
This looks correct but I want to make sure it's necessary so I left a question in JIRA.

PR #397, opened 2022-11-23
Add checking for optional support of track control commands

ctowns:
The next/prev track control commands are optional in the MediaPlayback cluster. This will check for the commands using the AcceptedCommandList attribute in the MediaPlayback cluster and then switch to an appropriate profile if these commands are supported.

https://smartthings.atlassian.net/browse/CHAD-9933

### DIFF
--- a/drivers/SmartThings/matter-media/src/init.lua
+++ b/drivers/SmartThings/matter-media/src/init.lua
@@ -16,6 +16,7 @@ local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
 local utils = require "st.utils"
+local MediaPlaybackClusterAcceptedCommandList = clusters.MediaPlayback.attributes.AcceptedCommandList
 
 local VOLUME_STEP = 5
 
@@ -25,6 +26,9 @@ end
 
 local configure_handler = function(self, device)
   local variable_speed_eps = device:get_endpoints(clusters.MediaPlayback.ID, {feature_bitmap = clusters.MediaPlayback.types.MediaPlaybackFeature.VARIABLE_SPEED})
+  local media_playback_eps = device:get_endpoints(clusters.MediaPlayback.ID)
+
+  device:send(MediaPlaybackClusterAcceptedCommandList:read(device, media_playback_eps[1]))
 
   if #variable_speed_eps > 0 then
     device:emit_event(capabilities.mediaPlayback.supportedPlaybackCommands({
@@ -42,11 +46,7 @@ local configure_handler = function(self, device)
     }))
   end
 
-
-  device:emit_event(capabilities.mediaTrackControl.supportedTrackControlCommands({
-    capabilities.mediaTrackControl.commands.previousTrack.NAME,
-    capabilities.mediaTrackControl.commands.nextTrack.NAME,
-  }))
+  -- TODO: when to emit supported commands for MediaTrackControl?
 
   device:emit_event(capabilities.keypadInput.supportedKeyCodes({
     "UP",
@@ -70,6 +70,7 @@ local configure_handler = function(self, device)
     "NUMBER8",
     "NUMBER9",
   }))
+
 end
 
 local function on_off_attr_handler(driver, device, ib, response)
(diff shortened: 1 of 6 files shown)


ctowns:
I left some TODO comments in the code just related how to handle the supported capabilities event for the media track control now.

Also note that I changed some of the profile names so that I they would all follow a pattern that I could use for the profile switching logic (based on https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/370). Let me know if there is a preferred method for doing that or if changing profile names once in production is problematic.

drivers/SmartThings/matter-media/src/init.lua
```diff
@@ -112,6 +113,29 @@ local function media_playback_state_attr_handler(driver, device, ib, response)
...
+    if accepted_command_id.value == clusters.MediaPlayback.commands.Next.ID then
+      if device:supports_capability(capabilities.audioMute, device:endpoint_to_component(ib.endpoint_id)) then
+        new_profile = new_profile .. "-speaker"
+      end
+      new_profile = new_profile .. "-track-control"
+
+      device.log.info(string.format("Updating device profile to %s.", new_profile))
+      device:try_update_metadata({profile = new_profile})
+
+      -- TODO: Should this be moved here or no? Should this event be emitted somewhere else now?
```
cjswedes, line 129: Yeah putting this here is correct

ctowns:
With the completion of [HCS-3895](https://smartthings.atlassian.net/browse/HCS-3895), this seems to be working for me now. I added the event for the supportedTrackControlCommands to the infoChanged event because in my testing, that seemed to be directly tied to the profile update.

drivers/SmartThings/matter-media/src/init.lua
```diff
@@ -15,7 +15,7 @@
 local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
-local utils = require "st.utils"
+local MediaPlaybackClusterAcceptedCommandList = clusters.MediaPlayback.attributes.AcceptedCommandList
```
cjswedes, line 18: I think this should be somewhat safe to merge at this point since 0.46.x has been distributed all the way to external beta populations and they are the only ones that would really be running the driver. But this will cause errors on any hub running 0.45.x because the attribute does not exist yet.

lelandblue:
Closing this PR due to time lapsed. If we need to reopen we can.

nickolas-deboom:
I re-opened this PR and added a new commit with a few changes:
(1) Rebase with latest changes from main
(2) Check the RPC version to determine if `AcceptedCommandList` is supported to address the issue brought up previously in this PR
(3) Add profiles with "-no-track-control" rather than "track-control" in order to avoid renaming profiles or removing capabilities from existing profiles that are currently in use by devices in the field. Note that "media-video-speaker.yml" can safely be renamed because there are 0 devices in the field using this profile currently.
(4) A few updates and minor formatting cleanup in the test cases

drivers/SmartThings/matter-media/profiles/media-video-player-no-track-control.yml
```diff
@@ -0,0 +1,16 @@
...
+  - id: keypadInput
+    version: 1
+  - id: switch
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Television
```
cjswedes, line 16: nit: newline at eof

drivers/SmartThings/matter-media/src/init.lua
```diff
@@ -19,6 +19,12 @@
 local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
+local version = require "version"
+
+local MediaPlaybackClusterAcceptedCommandList
+if version.rpc >= 1 then
```
cjswedes, line 25: Why is RPC version checked instead of the api version?

drivers/SmartThings/matter-media/src/init.lua
```diff
@@ -138,6 +142,28 @@ local function media_playback_state_attr_handler(driver, device, ib, response)
...
+  end
+  local track_control = false
+  for _, accepted_command_id in ipairs (ib.data.elements) do
+    if accepted_command_id.value == clusters.MediaPlayback.commands.Next.ID then
+      track_control = true
+      break
+    end
+  end
+  local new_profile = "media-video-player"
+  if device:supports_capability(capabilities.audioMute, device:endpoint_to_component(ib.endpoint_id)) then
```
cjswedes, line 157: maybe add a comment for why you're checking for audioMute capability. Its not super clear why that is a differentiator.

PR #402, opened 2022-11-28
temporarily remove synthetic events

greens:


KKlimczukS (commented):
@greens 
Is scripting engine going to take care of emitting such synthetic events ?

dljsjr:
I guess my only concern is, how does this impact fresh join vs migration? Are we going to end up in a situation where we get a bunch of bug reports for mesh edge devices that have done fresh join because of the OneApp changes around capabilities and completely disabling the controls for capabilities that haven't seen any activity?

dljsjr (approved):

kianooshkarami (approved):

cjswedes (approved):

PR #407, opened 2022-11-30
Matter thermostat setpoint limits

cjswedes:
This implements setpoint limits for thermostats according to the spec. 

When the device is configured, the driver reads the min/max heat/cool setpoint limits and the min setpoint deadband limit attributes; these values are persisted on the device object. Note that we only consider the AbsX variety of  setpoint limit attributes (i.e. AbsMinHeatingSetpoint not MinHeatingSetpoint).

If a set point command has a value outside the min/max or deadband, we will log a warning with the relevant information and then emit a capability event with the current cached setpoint value. This will prevent a timeout error in the app, and will indicate to the user that the setpoint did not actually change.

- [x] Unit test fixes and tests for changes.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -15,6 +15,7 @@
 local capabilities = require "st.capabilities"
 local log = require "log"
 local clusters = require "st.matter.clusters"
+local im = require "st.matter.interaction_model"
 
 local MatterDriver = require "st.matter.driver"
 local utils = require "st.utils"
@@ -34,6 +35,14 @@ local THERMOSTAT_OPERATING_MODE_MAP = {
   [2]		= capabilities.thermostatOperatingState.thermostatOperatingState.fan_only,
 }
 
+local setpoint_limit_device_field = {
+  MIN_HEAT = "MIN_HEAT",
+  MAX_HEAT = "MAX_HEAT",
+  MIN_COOL = "MIN_COOL",
+  MAX_COOL = "MAX_COOL",
+  MIN_DEADBAND = "MIN_DEADBAND",
+}
+
 local function device_init(driver, device)
   device:subscribe()
 end
@@ -41,6 +50,7 @@ end
 local function do_configure(driver, device)
   local heat_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.HEATING})
   local cool_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.COOLING})
+  local auto_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.AUTOMODE})
   local thermo_eps = device:get_endpoints(clusters.Thermostat.ID)
   local fan_eps = device:get_endpoints(clusters.FanControl.ID)
   local humidity_eps = device:get_endpoints(clusters.RelativeHumidityMeasurement.ID)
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -214,10 +249,94 @@ local function set_setpoint(setpoint)
...
+
+    local is_auto_capable = #device:get_endpoints(
+      clusters.Thermostat.ID,
+      {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.AUTOMODE}
+    ) > 0
+    print("is_auto_capable", is_auto_capable)
+    --Check setpoint limits for the device
+    local setpoint_type = string.match(setpoint.NAME, "Heat") or "Cool"
+    -- local deadband = device:get_field(setpoint_limit_device_field.MIN_DEADBAND) or 0
+    local deadband = 1.3
```
samadDotDev, line 278: Should we assume a deadband of `2.5` when the field is failed to be read from the device (assuming the above line will be uncommented) since the attribute ([MinSetpointDeadBand](https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/Thermostat.adoc#7-attributes)) is `2.5` by default and that is the maximum deadband value a device could have?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -77,6 +87,27 @@ local function do_configure(driver, device)
...
+  if #heat_eps ~= 0 and device:get_field(setpoint_limit_device_field.MIN_HEAT) == nil then
+    setpoint_limit_read:merge(clusters.Thermostat.attributes.AbsMinHeatSetpointLimit:read())
+  end
+  if #heat_eps ~= 0 and device:get_field(setpoint_limit_device_field.MAX_HEAT) == nil then
+    setpoint_limit_read:merge(clusters.Thermostat.attributes.AbsMaxHeatSetpointLimit:read())
+  end
+  if #cool_eps ~= 0 and device:get_field(setpoint_limit_device_field.MIN_COOL) == nil then
+    setpoint_limit_read:merge(clusters.Thermostat.attributes.AbsMinCoolSetpointLimit:read())
+  end
+  if #cool_eps ~= 0 and device:get_field(setpoint_limit_device_field.MIN_COOL) == nil then
```
tpmanley, line 102:
```suggestion
  if #cool_eps ~= 0 and device:get_field(setpoint_limit_device_field.MAX_COOL) == nil then
```

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -214,10 +245,86 @@ local function set_setpoint(setpoint)
       value = f_to_c(value)
     end
 
+    -- Gather cached setpoint values when considering setpoint limits
+    -- Note cached values should always exists, but defaults are chosen just in case to prevent
```
tpmanley, line 249:
```suggestion
    -- Note: cached values should always exist, but defaults are chosen just in case to prevent
```

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -214,10 +245,86 @@ local function set_setpoint(setpoint)
...
+    -- nil operation errors, and deadband logic from triggering.
+    local cooling_setpoint = device.state_cache.main.thermostatCoolingSetpoint.coolingSetpoint or { value = 100 }
+    if cooling_setpoint and cooling_setpoint.unit == "F"then
```
tpmanley, line 252:
```suggestion
    if cooling_setpoint and cooling_setpoint.unit == "F" then
```

tpmanley:
What are the pros/cons of this approach vs just attempting to set the setpoint using the response to determine whether the setpoint was in range or not? If a `CONSTRAINT_ERROR` is returned then you could log an error and generate an event with the current value and if a SUCCESS is returned then you can generate an event with the new value.

cjswedes:
Pros of this approach:
* We dont send messages to a device that we know will fail, and we are following what the spec says a controller should do.
* This approach is more flexible to taking different actions based on what limit is violated and we make the limit info available to diagnose why the CONSTRAINT_ERROR would have happened

Cons of this approach:
* The implementation adds complexity to cap cmd handlers.
* Users do not actually get any info about why their set point didn't change (unless they look at driver logs). This seems unavoidable and drivers do not have a mechanism to show errors in the app.

> If a CONSTRAINT_ERROR is returned then you could log an error and generate an event with the current value and if a SUCCESS is returned then you can generate an event with the new value.

This is possible, but there isn't a clean way to do this due to the async nature of our interaction model, and the existing dispatcher functionality. The matter dispatcher does not dispatch write responses atm, and we would need to work around this by using the dispatcher's fallback handler. Even if we do have handling for write responses implemented, we would have to set state on the device object for set point values, and check the state based on the write response. This state management is brittle IMO, and it is the reason we have opted for using attribute reports from subscriptions rather than command responses or write responses to emit capability events.

As an aside, it seems feasible to implement a synchronous API by developing a heuristic to associate interaction responses with requests. Then cosock can wrap the interaction model and allow a response to be handled in the same context as sending the request. This would be a cool feature, that would be a game changer for interaction error handling in matter drivers.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -214,10 +245,86 @@ local function set_setpoint(setpoint)
...
+      cooling_setpoint = f_to_c(cooling_setpoint.value)
```
greens, line 253: unless full precision is desired, you can probably switch this over to the `util` function now, which gives two digits of precision

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -214,10 +245,86 @@ local function set_setpoint(setpoint)
...
+    else
+      cooling_setpoint = cooling_setpoint.value
+    end
+    local heating_setpoint = device.state_cache.main.thermostatHeatingSetpoint.heatingSetpoint or { value = 0 }
```
greens, line 257: we prefer to use `device:get_latest_state` here rather than accessing the cache directly

the second return value has the unit you're looking for:

```
--- @return any, any The first return value is the state.value present for the attribute, the second return is the state
---                  table (e.g. it would include both the value and unit keys if both are present)
```

drivers/SmartThings/matter-thermostat/src/init.lua
cjswedes: Looks like that change is in the 0.45 release, so I will make it.

greens (approved):

PR #411, opened 2022-12-01
Modify temp_attr_handler function in matter-switch driver

HunsupJung:
Signed-off-by: Hunsup Jung <hunsup.jung@samsung.com>

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -177,8 +177,8 @@ local function temp_attr_handler(driver, device, ib, response)
     local most_recent_temp = device:get_field(MOST_RECENT_TEMP)
     -- this is to avoid rounding errors from the round-trip conversion of Kelvin to mireds
     if most_recent_temp ~= nil and
-      most_recent_temp >= utils.round(CONVERSION_CONSTANT/(ib.data.value - 1)) and
-      most_recent_temp <= utils.round(CONVERSION_CONSTANT/(ib.data.value + 1)) then
+      most_recent_temp <= utils.round(CONVERSION_CONSTANT/(ib.data.value - 1)) and
+      most_recent_temp >= utils.round(CONVERSION_CONSTANT/(ib.data.value + 1)) then
         temp = most_recent_temp
     end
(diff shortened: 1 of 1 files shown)


HunsupJung:
To keep setting value by customer, inequality signs supposed to be reversed.

cjswedes (commented):
I'm not sure this is correct. It goes against what is done in zigbee defaults also, which means it would be wrong for many devices. Hopefully @greens can definitively confirm, because I am struggling to prove it when I think through it.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -177,8 +177,8 @@ local function temp_attr_handler(driver, device, ib, response)
     local most_recent_temp = device:get_field(MOST_RECENT_TEMP)
     -- this is to avoid rounding errors from the round-trip conversion of Kelvin to mireds
     if most_recent_temp ~= nil and
-      most_recent_temp >= utils.round(CONVERSION_CONSTANT/(ib.data.value - 1)) and
-      most_recent_temp <= utils.round(CONVERSION_CONSTANT/(ib.data.value + 1)) then
+      most_recent_temp <= utils.round(CONVERSION_CONSTANT/(ib.data.value - 1)) and
+      most_recent_temp >= utils.round(CONVERSION_CONSTANT/(ib.data.value + 1)) then
```
greens, line 181: Doing some replacement here to make things easier for myself.

Let's say our most recent temp is 2700K. That's 370.37 mireds.

So let's say we get a report from the device of 371 mireds:

```
2700K >= round(1M/(370)) and 2700K <= round(1M/372)
2700K >= round(2702.7) and 2700 <= round(2688.17)
2700K >= 2703 and 2700 <= 2688
```

that leads to the direct conversion from 371 value being used, rather than 2700K

so @cjswedes it looks like they're probably right here.

cjswedes (approved):
@greens we need to make a change in the zigbee defaults as well then.

cjswedes:
I created an internal ticket for the zigbee default changes (CHAD-10229)

greens (approved):

PR #412, opened 2022-12-02
Add the abnormal model information of SONOFF's dev

guozi7788:
Add the abnormal model information of SONOFF's wireless button, temperature and humidity sensor, and motion sensor to the corresponding fingerprint file.


guozi7788:
Hi,
Due to our error, some batches of sensors (Button, Temperature And Humidity Sensor, Motion Sensor) model information is abnormal, which means that one sensor has two Zigbee Model IDs.  Since these abnormal model information is not registered to the fingerprint file, it will cause these sensors and Samsung Hub to not pair smoothly. 
In order to facilitate users to use these sensors with abnormal information under the Samsung Hub, we have submitted this PR.

PR #420, opened 2022-12-07
Allow labeling of PRs to select library version

greens:


greens:
PSA: The labeling feature will only work for PRs _not_ originating from a forked repo. If you'd like to use it, your branch will have to be on the main repo, and not a fork.

drivers/SmartThings/zwave-switch/src/test/test_qubino_temperature_sensor_with_power.lua
```diff
@@ -15,7 +15,6 @@
 local test = require "integration_test"
 local t_utils = require "integration_test.utils"
 local capabilities = require "st.capabilities"
-local constants = require "st.zwave.constants"
```
dljsjr, line 18: Assuming this is totally harmless but just wanted to ask about it since it seems to be totally unrelated.

dljsjr (approved):
Sure is yaml alright.

drivers/SmartThings/zwave-switch/src/test/test_qubino_temperature_sensor_with_power.lua
greens: changed something in the `drivers/` path to trigger the tests running

also it's an unused import so why not

PR #425, opened 2022-12-08
update thirdreality smart plug

weihuan1111:
@greens  
Hi, we have update our device, thirdreality smart plug, now, our plug have power meter. But in smartthings APP, it will not show the power.
Does this pull request will make our plug show the power? If not, what i need to change?

### DIFF
--- a/drivers/SmartThings/zigbee-switch/fingerprints.yml
+++ b/drivers/SmartThings/zigbee-switch/fingerprints.yml
@@ -820,7 +820,7 @@ zigbeeManufacturer:
     deviceLabel: ThirdReality Plug
     manufacturer: Third Reality, Inc
     model: 3RSP019BZ
-    deviceProfileName: basic-switch
+    deviceProfileName: switch-power-smartplug
   - id: "DAWON_DNS/PM-S140-ZB"
     deviceLabel: Dawon Switch
     manufacturer: DAWON_DNS


greens (commented):
Newly-installed devices will have the `Power Meter` capability displayed.

But I wonder what you mean by "updated the device"? Is there a new firmware? Devices on the old firmware probably cannot report power consumption still, is that right?

weihuan1111:
> Newly-installed devices will have the `Power Meter` capability displayed.
> 
> But I wonder what you mean by "updated the device"? Is there a new firmware? Devices on the old firmware probably cannot report power consumption still, is that right?

yes

weihuan1111:
@greens please help to review the merge require, thanks

greens:
@weihuan1111 unless someone manages to buy and install a device still using the v1 firmware. Or they have a v1 installed and remove and re-add it. Then they'll see a capability their device does not support. 

I'd recommend building in some functionality to handle the version of the device. You can see how to query this information and use a sub-driver based on it in this driver and its sub-driver: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/tree/main/drivers/SmartThings/zigbee-window-treatment/src/axis

weihuan1111:
@greens I have tested locally, both v1 and v2 are work normally in this pull request.

greens:
@weihuan1111 I don't understand how that can be. If you install a v1 that does not support power meter, but it uses a profile that includes it, the user will see and empty power meter tile. Is that not correct?

weihuan1111:
@greens But now, our plugv1 and plugv2 use the same version, the difference is that v2 has the power function, everything else is the same. And we have tested locally, v1's on/off function, v2's on/off and power function are both used normally on smartthings.

PR #433, opened 2022-12-14
Replaced USN to MAC for Wemo Device Identifier

satyajitanand:


satyajitanand:
@greens : Can you pls help to review it?

drivers/SmartThings/wemo/src/discovery.lua
```diff
@@ -1,7 +1,7 @@
 local cosock = require "cosock"
 local socket = require "cosock.socket"
 local http = cosock.asyncify "socket.http"
-local ltn12 = require "ltn12"
+local ltn12 = require "socket.ltn12"
```
greens, line 4: leave as is in the current file, please

drivers/SmartThings/wemo/src/discovery.lua
```diff
@@ -103,7 +104,12 @@ local function find(deviceid, callback)
     if val then
       local headers = process_response(val)
       local ip, port = headers["location"]:match("http://([^,/]+):([^/]+)")
-      local id = headers["usn"]
+      -- local id = headers["usn"]  --instead of usn, we need to use mac as id (to handle migration scenario)
+      local meta = fetch_device_metadata(headers["location"])
+      if not meta then
+        meta = {}
+      end
+      local id = meta.mac
```
naman-dixit-srib, line 112: LGTM. One question, wouldn't something like the following be better?

```lua
      local meta = fetch_device_metadata(headers["location"])
      local id = nil
      if meta then
        id = meta.mac
      end
```

This way, we'll still be able to test later down the line whether `fetch_device_metadata` returned nil by `if not meta ...` instead of something like `if next(meta) == nil ...` (i.e., a simple null-check instead of checking if the table is empty).

naman-dixit-srib (approved):

greens (approved):

PR #437, opened 2022-12-16
Fix multi-component issues in duplicate profile check

ctowns:
Fix logic in duplicate profile checking script to account for profiles with multiple components. Previously, the script would indicate a duplicate if just one of the components in a profile matches another profile.

Also, refactored some of the logic for readability and brevity


ctowns:
I needed to fix some of the behavior for multi component devices. Also, I switched my approach from kinda "trying to prove two profiles are duplicates" to "trying to prove two profiles are NOT duplicates" and I think it simplified the logic a little bit and made it more readable.

Good news is that this already helped me spot a duplicate profile on one of my other PRs... maybe I was the fool I was trying to foolproof all along 🤔

greens (approved):

PR #441, opened 2022-12-22
[Aqara] illuminance sensor

sbkimaqara:


drivers/SmartThings/zigbee-illuminance-sensor/capabilities/detectionFrequency.yaml
```diff
@@ -0,0 +1,34 @@
+id: stse.detectionFrequency
```
greens, line 1: I'm guessing the reason you can't just use a preference here is the same as in other cases: the user has to press a button on the device in order for the interval set command to be received?

drivers/SmartThings/zigbee-illuminance-sensor/profiles/battery-illuminance-aqara.yml
```diff
@@ -0,0 +1,15 @@
+name: battery-illuminance-aqara
+components:
+  - id: main
+    capabilities:
+      - id: illuminanceMeasurement
+        version: 1
+      - id: stse.detectionFrequency
```
greens, line 7: Can you elaborate a bit on what detection frequency means in this context? Is it how long a light has to be detected before the device reports or is it the amount of time in between reports?

If it's the amount of time in between reports, then 2 seconds as a minimum is entirely too low and would result in significant load, both on the zigbee network as well as the hub.

drivers/SmartThings/zigbee-illuminance-sensor/capabilities/detectionFrequency.yaml
sbkimaqara: Yes. This is a similar to:
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/208#issuecomment-1250581969

drivers/SmartThings/zigbee-illuminance-sensor/profiles/battery-illuminance-aqara.yml
sbkimaqara: The detection frequency means the amount of time in between reports. 
If I change the minimal detection frequency, can you recommand the value? (For example, 5 seconds)

greens: Most of our devices have a minimum reporting interval of 5 minutes. 1 minute could work if you're really intent on such a low number, but measuring the value in seconds seems likely to both quickly drain the battery and degrade communications with other zigbee devices on the network.

sbkimaqara: Thank you for your advice. 😀 I suggest:

- Change the default detection interval from 5 seconds to 30 seconds.
- Keep the current detection interval range.

In the Aqara app, the default detection interval is 5 seconds and the detection interval ranges are from 2 to 59 seconds. So it would be good to provide similar options to SmartThings app users.
Can I continue like this?

greens: Let me follow up with you on that.

lelandblue:
UL Testing has completed as communicated by team member on 19 Jan 2023

drivers/SmartThings/zigbee-illuminance-sensor/profiles/battery-illuminance-aqara.yml
greens: I think those changes will be acceptable for now. Please make them.

sbkimaqara: I'll check. Thanks :)

sbkimaqara: We discussed your comments with our engineers and tested the behavior of the device with the modified code.

"detectionFrequency" is the sampling rate of the device to measure the illuminance. This is not the reporting interval I mentioned before. The default preference is 5 seconds. The device's illuminance report policy is as follows.

- 0 ~ 50 Lux : △Lux 6
- 51~ 500 Lux : △Lux 14
- 501~ 83K Lux : △Lux 10

ΔLux is the difference from the previously measured illuminance value, and it is reported only when the illuminance range is exceeded.

In the code, I kept the detectionFrequency by default value and added the configuration of the IlluminanceMeasurement.MeasuredValue attribute for reporting on a device. The illuminance will be reported every hour, configuration's minimum interval.

Ref, ST history tab screenshot.
https://drive.google.com/file/d/1TO5gyutKwWMqFdNCfyqsVFeQm5M74wnn/view?usp=sharing

Thanks :)

greens (approved):

jsconstantelos:
ST broke the Zigbee Illuminance Sensor Edge driver with this latest update:

Details and images here:

https://community.smartthings.com/t/st-broke-the-zigbee-illuminance-sensor-edge-driver/258299

Today at 4:48pm EST this driver was updated on my hub, and ever since then my hub is flooded with device events and RAW messages from the device. Everything was working great until that happened. Heads up to anyone using the Zigbee Illuminance Sensor edge driver for the Xiaomi Mijia Smart Light Sensor. Mine has been working flawlessly with this edge drive up until now.

Rebooting the hub does nothing, and rejoining the device made it worse. Now it reports nothing at all, not even catchall messages when the button is pressed. It’s broke broke.

I checked my other light sensor, and it’s completely dead and not reporting and I didn’t reset and rejoin that one. It just stopped working the second the updated driver showed up.

PR #443, opened 2022-12-23
Add retry logic(bind and group) for ikea button

inasail:
[CHAD-9927](https://smartthings.atlassian.net/browse/CHAD-9927)


inasail:
Under testing, whether this commit is really necessary.

PR #452, opened 2023-01-06
Bugfix/matter lock max codes calculation

cjswedes:
Previously calculating the max credential index available for pin codes
we used the formula `NumberOfPINUsersSupported *
NumberOfCredentialsSupportedPerUser`, but this calculation was misguided
and results in an index out of range for some devices. The max credential
index available on a lock is actually bounded by
`NumberOfPINUsersSupported`. This is not super clear in the spec, but
has been confirmed by some of the lock companies, and other CSA members.

This change also checks the RequirePINForRemoteOperation attribute when determining if a COTA credential is needed.

An improvement that will be done in a follow up PR is to select the COTA credential index based on the scanned credentials rather than always picking the highest credential slot.

### DIFF
--- a/drivers/SmartThings/matter-lock/src/init.lua
+++ b/drivers/SmartThings/matter-lock/src/init.lua
@@ -29,6 +29,9 @@ local lock_utils = require "lock_utils"
 
 local YALE_LOCK_FINGERPRINT = {{vendorId = 0x101D, productId = 0x1}}
 
+--- If a device needs a cota credential this function attempts to clear the credential
+--- at the max credential index and then set the device's COTA credential 2 seconds later.
+--- It will delay doing so until we are no longer scanning credentials.
 local function set_cota_credential(device)
   local eps = device:get_endpoints(DoorLock.ID)
   local cota_cred = device:get_field(lock_utils.COTA_CRED)
@@ -37,13 +40,10 @@ local function set_cota_credential(device)
     device:send(DoorLock.attributes.RequirePINforRemoteOperation:read(device, #eps > 0 and eps[1] or 1))
     device.thread:call_with_delay(2, function(t) set_cota_credential(device) end)
   elseif not cota_cred then
-    -- Defensive, but shouldn't happen
-    device.log.debug("Device should not require PIN for remote operation. Not setting COTA credential")
+    device.log.debug("Device does not require PIN for remote operation. Not setting COTA credential")
     return
   end
 
-  -- If we are scanning codes, we should wait to set the cota credential until scanning completes
-  -- to help avoid replacing an existing code, and ensure we have queried the max codes on the device.
   if device:get_latest_state(
     "main", capabilities.lockCodes.ID, capabilities.lockCodes.scanCodes.NAME
   ) == "Scanning" then
@@ -455,9 +455,6 @@ local function device_init(driver, device) device:subscribe() end
 
 local function device_added(driver, device)
   device:emit_event(capabilities.tamperAlert.tamper.clear())
-end
-
-local function do_configure(driver, device)
   local eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
   if #eps == 0 then
     device.log.debug("Device does not support lockCodes")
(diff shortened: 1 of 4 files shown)


tpmanley:
I tested this with a lock that does require COTA pins and after switching to this driver the COTA PIN wasn't being included in the lock/unlock commands. I'll PM you the logs.

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -30,13 +30,17 @@ local lock_utils = require "lock_utils"
...
-  -- Device requires pin for remote operation if it supports COTA and PIN features.
-  local eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.CREDENTIALSOTA | DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
-  if #eps == 0 then
-    device.log.debug("Device should not require PIN for remote operation, so not setting COTA credential")
+  local eps = device:get_endpoints(DoorLock.ID)
+  local cota_cred = device:get_field(lock_utils.COTA_CRED)
+  if cota_cred == nil then
+    -- Shouldn't happen but defensive to try to figure out if we need the cota cred and set it.
+    device:send(DoorLock.attributes.RequirePINforRemoteOperation:read(device, #eps > 0 and eps[1] or 1))
+    device.thread:call_with_delay(2, function(t) set_cota_credential(device) end)
```
greens, line 38: I think this results in `set_cota_credentials` being called twice, since the handler for the read response of the `RequirePINforRemoteOperation` also calls it

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -113,24 +116,26 @@ end
...
 end
 
-local function num_creds_per_user_handler(driver, device, ib, response)
-  device:set_field(lock_utils.CREDENTIALS_PER_USER, ib.data.value)
-  local num_pin_users = device:get_field(lock_utils.TOTAL_PIN_USERS)
-  if num_pin_users and num_pin_users > 0 then
-    device:emit_event(capabilities.lockCodes.maxCodes(ib.data.value * num_pin_users, {visibility = {displayed = false}}))
+local function require_remote_pin_handler(driver, device, ib, response)
+  if ib.data.value then
+    --Process after all other info blocks have been dispatched to ensure MaxPINCodeLength has been processed
```
greens, line 124: does this delay(0) *guarantee* this, or just make it likely?

cjswedes: Yes, the device responds to our read of MinPINCodeLength, MaxPINCodeLength, NumberOfPINUsersSupported, and RequirePINForRemoteOperation, all in one interaction response (I think this is guaranteed by the SDK). Our dispatcher processes an entire interaction response as one driver "event", so all of those attribute reports will be processed by handlers before the driver does anything else. Doing the delay(0) queues up a new event which will be processed after all the attribute reports, which is needed since the order of attribute report info blocks is not guaranteed in the response.

cjswedes:
> I tested this with a lock that does require COTA pins and after switching to this driver the COTA PIN wasn't being included in the lock/unlock commands.

I cannot reproduce this when I tried going from the old driver to this version. If it does end up being reproducible, we can have a conversation about if such a breaking change is okay at this point, given the number of devices on the platform is so low.

greens (approved):

drivers/SmartThings/matter-lock/src/init.lua
dljsjr, line 38: Where does the 2 second heuristic come from?

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -319,7 +324,7 @@ local function lock_user_change_event_handler(driver, device, ib, response)
         lock_utils.code_deleted(device, cs)
       end
       lock_utils.lock_codes_event(device, {})
-      set_cota_credential(device)
+      if device:get_field(lock_utils.COTA_CRED) ~= nil then set_cota_credential(device) end
```
dljsjr, line 327: Since the `false` case in `set_cota_credential` is commented as defensive/shouldn't happen, should that also be checked here before calling?

dljsjr: Referring to this: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/69fa22cb492385b01a838318d79ac66ea9cf9011/drivers/SmartThings/matter-lock/src/init.lua#L39

drivers/SmartThings/matter-lock/src/init.lua
cjswedes: This is just a delay to give time for the read to complete before this function is called again. The same sort of delay is used later in this function to wait for the clear credential command to happen before we try to set it.

drivers/SmartThings/matter-lock/src/init.lua
cjswedes: Looks like my defense was needed...I could have detected that we dont need to do anything here, but I instead added the defensive checks in `set_cota_credential`. I will remove the comment there, and add a docstring for the function.

dljsjr (approved):
LGTM

PR #453, opened 2023-01-06
BUG-6478 Handle incorrectly typed OnOff attribute reports

greens:


drivers/SmartThings/zigbee-switch/src/bad_on_off_data_type/init.lua
```diff
@@ -0,0 +1,42 @@
...
+-- limitations under the License.
+
+local zcl_clusters = require "st.zigbee.zcl.clusters"
+local data_types = require "st.zigbee.data_types"
+local capabilities = require "st.capabilities"
+
+-- There are reports of at least one device (SONOFF 01MINIZB) which occasionally
+-- reports this value as an Int8, rather than a Boolean, as per the spec
+local function incorrect_data_type_detected(opts, driver, device, zb_rx, ...)
+  return opts.dispatcher_class == "ZigbeeMessageDispatcher" and zb_rx.body.zcl_body.attr_record.data_type ~= data_types.Boolean.ID
```
varzac, line 22: This can_handle can be called on any zigbee message, so there is no guarantee that `attr_record` will exist, so this could cause an error.  And _if_ we are going to compare against 0 explicitly below, we might want to test against specifically INT and UINT types here instead of just _not_ boolean.  My only hesitation with this approach is that this will catch _all_ non boolean messages from any device even though we are trying to handle a specific device.  But it might still be worth considering to avoid re-fixing for other devices.

greens: It doesn't look like any other sub-drivers override our handling for this attribute.

drivers/SmartThings/zigbee-switch/src/bad_on_off_data_type/init.lua
```diff
@@ -0,0 +1,45 @@
...
+local data_types = require "st.zigbee.data_types"
+local capabilities = require "st.capabilities"
+
+-- There are reports of at least one device (SONOFF 01MINIZB) which occasionally
+-- reports this value as an Int8, rather than a Boolean, as per the spec
+local function incorrect_data_type_detected(opts, driver, device, zb_rx, ...)
+  return opts.dispatcher_class == "ZigbeeMessageDispatcher" and
+    zb_rx.body.zcl_body and
+    zb_rx.body.zcl_body.attr_record and
+    zb_rx.body.zcl_body.attr_record.data_type.value ~= data_types.Boolean.ID
```
tpmanley, line 25: Should this be further scoped to make sure it's only applying this logic to the OnOff attribute or does that happen automatically because it's only registering the one attribute handler?

greens: @varzac can chime in, but my thinking was that, ultimately, it ends up being the same number of comparisons either way?

varzac: I would lean toward making this more restrictive to just the onoff attribute just because I think the can handle will be a bit clearer in the actual intent.  But otherwise that restriction will come with the handler check.

greens:
At this point, with this looking how it does now and the sheer number of comparisons, I'm thinking I should just go back to doing a model-based sub-driver here and then once it gets fixed in the defaults I can get rid of it.

tpmanley (approved):
It's be good to add a unit test for this too

PR #469, opened 2023-01-11
Add initial value event for colorTemp in zigbee bulbs

ctowns:
OneApp can have issues when an initial value is not reported for an attribute. Add an initial value for colorTemperature to prevent this issue.


ctowns:
@greens the zll-dimmer-bulb driver also handles some devices that have color temperature (as well as some that do not). I'm assuming I would not want to hardcode a color temp read to the add event in the zll-dimmer-bulb driver since not all device support color temp. However, there is a device:refresh() in the added event in that driver - is that enough to get a read event for the color temp (on devices that actually support it)?

greens (commented):
This read should already be happening in `configure`, which is called on add and migration. I think we need to investigate a bit more here...

PR #470, opened 2023-01-12
BUG-6683 Simplify CT100 driver

greens:
A lot of behavior was ported over from the DTH which, I think, in hindsight, could have been foregone. I've attempted to pare this driver back to the z-wave standard except for some polling that seems as though it might have been necessary. This also includes a new profile with a range defined to match the set point bounds in the DTH.

### DIFF
--- a/drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
+++ b/drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
@@ -13,43 +13,66 @@
 -- limitations under the License.
 
 local capabilities = require "st.capabilities"
-local utils = require "st.utils"
---- @type st.zwave.CommandClass
 local cc = require "st.zwave.CommandClass"
 --- @type st.zwave.CommandClass.SensorMultilevel
-local SensorMultilevel = (require "st.zwave.CommandClass.SensorMultilevel")({ version = 5 })
---- @type st.zwave.CommandClass.ThermostatFanMode
-local ThermostatFanMode = (require "st.zwave.CommandClass.ThermostatFanMode")({ version = 3 })
---- @type st.zwave.CommandClass.ThermostatFanState
-local ThermostatFanState = (require "st.zwave.CommandClass.ThermostatFanState")({ version = 2 })
+local SensorMultilevel = (require "st.zwave.CommandClass.SensorMultilevel")({ version = 2 })
 --- @type st.zwave.CommandClass.ThermostatMode
 local ThermostatMode = (require "st.zwave.CommandClass.ThermostatMode")({ version = 2 })
 --- @type st.zwave.CommandClass.ThermostatOperatingState
 local ThermostatOperatingState = (require "st.zwave.CommandClass.ThermostatOperatingState")({ version = 1 })
 --- @type st.zwave.CommandClass.ThermostatSetpoint
 local ThermostatSetpoint = (require "st.zwave.CommandClass.ThermostatSetpoint")({ version = 1 })
+--- @type st.zwave.CommandClass.ThermostatFanMode
+local ThermostatFanMode = (require "st.zwave.CommandClass.ThermostatFanMode")({ version = 1 })
+--- @type st.zwave.CommandClass.Battery
+local Battery = (require "st.zwave.CommandClass.Battery")({ version = 1 })
+local constants = require "st.zwave.constants"
+local utils = require "st.utils"
 
 local CT100_THERMOSTAT_FINGERPRINTS = {
   { manufacturerId = 0x0098, productType = 0x6401, productId = 0x0107 }, -- 2Gig CT100 Programmable Thermostat
   { manufacturerId = 0x0098, productType = 0x6501, productId = 0x000C }, -- Iris Thermostat
 }
 
--- Constants
-local TEMPERATURE_SCALE = "temperature_scale"
-local PRECISION = "precision"
-local CURRENT_HEATING_SETPOINT = "current_heating_setpoint"
-local CURRENT_COOLING_SETPOINT = "currnet_cooling_setpoint"
-local MODE = "mode"
-local TEMPERATURE = "temperature"
-local HEATING_SETPOINT_IS_LIMITED = "heating_setpoint_is_limited"
-local COOLING_SETPOINT_IS_LIMITED = "cooling_setpoint_is_limited"
+-- This old device uses separate endpoints to get values of temp and humidity
+-- DTH actually uses the old mutliInstance encap, but multichannel should be back-compat
+local TEMPERATURE_ENDPOINT = 1
+local HUMIDITY_ENDPOINT = 2
 
-local MIN_HEATING_SETPOINT = 35.0
-local MAX_HEATING_SETPOINT = 92.0
-local MIN_COOLING_SETPOINT = 38.0
-local MAX_COOLING_SETPOINT = 95.0
+--TODO: Update this once we've decided how to handle setpoint commands
+local function convert_to_device_temp(command_temp, device_scale)
+  -- under 40, assume celsius
+  if (command_temp <= 35 and device_scale == ThermostatSetpoint.scale.FAHRENHEIT) then
+    command_temp = utils.c_to_f(command_temp)
+  elseif (command_temp > 35 and (device_scale == ThermostatSetpoint.scale.CELSIUS or device_scale == nil)) then
+    command_temp = utils.f_to_c(command_temp)
+  end
+  return command_temp
+end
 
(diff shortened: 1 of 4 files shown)


greens:
very aware this is not passing tests right now, but hoping to get some folks who actually have this device to try it out a little bit

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
```diff
@@ -13,212 +13,73 @@
...
+local TEMPERATURE_ENDPOINT = 1
+local HUMIDITY_ENDPOINT = 2
 
-local function can_handle_ct100_thermostat(opts, driver, device, cmd, ...)
-  for _, fingerprint in ipairs(CT100_THERMOSTAT_FINGERPRINTS) do
-    if device:id_match( fingerprint.manufacturerId, fingerprint.productType, fingerprint.productId) then
-      return true
-    end
+--TODO: Update this once we've decided how to handle setpoint commands
+local function convert_to_device_temp(command_temp, device_scale)
```
ctowns, line 43: I know this is a workaround, but would it be better to set the cutoff to 35 in this case since that is the highest celcius value we would expect for this subdriver based on the profile?

greens: sure. that makes sense

greens: here's hoping nobody ever sets their thermostat to 35 C or F, or 40 C or F

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
```diff
@@ -13,212 +13,73 @@
...
-local utils = require "st.utils"
---- @type st.zwave.CommandClass
 local cc = require "st.zwave.CommandClass"
 --- @type st.zwave.CommandClass.SensorMultilevel
-local SensorMultilevel = (require "st.zwave.CommandClass.SensorMultilevel")({ version = 5 })
---- @type st.zwave.CommandClass.ThermostatFanMode
-local ThermostatFanMode = (require "st.zwave.CommandClass.ThermostatFanMode")({ version = 3 })
---- @type st.zwave.CommandClass.ThermostatFanState
-local ThermostatFanState = (require "st.zwave.CommandClass.ThermostatFanState")({ version = 2 })
+local SensorMultilevel = (require "st.zwave.CommandClass.SensorMultilevel")({ version = 2 })
```
cjswedes, line 18: image
Looking at the spec, one of the added sensor types in v5 is humidity, which is likely why that version was used. Not sure if this will break anything, likely the device will handle everything just fine since msg formats didn't change between versions, but we should probably keep v5.

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
```diff
@@ -237,19 +98,17 @@ local function thermostat_mode_report_handler(self, device, cmd)
...
-  local cooling_setpoint = device:get_field(CURRENT_COOLING_SETPOINT)
-  local current_temperature = device:get_field(TEMPERATURE)
+  local heating_setpoint = device:get_latest_state("main", capabilities.thermostatHeatingSetpoint.ID, capabilities.thermostatHeatingSetpoint.heatingSetpoint.NAME, 0)
+  local cooling_setpoint = device:get_latest_state("main", capabilities.thermostatCoolingSetpoint.ID, capabilities.thermostatCoolingSetpoint.coolingSetpoint.NAME, 0)
+  local current_temperature = device:get_latest_state("main", capabilities.temperatureMeasurement.ID, capabilities.temperatureMeasurement.temperature.NAME, 0)
 
   device:send(ThermostatOperatingState:Get({}))
   if mode == ThermostatMode.mode.COOL or
-    ((mode == ThermostatMode.mode.COOL or mode == ThermostatMode.mode.OFF) and (current_temperature > (heating_setpoint + cooling_setpoint) / 2)) then
+    ((mode == ThermostatMode.mode.AUTO or mode == ThermostatMode.mode.OFF) and (current_temperature > (heating_setpoint + cooling_setpoint) / 2)) then
```
cjswedes, line 111: Why do we need to read the setpoints in the `((mode == ThermostatMode.mode.AUTO or mode == ThermostatMode.mode.OFF) and (current_temperature > (heating_setpoint + cooling_setpoint) / 2))` case? I dont understand the `current_temperature > (heating_setpoint + cooling_setpoint) / 2)` at all.

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
```diff
@@ -258,40 +117,40 @@ local function thermostat_mode_report_handler(self, device, cmd)
...
 end
 
-local function temperature_report_handler(self, device, cmd)
-  if (cmd.args.sensor_type == SensorMultilevel.sensor_type.TEMPERATURE) then
-    local scale = 'C'
-    if (cmd.args.scale == SensorMultilevel.scale.temperature.FAHRENHEIT) then scale = 'F' end
-    device:emit_event_for_endpoint(cmd.src_channel, capabilities.temperatureMeasurement.temperature({value = cmd.args.sensor_value, unit = scale}))
-    device:set_field(TEMPERATURE, cmd.args.sensor_value, {persist = true})
-  end
+local function do_refresh(self, device, cmd)
```
cjswedes, line 120: nit: cmd is unused

drivers/SmartThings/zwave-thermostat/src/test/test_ct100_thermostat.lua
```diff
@@ -343,40 +301,42 @@ test.register_coroutine_test(
...
-          precision = 0,
-          value = 25
-        })
+          mock_device,
+          ThermostatSetpoint:Set({
+            setpoint_type = ThermostatSetpoint.setpoint_type.COOLING_1,
+            scale = 0,
+            precision = 0,
+            value = 25
+          })
```
cjswedes, line 316: nit: unnecessary indent

cjswedes (commented):
LGTM, except for the SensorMultiLevel version, but like you mention it would be great to get someone to actually test it with a real device, since we have been bitten in the past by these devices

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
greens: @L153 it's set as the capability command handler

greens: and called on L132

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
greens: me neither, but the DTH does it

```
	if (mode == "cool") {
		thermostatSetpoint = coolingSetpoint
	} else if (mode == "auto" || mode == "off") {
		// Set thermostatSetpoint to the setpoint closest to the current temperature
		def currentTemperature = getTempInLocalScale("temperature")
		if (currentTemperature > (heatingSetpoint + coolingSetpoint)/2) {
			thermostatSetpoint = coolingSetpoint
		}
	}
```

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
greens: the real frustrating thing is that, if they supported SMv5 why they put the sensors on two different endpoints instead of just using the sensor_type field in the get command

cjswedes: Ahh okay, then Ill leave it to your discretion then, but I think this is one of those things where we would need to test with a real device.

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
cjswedes: I meant the argument `cmd` is not used in the function, the only usage also passes in nil...

greens: ooooh

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
greens: I'm asking community members to test, since I don't think we have access to one of these anymore.

greens:
Merging this as-is. It seems there may still be some issues, but it at least seems like an improvement. Hoping to get more feedback as this works its way out.

PR #471, opened 2023-01-12
Matter lock cota edge cases

cjswedes:
When a lock is added to the driver, and we receive a RequirePINforRemoteOperation report set to true, we kick off the process of setting a COTA credential. First we SetCredential at index 1, if it fails because that slot is occupied we retry with the next available index which is reported in the SetCredentialResponse. Edge cases that are detected:
* The generated credential is a duplicate of a PIN that already exists
* We cannot set a credential because there is no space left on the device
* The user deletes the COTA credential

Work to be done:
- [x] Unit tests for set COTA credential cases
- [x] Add profile switching for situation where the COTA cred cannot be set
- [x] ~~Attempt to clear the COTA credential on device remove event.~~ Sending interactions from the removed handler does not work because the device is already being removed from the fabric by the time the driver gets the removed event. The impact of this is that eventually, after removing and re-adding the lock many times to our platform, we will no longer be able to set a COTA credential. This is an unlikely scenario, and the user would be able to clear the lock codes with our app if it is joined.

### DIFF
--- a/drivers/SmartThings/matter-lock/profiles/nonfunctional-lock.yml
+++ b/drivers/SmartThings/matter-lock/profiles/nonfunctional-lock.yml
@@ -1,0 +1,14 @@
+name: nonfunctional-lock
+components:
+- id: main
+  capabilities:
+  - id: lockCodes
+    version: 1
+  - id: tamperAlert
+    version: 1
+  - id: battery
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: SmartLock
--- a/drivers/SmartThings/matter-lock/src/init.lua
+++ b/drivers/SmartThings/matter-lock/src/init.lua
@@ -14,74 +14,62 @@
 
 local MatterDriver = require "st.matter.driver"
 local interaction_model = require "st.matter.interaction_model"
-local Status = interaction_model.InteractionResponse.Status
 local clusters = require "st.matter.clusters"
 
 local DoorLock = clusters.DoorLock
 local PowerSource = clusters.PowerSource
 
 local capabilities = require "st.capabilities"
-local log = require "log"
-local json = require "st.json"
 local im = require "st.matter.interaction_model"
-local utils = require "st.utils"
 local lock_utils = require "lock_utils"
 
-local YALE_LOCK_FINGERPRINT = {{vendorId = 0x101D, productId = 0x1}}
+local INITIAL_COTA_INDEX = 1
 
---- If a device needs a cota credential this function attempts to clear the credential
---- at the max credential index and then set the device's COTA credential 2 seconds later.
---- It will delay doing so until we are no longer scanning credentials.
-local function set_cota_credential(device)
+--- If a device needs a cota credential this function attempts to set the credential
+--- at the index provided. The set_credential_response_handler handles all failures
+--- and retries with the appropriate index when necessary.
+local function set_cota_credential(device, credential_index)
   local eps = device:get_endpoints(DoorLock.ID)
   local cota_cred = device:get_field(lock_utils.COTA_CRED)
   if cota_cred == nil then
     -- Shouldn't happen but defensive to try to figure out if we need the cota cred and set it.
     device:send(DoorLock.attributes.RequirePINforRemoteOperation:read(device, #eps > 0 and eps[1] or 1))
-    device.thread:call_with_delay(2, function(t) set_cota_credential(device) end)
+    device.thread:call_with_delay(2, function(t) set_cota_credential(device, credential_index) end)
   elseif not cota_cred then
     device.log.debug("Device does not require PIN for remote operation. Not setting COTA credential")
     return
   end
 
-  if device:get_latest_state(
-    "main", capabilities.lockCodes.ID, capabilities.lockCodes.scanCodes.NAME
-  ) == "Scanning" then
+  if device:get_field(lock_utils.SET_CREDENTIAL) ~= nil then
+    device.log.debug("delaying setting COTA credential since a credential is currently being set")
     device.thread:call_with_delay(2, function(t)
-      set_cota_credential(device)
+      set_cota_credential(device, credential_index)
     end)
     return
   end
 
-  --try to use last code slot in hopes that it wont overwrite existing codes on the device
-  local credential_index = device:get_field(lock_utils.TOTAL_PIN_USERS) or
-    device:get_latest_state("main", capabilities.lockCodes.ID, capabilities.lockCodes.maxCodes.NAME) or 1
+  device:set_field(lock_utils.COTA_CRED_INDEX, credential_index, {persist = true})
   local credential = {credential_type = DoorLock.types.DlCredentialType.PIN, credential_index = credential_index}
-
-  -- Clear the credential to make sure that we have an open slot for the cota credential
-  device.thread:call_with_delay(0, function(t)
-    --Note we dont set lock_utils.DELETEING_CODE field to avoid re-setting cota credential this time
-    device:send(DoorLock.server.commands.ClearCredential(
-      device,
-      #eps > 0 and eps[1] or 1,
-      credential
-    ))
-  end)
-
   -- Set the credential to a code
-  device.thread:call_with_delay(2, function(t)
-    device:set_field(lock_utils.SET_CREDENTIAL, credential_index)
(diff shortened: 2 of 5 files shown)


drivers/SmartThings/matter-lock/profiles/nonfunctional-lock.yml
```diff
@@ -0,0 +1,14 @@
+name: nonfunctional-lock
```
greens, line 1: this could just be "lock without codes" like we have for z-wave and zigbee, yeah?

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -29,59 +29,51 @@ local lock_utils = require "lock_utils"
...
+    DoorLock.types.DlDataOperationType.ADD,
+    credential,
+    device:get_field(lock_utils.COTA_CRED),
+    nil, -- nil user_index creates a new user
+    DoorLock.types.DlUserStatus.OCCUPIED_ENABLED,
+    DoorLock.types.DlUserType.REMOTE_ONLY_USER
+  ))
+end
+
+local function generate_cota_cred_for_device(device)
```
greens, line 72: should this be a no-op if there's already a generated credential?

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -186,10 +177,31 @@ local function set_credential_response_handler(driver, device, ib, response)
...
+  elseif device:get_field(lock_utils.COTA_CRED) and credential_index == device:get_field(lock_utils.COTA_CRED_INDEX) then
+    -- Handle failure to set a COTA credential
+    if status == DoorLock.types.DlStatus.OCCUPIED and elements.next_credential_index.value ~= nil then
+      --This credential index is unavailable, but there is another available
+      set_cota_credential(device, elements.next_credential_index.value)
+    elseif status == DoorLock.types.DlStatus.OCCUPIED and elements.next_credential_index.value == nil and credential_index == 1 then
+      --There are no credential indices available on the device
+      device.log.error("Device requires COTA credential, but has no credential indexes available!")
+      device.log.error("Lock and Unlock commands will no longer work!!")
+      device:try_update_metadata({profile = "nonfunctional-lock", provisioning_state = "NONFUNCTIONAL"})
```
greens, line 189: what does this provisioning state do, functionally?

greens: I didn't think about this during the discussion, but, if we kept the profile that includes the lock codes capability, presumably the user would be able to delete codes through LCM. (after we've implemented the scanning for codes on add)

greens: or, if we want to keep this profile switching, we should be able to switch back to allowing operation if we detect that a code has been deleted

drivers/SmartThings/matter-lock/profiles/nonfunctional-lock.yml
cjswedes: Well its actually "lock without lock". It is missing the Lock capability since the lock commands wont work. It has lock codes because as you mention elsewhere we want to allow the user to delete/manage codes.

drivers/SmartThings/matter-lock/src/init.lua
cjswedes: No. There is a case, although unlikely, where our generated credential is the same as a PIN that already exists on the lock. In this case we fail to set the credential, and need to regenerate, or else we would continue to fail to set the credential.

drivers/SmartThings/matter-lock/src/init.lua
cjswedes: > what does this provisioning state do, functionally?

I dont think provisioning state really does anything atm, but I thought it made sense to use it here: the device is non-functional. Probably not worth the trouble though.

cjswedes: > if we want to keep this profile switching, we should be able to switch back to allowing operation if we detect that a code has been deleted

I think I have overlooked the case where the user fixes the problem. Right now they would have to re-onboard to get a COTA cred set, but I think a periodic attempt to set the COTA cred or detecting a deleted code would be ways to fix this. Ill add that in today.

- [ ] If a user deletes codes, the COTA cred is reset and the profile is switched back to the normal profile

cjswedes: I have added functionality to detect when a code is deleted, and the device is nonfunctional; at which point it tries to set a COTA cred and make the device functional. Unit tests added also.

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -123,12 +115,11 @@ local function require_remote_pin_handler(driver, device, ib, response)
...
-      local len = device:get_latest_state("main", capabilities.lockCodes.ID, capabilities.lockCodes.maxCodeLength.NAME) or 4
-      local cred_data = math.floor(math.random() * (10 ^ len))
-      cred_data = string.format("%0" .. tostring(len) .. "d", cred_data)
-      device:set_field(lock_utils.COTA_CRED, cred_data, {persist = true})
+      generate_cota_cred_for_device(device)
+      -- delay needed to allow test to override the random credential data
       device.thread:call_with_delay(0, function(t)
-        set_cota_credential(device)
+        -- Attempt to set cota credential at the lowest index
+        set_cota_credential(device, 1)
```
tpmanley, line 122: Might be a good idea to make a variable called INITIAL_COTA_INDEX or something in order to avoid the magic numbers in the code

drivers/SmartThings/matter-lock/src/init.lua
dljsjr: > There is a case, although unlikely, where our generated credential is the same as a PIN that already exists on the lock

What are the security implications here? Checking to see if a code is already in use and then exhibiting specific behaviors on collision is an actionable form of brute force password sniffing in the general case. I don't know enough about Matter or what else is going on on the system to know if this is actionable in this specific case but I figure it was worth asking.

cjswedes: This came up in my conversation with Yale. There is a limit to the number of wrong codes that can be entered manually (WrongCodeEntryLimit) which does apply to PINs in remote commands. This mitigates one command that can be used to brute force attack, but there is still the SetCredential to discover duplicates in a brute force way... 

image

cjswedes: The device would need to be commissioned to a malicious fabric though, which requires user interaction to commission the device.

dljsjr (approved):

drivers/SmartThings/matter-lock/profiles/nonfunctional-lock.yml
greens: reading is hard

tpmanley (approved):

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -142,11 +129,12 @@ local function clear_credential_response_handler(driver, device, ib, response)
...
     return
   end
-  local max_codes = device:get_latest_state("main", capabilities.lockCodes.ID, capabilities.lockCodes.maxCodes.NAME)
   if ib.status == im.InteractionResponse.Status.SUCCESS then
     lock_utils.lock_codes_event(device, lock_utils.code_deleted(device, tostring(deleted_code_slot)))
-    if deleted_code_slot == max_codes then --make sure cota credential exists if the user deletes it
-      set_cota_credential(device)
+    --make sure cota credential exists if the user deletes it or if space was created for the COTA cred
+    if deleted_code_slot == device:get_field(lock_utils.COTA_CRED_INDEX) or
+      device:get_field(lock_utils.NONFUNCTIONAL) then
```
greens, line 136: you could also just check `~device.supports_capability(lock)` instead of managing a flag

cjswedes: ahh I was rushing a bit, this is much better

cjswedes: actually due to unit testing difficulties, I would prefer to leave as is: `expect_metadata_update` and `try_update_metadata` on the mock device don't actually do the updates to the mock device, and applying the update myself mid test is proving difficult/impossible without hacky accessing of the test environment.

PR #472, opened 2023-01-12
fix: Addresses several issues with the Hue driver discovered in beta testing

dljsjr:
This PR addresses multiple reported issues from QA, internal testing, and beta testing.

- The first commit addresses coroutine safety on the REST API; If multiple capability commands arrived in quick enough succession, it could cause a crash because multiple callsites would be waiting for a response on the TCP socket at the same time, leading to `select`ing from multiple callsites which cosock will properly and defensively treat as an error condition. This is done by adopting what is basically an actor/handle pattern for the Hue API instances.
- The second commit fixes an issue where certain network topologies or user configurations can mask the bridge's MAC Address at discovery time when processing mDNS responses, so we now always query the bridge directly for this information.
- The third commit fixes a regression from the DTH where dimming commands sent to an "off" bulb would not turn the bulb back on
- The fourth commit adds ranges for color temperature to the profile
- The fifth commit adds ranges to the dimmer profiles and fixes the way switch levels are rounded/clamped to be in line with how the DTH rounded/clamped at the bottom end of the dimming range.

Bugs fixed:
- [x] [CHAD-10182] ]Make the Hue REST API Coroutine Safe
- [x] [CHAD-10160] mDNS discovery assumes that the hostname will always be the bridgeId; we should query the bridge for this instead
- [x] [CHAD-10104] Adjusting the dimmer on a bulb that is "off" does not turn the bulb on 
- [x] [CHAD-10100] Establish ranges on profiles for Color Temperature values 
- [x] [CHAD-10099] Make dimmer capability behave like it did with the DTH

### DIFF
--- a/drivers/SmartThings/philips-hue/profiles/white-ambiance.yml
+++ b/drivers/SmartThings/philips-hue/profiles/white-ambiance.yml
@@ -6,6 +6,10 @@ components:
     version: 1
   - id: switchLevel
     version: 1
+    config:
+      values:
+        - key: "level.value"
+          range: [ 1, 100 ]
   - id: colorTemperature
     version: 1
     config:
--- a/drivers/SmartThings/philips-hue/profiles/white-ambiance.yml
+++ b/drivers/SmartThings/philips-hue/profiles/white-ambiance.yml
@@ -8,6 +8,10 @@ components:
     version: 1
   - id: colorTemperature
     version: 1
+    config:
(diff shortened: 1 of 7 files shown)


cjswedes (approved):
Nice use of the cosock channel to queue requests like this. It is a useful example to have in the repo, and will probably be useful for other lan drivers.

PR #473, opened 2023-01-12
fix: Perform explicit MAC lookup at discovery

dljsjr:
Some bridges won't use their MAC-based Bridge ID as the hostname when responding
to mDNS queries. We were relying on that functionality in our discovery loop so
we fix that by querying the bridge config info, which is an endpoint that does
not require auth and that has the MAC address as an explicitly reported field.

Fixes: CHAD-10160

### DIFF
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -69,7 +69,8 @@ local function emit_light_status_events(light_device, light)
     end
 
     if light.dimming then
-      light_device:emit_event(capabilities.switchLevel.level(math.floor(light.dimming.brightness)))
+      local adjusted_level = st_utils.clamp_value(light.dimming.brightness, 1, 100)
+      light_device:emit_event(capabilities.switchLevel.level(st_utils.round(adjusted_level)))
     end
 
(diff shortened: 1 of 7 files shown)


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -191,7 +195,8 @@ bridge_added = function(driver, device)
   local device_bridge_id = device.device_network_id
 
   if not driver.joined_bridges[device_bridge_id] then
-    Discovery.search_for_bridges(driver, function(driver, bridge_ip, bridge_id)
+    local known_macs = {}
+    Discovery.search_for_bridges(driver, known_macs, function(driver, bridge_ip, bridge_id)
```
greens, line 199: Couldn't you just pass in an explicit `{}`? I must be missing something.

greens (approved):

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: Absolutely could, wanted to leave this here to hang my hat off of if I find out later I could actually have done something to optimize this. I am *fatigued*.

PR #474, opened 2023-01-12
fix: Set bulbs on if off when handling switchLevel

dljsjr:
The established behavior for a Hue bulb on our platform is to turn on if it's off
and it receives a dimming event. This fixes that regression.

Fixes: CHAD-10104
Depends-On: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/473

### DIFF
--- a/drivers/SmartThings/philips-hue/src/handlers.lua
+++ b/drivers/SmartThings/philips-hue/src/handlers.lua
@@ -60,6 +60,22 @@ local function do_switch_level_action(driver, device, args)
     return
   end
 
+  local is_off = device:get_latest_state(
+    "main", capabilities.switch.ID, capabilities.switch.switch.NAME) == "off"
+
+  if is_off then
+    local resp, err = hue_api:set_light_on_state(light_id, true)
+    if not resp or (resp.errors and #resp.errors == 0) then
+      if err ~= nil then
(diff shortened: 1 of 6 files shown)


drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -60,6 +60,22 @@ local function do_switch_level_action(driver, device, args)
...
   end
 
+  local is_off = device:get_latest_state(
+    "main", capabilities.switch.ID, capabilities.switch.switch.NAME) == "off"
+
+  if is_off then
+    local resp, err = hue_api:set_light_on_state(light_id, true)
+    if not resp or (resp.errors and #resp.errors == 0) then
+      if err ~= nil then
+        log.error("Error performing switch level action: " .. err)
```
greens, line 70: should this error message indicate the capability command that had the error (as you've done here) or the api command that had the error? Fine with either answer.

greens (approved):

drivers/SmartThings/philips-hue/src/handlers.lua
dljsjr: No that's a copy paste oopsie, I'll address it.

cjswedes (approved):

PR #476, opened 2023-01-12
CHAD-10123 Fix Stelpro thermostat setpoint logic

greens:
Fahrenheit values were causing a double-conversion, resulting in a loss of precision that caused overly-large sets to be sent to the device.


greens:
I expect this still might not work as it seems the thermostat rejects anything that isn't a whole number.

drivers/SmartThings/zwave-thermostat/src/test/test_stelpro_ki_thermoestat.lua
```diff
@@ -69,16 +69,16 @@ test.register_coroutine_test(
...
-test.register_coroutine_test(
-  "Setting heating setpoint should not be handled",
-  function()
-    mock_device:set_field(constants.TEMPERATURE_SCALE, 1, {persist = true})
-
-    test.timer.__create_and_queue_test_time_advance_timer(1, "oneshot")
-    test.socket.capability:__queue_receive({ mock_device.id, { capability = "thermostatHeatingSetpoint", command = "setHeatingSetpoint", args = { 333.0 } } })
-  end
-)
+-- This test case should not be necessary as the bounds should be handled by the range in the profile
```
ctowns, line 72: Do you plan to eventually remove this or leave the comment here for posterity?

greens: open to either. what do you think?

ctowns: I'm fine with just removing the commented out test since it isn't needed anymore and the embedded configs are becoming more widespread so it's something I think someone would be familiar with when looking at the driver.

ctowns (approved):

greens:
aware of the test in ct100, but that DTH is being worked on here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/470 and I don't want to create conflicts for myself.

PR #483, opened 2023-01-16
Wemo Migration Troubleshooting

satyajitanand:
Troubleshooting for https://smartthings.atlassian.net/browse/CHAD-9965 (Wemo Migration Issue).

### DIFF
--- a/drivers/SmartThings/wemo/src/discovery.lua
+++ b/drivers/SmartThings/wemo/src/discovery.lua
@@ -26,7 +26,7 @@ local function fetch_device_metadata(url)
 
   local response = table.concat(responsechunks)
 
-  log.trace("metadata response status", body, status, headers)
+  -- log.trace("metadata response status", body, status, headers)
 
   -- vvvvvvvvvvvvvvvv TODO: errors are coming back as literal string "[string "socket"]:1239: closed"
   -- instead of just "closed", so do a `find` for the error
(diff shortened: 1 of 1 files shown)


dljsjr (approved):

greens (commented):
Does this need to be merged or can you just test using the channel created with the PR? I ask because there are no functional changes.

satyajitanand:
> 

@greens : yeah, you are right. This is targeted for migration testing purpose only (to identify the root cause and find solution). Can you kindly help to approve/merge it to beta.

cjswedes:
I think this PR might not be needed, as I have a PR out there that refactors the discovery loop (#501), and also because I have discovered a situation that is unhandled during migration (and likely the cause of many migration issues). The migration issue is that the DTH uses a MAC that is discovered on the hub via linux utilities, and the driver uses the MAC that is reported by the device in its metadata; these values are "off by one". This means that when we try to find a migrated device on the network with the driver we discover the device but it reports a different MAC. [Here](https://github.com/tigoe/WeMoExamples/blob/master/README.md#getting-the-wemo-mac-address) is a good explanation of why wemo devices have 2 different MAC addresses. The DTH discovers the MAC used when the device is operational (mac+1), and the driver discovers the MAC used when onboarding with the devices access point; not all wemos do this though, which is annoying, and only the MAC used during onboarding is discoverable via the device's endpoints.

lelandblue:
Closing per above.

PR #484, opened 2023-01-17
Fix multiple comments from duplicate profile script

ctowns:
Fixes an issue where multiple "No duplicate profiles detected." messages would be posted on PR's. The script will now edit the same comment everytime it is run.


greens (approved):

franciscot276 (changes requested):
Hacked

PR #491, opened 2023-01-18
fix: use EDGE_CHILD type for lights

dljsjr:
EDGE_CHILD didn't exist when I started this driver, and now it does.

Open question: How do we fix devices already in the wild?


dljsjr:
> > How do we fix devices already in the wild?
> 
> Do we need to fix them? I dont think this will break them, so ignoring should be fine. If they do cause issues, are there enough out there where it is worth it to handle rather than just have the user rejoin devices?

I checked with HCS and they said we don't need to fix them.

cjswedes (approved):

PR #492, opened 2023-01-18
fix: Forgot that EDGE_CHILD devices don't have DNIs

dljsjr:
Forgot to refactor all usages of `device_network_id` to `parent_assigned_child_key` with the previous `EDGE_CHILD` refactor.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/handlers.lua
+++ b/drivers/SmartThings/philips-hue/src/handlers.lua
@@ -9,14 +9,14 @@ local st_utils = require "st.utils"
 local handlers = {}
 
 ---@param driver HueDriver
----@param device HueDevice
+---@param device HueChildDevice
 local function do_switch_action(driver, device, args)
   local on = args.command == "on"
   local id = device:get_field(Fields.PARENT_DEVICE_ID)
   local bridge_device = driver:get_device_info(id)
 
   if not bridge_device then
-    log.warn("Couldn't get a bridge for light with DNI " .. device.device_network_id)
+    log.warn("Couldn't get a bridge for light with Child Key " .. device.parent_assigned_child_key)
     return
   end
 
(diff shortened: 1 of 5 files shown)


drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -42,13 +42,13 @@ local function do_switch_action(driver, device, args)
...
 ---@param driver HueDriver
----@param device HueDevice
+---@param device HueChildDevice
 local function do_switch_level_action(driver, device, args)
   local level = st_utils.clamp_value(args.args.level, 1, 100)
   local bridge_device = driver:get_device_info(device:get_field(Fields.PARENT_DEVICE_ID))
 
   if not bridge_device then
-    log.warn("Couldn't get a bridge for light with DNI " .. device.device_network_id)
+    log.warn("Couldn't get a bridge for light with Child Key " .. device.parent_assigned_child_key)
```
cjswedes, line 51: do we need the `device.device_network_id or device.parent_assigned_child_key` here?

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -511,7 +511,7 @@ local function init_light(driver, device)
       device:set_field(Fields.MIN_KELVIN, HueApi.MIN_TEMP_KELVIN_WHITE_AMBIANCE, { persist = true})
     end
   end
-  local device_light_resource_id = device:get_field(Fields.RESOURCE_ID) or device.device_network_id
+  local device_light_resource_id = device:get_field(Fields.RESOURCE_ID) or device.parent_assigned_child_key
```
cjswedes, line 514: do we need to consider old device's `device.device_network_id` here?

drivers/SmartThings/philips-hue/src/handlers.lua
dljsjr: Shouldn't need it, all of the handlers will only ever receive bulbs (bridges will be filtered out by now by pseudo-type checks). And lights will never have a DNI with this change.

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: Yes, good eyes.

cjswedes (approved):

PR #496, opened 2023-01-20
Aeotech Multisensor:added configuration get command

sjl2178:
Edited added handler to include a Configuration:Get() call to query for power source in Aeotech Multisensor 6 & 7

### DIFF
--- a/drivers/SmartThings/zwave-sensor/src/aeotec-multisensor/multisensor-6/init.lua
+++ b/drivers/SmartThings/zwave-sensor/src/aeotec-multisensor/multisensor-6/init.lua
@@ -18,7 +18,24 @@ local cc = require "st.zwave.CommandClass"
 --- @type st.zwave.CommandClass.Configuration
 local Configuration = (require "st.zwave.CommandClass.Configuration")({ version = 2 })
 
+local preferences = require "preferences" 
+
 local MULTISENSOR_6_PRODUCT_ID = 0x0064
+local PREFERENCE_NUM = 9
+
+local function preference_update(driver, device, args)
+  preferences.update_preferences(driver, device, args)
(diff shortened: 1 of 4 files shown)


sjl2178:
@greens Was this what you meant by updating init to call update_preference and then the Get call?

drivers/SmartThings/zwave-sensor/src/aeotec-multisensor/multisensor-6/init.lua
```diff
@@ -18,8 +18,20 @@ local cc = require "st.zwave.CommandClass"
...
 
+local function device_added(self, device)
+  device:send(Configuration:Get({parameter_number = 9})) 
+  device:refresh()
+end
+
+local function device_init(self, device)
+  device:set_update_preferences_fn(preferences.update_preferences)
+  device:send(Configuration:Get({parameter_number = 9}))
+end
```
greens, line 33: added is correct, but here what I would do is set the update preferences function to a new function that is a composition of the standard preferences update function + the sending of the configuration get

greens (approved):
Great. The one thing I would change is to pull out the specific preference number (10 or 9) to and define it at the top of the file, since we're re-using it three times now.

greens:
If you wanted to be super hardcore, you could probably use a device field set in the init function of the driver to define the specific preference we care about, and then refactor these two sub-drivers to be nested, but that's a lot of work for not much gain.

sjl2178:
> If you wanted to be super hardcore, you could probably use a device field set in the init function of the driver to define the specific preference we care about, and then refactor these two sub-drivers to be nested, but that's a lot of work for not much gain.

I thought about that, but then figured that was a lot of work for a relatively straightforward fix lol

PR #499, opened 2023-01-23
Fix z-wave siren to allow it to run on versions<45

greens:


greens:
missed this in my initial removal of synthetic events

ctowns (approved):

PR #501, opened 2023-01-24
Feature/wemo improvements

cjswedes:
This provides a few changes initially to make development easier, and then the main change is to refactor the discovery module to be a cosock task that can handle multiple discovery requests at once rather than spinning up a discovery loop for each request.

Other issues I have identified in the wemo driver that will be taken care of in separate PRs are:
* Device subscriptions not being setup properly where the listening server is getting updates
* Devices being rediscovered after failed communication if necessary (i.e. ip addr changes).
* Migration issue where DTHs have different macs than what is reported by the device explained more [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/483#issuecomment-1402773011). Fixed by #505

### DIFF
--- a/drivers/SmartThings/wemo/src/util.lua
+++ b/drivers/SmartThings/wemo/src/util.lua
@@ -13,4 +13,15 @@ function util.tablefind(t, path)
   return item
 end
 
+--- This alternate determination of MAC addrs being equal is
+--- needed since wemo devices usually have a MAC on the network
+--- one greater than what is reported by the device. Migrated
+--- devices use the real network MAC, and devices joined to the
+--- driver use the device reported MAC.
+function util.mac_equal(m1, m2)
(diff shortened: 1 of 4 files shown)


drivers/SmartThings/wemo/src/util.lua
```diff
@@ -13,4 +13,15 @@ function util.tablefind(t, path)
   return item
 end
 
+--- This alternate determination of MAC addrs being equal is
+--- needed since wemo devices usually have a MAC on the network
+--- one greater than what is reported by the device. Migrated
+--- devices use the real network MAC, and devices joined to the
+--- driver use the device reported MAC.
+function util.mac_equal(m1, m2)
```
dljsjr, line 21: Clever :)

dljsjr (approved):

PR #504, opened 2023-01-24
Enhance changed drivers detection

greens:


greens:
Gets rid of some of the hacky bash I was using to get the list of changed drivers. The hope is that the changes should now reflect the differences the PR has with the version of the base branch of the PR, not the top of tree of the base branch.

This was causing problems where stale PRs would think there were large number of changes due to the differences here.

Also updates checkout to v3.

PR #505, opened 2023-01-25
Wemo MAC differences on migrated devices

cjswedes:
The MAC printed on the device and advertised in device metadata is used for the access point that the wemo device sets up for onboarding, and the MAC that is used once on the homes network is one more than what is advertised in metadata. The DTH got the actual MAC because its discovery pulls this info from the network; the driver queries the device for this info because it is not available on the socket that we have in drivers. This change will consider MAC addrs that are off by one as equal when discovering devices on the LAN.

An alternative approach would be to "calculate" the actual mac address from what is reported from the device, but there is no guarantee that all wemo devices will actually being doing this MAC change. Documentation is very limited.

### DIFF
--- a/drivers/SmartThings/wemo/src/discovery.lua
+++ b/drivers/SmartThings/wemo/src/discovery.lua
@@ -4,6 +4,7 @@ local http = cosock.asyncify "socket.http"
 local ltn12 = require "ltn12"
 local log = require "log"
 local tablefind = require "util".tablefind
+local mac_equal = require "util".mac_equal
 local utils = require "st.utils"
 local xml2lua = require "xml2lua"
 local xml_handler = require "xmlhandler.tree"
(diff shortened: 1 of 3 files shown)


greens (approved):

cjswedes:
> I also added one improvement of persisting ip/port that makes the integration more stable across driver restarts, since I have seen wemos just stop responding to ssdp after awhile on the network even though they are still reachable.

So I did also encounter this in my migration test, where the migrated device didn't respond to ssdp, and therefore was non-functional in the driver since we never properly initialized the device the first time it was joined to the driver. Unplug and replug the device has always fixed this for me, when our endless retry does the next discovery attempt to initialize the device. I worry that migration may result in a lot of devices in this state until the user unplugs and replugs and waits 5 min (max retry interval). We could mitigate this potentially by extracting the ip and port from the DTH device data so the driver can at least try to communicate with the device until driver disco is successful.

PR #507, opened 2023-01-25
Fix typo in profile name on fingerprint

greens:


sjl2178 (approved):
Sometimes copy paste does not pay off

PR #456, opened 2023-01-09
Add step to aeotec thermostat setpoint; remove unneeded capabilities

ctowns:
https://smartthings.atlassian.net/browse/CHAD-8786

PR #508, opened 2023-01-25
CHAD-10123 Round fahrenheit values before sending them to z-wave ther…

greens:
…mostats

Occasionally the UI will send us fractional values which we were sending along dutifully, but thermostats will often ignore these.

### DIFF
--- a/drivers/SmartThings/zwave-thermostat/src/init.lua
+++ b/drivers/SmartThings/zwave-thermostat/src/init.lua
@@ -48,6 +48,11 @@ local function convert_to_device_temp(command_temp, device_scale)
   elseif (command_temp >= 40 and (device_scale == ThermostatSetpoint.scale.CELSIUS or device_scale == nil)) then
     command_temp = utils.f_to_c(command_temp)
   end
+
+  -- Lots of devices don't like being sent fractional F values, so let's just round them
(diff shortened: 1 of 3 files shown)


ctowns (commented):
There are some devices that will take fractional values if the "step" embedded config is a fractional number (see https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/456).

Is there a way that we only round on devices where the step config is not present?

greens:
@ctowns I _believe_ that's intended to be an EU-only device? And 1 C degree is 1.8 F degrees, so .5C is. roughly equal to 1F steps anyway, so I'm not _too_ concerned about it, but it's a valid concern and we may need to address this later.

sjl2178 (approved):

ctowns (approved):
@greens you are right, that device is intended for EU-only (see zwave device info [here](https://manual.zwave.eu/backend/make.php?lang=en&sku=AEOEZWA021)), so this is okay since celcius is the unit that is used and I believe this is the only zwave thermostat we intend to have the step value at the moment. But I agree we will probably want to address it later.

lelandblue:
Tested by Camille she had success

greens:
this change might not be necessary since I forgot to actually deploy to beta on Tuesday, double-checking with Camille

PR #513, opened 2023-01-27
Added samsung speaker model check

satyajitanand:
Added samsung speaker model check to filter SSDP response message during discovery to support the devices mentioned at https://github.com/SmartThingsCommunity/SmartThingsPrivate/blob/d03ee28441cea96245634faf7a59dc7bf36b2452/devicetypes/smartthings/lan-samsungaudio.src/lan-samsungaudio.groovy#L1093

### DIFF
--- a/drivers/SmartThings/samsung-audio/src/disco.lua
+++ b/drivers/SmartThings/samsung-audio/src/disco.lua
@@ -93,6 +93,18 @@ local function process_response(val)
   return info
 end
 
+local function check_samsung_model(val)
+  log.debug(string.format("Doing the Device Model check --> %s", val))
+   if string.find(val, 'WAM7500') or string.find(val, 'WAM6500') or string.find(val, 'WAM5500') or string.find(val, 'WAM3500') or string.find(val, 'WAM3501') or 
+      string.find(val, 'WAM1500') or string.find(val, 'WAM1501') or string.find(val, 'WAM1400') or string.find(val, 'WAM750') or string.find(val, 'WAM550') or
+      string.find(val, 'WAM350') or string.find(val, 'J8500') or string.find(val, 'J7500') or string.find(val, 'J6500') or string.find(val, 'J650') or string.find(val, 'H750') or
+      string.find(val, 'K650') or string.find(val, 'K850') or string.find(val, 'K950') or string.find(val, 'J6500R') or string.find(val, 'J7500R') or string.find(val, 'J8500R') then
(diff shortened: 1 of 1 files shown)


satyajitanand:
This PR is in regard to https://smartthings.atlassian.net/browse/CHAD-10254.
1. Samsung Speaker Model check applied.
2. Discovery logic enhanced further.

Test Result:
1. I tested onboarding locally with this PR code and it's working fine at my end.
2. I performed migration testing for samsung-audio also (with the current samsung code in Beta) and migration worked fine at my end. Yesterday samsung-audio PR merged to beta.

Can you kindly review.

drivers/SmartThings/samsung-audio/src/disco.lua
```diff
@@ -93,6 +93,18 @@ local function process_response(val)
   return info
 end
 
+local function check_samsung_model(val)
+  log.debug(string.format("Doing the Device Model check --> %s", val))
+   if string.find(val, 'WAM7500') or string.find(val, 'WAM6500') or string.find(val, 'WAM5500') or string.find(val, 'WAM3500') or string.find(val, 'WAM3501') or 
+      string.find(val, 'WAM1500') or string.find(val, 'WAM1501') or string.find(val, 'WAM1400') or string.find(val, 'WAM750') or string.find(val, 'WAM550') or
+      string.find(val, 'WAM350') or string.find(val, 'J8500') or string.find(val, 'J7500') or string.find(val, 'J6500') or string.find(val, 'J650') or string.find(val, 'H750') or
+      string.find(val, 'K650') or string.find(val, 'K850') or string.find(val, 'K950') or string.find(val, 'J6500R') or string.find(val, 'J7500R') or string.find(val, 'J8500R') then
```
greens, line 101: these model names should be stored in a list somewhere, preferably at the top of the file, and you should just loop through them and return true if there's a match

greens: unless there's a better pattern to use here @dljsjr

ctowns: I moved the model names to a list at the top of the file and then created a loop for matching with the model name 👍

dljsjr: > unless there's a better pattern to use here @dljsjr

Seems reasonable to me

dljsjr (approved):

greens (approved):

PR #514, opened 2023-01-27
Feature/wemo listening server subscription refactor

cjswedes:
The first commit shows the design I was going for before I found a few hub firmware bugs that resulted in a change of design. Unless there is an objection, I think I will leave it there to look back on once the fw bugs are fixed. The second commit contains the working subscription server.

I have done some testing with 5 switch devices (3 different models), and all seem to be maintaining their subscriptions. I will do further testing with migration and the motion devices prior to merging.

### DIFF
--- a/drivers/SmartThings/wemo/src/init.lua
+++ b/drivers/SmartThings/wemo/src/init.lua
@@ -21,6 +21,7 @@ local capabilities = require "st.capabilities"
 local command_handlers = require "command_handlers"
 local discovery = require "discovery"
 local protocol = require "protocol"
+local SubscriptionServer = require "subscription_server"
 
 local Driver = require "st.driver"
 
@@ -38,40 +39,7 @@ local profiles = {
   ["Lightswitch"] = "wemo.light-switch.v1",
 }
 
-local server = {}
-
-local function start_server(driver)
-  server.listen_sock = socket.tcp()
-
-  -- create server on IP_ANY and os-assigned port
-  assert(server.listen_sock:bind("*", 0))
-  assert(server.listen_sock:listen(1))
-  local ip, port, _ = server.listen_sock:getsockname()
-
-  if ip ~= nil and port ~= nil then
-    log.info_with({ hub_logs = true }, string.format("Started listening server on %s:%s", ip, port))
-    server.listen_port = port
-    server.listen_ip = ip;
-    driver:register_channel_handler(server.listen_sock, protocol.accept_handler)
-  else
-    log.error_with({ hub_logs = true }, "could not get IP/port from TCP getsockname(), not listening for device status")
-    server.listen_sock:close()
-    server.listen_sock = nil
-    --TODO schedule a retry if this happens?
-  end
-end
-
-local function stop_server()
-  log.info_with({ hub_logs = true },
-    string.format("Shutting down listening server @ %s:%s", server.listen_ip, server.listen_port))
-
-  if server.listen_sock ~= nil then
-    server.listen_sock:close()
-  end
-  --TODO unregister channel handler? Seems possible to get in a state
-  -- where sockets are never gc'd because they are held by the driver.
-  server = {}
-end
+local server
 
 -- build a exponential backoff time value generator
 --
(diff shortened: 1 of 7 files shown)


drivers/SmartThings/wemo/src/parser.lua
```diff
@@ -30,49 +30,41 @@ end
 
 local function handle_binary_state(device, value)
   -- parses wemo insight style ("8|1611850428|58|...") state, works just fine for a single value too
```
dljsjr, line 32: Might be nice if there was a link to a doc that describes what the format being parsed looks like but I'm assuming that document doesn't exist.

drivers/SmartThings/wemo/src/parser.lua
```diff
@@ -30,49 +30,41 @@ end
...
-  log.debug("binary state value=", value)
   local vals = string.gmatch(value, "%d+")
```
dljsjr, line 33: So is the idea that it's just a bunch of pipe-delimited numbers then?

drivers/SmartThings/wemo/src/init.lua
```diff
@@ -38,40 +39,7 @@ local profiles = {
...
-  if server.listen_sock ~= nil then
-    server.listen_sock:close()
-  end
-  --TODO unregister channel handler? Seems possible to get in a state
-  -- where sockets are never gc'd because they are held by the driver.
-  server = {}
-end
+local server
 
 -- build a exponential backoff time value generator
```
dljsjr, line 44: This function is in `st.utils` now.

drivers/SmartThings/wemo/src/init.lua
```diff
@@ -121,32 +89,39 @@ local function device_init(driver, device)
...
+  --Persist to avoid issues with driver restart
   device:set_field("ip", info.ip, { persist = true })
   device:set_field("port", info.port, { persist = true })
   device:set_field("serial_num", info.serial_num, { persist = true })
 
-  --TODO maybe we should call_on_schedule with the device thread, and the polling.
-  -- instead of doing the resubscribe for all devices at once.
-  protocol.subscribe(server, device)
-end
+  local jitter = 2 * math.random()
```
dljsjr, line 110: What's the variance for?

dljsjr (approved):
Some style comments/clarifying questions but LGTM otherwise

drivers/SmartThings/wemo/src/parser.lua
cjswedes: wemo? documentation? I wish. These numbers are still a mystery to me.

drivers/SmartThings/wemo/src/parser.lua
cjswedes: Yeah, my guess is that they are power/energy numbers from the insight plug, but I have yet to make sense of them, and there's no documentation that I have found.

drivers/SmartThings/wemo/src/init.lua
cjswedes: During Migration, driver startup, and polling intervals, it spreads network traffic out. If 10 wemo devices are added, we would send 10 http request in like a quarter second if all devices respond to the first discovery message. And then every 60 sec we would send 10 more. This is an attempt to spread that out a little bit. The same goes for re subscribing. It makes the biggest difference during driver restart.

Im not sure it solves a problem that exists, but these devices are unreliable and I want to give them the best chance.

drivers/SmartThings/wemo/src/init.lua
```diff
@@ -113,74 +52,85 @@ local function device_init(driver, device)
...
-    log.info_with({ hub_logs = true }, "[" .. device.device_network_id .. "] polling device")
-    protocol.poll(driver, device)
-  end
+    --Sometimes wemos just stop responding to ssdp even though they are connected to the network.
+    --Persist to avoid issues with driver restart
+    device:set_field("ip", info.ip, { persist = true })
+    device:set_field("port", info.port, { persist = true })
+    device:set_field("serial_num", info.serial_num, { persist = true })
+    if driver.server and (ip ~= info.ip or port ~= info.port) then
+      print("not working how expected!!!!!!!!!!1")
```
sjl2178: nit: I think 1 and ! are equivalent but not everyone will

sjl2178 (approved):

cjswedes:
Merging after doing a migration test with 7 switches and 3 motion sensors. There are some pain points in the integration, but nothing as bad as the DTH backed devices and the integration overall works much better in the driver IMO.

greens:
Squash your commits next time, please.

cjswedes:
I intentionally left multiple commits because they represent specific pieces of functionality, and improvements. It aids in review both in the PR and when looking back in git history imo.

PR #516, opened 2023-01-27
Add fingerprints for Eve Energy

greens:
This reverts commit 376cdbeed9c6e664942ebb19a1823e2bdbf204d7.

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -1,3 +1,26 @@
+matterManufacturer:
+#Eve
+  - id: "Eve/Energy/US"
+    deviceLabel: Eve Energy US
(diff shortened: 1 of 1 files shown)


lelandblue (approved):

drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -1,3 +1,26 @@
+matterManufacturer:
+#Eve
+  - id: "Eve/Energy/US"
+    deviceLabel: Eve Energy US
+    vendorId: 0x130a
```
tpmanley, line 5: Just a nit, but good to be consistent across the fingerprints. Two are currently lowercase and two are currently uppercase. I prefer uppercase personally.

```suggestion
    vendorId: 0x130A
```

PR #517, opened 2023-01-27
Add luacheck linting on pull requests

greens:


greens:
_Lots_ of changes in here. I recommend folks take a look at drivers/etc they care about and see if they notice anything particularly egregious.

dljsjr (approved):
Sonos and Hue: LGTM 👍

drivers/SmartThings/zwave-sensor/src/init.lua
```diff
@@ -67,14 +67,14 @@ local function do_configure(driver, device)
...
-  [capabilities.motionSensor.ID] = capabilities.motionSensor.motion.inactive()
-}
+-- local initial_events_map = {
+--   [capabilities.tamperAlert.ID] = capabilities.tamperAlert.tamper.clear(),
+--   [capabilities.waterSensor.ID] = capabilities.waterSensor.water.dry(),
+--   [capabilities.moldHealthConcern.ID] = capabilities.moldHealthConcern.moldHealthConcern.good(),
+--   [capabilities.contactSensor.ID] = capabilities.contactSensor.contact.closed(),
+--   [capabilities.smokeDetector.ID] = capabilities.smokeDetector.smoke.clear(),
+--   [capabilities.motionSensor.ID] = capabilities.motionSensor.motion.inactive()
+-- }
```
cjswedes, line 77: Did the luacheck comment these out rather than remove them or was that you?

drivers/SmartThings/zwave-siren/src/aeotec-doorbell-siren/init.lua
```diff
@@ -200,11 +200,11 @@ local function activateTamper(device)
...
-local function clearAlarmAndChimeStateOfSoundComponents(device)
-  for endpoint = 1, NUMBER_OF_SOUND_COMPONENTS do
-    clearAlarmAndChime(device, endpoint)
-  end
-end
+-- local function clearAlarmAndChimeStateOfSoundComponents(device)
+--   for endpoint = 1, NUMBER_OF_SOUND_COMPONENTS do
+--     clearAlarmAndChime(device, endpoint)
+--   end
+-- end
```
cjswedes, line 207: nit: can we remove this entirely?

cjswedes (approved):
I reviewed the matter driver, bose, wemo, and all the zwave drivers. LGTM.

Is it possible for us to add custom lints in the future? The use case I have in mind is for replacing usages of things like `require "dkjson"` with `require "st.json"`.

Not sure on why there were a couple commented out blocks rather than just removing them entirely.

drivers/SmartThings/zwave-sensor/src/init.lua
greens: I commented those out because that code is currently unused, but will be re-enabled after migration is done so that newly-joined devices once again emit synthetic events.

drivers/SmartThings/zwave-siren/src/aeotec-doorbell-siren/init.lua
greens: same as above here, this code will be turned back on eventually

drivers/SmartThings/zwave-sensor/src/init.lua
cjswedes: ahh okay, forgot about this.

satyajitanand (approved):
For samsung-audio: LGTM

PR #519, opened 2023-01-30
CHAD-10275 Change the device type of zwave dawon smartPlug(switch -> smartPlug)

inasail:
[CHAD-10275](https://smartthings.atlassian.net/browse/CHAD-10275)


greens (approved):
@inasail looks like an issue with that workflow around commenting on PRs from forks. @ctowns is working on it.

PR #520, opened 2023-01-30
zwave-sensor: Homeseer Multisensor

sjl2178:
Edited homeseer multisensor sub-driver to support device refresh specified from DTH.

### DIFF
--- a/drivers/SmartThings/zwave-sensor/src/homeseer-multi-sensor/init.lua
+++ b/drivers/SmartThings/zwave-sensor/src/homeseer-multi-sensor/init.lua
@@ -19,7 +19,9 @@ local cc = require "st.zwave.CommandClass"
 local Basic = (require "st.zwave.CommandClass.Basic")({ version = 1 })
 --- @type st.zwave.CommandClass.WakeUp
 local WakeUp = (require "st.zwave.CommandClass.WakeUp")({ version = 1 })
-
+local SensorMultilevel = (require "st.zwave.CommandClass.SensorMultilevel")({version = 5})
+local Battery = (require "st.zwave.CommandClass.Battery")({ version = 1})
+local utils = require "st.utils"
 local HOMESEER_MULTI_SENSOR_FINGERPRINTS = {
   { manufacturerId = 0x001E, productType = 0x0002, productId = 0x0001 }, -- Homeseer multi sensor HSM100
 }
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zwave-sensor/src/homeseer-multi-sensor/init.lua
```diff
@@ -54,11 +56,18 @@ local function update_preferences(self, device, args)
   end
 end
 
-local function device_init(self, device)
-  device:set_update_preferences_fn(update_preferences)
+local function device_refresh(driver, device, args)
+  update_preferences(driver, device, args)
+  local get_temp = SensorMultilevel:Get({sensor_type = SensorMultilevel.sensor_type.TEMPERATURE})
```
greens, line 61:
```suggestion
  local get_temp = SensorMultilevel:Get({sensor_type = SensorMultilevel.sensor_type.TEMPERATURE}, {dst_channels = {3}})
```

drivers/SmartThings/zwave-sensor/src/homeseer-multi-sensor/init.lua
```diff
@@ -54,11 +56,18 @@ local function update_preferences(self, device, args)
   end
 end
 
-local function device_init(self, device)
-  device:set_update_preferences_fn(update_preferences)
+local function device_refresh(driver, device, args)
```
greens, line 59: DTH also includes a battery_get in here, I think

sjl2178: It does, but it is also being sent here, I'm not entirely sure where it gets sent but according to the tests a get battery does get sent on init.

dljsjr: The shared functionality between init and refresh is `update_preferences`, which makes me think that it's the update of the wake interval that's causing a battery report to come in (somehow). I don't know enough about z-wave to know if that makes sense.

> according to the tests a get battery does get sent on init
Only thing that makes me concerned is that `update_preferences` appears to be a no-op if the preference hasn't actually changed. I would say that it makes sense for it to fire on init in the test suite but if somebody triggers a refresh via pull to refresh on a device card then the value won't change so that event might not get emitted. Which in turns means that the pull to refresh won't update the battery.

Also means that we probably need to add another test case since it was pretty easy to construct an adversarial scenario that the existing test wouldn't catch (assuming I'm right which I very well may not be)

greens: So I believe the battery check is just built in to some generic wakeup handling and I forgot about it, but doug makes a good point about the update_preferences mechanism that I hadn't considered.

Good to verify with a test of preferences not being changed, but I suspect he is right and we may need to go back to the drawing board.

greens: nvm, the battery get is in the battery defaults

we send one when we get a wakeup notification

sjl2178: Can confirm after a quick test that without `info_changed` `update_preferences` does appear to be a no-op.

dljsjr: > we send one when we get a wakeup notification

So that means this code's good? I need to learn much more about Z-Wave 😅

greens: setting the wakeup interval doesn't necessarily trigger a wakeup notification

I think what you may want to do is change the wakeup notification handler to send the refresh command (and also a battery refresh)

the update preferences function will just add itself on to any existing wakeup notification handler

drivers/SmartThings/zwave-sensor/src/homeseer-multi-sensor/init.lua
```diff
@@ -54,23 +56,30 @@ local function update_preferences(self, device, args)
...
+  local get_temp = SensorMultilevel:Get({sensor_type = SensorMultilevel.sensor_type.TEMPERATURE}, {dst_channels = {3}})
+  device:send(get_temp) 
+  local get_luminance = SensorMultilevel:Get({sensor_type = SensorMultilevel.sensor_type.LUMINANCE}, {dst_channels = {2}})
+  device:send(get_luminance) 
+  device:send(Battery:Get({}))
 end
 
-local function info_changed(self, device, event, args)
+local function device_init(self, device)
+  device:set_update_preferences_fn(update_preferences)
```
greens, line 68: this should be removed

greens (approved):

PR #525, opened 2023-01-31
Fix/hue-join-improvements-and-CHAD-10213

dljsjr:
## [fix CHAD-10213: Legacy color profile](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/525/commits/8952caa2207b2736155cdfb8182ed72961096e88)

Adds a profile for legacy Hue Color devices that support color
control without white/temperature control.

Fixes: CHAD-10213

## [fix: Don't send join message for joined bulbs](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/525/commits/31c489f84ece6c86242dae28f303850646651277)

EDGE_CHILD devices get an auto-assigned UUID DNI so we weren't
using the correct value for child devices when trying to ascertain
already joined devices.

This was leading to a flood of try_create_device messages during
discovery for already joined devices.

## [fix: Eliminate several spurious refreshes](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/525/commits/d09fc6d42a66956f6404bc5ba08fd31c0224616f)

The logic revised in this commit was causing a lot of spurious
events to be emitted, so we revise some things.

First is that we only refresh after init on fresh join, not general init.

Second is that we don't do a forced refresh whenever the SSE stream connects.
This is fine for fresh join because we do a forced refresh anyway, and it's
fine for reconnects because the SSE implementation will request any messages
that it has missed in the interrim whenever it performs a reconnect.

### DIFF
--- a/drivers/SmartThings/philips-hue/profiles/legacy-color.yml
+++ b/drivers/SmartThings/philips-hue/profiles/legacy-color.yml
@@ -1,0 +1,20 @@
+name: legacy-color
+components:
+- id: main
+  capabilities:
+  - id: switch
+    version: 1
+  - id: switchLevel
+    version: 1
+    config:
+      values:
+        - key: "level.value"
+          range: [ 1, 100 ]
+  - id: colorControl
+    version: 1
+  - id: samsungim.hueSyncMode
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Light
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -237,7 +237,9 @@ bridge_added = function(driver, device)
   end
 
   local bridge_ip = bridge_info.ip
-
+  if device:get_field(Fields._REFRESH_AFTER_INIT) == nil then
+    device:set_field(Fields._REFRESH_AFTER_INIT, true, { persist = true })
+  end
   device:set_field(Fields.DEVICE_TYPE, "bridge", { persist = true })
   device:set_field(Fields.IPV4, bridge_ip, { persist = true })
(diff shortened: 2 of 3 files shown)


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -237,7 +237,10 @@ bridge_added = function(driver, device)
   end
 
   local bridge_ip = bridge_info.ip
-
+  if device:get_field(Fields._REFRESH_AFTER_INIT) == nil then
+    device:set_field(Fields._REFRESH_AFTER_INIT, true, { persist = true })
+  end
+  device:set_field(Fields._REFRESH_AFTER_INIT, true, { persist = true })
```
sjl2178, line 243: Out of curiosity Doug, if you set Fields._REFRESH_AFTER_INIT after a nil check, why do you set it again? Wouldn't it get set to the same value whether the check returned nil or not? Or is the value getting set different?

dljsjr: That is a copy paste bug.

sjl2178 (approved):
LGTM :)

drivers/SmartThings/philips-hue/profiles/legacy-color.yml
```diff
@@ -0,0 +1,20 @@
...
+    version: 1
+  - id: switchLevel
+    version: 1
+    config:
+      values:
+        - key: "level.value"
+          range: [ 1, 100 ]
+  - id: colorControl
+    version: 1
+  - id: samsungim.hueSyncMode
```
greens, line 15: Where does this live?

dljsjr: No clue. HQ owns it. It's in Prod though.

PR #528, opened 2023-02-02
Port HSV < - > xy math from the Protocol Handler

dljsjr:
When this driver was originally published it used the HSV < - > CIE 1931 XY math in `st.utils` which is based on the math from `zll-rgbw-bulb` but it turns out that the transform for Hue isn't actually the same.

This PR introduces the same color conversion math that was used in the Protocol Handler for Hue, making the edge driver behave more like existing Hue integrations.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/handlers.lua
+++ b/drivers/SmartThings/philips-hue/src/handlers.lua
@@ -1,6 +1,6 @@
-local cosock = require "cosock"
 local Fields = require "hue.fields"
 local HueApi = require "hue.api"
+local HueColorUtils = require "hue.cie_utils"
 local log = require "log"
 
 local capabilities = require "st.capabilities"
@@ -91,7 +91,7 @@ end
 ---@param driver HueDriver
 ---@param device HueChildDevice
 local function do_color_action(driver, device, args)
-  local hue, sat = args.args.color.hue, args.args.color.saturation
+  local hue, sat = (args.args.color.hue / 100), (args.args.color.saturation / 100)
   local bridge_device = driver:get_device_info(device:get_field(Fields.PARENT_DEVICE_ID))
 
   if not bridge_device then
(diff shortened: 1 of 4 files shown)


greens:
bro are you gonna put emojis in our git history

dljsjr:
> bro are you gonna put emojis in our git history

Real talk I actually just inserted a unicode two headed arrow and GitHub replaced it with the Emoji.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -86,11 +87,13 @@ local function emit_light_status_events(light_device, light)
     end
 
     if light.color then
-      local x, y = light.color.xy.x * 65536, light.color.xy.y * 65536
-      local hue, sat = st_utils.safe_xy_to_hsv(x, y, 1)
+      light_device:set_field(Fields.GAMUT, light.color.gamut, { persist = true })
+      log.trace(st_utils.stringify_table(light.color.xy, "Received Color XY:\n", true))
```
greens, line 91: keeping this in?

dljsjr: No that log message should be removed.

drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -229,6 +226,7 @@ local function do_refresh_light(driver, light_device)
       else
         for _, light_info in ipairs(light_resp.data) do
           if light_info.id == light_resource_id then
+            light_device:set_field(Fields.COLOR_GAMUT, light_info.color.gamut_type, { persist = true })
```
greens, line 229: this doesn't have to be translated from JSON or anything?

dljsjr: It's already translated by the time it gets here. The `api` module does that processing here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/ecfa1ffdb9c52a0a166c878492ec9f0f7e11e7c6/drivers/SmartThings/philips-hue/src/hue/api.lua#L57

greens (approved):

sjl2178 (approved):

PR #533, opened 2023-02-05
fix: Responsive attr. emit after handling commands

dljsjr:
Previously we relied on the eventual update from the SSE stream to emit capability attribute updates that result from command handling. This is subject to latency spikes and responsiveness jitter due to encapsulation and rate limiting on the Hue bridge itself that it uses to conserve its own resources.

This change adopts a pattern found in other drivers where the endpoint provides an ACK which we use as a signal that we can emit and update that attribute update that matches the args to command we just sent. We then cache that attribute update, and we check against that for similarity before performing any additional attribute event emissions for that capability.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/handlers.lua
+++ b/drivers/SmartThings/philips-hue/src/handlers.lua
@@ -8,6 +8,25 @@ local st_utils = require "st.utils"
 
 local handlers = {}
 
+local function check_response_then(resp, err, on_success)
+  if not resp or (resp.errors and #resp.errors == 0) then
+    if err ~= nil then
+      log.error("Error performing on/off action: " .. err)
+      return
+    elseif resp and #resp.errors > 0 then
+      for _, error in ipairs(resp.errors) do
+        log.error("Error returned in Hue response: " .. error.description)
+      end
+      return
+    end
+  end
+  if type(on_success) == "function" then
+    on_success(resp)
+  else
+    log.warn("received argument in `on_success` position that is not a function")
+  end
+end
+
 ---@param driver HueDriver
 ---@param device HueChildDevice
 local function do_switch_action(driver, device, args)
@@ -30,15 +49,9 @@ local function do_switch_action(driver, device, args)
 
   local resp, err = hue_api:set_light_on_state(light_id, on)
 
-  if not resp or (resp.errors and #resp.errors == 0) then
-    if err ~= nil then
-      log.error("Error performing on/off action: " .. err)
-    elseif resp and #resp.errors > 0 then
-      for _, error in ipairs(resp.errors) do
-        log.error("Error returned in Hue response: " .. error.description)
-      end
-    end
-  end
+  check_response_then(resp, err, function(...)
+    driver.emit_light_status_events(device, { on = { on = on } })
+  end)
 end
 
 ---@param driver HueDriver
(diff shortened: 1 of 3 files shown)


dljsjr:
This was code complete on Friday, but I forgot to open the PR before taking flex time.

tpmanley:
Have you tested how this works when the light is unreachable/offline? I think the ACK just means the command was valid but doesn't mean the light is necessarily able to respond to the command. Even though it takes longer, we usually choose to wait for the device to positively confirm it's new state.

dljsjr:
> Have you tested how this works when the light is unreachable/offline? I think the ACK just means the command was valid but doesn't mean the light is necessarily able to respond to the command.

This is correct. In this case we would rely on the eventual Zigbee Connectivity update to come in eventually and just mark the entire light bulb as offline:

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/1c0bbde8eb74883663b64601e126315612a1beab/drivers/SmartThings/philips-hue/src/init.lua#L80

> Even though it takes longer, we usually choose to wait for the device to positively confirm it's new state. 

That's how the driver worked originally. This change was made because the timeouts in the plugin for some stuff are way tighter so without doing this change there is a lot of UI jitter/jank in the controls, and we're getting mad bug reports for it.

This is not my favorite PR, these are "latency optimizations" to improve the way the OneApp UI handles things. I'm open to other suggestions for improvement. If there's really high rate activity from the bridge, there can be up to a second of delay on the SSE event stream.

sjl2178 (approved):
Codewise LGTM :)

ctowns (approved):

drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -8,6 +8,25 @@ local st_utils = require "st.utils"
 
 local handlers = {}
 
+local function check_response_then(resp, err, on_success)
+  if not resp or (resp.errors and #resp.errors == 0) then
+    if err ~= nil then
+      log.error("Error performing on/off action: " .. err)
```
greens, line 14: nit: not all of these are on/off actions

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -83,16 +117,24 @@ local function emit_light_status_events(light_device, light)
...
-      light_device:emit_event(capabilities.colorControl.saturation(st_utils.round(sat * 100)))
+      if not _xy_nearly_equal(light.color.xy, last_emitted.color.xy) then
+        local r, g, b = HueColorUtils.safe_xy_to_rgb(light.color.xy, light.color.gamut)
+        local hue, sat, _ = st_utils.rgb_to_hsv(r, g, b)
+
+        local hue_emit = st_utils.round(hue * 100)
+        local sat_emit = st_utils.round(sat * 100)
+        light_device:emit_event(capabilities.colorControl.hue(hue_emit))
+        light_device:emit_event(capabilities.colorControl.saturation(sat_emit))
+        last_emitted.color.xy = light.color.xy
```
greens, line 136: Why are you setting this, won't it go out of scope at the end of this function? Should you be setting a field?

dljsjr:
I think we've decided that, while it would improve responsiveness, the "correct" thing to do is wait for the device to respond with the state change. So I'll close this PR without merging. We can always revisit this concept if we need to as it's relatively straight forward.

PR #534, opened 2023-02-06
Fix motion threshold on SmartThings multisensor

ctowns:
The motion threshold was being passed with incorrect endianess, causing the threshold for motion detection to be higher than expected.

Check here for more info on the bug: https://smartthings.atlassian.net/browse/DVCSMP-4739


ctowns:
@greens Do you know if this workaround would still be required? Based on what I can see in the https://smartthings.atlassian.net/browse/DVCSMP-4739 ticket, it seems like the issue was not on our side, but within the sensor itself, so I believe this workaround would still be required for the old sensors then, right?

greens (approved):
I assume this is on the device fw side, but I'm not sure. @tpmanley ?

Great catch.

tpmanley:
The ticket referenced was an issue specific to Groovy DTHs and the handling of `wattr` commands. That same issue should not exist for Lua drivers so I think the code should be correct as is. You can verify by reading that attribute back or watching the write on a sniffer.

ctowns:
This was not the actual issue (see Tom's comment above). The real issue was fixed here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/548

PR #536, opened 2023-02-07
Fix Sonos issues related to capability mismatches from DTH to Edge Driver

dljsjr:
## [Fix CHAD-10316 with embedded device config](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/a6c48c4cabd91f6ee93ff44a0e78271c70c0a4d0)

Remove the `Rewind` and `Fast Forward` capabilities because not everything in the plugin is respecting the supported command event emission.

Fixes: CHAD-10316

## [Fix: disconnect detection bugs](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/c485252110aea0f94cce83e2213e585925300854)

No ticket, just some issues that were found during development that would result in crashes sometimes during disconnect.

## [Fix: Restore `mediaGroup` capability functionality](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/c1a314d9349ea142034825f084a3bbd0b5788ef5)

Add the `mediaGroup` capability and its handlers/emissions back to Sonos. Erroneously dropped when porting.

Fixes: CHAD-10310

## [Fix: restore `audioNotification` capability](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/2c3d8a02e46de21f238b160c3110602ea40b0050)

Add the `audioNotification` and its handlers back to Sonos. Erroneously dropped when porting.

Fixes: CHAD-10289

### DIFF
--- a/drivers/SmartThings/sonos/src/api/sonos_connection.lua
+++ b/drivers/SmartThings/sonos/src/api/sonos_connection.lua
@@ -28,7 +28,7 @@ local SonosRestApi = require "api.rest"
 local SonosConnection = {}
 SonosConnection.__index = SonosConnection
 
-local self_subscriptions = { "groups", "playerVolume" }
+local self_subscriptions = { "groups", "playerVolume", "audioClip" }
 local coordinator_subscriptions = { "groupVolume", "playback", "favorites", "playbackMetadata" }
 local favorites_version = ""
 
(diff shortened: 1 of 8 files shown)


drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -248,7 +256,9 @@ end
 --- Whether or not the connection has all of the live websocket connections it needs to function
 --- @return boolean
 function SonosConnection:is_running()
-  return self:self_running() and self:coordinator_running()
+  local self_running = self:self_running()
+  local coord_running = self:coordinator_running()
+  return  self_running and coord_running
```
greens, line 261: what's different here?

greens (approved):

drivers/SmartThings/sonos/src/api/sonos_connection.lua
dljsjr: I refactored it for debugging and left it in. There's no functionality changes. It could possibly be a perf improvement with the way Lua looks for upvalues but nothing excited.

PR #539, opened 2023-02-07
Fix: Set the parent assigned key on migration

dljsjr:


dljsjr:
This might not be necessary, but I'm putting a draft PR in the wings in case we do need to make some changes here. Currently talking through stuff w/ Nate from HCS.

dljsjr:
This isn't a valid field to set via try update metadata and the issue itself was upstream. Closing this PR.

PR #542, opened 2023-02-08
Prevent migrating parent-child devices from creating children

greens:


greens:
If @varzac thinks this is correct (mostly concerned about the field addressing), I can put this on a channel Nate can test with.

drivers/SmartThings/zwave-switch/src/test/test_zwave_dual_switch_migration.lua
```diff
@@ -0,0 +1,113 @@
...
+      )
+    )
+  end
+)
+
+test.register_coroutine_test(
+  "Adding a device that doesn't have childen should create more",
+  function()
+    test.socket.device_lifecycle:__queue_receive({ mock_parent_no_data.id, "added" })
+    mock_parent_no_data:expect_device_create({
```
varzac, line 93: It's kinda weird to use this function is on a device object.  I think it would read easier if we directly use the mock device api: https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/main/lua_libs/integration_test/mock_devices_api.lua#L70

varzac (approved):

drivers/SmartThings/zwave-switch/src/test/test_zwave_dual_switch_migration.lua
greens: just using the same pattern as with `expect_metadata_update`

varzac: The difference is that the `expect_metadata_update` is related to a specific device, where the creation of a new device isn't associated with the existing device so it feels kinda weird.

greens: it _kinda_ is!

PR #546, opened 2023-02-09
Add firmwareUpdate capability to matter and zb profiles

samadDotDev:


ctowns:
Please ignore the failure in the "Check for duplicate profiles" step below. This is due to an issue with the duplicate profile script not handling profiles without categories and not properly handling preferences. It is an issue with the script and not with this PR.

@greens do you know if there was a specific reason why `firmwareUpdate` was excluded from all matter drivers?

greens:
@ctowns I think we just didn't know it should have been included. As far as I knew, `firmwareUpdate` was just for Zigbee and no one told us otherwise.

greens (approved):

ctowns (approved):

hubTab (approved):

Con-Tejus (approved):

cjswedes (approved):
Is it true that going forward this capability should be included in every matter profile without exceptions?

samadDotDev:
> Is it true that going forward this capability should be included in every matter profile without exceptions?

I believe so. The hubs running 0.47.x will emit `firmwareUpdate` capability with the current software version post device-onboarding for all matter devices. Additionally, if they undergo an update whether through our hub or any other fabric, the `availableVersion` attribute and update statuses will be reflected in the capability throughout the process.

greens:
feel free to merge @samadDotDev

PR #547, opened 2023-02-09
[Aqara] Wall Switch H1 Pro, Single Switch Module

sbkimaqara:
New switches
- Aqara Smart Wall Switch H1 Pro (With Neutral, Single Rocker)
- Aqara Smart Wall Switch H1 Pro (With Neutral, Double Rocker) 1
- Aqara Smart Wall Switch H1 Pro (With Neutral, Triple Rocker) 1
- Aqara Single Switch Module T1 (With Neutral)

~~- Aqara Smart Wall Switch (No Neutral, Double Rocker) 1~~
~~- Aqara Smart Dimmer Controller T1 Pro~~


drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -70,140 +116,251 @@ local function is_aqara_products(opts, driver, device)
...
-end
-
 local function present_value_handler(driver, device, value, zb_rx)
+  if device:get_model() == "lumi.plug.maeu01" then
+    local version = device:get_field(APPLICATION_VERSION) or 0
+    if version == 41 then
+      -- not allowed
+      return
+    end
+  end
```
greens, line 202: This sort of thing should have been left as a sub-driver how it was. If you want one sub-driver that handles both of the version-related commands, that's fine, but in general, for handling of events, if you've included a model check, it should probably be a sub-driver.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -70,140 +116,251 @@ local function is_aqara_products(opts, driver, device)
...
+        driver:try_create_device(metadata)
+      end
     end
 
-    if device.preferences[RESTORE_STATE_ID] ~= nil and
-        device.preferences[RESTORE_STATE_ID] ~= args.old_st_store.preferences[RESTORE_STATE_ID] then
-      device:send(cluster_base.write_manufacturer_specific_attribute(device, PREF_CLUSTER_ID,
-        PREF_RESTORE_STATE_ATTR_ID, MFG_CODE, data_types.Boolean, device.preferences[RESTORE_STATE_ID]))
+    -- for wireless button
+    if device:supports_capability(capabilities.button) then
```
greens, line 291: `supports_capability` checks are also a good indicator that we'd prefer a sub-driver, for readability's sake

greens (commented):
My best suggestion would be to have the acn004 and b2laus01 as separate PRs, and to leave the remaining 4 devices together.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: Moved plug's code to the sub-driver.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: Moved multi-switch's code to the sub-driver.

sbkimaqara:
Driver removed
- acn004
- b2laus01

New PR
- acn004 : https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/554

Thanks 👍

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -70,9 +111,25 @@ local function is_aqara_products(opts, driver, device)
...
+  local current_time = os.time()
+  local last_time = device:get_field(LAST_REPORT_TIME) or 0
+  local next_time = last_time + 60 * 15 -- 15 mins, the minimum interval allowed between reports
+  if current_time < next_time then
+    return
+  end
+  device:set_field(LAST_REPORT_TIME, current_time, { persist = true })
+
+  -- energy meter
+  device:emit_event(capabilities.energyMeter.energy({ value = raw_value, unit = "Wh" }))
```
greens, line 130: Shouldn't this be emitted regardless of the restrictions on `powerConsumptionReport`?

greens (commented):
I'm confused why you deleted the version-based handler when it looks like all of the handlers in `smart-plug` are based on the version of the device, and are otherwise identical to the parent driver.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: The reported power consumption value not only used the energyMeter capability but also used the powerConsumptionReport capability. So I thought it would be good to display the event to the user at the same interval. Code changes required?

sbkimaqara:
I written sub-driver handlers by the model because there are functional differences between models. I'll take your comments and rewrite the code with a version-based handler.

drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
```diff
@@ -0,0 +1,120 @@
...
+  { model = "lumi.switch.n2acn1", children = 2, child_profile = "aqara-switch-child" },
+  { model = "lumi.switch.n3acn1", children = 3, child_profile = "aqara-switch-child" },
+}
+
+local function do_configure(self, device)
+  if device.network_type == device_lib.NETWORK_TYPE_ZIGBEE then
+    device:configure()
+    device:send(cluster_base.read_manufacturer_specific_attribute(device,
+      PRIVATE_CLUSTER_ID, PRIVATE_ATTRIBUTE_ID, MFG_CODE))
+    device:send(SimpleMetering.attributes.CurrentSummationDelivered:configure_reporting(device, 900, 3600, 1)) -- minimal interval : 15min
```
sbkimaqara, line 25: Set the minimum interval allowed between reports of SimpleMetering to 15 minutes like the reporting interval of powerConsumptionReport.

sbkimaqara:
```
# drivers/SmartThings/zigbee-switch/src/aqara/init.lua
# Line 242
device:send(SimpleMetering.attributes.CurrentSummationDelivered:configure_reporting(device, 900, 3600, 1))
```

Set the minimum interval allowed between reports of the SimpleMetering to 15 minutes like the reporting interval of the powerConsumptionReport.

sbkimaqara:
```
# drivers/SmartThings/zigbee-switch/src/aqara/init.lua
# Line 240
device:send(cluster_base.read_manufacturer_specific_attribute(device,
    PRIVATE_CLUSTER_ID, PRIVATE_ATTRIBUTE_ID, MFG_CODE))
# Line 285
[PRIVATE_CLUSTER_ID] = {
  [PRIVATE_ATTRIBUTE_ID] = private_mode_handler
}
```
Instead of reading version, read the private-defined attribute's value and use it for electric measurement event. If the value from the device is 1, the AnalogInput cluster is used; otherwise, the ElectricalMeasurement is used.

sbkimaqara:
```
# drivers/SmartThings/zigbee-switch/src/aqara/init.lua
# Line 125, 155, 164, 177, 210
local private_mode = device:get_field(PRIVATE_MODE) or 0
```
The value read from the device determines which cluster to use.

sbkimaqara:
```
# drivers/SmartThings/zigbee-switch/src/aqara/init.lua
# Line 195
if current_time < next_time then
  return
end
```
The timer was only used for AnalogInput cluster reporting events. The SimpleMetering cluster was configured by configure_reporting().

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
greens: `powerConsumptionReport` is more focused on historical consumption over periods of time, but for `energyMeter` you should just report the most up-to-date info you have

sbkimaqara: I agree. However, I set it to 15 minute intervals in the configuration and timer condition because I need to consider for both current value and traffic.

```
local function present_value_handler(driver, device, value, zb_rx)
  # 195
  if current_time < next_time then
    return
  end
end

local function do_configure(self, device)
  # Line 242
  device:send(SimpleMetering.attributes.CurrentSummationDelivered:configure_reporting(device, 900, 3600, 1)) -- minimal interval : 15min
end
```

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -70,9 +112,36 @@ local function is_aqara_products(opts, driver, device)
...
+  local private_mode = device:get_field(PRIVATE_MODE) or 0
+  if private_mode == 1 then
+    -- read power meter
+    device:send(AnalogInput.attributes.PresentValue:read(device):to_endpoint(POWER_METER_ENDPOINT))
+  end
+end
+
+local function wireless_switch_handler(driver, device, value, zb_rx)
+  local endpoint = wireless_switch_endpoint_map[zb_rx.address_header.src_endpoint.value]
+  local event = wireless_switch_button_event_map[value.value]
```
greens, line 134: Is the plan to extend this at some point?

greens: (beyond the one value it has now)

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -70,9 +112,36 @@ local function is_aqara_products(opts, driver, device)
...
+  if event ~= nil then
+    device:emit_event_for_endpoint(endpoint, event)
```
greens, line 136: So the child devices emit events from _both_ endpoints `1,2,3` but also `0x29, 0x2A, 0x2B`? You could just change your find_child function to subtract 40 for endpoint sources over 3 and then avoid this extra mapping work.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -70,9 +112,36 @@ local function is_aqara_products(opts, driver, device)
   return false
 end
 
-local function emit_power_consumption_report_event(device, value)
-  local raw_value = value.value -- 'Wh'
+local function private_mode_handler(driver, device, value, zb_rx)
```
greens, line 115: can you help me understand what `private mode` is intended to do?

sbkimaqara: The device has an operating mode internally. Depending on this value, the cluster that the device reports for the event is determined. 
Let's take a smart plug's power consumption report as an example. 
- If the private mode value is 1, it is reported to the AnalogInput cluster. 
- If the value is 0, it is reported to the ElectricalMeasurement cluster.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: No. Only switch 1, 2 and 3.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: I didn't understand what you said. Can you tell me more specific?
When the user presses a switch (button) of the device, a switch press event is received. Users can create automations using these events as triggers.
The log is below.

Button 1:

received Zigbee message: 
< ZigbeeMessageRx || type: 0x00, < AddressHeader || src_addr: 0x999F, src_endpoint: 0x2A, dest_addr: 0x0000, dest_endpoint: 0x01, profile: 0x0104, cluster: 0x0012 >, lqi: 0xC8, rssi: -80, body_length: 0x0008, < ZCLMessageBody || < ZCLHeader || frame_ctrl: 0x18, seqno: 0x08, ZCLCommandId: 0x0A >, < ReportAttribute || < AttributeRecord || AttributeId: 0x0055, DataType: Uint16, Uint16: 0x0001 > > > >

emitting event: 
{"attribute_id":"button","capability_id":"button","component_id":"main","state":{"value":"pushed"},"state_change":true}
2023-02-22T03:42:52.761736341+00:00 DEBUG Zigbee Switch  Aqara Smart Wall Switch (No Neutral, Double Rocker) 1 device thread event handled

Button 2:

received Zigbee message: 
< ZigbeeMessageRx || type: 0x00, < AddressHeader || src_addr: 0x999F, src_endpoint: 0x29, dest_addr: 0x0000, dest_endpoint: 0x01, profile: 0x0104, cluster: 0x0012 >, lqi: 0xCA, rssi: -64, body_length: 0x0008, < ZCLMessageBody || < ZCLHeader || frame_ctrl: 0x18, seqno: 0x0A, ZCLCommandId: 0x0A >, < ReportAttribute || < AttributeRecord || AttributeId: 0x0055, DataType: Uint16, Uint16: 0x0001 > > > >

emitting event: 
{"attribute_id":"button","capability_id":"button","component_id":"main","state":{"value":"pushed"},"state_change":true}

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
greens: is there a reason I'd want this reported on one cluster or another?

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
greens: then it seems strange to use a map in only one place with only one value

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
greens: The `emit_event_for_endpoint` function first tries to use the device's defined `find_child` function, and then falls back to the `get_component_for_endpoint` function (for multi-component devices). You have your regular mapping of children, which seems to be `endpoint 1 -> switch 1`, `endpoint 2 -> switch 2`, etc. But you're also mapping these events from other endpoints to these same switch children. If you changed your `find_child` function to also map events from `endpoint 0x29 -> switch 1`, you wouldn't have to do this separate lookup here, you could just call `emit_event_for_endpoint([button event], 0x29)` and have it automatically be emitted by switch 1.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: Are you suggesting using an if-else conditional inside a function?

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: Yes. By this value, a device report power usage and power consumption, depending on which clusters are reported. This is a operation condition of the device.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: Good idea 👍 I updated code. Please review it. Thanks.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: I updated code. Please review it. Thanks. 😀

drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
```diff
@@ -0,0 +1,121 @@
...
+local function find_child(parent, ep_id)
+  if ep_id == 0x29 then
+    return parent:get_child_by_parent_assigned_key("01")
+  elseif ep_id == 0x2A then
+    return parent:get_child_by_parent_assigned_key("02")
+  elseif ep_id == 0x2B then
+    return parent:get_child_by_parent_assigned_key("03")
+  else
+    return parent:get_child_by_parent_assigned_key(string.format("%02X", ep_id))
+  end
```
greens, line 57:
```suggestion
  -- Buttons 1-3 report using endpoints 0x29, 0x2A, 0x2B, respectively
  if ep_id >= 0x29 then
    ep_id = ep_id - 0x28
  end
  return parent:get_child_by_parent_assigned_key(string.format("%02X", ep_id))
  end
```

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -2,63 +2,95 @@ local capabilities = require "st.capabilities"
...
+    data_type = data_types.Boolean,
+  },
+  [changeToWirelessSwitch.ID] = {
+    cluster_id = PRIVATE_CLUSTER_ID,
+    attribute_id = CHANGE_TO_WIRELESS_SWITCH_ATTRIBUTE_ID,
+    mfg_code = MFG_CODE,
+    data_type = data_types.Uint8,
+    value_map = { [true] = 0x00,[false] = 0x01 },
+  },
+  [maxPower.ID] = {
```
greens, line 56: I don't see this as a preference or capability in any of the profiles

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -2,63 +2,95 @@ local capabilities = require "st.capabilities"
...
+      ["22"] = SinglePrecisionFloat(0, 11, 0.07421875),
+      ["23"] = SinglePrecisionFloat(0, 11, 0.123046875)
+    },
+  },
+  [electricSwitchType.ID] = {
+    cluster_id = PRIVATE_CLUSTER_ID,
+    attribute_id = ELECTRIC_SWITCH_TYPE_ATTRIBUTE_ID,
+    mfg_code = MFG_CODE,
+    data_type = data_types.Uint8,
+    value_map = { ['rocker'] = 0x01,['rebound'] = 0x02 },
```
greens, line 92: If your keys are strings, they don't need to be wrapped in brackets

```suggestion
    value_map = { rocker = 0x01, rebound = 0x02 },
```

greens (commented):
All of the methods/handlers that don't do _anything_ unless private mode is enabled should be moved to a sub-driver whose `can_handle` function checks whether or not it is enabled

drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
sbkimaqara: Will do.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: Will do.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
sbkimaqara: maxPower is used in the smart plug setting and is described in the "switch-power-energy-consumption-report-aqara.yml" profile.

sbkimaqara:
> All of the methods/handlers that don't do _anything_ unless private mode is enabled should be moved to a sub-driver whose `can_handle` function checks whether or not it is enabled

Thanks for the suggestion. However, I used an if-conditional function because the multi-switch driver must include parent's methods/handlers. If sub-drivers are configured by private-mode variable value, how configure can_handle function of multi-switch sub-driver?

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
greens: that's not included in this pull request

greens:
@sbkimaqara If I understand your question correctly, the parent driver could define the handlers that do nothing if "private mode" is not enabled as no-ops, while the sub-driver defines the behavior.

sbkimaqara:
> @sbkimaqara If I understand your question correctly, the parent driver could define the handlers that do nothing if "private mode" is not enabled as no-ops, while the sub-driver defines the behavior.

I updated code. Please review it.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
greens: Ah. I see.

greens (approved):

greens (approved):

lelandblue:
Hello @sbkimaqara . There are conflicts on this PR that need to be resolved before the PR can be merged. Can you please urgently resolve those conflicts before end of day Monday? 

My understanding is that this has passed ul testing. Great news.

sbkimaqara:
> Hello @sbkimaqara . There are conflicts on this PR that need to be resolved before the PR can be merged. Can you please urgently resolve those conflicts before end of day Monday?
> 
> My understanding is that this has passed ul testing. Great news.

@lelandblue I merged. Thanks.

PR #548, opened 2023-02-09
zigbee-contact: Fix vibration/3-axis reporting for original SmartThings multi sensor

ctowns:
The original `can_handle` function was not working, so anytime one of the original SmartThings multi sensors would join, it would not correctly map to the smartsense-multi subdriver. This is due to the fact that the multisensors don't report that they support the cluster 0xFC03 (custom smartsense multi cluster), so the previous `can_handle` implementation would never return true. Now, we just check the manufacturer and the model in the `can_handle` function to ensure that the function returns true for the appropriate sensors.


ctowns:
It appears that the kickstarter multi sensors emit threeAxis and accleration events on ep2, but the `can_handle` function currently checks for that cluster on ep1. This changes `can_handle` so that it checks all endpoints to for the custom kickstarter cluster to make sure that the correct subdriver is used.

tpmanley (approved):
I successfully tested this change with a newly joined multi sensor

greens (approved):

PR #549, opened 2023-02-09
Handle cases where there is no category in duplicate profile check

ctowns:
Some profiles don't have a category (such as "thing" profiles like matter-thing.yml and zigbee-thing.yml). This addresses that by using the `get()` function to access the category in the python dictionary so that if no category is found, then `get()` will return `None` instead of a `KeyError` that would happen when using `["key"]`.


ctowns:
I thought about changing all the `[]` to `get()`, but I think the remaining options are actually required in a profile and if there is another case where a `KeyError` is raised due to a missing `component`/`capability`/etc. field, then it would probably be easier to just let the script fail with a `KeyError` and then address the unique situation then rather than trying to handle every case now.

tpmanley (approved):

greens (approved):

PR #554, opened 2023-02-13
[Aqara] Aqara Smart Dimmer Controller T1 Pro

sbkimaqara:
Aqara Smart Dimmer Controller T1 Pro

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
@@ -1,4 +1,3 @@
-local capabilities = require "st.capabilities"
 local clusters = require "st.zigbee.zcl.clusters"
 local cluster_base = require "st.zigbee.cluster_base"
 local data_types = require "st.zigbee.data_types"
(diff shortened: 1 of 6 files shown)


drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
```diff
@@ -0,0 +1,99 @@
...
+local function do_configure(self, device)
+  device:send(ColorControl.commands.MoveToColorTemperature(device, 200, 0x0000))
+  device:configure()
+  do_refresh(self, device)
+end
+
+local function device_added(driver, device, event)
+  device:emit_event(capabilities.switch.switch.on())
+  device:emit_event(capabilities.switchLevel.level(100))
+  device:emit_event(capabilities.colorTemperature.colorTemperature(5000))
```
greens, line 80: shouldn't events for all of these be generated organically by the device responding to the `refresh` command?

drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
```diff
@@ -0,0 +1,99 @@
...
+      device:send(Level.attributes.OnTransitionTime:write(device, raw_value))
+    end
+
+    if device.preferences[lightFadeOutTimeInSec.ID] ~= nil and
+        device.preferences[lightFadeOutTimeInSec.ID] ~= args.old_st_store.preferences[lightFadeOutTimeInSec.ID] then
+      local raw_value = device.preferences[lightFadeOutTimeInSec.ID] -- seconds
+      raw_value = raw_value * 10 -- unit: 100ms
+      device:send(Level.attributes.OffTransitionTime:write(device, raw_value))
+    end
+  end
```
greens, line 68: you repeat this pattern a lot, perhaps creating a mapping of all of the preferences to a function to call if the value has changed would reduce some of the boilerplate you're using in your drivers?

e.g. (in your base driver function, with switching based on model name, see https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zwave-sensor/src/preferences.lua for a similar idea):

```
LIGHT = {
  MODELS = { mfr = ..., model = ...},
  PREFERENCES = {
    lightFadeInTimeInSec = function(device, new_value) ... end,
    ...
  }
},
...
```

then you could just use one generic configure lifecycle handler across all your devices and adding devices with new preferences would be simpler.

drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
sbkimaqara: OK, I will delete the code.

drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
sbkimaqara: OK, I will update the code.

greens (commented):
Almost. I was suggesting you use the map and the `device_info_changed` function into a separate file so that all of your drivers can use the same pattern.

sbkimaqara:
Is it correct to write the code by creating a new file in "/drivers/SmartThings/zigbee-switch/src/preferences.lua" path?

greens:
Yes. I think that would work.

sbkimaqara:
> Yes. I think that would work.

OK, I updated. Please review my code.

greens (approved):

ctowns (approved):

PR #558, opened 2023-02-14
String data type handling for frequency duration in Aqara Motion Sensor 

satyajitanand:
Fix for the error:
2023-02-09_11:45:27.134 | E | edi        | <Aqara motion Sensor (0cf7d3b5)> 동작센서 thread encountered error: [string "st/dispatcher.lua"]:233: Error encountered while processing event for <ZigbeeDevice: 6de00bca-*-*-*-dbe918c4937c [0x2B86] (동작센서)>:
[string "init.lua"]:46: Timer delay must be a number

### DIFF
--- a/drivers/SmartThings/zigbee-motion-sensor/src/aqara/aqara_utils.lua
+++ b/drivers/SmartThings/zigbee-motion-sensor/src/aqara/aqara_utils.lua
@@ -24,6 +24,10 @@ local function motion_detected(device)
   end
 
   local detect_duration = device:get_field(PREF_FREQUENCY_KEY) or PREF_FREQUENCY_VALUE_DEFAULT
+  if type(detect_duration) == "string"  then
+    detect_duration = tonumber(detect_duration)
+  end
+
   local inactive_state = function()
     device:emit_event(capabilities.motionSensor.motion.inactive())
   end


amit3barthwal (commented):
Value is coming in string instead of number. Is this a violation or sensors are allowed to send in string or any other data type?

satyajitanand:
> Value is coming in string instead of number. Is this a violation or sensors are allowed to send in string or any other data type?

We are unable to check here what data the device is exactly sending for "Frequency" during error case (as we don't have that device). This commit is to handle such cases if it happens. This will have no side effect.

drivers/SmartThings/zigbee-motion-sensor/src/aqara/aqara_utils.lua
```diff
@@ -24,6 +24,10 @@ local function motion_detected(device)
   end
 
   local detect_duration = device:get_field(PREF_FREQUENCY_KEY) or PREF_FREQUENCY_VALUE_DEFAULT
+  if type(detect_duration) == "string"  then
+    detect_duration = tonumber(detect_duration)
```
naman-dixit-srib, line 28: `tonumber` can return `nil` too. If we are having issue with protocol compliance, maybe we should handle that case too, in case it sends a string that can not be converted to a number for some reason.

naman-dixit-srib (approved):

r-lanka (approved):

drivers/SmartThings/zigbee-motion-sensor/src/aqara/aqara_utils.lua
raj-1219: In case tonumber return nil, we modify line 28 as below.
```suggestion
    detect_duration = tonumber(detect_duration) or PREF_FREQUENCY_VALUE_DEFAULT
```

raj-1219 (approved):
It may be protocol compliance issue from device side. As above change will not have any side effects, it is safe to handle this scenario.

KishorSarswatSam (approved):

drivers/SmartThings/zigbee-motion-sensor/src/aqara/aqara_utils.lua
hrithik3322: This suggestion looks good

hrithik3322 (approved):
This should work

amit3barthwal (approved):

syamsidhardh (approved):
Is ok to mention the PLM no as teh commit msg, as ST do not have access to this PLM?

r-lanka:
String to number conversion looks good

drivers/SmartThings/zigbee-motion-sensor/src/aqara/aqara_utils.lua
greens: I also agree with the above suggestion.

abhijit-sahu (approved):
Now nil case also handled. Looks ok.

lelandblue:
@satyajitanand Is this still a valid PR that is needed or can we close?

lelandblue:
Closing, please see comment above.

PR #562, opened 2023-02-15
Play and Volume Delay Fix

satyajitanand:
"https://smartthings.atlassian.net/browse/CHAD-10011" -> Fix for delay in Play and Volume command.

Power On/Off command is not responsive for samsung-audio (known issue with cloud DTH case too).
Play is using power command which is causing delay in command execution. So, commented the power command.

For Volume setting, now emitting the new set volume (received in device response) back to ST app.

### DIFF
--- a/drivers/SmartThings/samsung-audio/src/handlers.lua
+++ b/drivers/SmartThings/samsung-audio/src/handlers.lua
@@ -16,6 +16,7 @@ local command = require "command"
 local log = require "log"
 local utils = require "st.utils"
 local capabilities = require "st.capabilities"
+local VOL_STEP = 5
 
 --- @module Samsung-audio.CapabilityHandlers
 local CapabilityHandlers = {}
@@ -106,7 +107,7 @@ function CapabilityHandlers.handle_volume_up(driver, device, cmd)
   local ip = device:get_field("ip")
   local vol = command.volume(ip)
   if vol then
-    local set_vol = command.set_volume(ip, tonumber(vol.volume) + 5)
+    local set_vol = command.set_volume(ip, tonumber(vol.volume) + VOL_STEP)
     if set_vol then
       device:emit_event(capabilities.audioVolume.volume(tonumber(set_vol.volume)))
       device:emit_event(capabilities.audioMute.mute.unmuted())
@@ -118,7 +119,7 @@ function CapabilityHandlers.handle_volume_down(driver, device, cmd)
   local ip = device:get_field("ip")
   local vol = command.volume(ip)
   if vol then
-    local set_vol = command.set_volume(ip, tonumber(vol.volume) - 5)
+    local set_vol = command.set_volume(ip, tonumber(vol.volume) - VOL_STEP)
     if set_vol then
       device:emit_event(capabilities.audioVolume.volume(tonumber(set_vol.volume)))
       device:emit_event(capabilities.audioMute.mute.unmuted())
--- a/drivers/SmartThings/samsung-audio/src/handlers.lua
+++ b/drivers/SmartThings/samsung-audio/src/handlers.lua
@@ -39,7 +39,7 @@ end
 
 function CapabilityHandlers.handle_play(driver, device, cmd)
   local ip = device:get_field("ip")
-  CapabilityHandlers.handle_on(driver, device, nil) --turn on if device is off
+  -- CapabilityHandlers.handle_on(driver, device, nil) --on/off is not working for samsung-audio (same issue with cloud DTH)
   local ret = command.play(ip)
   if ret then
    device:emit_event(capabilities.mediaPlayback.playbackStatus.playing())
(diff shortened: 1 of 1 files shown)


KishorSarswatSam (approved):

amit3barthwal (approved):

abhijit-sahu (approved):
Can we add your below same comment as a comment in handlers.lua..

--[[ Power On/Off command is not responsive for samsung-audio (known issue with cloud DTH case too). Play is using power command which is causing delay in command execution. So, commented the power command. For Volume setting, now emitting the new set volume (received in device response) back to ST app.--]]

hrithik3322 (approved):

naman-dixit-srib (approved):

drivers/SmartThings/samsung-audio/src/handlers.lua
```diff
@@ -39,7 +39,7 @@ end
 
 function CapabilityHandlers.handle_play(driver, device, cmd)
   local ip = device:get_field("ip")
-  CapabilityHandlers.handle_on(driver, device, nil) --turn on if device is off
+  -- CapabilityHandlers.handle_on(driver, device, nil) --on/off is not working for samsung-audio (same issue with cloud DTH)
```
greens, line 42: So the turn on command doesn't work? Can we investigate this further?

greens: Because if it really doesn't work at all, we should disable the 'on' command in the device's profile.

drivers/SmartThings/samsung-audio/src/handlers.lua
```diff
@@ -106,7 +106,10 @@ end
 
 function CapabilityHandlers.handle_set_volume(driver, device, cmd)
   local ip = device:get_field("ip")
-  command.set_volume(ip, cmd.args.volume)
+  local vol = command.set_volume(ip, cmd.args.volume)
+  if vol then
+    device:emit_event(capabilities.audioVolume.volume(tonumber(vol.volume)))
+  end
```
greens, line 112: Could this same pattern be used with other commands? Emitting the event if we get a response?

greens (commented):
While you're working on this driver, please run luacheck on it. As I discovered here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/517 there are a lot of unused imports and globals that should be local in this driver.

drivers/SmartThings/samsung-audio/src/handlers.lua
satyajitanand: > Because if it really doesn't work at all (and there's no way to fix it), we should disable the 'on' command in the device's profile.

It's the same behavior for cloud DTH case. Tester had advised the same a few weeks earlier (to disable 'switch' command), however we want to copy the cloud DTH behavior for user.  So, switch option still there. @lelandblue is aware of it.

drivers/SmartThings/samsung-audio/src/handlers.lua
satyajitanand: > Could this same pattern be used with other commands? Emitting the event if we get a response?

Done

amit3barthwal (approved):
approve - unmuting during volume up/down.

r-lanka (approved):

KishorSarswatSam (approved):

syamsidhardh (approved):
Changes looks ok.

raj-1219 (approved):

hrithik3322 (approved):
This fixes the volume delay issue

naman-dixit-srib (approved):

drivers/SmartThings/samsung-audio/src/handlers.lua
```diff
@@ -80,33 +80,59 @@ end
...
+    end
+  end
 end
 
 function CapabilityHandlers.handle_volume_up(driver, device, cmd)
   local ip = device:get_field("ip")
   local vol = command.volume(ip)
   if vol then
-    command.set_volume(ip, vol.volume + 5)
+    local set_vol = command.set_volume(ip, tonumber(vol.volume) + 5)
```
greens, line 109: sorry, nit, but can you make this volume delta a constant defined at the top of the file?

`local VOLUME_STEP = 5` or something like that?

r-lanka:
LGTM

drivers/SmartThings/samsung-audio/src/handlers.lua
satyajitanand: > sorry, nit, but can you make this volume delta a constant defined at the top of the file?
> 
> `local VOLUME_STEP = 5` or something like that?

Done.

satyajitanand:
@greens : Can you please help to review.

amit3barthwal (approved):
looks good to me for this variable VOL_STEP set.

abhijit-sahu (approved):
Volume up started value is 5.

abhijit-sahu (approved):
emit_event are handled for all has done for handlers.lua.

greens (approved):

satyajitanand:
@lelandblue : @greens has approved. Should this PR be merged?

dljsjr (approved):

PR #567, opened 2023-02-15
Refresh of lock with codes causes too many events

greens:
This brings behavior back in line with the DTH.


ctowns (approved):

varzac (approved):
Do you want to make a ticket to update the defaults for lockCodes as well to not call `reloadAllCodes` on refresh?

PR #573, opened 2023-02-16
CHAD-10232 Handle overlap of 0/360 degrees in HSV color space

greens:
Our color conversions don't properly wrap the value of 100% hue (360 degrees) around to be equivalent to 0% hue (0 degrees). This workaround does that wrapping, and reverses it when we receive a device update.


cjswedes (approved):
It'd be nice if we could get the wraparound logic into the conversion functions, but I dont really see a clean way to do it, and the wraparound case will always have to be exposed to the driver anyway to handle emitting the correct events.

sjl2178 (approved):

dljsjr (approved):
I took the liberty of rebasing this to get it up to date and testing it on my hub, and it does in fact seem to address the angle wrapping issue for hue on Hue! Thanks @greens!

PR #575, opened 2023-02-17
Align mediaGroup capabilites & CmdHandlers

TheMegamind:
mediaGroup capabilities & CmdHandlers are misaligned

### DIFF
--- a/drivers/SmartThings/bose/src/init.lua
+++ b/drivers/SmartThings/bose/src/init.lua
@@ -198,6 +198,14 @@ local function device_init(driver, device)
   device:set_field("init_started", true)
   device.log.info_with({ hub_logs = true }, "initializing device")
   local serial_number = bose_utils.get_serial_number(device)
+  -- Carry over DTH discovered ip during migration to enable some communication
+  -- in cases where it takes a long time to rediscover the device on the LAN.
+  if not device:get_field("ip") and device.data and device.data.ip then
+    local nu = require "st.net_utils"
+    local ip = nu.convert_ipv4_hex_to_dotted_decimal(device.data.ip)
+    device:set_field("ip", ip, { persist = true })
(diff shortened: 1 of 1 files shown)


TheMegamind:
Is anybody available to review this PR?  While this is one of at least three significant issues with the Edge integration, this one is the most obvious and should be the easiest to fix.  Perhaps @dljsjr ?

dljsjr:
@TheMegamind Thanks for the PR, I'll take a look tomorrow. I was out of the office last week and still getting spun back up. I think we identified this bug internally already as well so should be easy enough to verify the fix.

TheMegamind:
Thanks, I've got a related project I'm working on, and this issue has essentially brought things to a halt. :)

There's also a [PR to include the `replace` action with](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/577) the playPreset command if you wouldn't mind looking at that as well.

dljsjr:
@TheMegamind Sorry for the delay, I needed to investigate the `replace` action change from your other PR as it is present here as well. I provided feedback there on that change.

That said, if you were to remove that change from this PR and squash your commits down, the actual bugfix itself looks good to me.

dljsjr (changes requested):
@TheMegamind Sorry for the delay, I needed to investigate the `replace` action change from your other PR as it is present here as well. I provided feedback there on that change.

That said, if you were to remove that change from this PR and squash your commits down, the actual bugfix itself looks good to me.

TheMegamind:
@dljsjr Ok, I knew I made a new kind of mess when I added the "replace" PR, but was traveling and couldn't remember how to undo the damage. I haven't done that in a while, but will figure it out.

TheMegamind:
@dljsjr Lol, called in the local cavalry to help me fix my mess and left two guys who could build rocket ships scratching their heads. Maybe we should just close 575 and let you do it the right way.

PR #576, opened 2023-02-17
Bose discovery hardening

cjswedes:
Harden the SSDP header regex to handle more valid HTTP headers. Also provide more explicit logging for the various discovery conditions.

Spawn device init as a cosock task in case it takes a long time, so the device thread is not blocked.

Device added and device init result in initialization code running once to account for existing lan bugs where sometimes init is not run when newly joined devices are added.

Make disco loop more robust to avoid overuse of sockets when multiple devices are being discovered/migrated at one time. This will follow the pattern I used for Wemo where the disco loop can service multiple discovery requests from the driver (due to init or scan nearby) with a single multicast discovery request.

Refresh should restart the websocket listener if it needs to be. This is a defensive change that might help users resolve state update issues if they are occurring, but I have yet to see it be needed in the field.

### DIFF
--- a/drivers/SmartThings/bose/src/disco.lua
+++ b/drivers/SmartThings/bose/src/disco.lua
@@ -9,95 +9,193 @@
 --  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
 --  either express or implied. See the License for the specific language governing permissions
 --  and limitations under the License.
---
 
 local cosock = require "cosock"
 local socket = require "cosock.socket"
-local http = cosock.asyncify "socket.http" -- TODO use luncheon instead
-local ltn12 = require "ltn12"
 local log = require "log"
-local xml2lua = require "xml2lua"
-local xml_handler = require "xmlhandler.tree"
 local utils = require "st.utils"
-local command = require "command"
 
---- @module bose.Disco
-local Disco = {}
+local ControlMessageTypes = {
+  Scan = "scan",
+  FindDevice = "findDevice",
+}
+
+local ControlMessageBuilders = {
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/bose/src/init.lua
```diff
@@ -233,6 +240,7 @@ local bose = Driver("bose", {
     init = device_init,
     removed = device_removed,
     infoChanged = info_changed,
+    added = device_added,
```
greens, line 243: why not just `added = device_init`?

drivers/SmartThings/bose/src/init.lua
```diff
@@ -147,71 +148,77 @@ local function do_refresh(driver, device, cmd)
...
-    local randval = 0
-    if rand then
-      -- random value in range (-rand, rand)
-      randval = math.random() * rand * 2 - rand
+  -- restart listener if needed
+  local listener = device:get_field("listener")
+  local success = false
+  if listener and (listener:is_stopped() or listener.websocket == nil)then
+    device.log.info("Restarting listening websocket client for device updates")
+    listener:stop()
```
greens, line 161: does this do anything if `is_stopped` is already true or the `websocket` is `nil`?

drivers/SmartThings/bose/src/disco.lua
```diff
@@ -9,95 +9,193 @@
...
       end
-      break
-    else
-      error(string.format("[%s]error receving discovery replies: %s", deviceid, rip))
     end
   end
-  s:close()
 end
 
-return Disco
```
greens, line 103: Disco is dead.

greens (approved):

PR #577, opened 2023-02-17
Include "replace" action to playPreset Command

TheMegamind:
In the absence of the `replace` parameter shown here, if a user selects a Favorite that is a playlist (as opposed to a station), the `playPreset` command _appends_ the Favorite selection to the end of any existing queue rather than replacing that queue as expected. This not only causes the queue to continue growing in size until manually cleared by the user via the Sonos app or Desktop interface, but if "shuffle" is enabled, the intended playlist is shuffled into the larger queue and does not play directly in response to the request.    

**Note**: The action:replace and shuffle toggle only affect playlists, not stations.  While I also believe the option to toggle "shuffle" on and off for playlists (see payload below) should also be present, lua is not my native language and therefore I'm less confident in my ability to submit a proper pull request. 

```
{
  "favoriteId" : "223",
  "playOnCompletion" : true,
  "action" : "replace"
  },
  "playModes" : {
    "shuffle" : false
  }
}
```

### DIFF
--- a/drivers/SmartThings/sonos/src/api/cmd_handlers.lua
+++ b/drivers/SmartThings/sonos/src/api/cmd_handlers.lua
@@ -151,7 +151,8 @@ function CapCommandHandlers.handle_play_preset(driver, device, cmd)
     { namespace = "favorites", command = "loadFavorite" },
     {
       favoriteId = cmd.args.presetId,
-      playOnCompletion = true
+      playOnCompletion = true,
+      action = "replace"
     }
   }
   _do_send_to_group(driver, device, payload)
--- a/drivers/SmartThings/sonos/src/api/cmd_handlers.lua
+++ b/drivers/SmartThings/sonos/src/api/cmd_handlers.lua
@@ -151,8 +151,7 @@ function CapCommandHandlers.handle_play_preset(driver, device, cmd)
     { namespace = "favorites", command = "loadFavorite" },
     {
       favoriteId = cmd.args.presetId,
-      playOnCompletion = true,
-      action = "replace"
+      playOnCompletion = true
     }
   }
   _do_send_to_group(driver, device, payload)


dljsjr:
@TheMegamind This should definitely be the preferred behavior, but I looked into what the behavior of our existing integration uses, and it behaves like this as well. We're working our hardest to avoid introducing breaking changes for folks who migrate existing devices over to Edge Drivers, so we won't be able to pull this in as-is.

I think what we'd look at here is introducing a preference that defaults to the existing behavior so that folks can choose how they want their system to behave.

TheMegamind:
@dljsjr A preference, as long as it is exposed within the API, would be fine.  I suspected you might go that direction because of the potential for "breakage,", even though "replace" is what I'm certain the vast majority of users would prefer.   

When you do this, please consider adding the preference for "shuffle" as well.  Thx.

dljsjr:
>A preference, as long as it is exposed within the API, would be fine

Preferences are indeed part of the API, if you're a https://github.com/SmartThingsCommunity/smartthings-cli user you can get a feel for the shape of that endpoint's responses under the `devices:preferences` command.

TheMegamind:
Thanks. I assumed as much, but mentioned it only because while playText is available for Sonos within routines (albeit without a volume preference, unfortunately) it is not available via the API and thus Rules, which I prefer over routines.  It just made me uncertain how or why that happens, although there may be a good reason or it.  

With regard to "Shuffle" and "Replace," I’m not sufficiently versed in lua to feel comfortable posting a PR. Any recommendations on moving this forward?

dljsjr:
There are actually 4 possible queue actions that the Sonos speakers support. I'll probably generate a new PR myself, and I'll make sure to include you on the reviewer list, so you can see how to integrate preferences with an Edge Driver in Lua.

TheMegamind:
@dljsjr Thanks, I appreciate being included as a reviewer. I am interested to see how this done.

dljsjr:
Closing in favor of https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/600

PR #592, opened 2023-02-27
Add special handling for fractional endpoints

ctowns:
This is a workaround for handling fractional setpoint in zwave thermostats while a more permanent fix is added to the lua libs. Fractional values such as 21.5 are expected to be represented with an int value and a corresponding precision value such that the temperature = value * 10^(-precision).

### DIFF
--- a/drivers/SmartThings/zwave-thermostat/src/init.lua
+++ b/drivers/SmartThings/zwave-thermostat/src/init.lua
@@ -56,11 +56,27 @@ local function set_setpoint_factory(setpoint_type)
     local scale = device:get_field(constants.TEMPERATURE_SCALE)
     local value = convert_to_device_temp(command.args.setpoint, scale)
 
-    local set = ThermostatSetpoint:Set({
-      setpoint_type = setpoint_type,
-      scale = scale,
-      value = value
-    })
+    -- Zwave thermostat devices expect to get fractional values as an integer value
+    -- with a provided precision such that the temp is value * 10^(-precision)
+    -- See section 2.2.113.2 of the Zwave Specification for more info
+    -- This is a temporary workaround for the Aeotec Thermostat device while
+    -- more permanent fixes are added to scripting-engine
+    local set
+    if value % 1 == 0.5 then
+      set = ThermostatSetpoint:Set({
+        setpoint_type = setpoint_type,
+        scale = scale,
+        value = value,
+        precision = 1,
+        size = 2
+      })
+    else
+      set = ThermostatSetpoint:Set({
+        setpoint_type = setpoint_type,
+        scale = scale,
+        value = value
+      })
+    end
     device:send_to_component(set, command.component)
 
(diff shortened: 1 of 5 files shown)


drivers/SmartThings/zwave-thermostat/src/aeotec-radiator-thermostat/init.lua
```diff
@@ -97,6 +99,52 @@ local function device_added(self, device)
...
+local function set_setpoint_factory(setpoint_type)
+  return function(driver, device, command)
+    local scale = device:get_field(constants.TEMPERATURE_SCALE)
+    local value = convert_to_device_temp(command.args.setpoint, scale)
+
+    local set
+    -- This device expects to get fractional values as a uint16 with
+    -- a provided "precision" of 1 to indicate that the integer should
+    -- be divided by 10. i.e. the value of 21.5 is 215 with precision 1
+    if value % 0.5 == 0 then
```
greens, line 122: Isn't this true for whole numbers too? I think you probably want `value % 1 == 0.5`, right?

ctowns:
@greens so this device expects fractional temperatures to be expressed as 2 bytes as explained above. However, the code we that is generated for the lua libs will always interpret a float value as 4 bytes when serializing the data for the set ThermostatSetpoint command unless a size value is provide in the args. So I guess my question is, is there a scripting-engine change that should be made? I guess a floating point should really by 4 bytes by default, and this works as expected (at least in unit tests so far) if I manually provide the precision/size. What do you think?

drivers/SmartThings/zwave-thermostat/src/aeotec-radiator-thermostat/init.lua
ctowns: Whoops, yes you are right!

drivers/SmartThings/zwave-thermostat/src/aeotec-radiator-thermostat/init.lua
```diff
@@ -97,6 +99,52 @@ local function device_added(self, device)
...
+    -- a provided "precision" of 1 to indicate that the integer should
+    -- be divided by 10. i.e. the value of 21.5 is 215 with precision 1
+    if value % 1 == 0.5 then
+      set = ThermostatSetpoint:Set({
+        setpoint_type = setpoint_type,
+        scale = scale,
+        value = value,
+        precision = 1,
+        size = 2
+      })
```
greens, line 129: I'd honestly include this in the base driver's method, since this seems like an error we'd want to address everywhere

ctowns:
@greens based on these units tests, it seems like there are other thermostats that can handle fractional setpoint values. However, none of them have the required embedded "step" configs to allow fractional values and none of them have been marked by QA for not handling this functionality. I will put up a separate PR to add step configs to the other thermostats where it seems applicable

greens (approved):

PR #597, opened 2023-03-02
Bose: use DTH discovered IP upon first migration

cjswedes:
This mitigates situations where it takes a long time for a migrated device to be discovered on the LAN. It is possible the migrated ip is not accurate, but this is unlikely and will be corrected once we do rediscover the device on the LAN.

### DIFF
--- a/drivers/SmartThings/bose/src/init.lua
+++ b/drivers/SmartThings/bose/src/init.lua
@@ -198,6 +198,14 @@ local function device_init(driver, device)
   device:set_field("init_started", true)
   device.log.info_with({ hub_logs = true }, "initializing device")
   local serial_number = bose_utils.get_serial_number(device)
+  -- Carry over DTH discovered ip during migration to enable some communication
+  -- in cases where it takes a long time to rediscover the device on the LAN.
+  if not device:get_field("ip") and device.data and device.data.ip then
+    local nu = require "st.net_utils"
+    local ip = nu.convert_ipv4_hex_to_dotted_decimal(device.data.ip)
+    device:set_field("ip", ip, { persist = true })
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/bose/src/init.lua
```diff
@@ -198,6 +198,14 @@ local function device_init(driver, device)
   device:set_field("init_started", true)
   device.log.info_with({ hub_logs = true }, "initializing device")
   local serial_number = bose_utils.get_serial_number(device)
+  -- Carry over DTH discovered ip during migration to enable some communication
+  -- in cases where it takes a long time to rediscover the device on the LAN.
+  if not device:get_field("ip") and device.data and device.data.ip then
+    local nu = require "st.net_utils"
+    local ip = nu.convert_ipv4_hex_to_dotted_decimal(device.data.ip)
+    device:set_field("ip", ip, { persist = true })
```
sjl2178, line 206: So in this case the ip address field is not set but this will catch it and set the ip address for the device?

sjl2178 (approved):

drivers/SmartThings/bose/src/init.lua
cjswedes: Yeah, when a device is first added to the driver and initialized it will not have an ip address (even the try_create_device from our disco loop will add the device without an ip and the init handler will rediscover the ip). This detects if the device is being added due to a migration (it has `device.data`) in which case we know it has the ip field populated in device data.

greens (approved):

ctowns (approved):

dljsjr (approved):

PR #600, opened 2023-03-04
[Sonos] Add preference for the load faves queue behavior

dljsjr:
This is a PR to address a separate community PR that wanted to shore up behavior for loading favorites to a Sonos queue: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/577

The linked PR definitely presents a better take on what the default behavior should be, however, the existing integration we provide uses the speaker's default behavior which is equivalent to the `Append` preference in this PR.

cc @TheMegamind

### DIFF
--- a/drivers/SmartThings/sonos/src/api/cmd_handlers.lua
+++ b/drivers/SmartThings/sonos/src/api/cmd_handlers.lua
@@ -9,6 +9,8 @@ local st_utils = require "st.utils"
 --- https://developer.sonos.com/reference/control-api-examples-lan/
 local CapCommandHandlers = {}
 
+local QUEUE_ACTION_PREF = "queueAction"
+
 local function _do_send(device, payload)
   local conn = device:get_field(PlayerFields.CONNECTION)
   if conn and conn:is_running() then
@@ -151,7 +153,8 @@ function CapCommandHandlers.handle_play_preset(driver, device, cmd)
     { namespace = "favorites", command = "loadFavorite" },
     {
       favoriteId = cmd.args.presetId,
-      playOnCompletion = true
+      playOnCompletion = true,
+      action = (device.preferences[QUEUE_ACTION_PREF] or "APPEND")
     }
   }
   _do_send_to_group(driver, device, payload)
--- a/drivers/SmartThings/sonos/profiles/sonos-player.yml
+++ b/drivers/SmartThings/sonos/profiles/sonos-player.yml
@@ -29,3 +29,16 @@ components:
     version: 1
   categories:
   - name: Speaker
+preferences:
+- name: "queueAction"
+  title: "\"Load Favorites\" Queuing Action"
+  description: Determines the playback behavior when a Favorite is loaded while content is already queued on the speaker.
+  required: false
+  preferenceType: enumeration
+  definition:
+    options:
+      Append: "Add to End of Queue"
+      Insert: "Play Now"
+      Insert_Next: "Play Next"
+      Replace: "Replace Queue"
+    default: Append


drivers/SmartThings/sonos/src/api/cmd_handlers.lua
```diff
@@ -9,6 +9,8 @@ local st_utils = require "st.utils"
 --- https://developer.sonos.com/reference/control-api-examples-lan/
 local CapCommandHandlers = {}
 
+local QUEUE_ACTION_PREF = "queueAction"
```
dljsjr, line 12: @TheMegamind The current preference value is stored on the device under the `preferences` table. The `name` defined in the .yml is the key. **The Lua value** will be **The YAML Key**. The YAML Value is used as the Display Name for the preference in the mobile phone app.

TheMegamind: @dljsjr Interesting. Different but less complicated than I expected.

TheMegamind:
@dljsjr Can we expand this to include the related playModes preferences [shuffle, crossfade, repeat, repeat-one] as well? Those are very commonly applied preferences, and adding them would complete the playback behavior options for favorites.

drivers/SmartThings/sonos/src/api/cmd_handlers.lua
```diff
@@ -151,7 +153,8 @@ function CapCommandHandlers.handle_play_preset(driver, device, cmd)
     { namespace = "favorites", command = "loadFavorite" },
     {
       favoriteId = cmd.args.presetId,
-      playOnCompletion = true
+      playOnCompletion = true,
+      action = (device.preferences[QUEUE_ACTION_PREF] or "APPEND")
```
cjswedes, line 157: Is there capitalization that needs to be accounted for when using the ST preference value? Looks like the default `"APPEND"` is gonna be different that the actual table value.

dljsjr: The API itself is case-insensitive on stringly-defined types.

cjswedes (approved):

TheMegamind:
@dljsjr I think it would be beneficial to have the playMode preferences added, but understand the need to get to a certain baseline first. As it is, this new integration is unable to maintain reliable connections, which often render the speakers unresponsive to commands from the app or the API.  Also, some commands, such as setVolume, are recognized by the group coordinator but ignored by auxiliary speakers. Whether those are backend or driver issues is well above my pay grade, of course.  Another gentleman from the developer support team was supposed to be looking into all that, but I haven't heard back in a while.  

In any event, I just figured I'd take a crack at enhancing the driver's capabilities in the meantime. The one here is a nice start. Thanks.

TheMegamind:
Nothing better than having someone who uses Sonos trying to figure it out!  

I can find the name of the developer support who was supposedly working on it if it makes any difference.

dljsjr:
I've already done the follow-up there, thanks though!

TheMegamind:
> > As it is, this new integration is unable to maintain reliable connections, which often render the speakers unresponsive to commands from the app or the API
> 
> Yeah, I'm seeing the same thing on my speakers at home. I'm trying to figure it out myself :)

Hi Douglas.  Sorry for the intrusion here, but I didn't know how else to message you.  Can you tell me if any progress is being made with this issue?  Now that the driver has been released out of beta and everyone has been migrated, the user complaints are piling up fast. Like many others, I've got a very active Sonos system, and I've had to move all my Sonos-related automations out of SmartThings.  

I've been tracking the websocket dropouts—or whatever the right term for the problem  is—by comparing events reported by the UPnP service and SmartThings and mine drops off every day, sometimes multiple times, and in fact dropped off again about 40 minutes ago. I've also noted that when it "reconnects," usually through outside intervention, a flood of old events is reported as if they just happened. I don't know if that helps or just adds to the confusion, but if there's anyway I can contribute (hah!), I'd be happy to try.

PR #604, opened 2023-03-07
edge drivers added for ROB_200-007-0 and ROB_200-007-0

zwave-mke:


drivers/SmartThings/zigbee-button/src/zigbee-multi-button/init.lua
```diff
@@ -57,7 +59,7 @@ local function added_handler(self, device)
     end
     device:emit_component_event(component, capabilities.button.numberOfButtons({value = number_of_buttons}, {visibility = { displayed = false }}))
   end
-  -- device:emit_event(capabilities.button.button.pushed({state_change = false}))
+  device:emit_event(capabilities.button.button.pushed({state_change = false}))
```
greens, line 62: This was intentionally commented out for the migration period and needs to stay that way for now, sorry.

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
```diff
@@ -0,0 +1,183 @@
...
+  local number_of_buttons
+  local comp_id
+
+  for _, comp in pairs(device.profile.components) do
+    comp_id = comp.id
+    -- Special handeling of ON buttons (left column)
+    if comp_id == "button1" or comp_id == "button3" or comp_id == "button5" or comp_id == "button7" then
+      supported_button_values = { "pushed", "up_hold" }
+      number_of_buttons = 1
+    -- Special handeling of OFF buttons (right column)
```
greens, line 143: nit: typo: `handling`

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
```diff
@@ -0,0 +1,183 @@
...
+  [0x03] = "button6",
+  [0x04] = "button8"
+}
+
+local build_button_handler = function(driver, device, zb_rx)
+    local button_name
+    local additional_fields = { state_change = true }
+
+    -- Fetch correct button name
+    if zb_rx.body.zcl_header.cmd.value == 0x01 then
```
greens, line 62: So, if you'll notice in the `button_utils` function you've based this off of, it's a function _factory_. I expect what you want to do is also have a factory, but pass in the correct component map to be used.

But, if you wanted to leverage the existing systems a bit more, I might suggest setting the `component_to_endpoint` function up like this:

```
function endpoint_to_component(device, ep)
local EP_MAP = {
  [0x01] = "button1",
  [0x02] = "button3",
  [0x03] = "button5",
  [0x04] = "button7",
  [0xF1] = "button2",
  [0xF2] = "button4",
  [0xF3] = "button6",
  [0xF4] = "button8"
}
return device.profile.components[EP_MAP[ep]]
end
```
and then, make `build_button_handler` actually a factory that takes one flag `addF` that is a boolean used to determine whether you add `0xF0` to the value of `zb_rx.address_header.src_endpoint.value` before calling `device:emit_event_for_endpoint()` or not.

so your handlers would look like:

```
      [OnOff.ID] = {
        [OnOff.server.commands.Off.ID] = build_button_handler(true),
        [OnOff.server.commands.On.ID] = build_button_handler(false)
      }
```

does that make sense?

Another option is to leave the function largely as is, but eliminate the maps and just use either 

`"switch"..(zb_rx.address_header.src_endpoint.value * 2)`

or

`"switch"..(zb_rx.address_header.src_endpoint.value * 2 - 1)`

depending on which is relevant

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/supported_values.lua
```diff
@@ -17,30 +17,38 @@ local devices = {
...
+    SUPPORTED_BUTTON_VALUES = { "pushed", "held" },
     NUMBER_OF_BUTTONS = 4
   },
+  BUTTON_PUSH_HELD_8 = {
+    MATCHING_MATRIX = {
+      { mfr = "ROBB smarrt", model = "ROB_200-007-0" },
+    },
+    SUPPORTED_BUTTON_VALUES = { "pushed", "held" },
+    NUMBER_OF_BUTTONS = 8
+  },
```
greens, line 38: since you're not using any of the code you added here (and it lists the wrong events anyway (held instead of up_held and down_held) please remove it

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
```diff
@@ -0,0 +1,183 @@
...
+
+local SWITCH_GROUP_CONFIGURE = "is_group_configured"
+local SWITCH8_NUM_ENDPOINT = 4
+local SWITCH8_NUM_BUTTONS = 8
+local SWITCH4_NUM_ENDPOINT = 2
+local SWITCH4_NUM_BUTTONS = 4
+
+local WIRELESS_REMOTE_FINGERPRINTS = {
+  { mfr = "ROBB smarrt", model = "ROB_200-007-0" },
+  { mfr = "ROBB smarrt", model = "ROB_200-008-0" }
```
greens, line 29: I would restructure this map to be more useful, and change the `CAN_HANDLE` appropriately.

```
local ROBB_MFR_STRING = "ROBB smarrt"
local WIRELESS_REMOTE_FINGERPRINTS = {
  "ROB_200-007-0" = { endpoints = 2, buttons = 4},
  "ROB_200-008-0" = { endpoints = 4, buttons = 8}
}
```

this way, in other places where you're doing this:

`number_of_buttons = device:get_model() == "ROB_200-007-0" and SWITCH8_NUM_BUTTONS or SWITCH4_NUM_BUTTONS`

you can just do

`number_of_buttons = WIRELESS_REMOTE_FINGERPRINTS[device:get_model()].buttons`

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
```diff
@@ -0,0 +1,183 @@
...
+
+  for endpoint = 1, endpoints do
+    device:send(device_management.build_bind_request(device, Level.ID, driver.environment_info.hub_zigbee_eui, endpoint))
+    device:send(device_management.build_bind_request(device, OnOff.ID, driver.environment_info.hub_zigbee_eui, endpoint))
+  end
+
+  device:send(OnOff.attributes.OnOff:configure_reporting(device, 0, 600, 1))
+
+  device:send(Basic.attributes.DeviceEnabled:write(device, true))
+  if not driver.datastore[SWITCH_GROUP_CONFIGURE] then
```
greens, line 119: We prefer you use `device:set_field` and `device:get_field` here rather than access the datastore directly.

You may also want to specify `{persist = true}` if you'd like those changes to be durable.

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
zwave-mke: solved with 714d055

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
zwave-mke: solved with 714d055

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/supported_values.lua
zwave-mke: solved with 714d055

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
zwave-mke: solved with 714d055

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
zwave-mke: solved with 714d055

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
```diff
@@ -0,0 +1,226 @@
...
+    ["button6"] = 3,
+    ["button8"] = 4
+  }
+
+  if COMP_MAP[component_id] ~= nil then
+    return COMP_MAP[component_id]
+  else
+    return 1
+  end
+end
```
greens, line 173: I think this is unused? Since you're not receiving any commands that have to be sent to a specific component, you don't need this mapping. This is for things like switches where you need to be able to translate "turn switch1 on" to "send on command to endpoint 2" and such

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
```diff
@@ -0,0 +1,226 @@
...
+      capabilities.button.supportedButtonValues(supported_button_values, { visibility = { displayed = false } }))
+    device:emit_component_event(comp,
+      capabilities.button.numberOfButtons({ value = number_of_buttons }, { visibility = { displayed = false } }))
+  end
+  device:emit_event(capabilities.button.button.pushed({ state_change = false }))
+  device:send(PowerConfiguration.attributes.BatteryPercentageRemaining:read(device))
+end
+
+local battery_perc_attr_handler = function(driver, device, value, zb_rx)
+  device:emit_event(capabilities.battery.battery(utils.clamp_value(value.value, 0, 100)))
```
greens, line 152: the zigbee spec says this attribute should be reported in half percents, i.e. 200 for 100% battery. Are you reporting whole percents? just curious because you're clamping to 100, implying you're getting values higher than 100

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
```diff
@@ -0,0 +1,226 @@
...
+
+    local event = capabilities.button.button.pushed(additional_fields)
+    local comp = device.profile.components[comp.id]
+    -- Emit events
+    if comp ~= nil then
+      device:emit_component_event(comp, event)
+      if comp.id ~= "main" then
+        device:emit_event(event)
+      end
+    end
```
greens, line 63:
```suggestion
    local offset = 0
    if addF == true then
      offset = 0xF0
    end

    local event = capabilities.button.button.pushed(additional_fields)
    device:emit_event_for_endpoint(ep + offset, event)
    device:emit_component_event("main", event)
```

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
```diff
@@ -0,0 +1,226 @@
...
+  local additional_fields = { state_change = true }
+  local ep = zb_rx.address_header.src_endpoint.value
+
+  -- Handle MoveStepMode.UP
+  if zcl_clusters.Level.types.MoveStepMode.UP == zb_rx.body.zcl_body.move_mode.value then
+    comp = device:get_component_id_for_endpoint(ep)
+    pressed_type = "up_hold"
+  -- Handle MoveStepMode.DOWN
+  else
+    comp = device:get_component_id_for_endpoint(ep + 240)
```
greens, line 79: use the same `offset` and `emit_event_for_endpoint` pattern as above

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/supported_values.lua
```diff
@@ -17,30 +17,31 @@ local devices = {
...
-    SUPPORTED_BUTTON_VALUES = {"pushed", "held"},
+    SUPPORTED_BUTTON_VALUES = { "pushed", "held" },
     NUMBER_OF_BUTTONS = 2
   },
   BUTTON_PUSH_HELD_4 = {
     MATCHING_MATRIX = {
       { mfr = "CentraLite", model = "3450-L" },
-      { mfr = "CentraLite", model = "3450-L2" }
+      { mfr = "CentraLite", model = "3450-L2" },
+      { mfr = "ROBB smarrt", model = "ROB_200-008-0" }
```
greens, line 27: remove

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
zwave-mke: You're right it's not necessary. It was added to face some upcoming battery trouble after adding endpoint_to_component() but now it is not used anymore. Removed with [8fb92ba](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/604/commits/8fb92ba337ac7bcc1e11c99e0dd85f5e5295cac0)

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
zwave-mke: Yes, you're right. The spec says 0 to 200 in 0.5 % - sorry for that. I just wonder if it is the same case [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-button/src/zigbee-multi-button/ikea/init.lua#L108) and [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-dimmer-remote/src/zigbee-battery-accessory-dimmer/IKEAofSweden/init.lua#L99) ?

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
zwave-mke: Changed with [8fb92ba](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/604/commits/8fb92ba337ac7bcc1e11c99e0dd85f5e5295cac0)

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
zwave-mke: Changed with [8fb92ba](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/604/commits/8fb92ba337ac7bcc1e11c99e0dd85f5e5295cac0)

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/supported_values.lua
zwave-mke: Removed with [8fb92ba](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/604/commits/8fb92ba337ac7bcc1e11c99e0dd85f5e5295cac0)

greens (approved):

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/robb/init.lua
```diff
@@ -0,0 +1,191 @@
...
+    device:emit_component_event(comp,
+      capabilities.button.supportedButtonValues(supported_button_values, { visibility = { displayed = false } }))
+    device:emit_component_event(comp,
+      capabilities.button.numberOfButtons({ value = number_of_buttons }, { visibility = { displayed = false } }))
+  end
+  device:emit_event(capabilities.button.button.pushed({ state_change = false }))
+  device:send(PowerConfiguration.attributes.BatteryPercentageRemaining:read(device))
+end
+
+local battery_perc_attr_handler = function(driver, device, value, zb_rx)
```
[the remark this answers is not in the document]
zwave-mke: I have now tested both devices again with a battery dummy, where I can set the voltage.
I am really sure that the device does not follow the Zigbee specification and instead reports whole percentages.
If I implement it like this "value.value / 2.0 + 0.5" it starts with 50% at the maximum battery voltage of 3V.

The capability handler is a workaround to show battery capacity only in the main component. For the 4-button remote, the battery level is reported via EP2, for the 8-button remote via EP4. The default battery capability doesn't seem to handle this correctly and expects a capability in one of the other components 3 or 7, but UX technically it doesn't make sense there.

```2023-05-03T05:55:59.373744354+00:00 INFO Zigbee Button  <ZigbeeDevice: 557b9977-fcd4-4883-8578-7b0d7c7fc8d9 [0xF93B] (ROBB 8x Wireless Smarrt Switch)> received Zigbee message: < ZigbeeMessageRx || type: 0x02, < AddressHeader || src_addr: 0xF93B, src_endpoint: 0x04, dest_addr: 0x0000, dest_endpoint: 0xFF, profile: 0x0104, cluster: PowerConfiguration >, lqi: 0xC6, rssi: -72, body_length: 0x0007, < ZCLMessageBody || < ZCLHeader || frame_ctrl: 0x08, seqno: 0x7B, ZCLCommandId: 0x0A >, < ReportAttribute || < AttributeRecord || AttributeId: 0x0021, DataType: Uint8, BatteryPercentageRemaining: 0x64 > > > >
2023-05-03T05:55:59.393409021+00:00 TRACE Zigbee Button  Found ZigbeeMessageDispatcher handler in zigbee_button
2023-05-03T05:55:59.396347687+00:00 INFO Zigbee Button  Executing ZclClusterAttributeValueHandler: cluster: PowerConfiguration, attribute: BatteryPercentageRemaining
2023-05-03T05:55:59.404658354+00:00 WARN Zigbee Button  Attempted to generate event for 557b9977-fcd4-4883-8578-7b0d7c7fc8d9.button7 but it does not support capability Battery
```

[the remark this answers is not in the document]
ctowns: @zwave-mke based on the results from the cert testing, it seems like the device is sending `200 (0xC8)` when reporting 100% battery. This is what we would expect from a device that follows the zigbee spec, so it seems this device is following the zigbee spec. Based on the testing, using the `capabilities.battery.battery(math.floor(value.value / 2.0 + 0.5))` equation should fix the reporting to be 0-100%. 

You are correct that you need to override the default handler because of the unique endpoint mapping on this device. So, you should still have the `battery_perc_attr_handler`, but it should look like this with the new equation that divides the value by 2:

```
local battery_perc_attr_handler = function(driver, device, value, zb_rx)
  device:emit_event(capabilities.battery.battery(math.floor(value.value / 2.0 + 0.5)))
end
```

> If I implement it like this "value.value / 2.0 + 0.5" it starts with 50% at the maximum battery voltage of 3V.
I am wondering if there might have been an issue with the test setup? Based on the results from the cert testing, the device was reporting 200% when the battery was at 100%. Please let me know if you have further questions, but based on the cert test results we believe that the suggested fix above should address the battery reporting issue

PR #610, opened 2023-03-08
HCS-4126 Correct several issues with Danfoss thermostat presentation

greens:


sjl2178 (approved):

ctowns (approved):
It looks like these are both EU thermostats again - should we assume all EU thermostats should have the 0.5 step?

greens:
@ctowns Yes, but I think that the radiator thermostats might be the only ones.

ctowns (approved):

PR #613, opened 2023-03-09
Add zigbee-thermostat for Vimar 02973 product

maxelweb-vimar:
Add support with a custom sub-driver for:

- Vimar Smart Thermostat (02973)


drivers/SmartThings/zigbee-thermostat/src/vimar/init.lua
```diff
@@ -0,0 +1,79 @@
...
+}
+
+
+local vimar_thermostat_can_handle = function(opts, driver, device)
+  for _, fingerprint in ipairs(VIMAR_THERMOSTAT_FINGERPRINTS) do
+    if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
+        return true
+    end
+  end
+  return false
```
greens, line 40: just a nit, but since this sub-driver only has one FP, you don't need to loop

you can just do:
```suggestion
  return device:get_manufacturer() == VIDAR_THERMOSTAT_FINGERPRINTS.mfr and device:get_model() == VIDAR_THERMOSTAT_FINGERPRINTS.model
```
with appropriate restructuring of `VIMAR_THERMOSTAT_FINGERPRINTS`

drivers/SmartThings/zigbee-thermostat/src/vimar/init.lua
```diff
@@ -0,0 +1,79 @@
+-- Copyright 2022 SmartThings
```
greens, line 1: 2023, here and other files

greens:
the luacheck errors are all about trailing whitespace in the test file

drivers/SmartThings/zigbee-thermostat/src/vimar/init.lua
```diff
@@ -0,0 +1,213 @@
...
+      break
+    end
+  end
+end
+
+
+local function vimar_thermostat_heating_setpoint_handler(driver, device, value, zb_rx)
+  local setpoint = value.value / 100.0
+  local unit_type = "C"
+  if (setpoint >= 41.0) then
```
greens, line 156: as far as I know, the zigbee spec requires these reports to be in Celsius

you can just leave these handlers as the defaults

drivers/SmartThings/zigbee-thermostat/src/vimar/init.lua
```diff
@@ -0,0 +1,213 @@
...
+  for zigbee_attr_val, st_cap_val in pairs(VIMAR_THERMOSTAT_MODE_MAP) do
+    if command.args.mode == st_cap_val.NAME then
+      device:send_to_component(command.component, Thermostat.attributes.SystemMode:write(device, zigbee_attr_val))
+      device.thread:call_with_delay(1, function(d)
+        device:send_to_component(command.component, Thermostat.attributes.SystemMode:read(device))
+      end)
+      break
+    end
+  end
+end
```
greens, line 150: Because you're setting your supported modes attribute, users shouldn't be able to set unsupported modes unless they're specifically trying to by going through the API. If that's something you're worried about you can leave this in, but you could also take it out and be fine.

drivers/SmartThings/zigbee-thermostat/src/vimar/init.lua
greens: here's the source from the default implementation:

```
function ths_defaults.thermostat_heating_setpoint_handler(driver, device, value, zb_rx)
  local raw_temp = value.value
  local celc_temp = raw_temp / 100.0
  local temp_scale = "C"
  device:emit_event_for_endpoint(zb_rx.address_header.src_endpoint.value, capabilities.thermostatHeatingSetpoint.heatingSetpoint({value = celc_temp, unit = temp_scale }))
end
```

I believe there's some code in the hub that rounds to tenths of a degree, but that would be hit regardless of whether you're using the default or not.

greens (approved):

PR #622, opened 2023-03-13
fix(Sonos): Streamline device initialization process

dljsjr:
The initialization routine opted for a pretty heavy-weight eventual consistency model with a bunch of repeated/duplicated checks on speakers as they onboarded. It was also doing an extra round or two of re-scanning.

This tightens up some of the checks to prevent doing a bunch of un-needed work and introducing a bunch of un-needed network traffic.

### DIFF
--- a/drivers/SmartThings/sonos/src/disco.lua
+++ b/drivers/SmartThings/sonos/src/disco.lua
@@ -37,18 +37,17 @@ end
 
 function Discovery.discover(driver, _, should_continue)
   log.info("Starting Sonos discovery")
-  local known_devices_dnis = {}
-  local found_ip_addrs = {}
-
-  local device_list = driver:get_devices()
-  for _, device in ipairs(device_list) do
-    local id = device.device_network_id
-    known_devices_dnis[id] = true
-  end
-
-  driver.found_ips = found_ip_addrs
+  driver.found_ips = {}
 
   while should_continue() do
+    local known_devices_dnis = {}
+
+    local device_list = driver:get_devices()
+    for _, device in ipairs(device_list) do
+      local id = device.device_network_id
+      known_devices_dnis[id] = true
+    end
+
     ssdp.search(SONOS_SSDP_SEARCH_TERM, function(group_info)
       ssdp_discovery_callback(driver, group_info, known_devices_dnis, driver.found_ips)
     end)
--- a/drivers/SmartThings/sonos/src/init.lua
+++ b/drivers/SmartThings/sonos/src/init.lua
@@ -73,17 +73,36 @@ local function _initialize_device(driver, device)
     end
 
     local fields = driver._field_cache[device.device_network_id]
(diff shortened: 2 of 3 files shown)


drivers/SmartThings/sonos/src/disco.lua
```diff
@@ -37,18 +37,17 @@ end
...
-  end
+  while should_continue() do
+    local known_devices_dnis = {}
 
-  driver.found_ips = found_ip_addrs
+    local device_list = driver:get_devices()
+    for _, device in ipairs(device_list) do
+      local id = device.device_network_id
+      known_devices_dnis[id] = true
+    end
```
greens, line 49: would it be worthwhile to have `know_devices_dnis` declared outside this loop so it doesn't get re-built every time, or is that intentional?

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -73,17 +73,36 @@ local function _initialize_device(driver, device)
...
     local fields = driver._field_cache[device.device_network_id]
-    driver._player_id_to_device[fields.player_id] = device -- quickly look up device from player id string
-    driver.sonos:mark_player_as_joined(fields.player_id)
 
     log.trace("Setting persistent fields")
     device:set_field(PlayerFields.WSS_URL, fields.wss_url, { persist = true })
     device:set_field(PlayerFields.HOUSEHOULD_ID, fields.household_id, { persist = true })
     device:set_field(PlayerFields.PLAYER_ID, fields.player_id, { persist = true })
 
-    device:set_field(PlayerFields._IS_INIT, true)
```
cjswedes, line 84: Will this result in failure to rediscover a device whose ip address has changed since restarts?

drivers/SmartThings/sonos/src/disco.lua
dljsjr: I intentionally moved it inside the loop; I had the same thought originally which is why it was outside the loop but if you think about it, a device can join while discovery is still happening meaning that the `known_dnis` won't update and you'll re-scan that device.

And for the sake of consistency and correctness I don't want to populate that `known_dnis` table with DNIs from anywhere other than the driver's canonical list of known DNI's, so I don't want to update them from the discovery code manually.

drivers/SmartThings/sonos/src/init.lua
dljsjr: My instinct says that this should become eventually consistent as the group info comes in over the websocket subscriptions, but looking at this, there could definitely be edge cases where if the IP does change it may fail to make a websocket connection to a coordinator with visibility in to the speaker whose IP has changed. I'll have to think about this change hunk a bit more.

dljsjr:
I'm closing this PR as the onboarding process for Sonos has diverged quite a bit. There are still opportunities to optimize it, but they would look very different now.

PR #634, opened 2023-03-17
Update matter-sensor temperature sensor device category

cjswedes:
Use Thermostat category because it results in a better icon than the GenericSensor category.


cjswedes:
I also changed HumiditySensor categories to GenericSensor, since currently that category does not have icons.

PR #646, opened 2023-03-23
Wemo disco http response processing improvement

cjswedes:
Several crashes are occurring because the location header is not always parsed. This could be because the regex was not considering all valid headers, or because the header didn't exist. This change makes sure no crash occurs in either case.

- [x] Manual test wemo driver with this change

### DIFF
--- a/drivers/SmartThings/wemo/src/discovery.lua
+++ b/drivers/SmartThings/wemo/src/discovery.lua
@@ -49,7 +49,7 @@ end
 local function process_response(val)
   local info = {}
   val = string.gsub(val, "HTTP/1.1 200 OK\r\n", "", 1)
-  for k, v in string.gmatch(val, "([%g]+): ([%g ]*)\r\n") do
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/wemo/src/discovery.lua
```diff
@@ -49,7 +49,7 @@ end
 local function process_response(val)
   local info = {}
   val = string.gsub(val, "HTTP/1.1 200 OK\r\n", "", 1)
-  for k, v in string.gmatch(val, "([%g]+): ([%g ]*)\r\n") do
+  for k, v in string.gmatch(val, "([%w_-]+): *([%g ]*)\r\n") do
```
posborne, line 52: nit: For zero or more spaces, this is typically written as the following in regex (`[ ]*`). I think it may be required in some languages but not lua.  At least for me, that threw me off at first.

posborne (approved):
LGTM, this patch resolves one of the issues I was seeing with my wemo setup (the other relates to a pre-first-publish data problem).

PR #648, opened 2023-03-24
Better matter multi-switch handling

cjswedes:
Since the generic fingerprinting is lacking in our ability to support multi-switch devices, we must detect them when being configured and profile switch if necessary.

Only devices that will work with the default handling will be switched:
1. They have multiple endpoints that support the OnOff cluster
2. Those endpoints have IDs that are consecutive to line up with our profile naming scheme to allow for component <-> endpoint mappings

I added profiles for multi switches up to 7 switches. Im not aware of even one powerstrip that is certified that has multiple endpoints.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -29,6 +29,7 @@ local function convert_huesat_st_to_matter(val)
 end
 
 local function component_to_endpoint(device, component_id)
+  -- Assumes matter endpoint layout is sequentional starting at 1.
   local ep_num = component_id:match("switch(%d)")
   return ep_num and tonumber(ep_num) or device.MATTER_DEFAULT_ENDPOINT
 end
@@ -51,6 +52,28 @@ end
 
 local function device_removed(driver, device)
   log.info("device removed")
+end
+
+local function do_configure(driver, device)
+  -- New profiles need to be added for devices that have more switch endpoints
+  local MAX_MULTI_SWITCH_EPS = 7
+  -- Note: This profile switching is needed because of shortcoming in the generic fingerprints
+  -- where devices with multiple endpoints with the same device type cannot be detected
+  local switch_eps = device:get_endpoints(clusters.OnOff.ID)
+  local num_switch_eps = #switch_eps
+  table.sort(switch_eps)
+  --Default MCD switch handling depends on consecutive endpoint numbering
+  if num_switch_eps == switch_eps[num_switch_eps] then
+    if num_switch_eps > 1 then
+      device:try_update_metadata({profile = string.format("switch-%d", math.min(num_switch_eps, MAX_MULTI_SWITCH_EPS))})
+    end
+    if num_switch_eps > MAX_MULTI_SWITCH_EPS then
+      error(string.format(
+        "Matter multi switch device will not function. Profile doesn't exist with %d components",
+        num_switch_eps
+      ))
+    end
+  end
 end
 
 local function handle_switch_on(driver, device, cmd)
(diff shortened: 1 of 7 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -53,6 +54,27 @@ local function device_removed(driver, device)
...
+  local num_switch_eps = #switch_eps
+  table.sort(switch_eps)
+  --Default MCD switch handling depends on consecutive endpoint numbering
+  if num_switch_eps == switch_eps[num_switch_eps] then
+    if num_switch_eps > 1 and num_switch_eps <= MAX_MULTI_SWITCH_EPS then
+      device:try_update_metadata({profile = string.format("switch-%d", num_switch_eps)})
+    elseif num_switch_eps > MAX_MULTI_SWITCH_EPS then
+      error(string.format(
+        "Matter multi switch device will not function. Profile doesn't exist with %d components",
+        num_switch_eps
```
greens, line 72: In theory we could still support the first 7. I'm guessing you thought about this and decided not to, but I was wondering your reasoning.

cjswedes: My main reason was to emit an error that would be easy to spot to know if we should add more profiles. But considering what you're suggesting, we could emit the error and update the profile to something that would work more. I will double check what would happen in a case where we support the first 7 and start getting reports from ep num 8. I want to make sure it doesn't cause more problems than it solves

cjswedes: I did what you suggested. What will happen is that reports from the endpoints higher than the max we support will be mapped to the main component. The error will be in the logs and hub_logs, so that will at least give us a clue tht it is happening

greens: your error message will never fire, as is

`num_switch_eps > 1` is still true for `num_switch_eps > 8`, so the second branch is never entered

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -53,6 +54,28 @@ local function device_removed(driver, device)
...
+  local num_switch_eps = #switch_eps
+  table.sort(switch_eps)
+  --Default MCD switch handling depends on consecutive endpoint numbering
+  if num_switch_eps == switch_eps[num_switch_eps] then
+    if num_switch_eps > 1 then
+      device:try_update_metadata({profile = string.format("switch-%d", math.min(num_switch_eps, MAX_MULTI_SWITCH_EPS))})
+    end
+    if num_switch_eps > MAX_MULTI_SWITCH_EPS then
+      error(string.format(
+        "Matter multi switch device will not function. Profile doesn't exist with %d components",
```
greens, line 72: I think the message should indicate that it's only endpoints beyond X that won't work

cjswedes: Well it would actually affect the main components endpoint, so that endpoint also wouldn't work.

greens: what do you mean?

cjswedes: > What will happen is that reports from the endpoints higher than the max we support will be mapped to the main component.

This means manual changes to those switches will be displayed on the main state. So not only will the higher endpoints not work, their presence will make the main component change state when it shouldn't in some cases.

greens: Ah right. Because even if `endpoint_to_component` returns `nil` we throw in a `main`. Ran into some annoyances with that earlier this week as well.

greens (approved):

PR #649, opened 2023-03-24
Add generic fingerprint for matter bridges

cjswedes:
This maps to the matter-thing fingerprint since the platform does not yet support matter bridges.

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -21,7 +21,51 @@ matterManufacturer:
     productId: 0x5E
     deviceProfileName: plug-binary
 
+#Bridge devices need manufacturer specific fingerprints until
+#bridge support is released to all hubs. This is because of the way generic
+#fingerprints work for bridges joined prior to hubs being able to support them
+  - id: "Aqara/m2/hub"
+    deviceLabel: Matter Bridge
+    vendorId: 0x115F
+    productId: 0x0802
+    deviceProfileName: matter-bridge
+  - id: "Ubisys/G1"
+    deviceLabel: Matter Bridge
+    vendorId: 0x10F2
+    productId: 0x6100
+    deviceProfileName: matter-bridge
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -22,6 +22,11 @@ matterManufacturer:
     deviceProfileName: plug-binary
 
 matterGeneric:
+  - id: "matter/bridge"
```
cjswedes, line 25: I put it first in this list because I think that has an affect on tie breaks where it matches other device types. @tpmanley do you think I should add a couple more generic fingerprints that also include On/Off and Dimmer device types? This might help ensure the bridges get the right fingerprint.

cjswedes: We determined that the number of permutations needed for this to be effective would be untenable. I am leaving the generic fingerprint in for when official bridge support is released to hubs, and existing bridge devices will be identified with manufactuer specific fingerprints.

drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -21,7 +21,41 @@ matterManufacturer:
     productId: 0x5E
     deviceProfileName: plug-binary
 
+#Bridge devices need manufacturer specific fingerprints until
+#bridge support is released to all hubs. This is because of the way generic
+#fingerprints work for bridges joined prior to hubs being able to support them
```
tpmanley, line 26: Can you add Philips Hue (0x100B/0x0002) and Switchbot (0x1397/0x07e7) as well?

cjswedes: Interesting, the Switchbot product has a device type of 0 in the DCL. Ill add it anyway.

tpmanley: Yeah, we just know it's a Matter bridge from their public communication

tpmanley (approved):

PR #655, opened 2023-03-28
HCS-4170 HCS-4173 Zooz Zen 30 fixes

greens:
Adds "stateChange: true" for all button events.

Fixes some preferences that were integers instead of enums


ctowns (approved):

dljsjr (approved):

cjswedes (approved):
I assume that some events would be missed if sent twice in a row? Is that the reason for state_change = true?

sjl2178 (approved):

PR #658, opened 2023-03-29
CHAD-10712 Multi metering switch fixes

greens:
- Fixes the method used for calculating the number of child devices
- Triggers a refresh when the root node sends updates, otherwise ignores them


greens:
yeah, as expected, there's a disconnect somewhere between the tests and the actual devices that we need to figure out

ctowns (approved):

PR #661, opened 2023-03-31
edited how ecolink wireless siren handles on/off and status reporting

sjl2178:
on/off capability handlers were not handled by sub driver, results in incorrect status reporting, changed to reflect DTH behavior

### DIFF
--- a/drivers/SmartThings/zwave-siren/src/ecolink-wireless-siren/init.lua
+++ b/drivers/SmartThings/zwave-siren/src/ecolink-wireless-siren/init.lua
@@ -68,6 +68,22 @@ local do_refresh = function(self, device)
   end
 end
 
+local function siren_on(self, device, cmd)
+  device:send_to_component(Basic:Set({value = 0xFF}), cmd.component)
+  device:send_to_component(SwitchBinary:Get({}), cmd.component)
+  if cmd.component == "siren1" then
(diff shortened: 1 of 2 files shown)


cjswedes (commented):
Can you explain a little bit how the alarm commands are supposed to work? I assume this is just a weird device, but I would think that the Both/Off alarm commands would result in SET zw cmds being sent to the device rather than GETs. Is this device just a unicorn?

cjswedes (approved):

drivers/SmartThings/zwave-siren/src/ecolink-wireless-siren/init.lua
```diff
@@ -68,6 +68,23 @@ local do_refresh = function(self, device)
   end
 end
 
+local function siren_on(self, device, cmd)
+  device:send_to_component(Basic:Set({value = 0xFF}), cmd.component)
+  device:send_to_component(SwitchBinary:Get({}), cmd.component)
+  local endpoint = component_to_endpoint(device, cmd.component)
+  if endpoint[1] == 2 then
```
greens, line 75: I think it would look better to switch this on `cmd.component` (`"siren1"` I guess?) rather than the endpoint number

dljsjr (approved):

greens (approved):

PR #669, opened 2023-04-05
BUG-7315 Change EZMultiPli set color commands

greens:
Changes these commands to be more similar to the DTH commands (removes extra fields not used in DTH).


greens:
These changes have been confirmed by the user who reported the issue.

ctowns (approved):

PR #670, opened 2023-04-06
Fix configuration for ezexl power meter

kdbang:
Fix configuration for ezexl power meter

 - SimpleMetering::InstantaneousDemand
 - from 5 to 500(0.5W)


drivers/SmartThings/zigbee-power-meter/src/ezex/init.lua
```diff
@@ -39,6 +48,11 @@ end
 local device_init = function(self, device)
   device:set_field(constants.SIMPLE_METERING_DIVISOR_KEY, 1000000, {persist = true})
   device:set_field(constants.ELECTRICAL_MEASUREMENT_DIVISOR_KEY, 10, {persist = true})
+
+  device:remove_monitored_attribute(SimpleMetering.ID, SimpleMetering.attributes.InstantaneousDemand.ID)
+  device:remove_configured_attribute(SimpleMetering.ID, SimpleMetering.attributes.InstantaneousDemand.ID)
+  device:add_monitored_attribute(instantaneous_demand_configuration)
+  device:add_configured_attribute(instantaneous_demand_configuration)
```
greens, line 55: I would have expected that you only need to call `add` and that the new config values would overwrite the default ones. Is that not the case?

kdbang: @greens I removed the line calling remove_XXX api!

varzac:
it's important to note that just merging this won't affect devices that are already joined. The device will have to be removed and re-added to apply this change.

varzac (approved):

kdbang:
> it's important to note that just merging this won't affect devices that are already joined. The device will have to be removed and re-added to apply this change.

Yes, we need let user re-onboarding.

kdbang:
@greens @varzac I can't see "Merge" button, Do I need anthing else to resolve?

PR #675, opened 2023-04-13
[Bug] Cync bulb turns off, when brightness level set to "1" in ST App.

sanghyukko:
In case of Apple, Alexa ecosystem, the Cync bulb didn't turn off. 
There is a problem with the level value sent from the edge driver to the device.


cjswedes (approved):

sanghyukko:
If there is a problem with other manufacturers due to issues that occur only in certain manufacturers, we will check again.

PR #684, opened 2023-04-20
Matter-thermostat add support for mains powered thermostats

cjswedes:
I only added profiles for the nostate profiles since currently all thermostats are joined using those profiles.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -43,7 +43,51 @@ local setpoint_limit_device_field = {
   MIN_DEADBAND = "MIN_DEADBAND",
 }
 
+local subscribed_attributes = {
+  [capabilities.temperatureMeasurement.ID] = {
+    clusters.Thermostat.attributes.LocalTemperature,
+    clusters.TemperatureMeasurement.attributes.MeasuredValue
+  },
+  [capabilities.relativeHumidityMeasurement.ID] = {
+    clusters.RelativeHumidityMeasurement.attributes.MeasuredValue
+  },
+  [capabilities.thermostatMode.ID] = {
+    clusters.Thermostat.attributes.SystemMode,
+    clusters.Thermostat.attributes.ControlSequenceOfOperation
+  },
+  [capabilities.thermostatOperatingState.ID] = {
+    clusters.Thermostat.attributes.ThermostatRunningState
+  },
+  [capabilities.thermostatFanMode.ID] = {
+    clusters.FanControl.attributes.FanModeSequence,
+    clusters.FanControl.attributes.FanMode
+  },
+  [capabilities.thermostatCoolingSetpoint.ID] = {
+    clusters.Thermostat.attributes.OccupiedCoolingSetpoint
+  },
+  [capabilities.thermostatHeatingSetpoint.ID] = {
+    clusters.Thermostat.attributes.OccupiedHeatingSetpoint
+  },
+  [capabilities.battery.ID] = {
+    clusters.PowerSource.attributes.BatPercentRemaining
+  }
+}
+
 local function device_init(driver, device)
+  device:subscribe()
+end
+
+local function info_changed(driver, device, event, args)
+  --Note this is needed because device:subscribe() does not recalculate
+  -- the subscribed attributes each time it is run, that only happens at init.
+  -- This will change in the 0.48.x release of the lua libs.
+  for cap_id, attributes in pairs(subscribed_attributes) do
+    if device:supports_capability_by_id(cap_id) then
+      for _, attr in ipairs(attributes) do
+        device:add_subscribed_attribute(attr)
+      end
+    end
+  end
   device:subscribe()
 end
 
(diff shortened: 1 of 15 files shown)


greens:
looks like we might have a duplicate profile. You can _probably_ just `git mv` the existing one to match the naming scheme, although I'd double-check locally that you're allowed to "delete" a profile. At some point in the past it wasn't allowed. Might not be still.

sathvik-gatti:
Hello, 
I was going through the matter-thermostat driver and came across this issue where few attributes are not being subscribed by the hub, hence ST App controls are disabled for cluster's corresponding capabilities.

Currently as observed, driver subscribes to [attributes](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/matter-thermostat/src/init.lua#L50:~:text=clusters.FanControl.attributes.FanModeSequence%2C,clusters.FanControl.attributes.FanMode) only when the corresponding [capability](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/227eeebff9c4f267a994b17f63e06b0a3c1e69c3/drivers/SmartThings/matter-thermostat/src/init.lua#L378) is also listed in the profile applied during driver's matter device instance initialization (populate_subscribed_attributes_from_driver_defaults) in st/matter/device.lua

Considering the device being onboarded has both fan control and humidity clusters along with thermostat cluster on endpoint 1 (thermostat device type 0x301), the profile applied would be thermostat.yml, but since [init](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/227eeebff9c4f267a994b17f63e06b0a3c1e69c3/drivers/SmartThings/matter-thermostat/src/init.lua#L46)(subscribe) occurs before [do_configure](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/227eeebff9c4f267a994b17f63e06b0a3c1e69c3/drivers/SmartThings/matter-thermostat/src/init.lua#L50)(profile update), fanControl and humidity cluster's attributes are not being subscribed. Please correct me if I am missing something in my understanding.

cjswedes:
@sathvik-gatti thanks for bringing finding this bug. I added a fix to subscribe to the device during the info_changed event. I did it here rather than do_configure because during do_configure, the profile has not been updated on the device record. The info_changed event will fire when the profile has updated and therefore we can re-subscribe at a point when the new capabilities can be considered.

cjswedes:
> looks like we might have a duplicate profile.

I missed removing the battery cap from this profile when doing my copy pastas

greens (approved):

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -76,8 +81,13 @@ local function do_configure(driver, device)
...
     -- TODO remove this in favor of reading Thermostat clusters AttributeList attribute
     -- to determine support for ThermostatRunningState
+    -- Add nobattery profiles if updated
     profile_name = profile_name .. "-nostate"
 
+    if #battery_eps == 0 then
+      profile_name = profile_name .. "-nobattery"
+    end
+
     log.info_with({hub_logs=true}, string.format("Updating device profile to %s.", profile_name))
```
sathvik-gatti, line 91: Reference
```suggestion
    log.info_with({hub_logs=true}, string.format("Updating device profile to %s.", profile_name))
    local add_attributes = function(attrs)
        for _, attr in ipairs(attr) do
            device:add_subscribed_attribute(attr)
        end
    end
    if #fan_eps > 0 then add_attributes(driver.subscribed_attributes[capabilities.thermostatFanMode.ID]) end
    if #humidity_eps > 0 then add_attributes(driver.subscribed_attributes[capabilities.relativeHumidityMeasurement.ID]) end
    if #battery_eps > 0 then add_attributes(driver.subscribed_attributes[capabilities.battery.ID]) end
```

sathvik-gatti:
> @sathvik-gatti thanks for bringing finding this bug. I added a fix to subscribe to the device during the info_changed event. I did it here rather than do_configure because during do_configure, the profile has not been updated on the device record. The info_changed event will fire when the profile has updated and therefore we can re-subscribe at a point when the new capabilities can be considered.

Thanks @cjswedes for considering this issue, but it seems like `device:subscribe()` fetches a subscribed attributes list from lua device's store `(key=SUBSCRIBED_ATTRIBUTES_KEY)` and subscribes to those attributes present in it. This list as last seen, is populated when the lua matter device is instantiated. I was thinking maybe we can use `device:add_subscribed_attributes()` to add updated attributes to the list if their cluster was found on any endpoint right before profile update. Once the profile updates then info_changed would subscribe. I tried suggesting [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/684/files#r1174610701), please let me know what you think or if there is any other way to handle this as well.

cjswedes:
Looks like my fixup doesn't actually fix the issue of subscribed attributes since `MatterDevice:subscribe()` does not reconsider the devices capabilities and drivers handlers each time it is called. Ill manually do this in the info_changed handler for now, but I will get this changed in the next Lua library release to always recalculate the subscribed attributes/events for a device prior to issuing the subscribe request to the hub.

cjswedes:
@sathvik-gatti Thanks for the help on this one, I think it should be good now, but Ill wait for your 👍🏼 since I think you can run a manual test, which is always useful.

sathvik-gatti:
Thank you @cjswedes , for your response on this issue. I tested your fix from my side and it is working fine. The hub is subscribing to all the required attributes :) 👍

PR #685, opened 2023-04-20
Use matter color control option to execute commands if off

cjswedes:
Currently color control commands will fail if the switch/light is in the off state, but setting an options override for this will allow them to be successful while in the off state.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -106,6 +106,12 @@ local tbl_contains = function(t, val)
   return false
 end
 
+local TRANSITION_TIME = 0 --1/10ths of a second
+-- When sent with a command, these options mask and override bitmaps cause the command
+-- to take effect when the switch/light is off.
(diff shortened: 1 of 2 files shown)


greens (commented):
Does this override result in the light being turned on, or just mean it will be set to that color the next time it is turned on?

cjswedes:
The light will not turn on, it just allows the color change to take effect when the light is off.

cjswedes:
> Is there any way, with one command, to "turn on and move to color"

No, that would take two commands. I do think this change is more intuitive than not having the command take effect, but we could implement a multi command solution to turn the light on as well. It's not clear to me what the behavior should be on our platform.

greens (approved):

tpmanley (approved):
I tested this with the Nanoleaf bulb and confirmed it's setting both color and color temperature correctly even when the light is off. I think we should go forward with this change as is and create a new ticket to track matching the behavior of Zigbee/Z-Wave.

lelandblue:
I also tested this with my Matter bulbs successfully.

PR #687, opened 2023-04-20
Hue: Fixes for CHAD-10827, CHAD-10798, general reliability/stability improvements

dljsjr:
PR contains 3 commits addressing some general edge cases and stability issues with the Hue driver.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/disco.lua
+++ b/drivers/SmartThings/philips-hue/src/disco.lua
@@ -20,6 +20,8 @@ local supported_resource_types = {
   light = true
 }
 
+local try_create_count = {}
+
 -- "forward declarations"
 local discovered_bridge_callback, discovered_device_callback, is_device_service_supported,
 process_discovered_light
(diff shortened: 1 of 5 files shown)


cjswedes (approved):
LGTM, although I didn't put time in to fully understand the chunk parsing and line ending changes made to the sse event source; I trust your testing.

drivers/SmartThings/philips-hue/src/disco.lua
```diff
@@ -20,6 +20,8 @@ local supported_resource_types = {
   light = true
 }
 
+local try_create_count = {}
```
greens, line 23: does this get used for anything except logging?

dljsjr: This is only used for logging, but it's *really* easy for a Hue setup to be in the rate limiting regime in the hub-core firmware for `try_create_device` (1-per-second sustained, up to 10 per second in a burst). It's an odd heuristic so I was wanting to leave this in to capture some data about how this impacts people.

The impetus is that we get a lot of reports of "not all devices get added when I scan nearby" if you have a metric ton of devices attached to your bridge, and I want to get metrics on how to make that process smoother.

greens (approved):

PR #690, opened 2023-04-24
CHAD-10714 Reset works inconsistently for aeotec multi switch

greens:
This change attempts to add a delay between the reset and the get, as was present in the DTH


greens:
So, based on testing, this sorta kinda works. We've at least narrowed it down to know pretty much for certain that this is due to message volume coming in to the device during a simultaneous on/off and meter reset event, and not anything we're doing wrong.

cjswedes (approved):

PR #692, opened 2023-04-24
fix(sonos): Properly remove FF and Rewind.

dljsjr:
We disabled the enum commands, but did not fully
disable the playbackStatus capability command values.

Fixes: CHAD-10815

### DIFF
--- a/drivers/SmartThings/sonos/profiles/sonos-player.yml
+++ b/drivers/SmartThings/sonos/profiles/sonos-player.yml
@@ -6,11 +6,16 @@ components:
     version: 1
     config:
       values:
-        - key: "{{enumCommands}}"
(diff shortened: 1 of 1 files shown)


cjswedes (commented):
FWIW I tested the same profile change with Bose, and it did not change Fast Forward/Rewind from being options as a trigger for a routine.

cjswedes (approved):

dljsjr:
@cjswedes With more testing, I see the same thing as you see, the profile still doesn't update reliably.

This change should still go in, but I've kicked the bug back to CODA.

dljsjr:
I got confirmation that this profile is accurate and the issue is probably upstream, so I'm gonna merge.

PR #695, opened 2023-04-25
BUG-7174 Invert level commands for Yoolax shades

greens:


greens:
This will work for now but I think, based on the fact that almost all of our zigbee window shade drivers invert commands, we might want to make a more general solution here.

greens:
also note that I tested this locally and it worked on my device

PR #696, opened 2023-04-26
BUG-7461 Consolidate automatic tamper clear behavior for z-wave sensors

greens:


drivers/SmartThings/zwave-sensor/src/fibaro-door-window-sensor/init.lua
```diff
@@ -120,8 +120,6 @@ local function notification_report_handler(self, device, cmd)
   if notificationType == Notification.notification_type.HOME_SECURITY then
     if event == Notification.event.home_security.STATE_IDLE then
       device:emit_event(capabilities.tamperAlert.tamper.clear())
-    elseif event == Notification.event.home_security.TAMPERING_PRODUCT_COVER_REMOVED then
-      device:emit_event(capabilities.tamperAlert.tamper.detected())
```
cjswedes, line 124: Since fibaro doesn't rely on the automatic tamper clear, I think our tamper capability would become less accurate for these devices since they could have tamper enabled for longer than 10 sec. Prior to this change we would honor that, but after, the tamper would only be there for 10 seconds. Is it worth adding some manufacturer exclusions to the can_handle for `timed-tamper-clear`?

greens: Yeah. Makes sense.

cjswedes (approved):

PR #697, opened 2023-04-26
Add ability to deploy drivers from multiple partners

greens:


tools/deploy.py
```diff
@@ -70,58 +64,72 @@
...
-          print("Uploaded package successfully: "+driver)
-          drivers_updated.append(driver)
-          response_json = json.loads(response.text)
-          uploaded_drivers[package_key] = {DRIVERID: response_json[DRIVERID], VERSION: response_json[VERSION]}
+      while not os.path.exists("edge.zip") or retries >= 5:
+        try:
+          subprocess.run(["zip -r ../edge.zip config.yml fingerprints.yml $(find profiles -name \"*.y*ml\") $(find . -name \"*.lua\") -x \"*test*\""], cwd=driver, shell=True, capture_output=True, check=True)
+        except subprocess.CalledProcessError as error:
+          print(error.stderr)
+        retries += 1
```
cjswedes, line 98: Whats the reason for 5 retries on the command to zip up the driver? I would think if it fails that retries wouldn't actually help.

cjswedes (approved):
LGTM although the retries on the zip command seem unnecessary. You know better than me though if there are failures.

tools/deploy.py
greens: there was a race condition at some point around zipping up the first driver, this was the sledgehammer I used

cjswedes: 🔨

PR #698, opened 2023-04-27
Add ability to deploy drivers from multiple partners

greens:


dljsjr (approved):

drivers/Aqara/aqara-feeder/src/init.lua
```diff
@@ -15,6 +15,7 @@ local PRIVATE_CLUSTER_ID = 0xFCC0
 local PRIVATE_ATTR_ID = 0xFFF1
 local MFG_CODE = 0x115F
 
+
```
dljsjr, line 18: :eyes:

greens: triggered a "change" so that the driver would get deployed

only changed drivers get deployed

PR #699, opened 2023-04-28
Icon of aeotec-multipurpose-6 sensor was wrong.

kdbang:
Icon of aeotec-multipurpose-6 sensor was wrong.

should be displayed as MotionSensor

https://aeotec.com/products/aeotec-multi-sensor-6

esp7rizr


varzac:
This seems fine to me.  But as I am no longer on the drivers team, I will let Stephen give official approval.

greens (approved):

PR #701, opened 2023-04-28
Workaround to allow checks run against forked repos

greens:


greens:
Some of this is restoring functionality that was removed here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/a8953b4ed1c854a0fe299afa795286b0e77d6e00

but some of it is new. Not positive how the test coverage thing will work since every action does things a bit differently

dljsjr (approved):

PR #704, opened 2023-05-01
fix: Address stray bulbs at init.

dljsjr:
We previously only addressed stray bulbs during the "added" codepaths that are hit during fresh join and during migration.

I'm not sure how one can get in to a situation where they have stray bulbs post-added, but we've seen it, so this is a fix for it.

Fixes: CHAD-10941

### DIFF
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -268,12 +268,6 @@ bridge_added = function(driver, device)
   device:set_field(Fields.API_KEY, Discovery.api_keys[device_bridge_id], { persist = true })
   device:set_field(Fields._ADDED, true, { persist = true })
   driver.api_key_to_bridge_id[Discovery.api_keys[device_bridge_id]] = device_bridge_id
-
-  driver.stray_bulb_tx:send({
-    type = StrayDeviceMessageTypes.FoundBridge,
-    driver = driver,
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -641,31 +667,27 @@ cosock.spawn(function()
...
+      msg_device.parent_device_id or msg_device:get_field(Fields.PARENT_DEVICE_ID)
+  local maybe_bridge = found_bridges[maybe_bridge_id]
+
+  if maybe_bridge ~= nil then
+    local bridge_ip = maybe_bridge:get_field(Fields.IPV4)
+    local api_instance =
+        maybe_bridge:get_field(Fields.BRIDGE_API) or
+        HueApi.new_bridge_manager("https://" .. bridge_ip, msg_device:get_field(Fields.API_KEY))
+    process_strays(driver, api_instance, stray_lights, maybe_bridge.id)
+  end
```
cjswedes, line 690: nit: indent is confusing here. I think this is a part of the` elseif msg.type == StrayDeviceMessageTypes.NewStrayLight` branch.

dljsjr: Oof yeah the formatting got messed up here. Thanks.

cjswedes (approved):

PR #708, opened 2023-05-02
Bugfix matter color control error on XY values of (0, 0)

cjswedes:
The lua libs utils function returns NAN for these inputs which results failed capability event serialization when they are reported by a device. The utils function itself will be fixed in hub FW 0.48.x

- [x] Test with the device to see that these values are actually making sense for the user in this case.


greens:
Oh. Just saw your comment. Color spaces are very confusing. Is the answer to prioritize HSV over XY?

cjswedes:
I think the answer might be to just ignore 0 values for CurrentX and CurrentY

PR #715, opened 2023-05-05
Bugfix/wemo crashes

cjswedes:
See commit messages for explanations of the crashes that are fixed. I recommend reviewing each commit individually for this PR.


drivers/SmartThings/wemo/src/discovery.lua
```diff
@@ -42,6 +42,11 @@ local function send_disco_request()
   local sock = assert(socket.udp(), "create discovery socket")
   assert(sock:setsockname(listen_ip, listen_port), "disco| socket setsockname")
   local timeouttime = socket.gettime() + 5 -- 5 second timeout, `MX` + 1 for network delay
+
+  -- TODO Sometimes this is failing because the OS reports the "Network is unreachable", and
+  -- it causes a crash/restart loop. Is this acceptable? Should we stop disco but not crash?
+  -- Some hubs recover when the network becomes available, but some never do, presumably because
+  -- of LAN setups with blocked ports from the hub.
```
greens, line 49: Will the symptoms here improve once backoff is implemented?

greens (approved):

drivers/SmartThings/wemo/src/discovery.lua
dljsjr: My personal preference here is to avoid `assert` as much as possible in the discovery loop and just have it end the iteration early and try again next go-around in the case that a network operation fails since it's a cyclic operation. If the network is messed up, then the hub will simply not be able to discover devices, which seems fine.

cjswedes: This seems reasonable. Repeated logs rather than repeated crashes has way less of a chance of having side affects on the hub. I will add though that crashes are our main way of discovering these issues, so if we prevent the crash we will no longer have an idea of the hubs that are experiencing it without gathering user feedback.

drivers/SmartThings/wemo/src/subscription_server.lua
```diff
@@ -29,11 +29,19 @@ SubscriptionServer.mt = {
...
+  if device ~= nil then
+    self.ctrl_tx:send(ControlMessage.Subscribe(device))
+  else
+    log.warn("serve| invalid device, cannot subscribe")
+  end
 end
 
 function SubscriptionServer:unsubscribe(device)
-  self.ctrl_tx:send(ControlMessage.Unsubscribe(device))
+  if device == nil then
```
ctowns, line 40: Is this intended to be `device==nil` or should it be `device~=nil` since `device` is used in the `send` below? If it should be `nil`, would changing this to `ControlMessage.Unsubscribe(nil)` make that more clear?

cjswedes: 🤦🏼 should be ~= nil

drivers/SmartThings/wemo/src/discovery.lua
cjswedes: I pushed a change to log errors to hub logs and end the iteration of the discovery loop should we hit this issue. It will result in many logs to hub logs, but the driver shouldn't crash. Is it worth adding a `socket.sleep(2)` to limit the amount of logs?

cjswedes: Note I also made the change for bose.

greens: good thinking

cjswedes: @dljsjr  Im gonna wait for your approval to merge on this one.

dljsjr:
> I will add though that crashes are our main way of discovering these issues, so if we prevent the crash we will no longer have an idea of the hubs that are experiencing it without gathering user feedback.

I'm going to capture this in an internal JIRA for us to have a discussion about it. It's true that losing this signal is bad, but it's also true that we should probably have a way to get at this signal without needing to do something as extreme as forcing the driver to crash.

dljsjr (approved):

PR #719, opened 2023-05-09
Handle non-sequential matter endpoints in matter-switch

cjswedes:
This functionality will be in the lua libs in 0.48 but only for devices that dont override the component_to_endpoint function.


Con-Tejus (approved):
Changes look good, thanks @cjswedes for the quick fix.

PR #723, opened 2023-05-12
BUG-7616 Add delays to gets after setpoint sets for CT100

greens:
customer devices were failing to respond. The DTH would have included a 500ms wait between each command, so this aims to restore that in hopes of the device being capable of responding.


dljsjr:
Recognizing in advance that these devices tend to be really poorly documented, is there some kind of documentation on the existence of this rate limit somewhere? Or is this just something that we're hoping comes out in the wash by making the behavior more similar to the DTH?

greens:
@dljsjr no and yes, respectively

dljsjr (approved):

cjswedes (approved):

PR #726, opened 2023-05-16
CHAD-10802 Fibaro Single Switch: Handle reports from all endpoints

greens:
Users have reported that the DTH behavior is incorrect, and reports from endpoints other than 1 should be handled.


lelandblue:
Closing - as we did not confirm this fix resolved the issue. We can re-open if need be.

PR #736, opened 2023-05-19
[Aqara] Aqara pet feeder

seojune79:
adjusting the "feeding" status time


greens (changes requested):
I believe you need to rebase your fork, as the version you're working off of seems to think this  is an all new driver.

Kwang-Hui:
It seems rebased now

greens (approved):

PR #739, opened 2023-05-23
Bugfix sonos listener receives message for unknown household

cjswedes:
This is mainly just a mitigation to unblock internal users experiencing crashes while testing other features. Im not sure when this would happen, but it seems like something in the device initialization failed to collect household info for the device. Ill try to get some more logs to investigate why this is happening.

### DIFF
--- a/drivers/SmartThings/sonos/src/api/sonos_connection.lua
+++ b/drivers/SmartThings/sonos/src/api/sonos_connection.lua
@@ -225,7 +225,7 @@ function SonosConnection.new(driver, device)
         if body.version ~= favorites_version then
           favorites_version = body.version
 
-          local household = self.driver.sonos:get_household(header.householdId)
+          local household = self.driver.sonos:get_household(header.householdId) or { groups = {} }
 
           for group_id, group in pairs(household.groups) do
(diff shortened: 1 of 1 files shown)


ctowns (approved):

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -225,7 +225,7 @@ function SonosConnection.new(driver, device)
         if body.version ~= favorites_version then
           favorites_version = body.version
 
-          local household = self.driver.sonos:get_household(header.householdId)
+          local household = self.driver.sonos:get_household(header.householdId) or { groups = {} }
```
tpmanley, line 228: Would it be helpful to log an error when hitting the default case?

cjswedes: I toyed with this idea, but didn't because the other branches dont. It is also only helpful if it goes to hub logs. I am hoping to get some more logs internally that can show me what is going on to make the household not available at this point. So not i dont think its particularly helpful.

tpmanley (approved):

greens (approved):

lelandblue:
Do we need the driver sharing link on this PR so that @doua can test?

PR #742, opened 2023-05-25
Update init.lua

cmiranti:
Fixing issue where mimolite garage doors would not close.

Door would close if smartthings showed as 'closed' when button was pushed - Momentary switch would be turned on.


wkhenon (commented):
Changes seem okay for a momentary switch ✅
Would this behavior hold if it was a basic switch instead of momentary though?

lelandblue:
I am going to close this PR as I believe this was resolved else where. If I am mistaken we can reopen.

PR #744, opened 2023-05-29
add savant color light

linus2017:
for new vid/pid, and cct range: 2000K-7000K.

### DIFF
--- a/drivers/SmartThings/matter-switch/profiles/light-color-level-savant.yml
+++ b/drivers/SmartThings/matter-switch/profiles/light-color-level-savant.yml
@@ -1,0 +1,21 @@
+name: light-color-level-savant
+components:
+- id: main
+  capabilities:
+  - id: switch
+    version: 1
+  - id: switchLevel
+    version: 1
+  - id: colorTemperature
+    version: 1
+  - id: colorControl
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Light
+metadata:
+  mnmn: SmartThingsEdge
+  vid: generic-rgbw-color-bulb-2000K-7000K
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/profiles/light-color-level-savant.yml
```diff
@@ -0,0 +1,21 @@
+name: light-color-level-savant
```
ctowns, line 1: @linus2017 I would suggest a few changes to this custom profile. We are trying to move away from using metadata like `vid` in profiles, so there is a new method of `embedded profile configs` that we can use to define ranges on these capabilities. For this case, I would suggest changing the `filename` and `name:` field to something like `light-color-level-2000K-7000K` and then using an embedded configuration on the `colorTemperature` capability like this:

```
name: light-color-level-2000K-7000K
components:
- id: main
  capabilities:
  - id: switch
    version: 1
  - id: switchLevel
    version: 1
  - id: colorTemperature
    version: 1
    config:
      values:
        - key: "colorTemperature.value"
          range: [ 2000, 7000 ]
  - id: colorControl
    version: 1
  - id: firmwareUpdate
    version: 1
  - id: refresh
    version: 1
  categories:
  - name: Light
```

You can see other examples of this in the `zigbee-switch` profiles, for example, the `color-temp-bulb-2200K-4000K.yml` profile [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/a4d1cb20005ed92bb03a7ad7da6334cbcfdda6c1/drivers/SmartThings/zigbee-switch/profiles/color-temp-bulb-2200K-4000K.yml)

Let me know if you have any questions!

ctowns:
Moved to https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/750

drivers/SmartThings/matter-switch/profiles/light-color-level-savant.yml
linus2017: thanks a lot! I see Pull 750 instead of this Pull 744.

PR #745, opened 2023-05-30
Roll up beta to production

greens:


.github/workflows/upload-driver-packages.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           files: "drivers"
           json: true
       - id: set-outputs
-        run: echo "matrix=${{ steps.changed-drivers.outputs.all_modified_files }}" >> "$GITHUB_OUTPUT"
+        run: echo "matrix=${{ steps.changed-drivers.outputs.all_modified_files }}" >> $GITHUB_OUTPUT
```
dljsjr, line 23: Why are we no longer quoting the output here?

dljsjr (approved):

.github/workflows/upload-driver-packages.yml
greens: shouldn't have been there in the first place

greens: just decided to make it consistent with elsewhere

ctowns (approved):

PR #759, opened 2023-06-07
BUG-7616 Wait for response before sending follow-up polling with setp…

greens:
…oint gets

Users have reported issues controlling the setpoints of CT100 thermostats. Inspecting logs revealed that the device was not responding to setpoint gets. This was presumably due to issues with the device being overloaded by the number of z-wave messages in a short timeframe. Timers were implemented to attempt to mitigate this, but on busy networks, message queueing means that delays between commands may not be preserved when they're sent by the radio. This approach waits until we have received a response and then sends any subsequent follow-up polls.

### DIFF
--- a/drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
+++ b/drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
@@ -28,6 +28,8 @@ local ThermostatFanMode = (require "st.zwave.CommandClass.ThermostatFanMode")({ 
 local Battery = (require "st.zwave.CommandClass.Battery")({ version = 1 })
 --- @type st.zwave.CommandClass.MultiChannel
 local MultiChannel = (require "st.zwave.CommandClass.MultiChannel")({ version = 4 })
+local heating_setpoint_defaults = require "st.zwave.defaults.thermostatHeatingSetpoint"
+local cooling_setpoint_defaults = require "st.zwave.defaults.thermostatCoolingSetpoint"
 local constants = require "st.zwave.constants"
 local utils = require "st.utils"
 
@@ -40,6 +42,7 @@ local CT100_THERMOSTAT_FINGERPRINTS = {
 -- DTH actually uses the old mutliInstance encap, but multichannel should be back-compat
 local TEMPERATURE_ENDPOINT = 1
 local HUMIDITY_ENDPOINT = 2
+local SETPOINT_REPORT_QUEUE = "_setpoint_report_queue"
 
 --TODO: Update this once we've decided how to handle setpoint commands
 local function convert_to_device_temp(command_temp, device_scale)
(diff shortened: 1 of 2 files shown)


greens:
This has renewed my desire to implement mailboxing for z-wave devices. Or to enable some sort of transactional relationships.

drivers/SmartThings/zwave-thermostat/src/ct100-thermostat/init.lua
```diff
@@ -108,13 +113,31 @@ local function thermostat_mode_report_handler(self, device, cmd)
...
-    device:send(ThermostatSetpoint:Get({setpoint_type = ThermostatSetpoint.setpoint_type.HEATING_1}))
+    device:set_field(SETPOINT_REPORT_QUEUE, function ()
+      device:send(ThermostatSetpoint:Get({setpoint_type = ThermostatSetpoint.setpoint_type.HEATING_1}))
+    end)
   else
     device:send(ThermostatSetpoint:Get({setpoint_type = ThermostatSetpoint.setpoint_type.HEATING_1}))
-    device:send(ThermostatSetpoint:Get({setpoint_type = ThermostatSetpoint.setpoint_type.COOLING_1}))
+    device:set_field(SETPOINT_REPORT_QUEUE, function ()
+      device:send(ThermostatSetpoint:Get({setpoint_type = ThermostatSetpoint.setpoint_type.COOLING_1}))
+    end)
```
cjswedes, line 123: Is the intention for `SETPOINT_REPORT_QUEUE` to be overridden or appended to here and in the `set_setpoint_factory` closure? I think there is a case where the driver receives a setpoint set cap cmd and the multilevel get and operating state get commands are queued, but before sending the commands a mode report is received and the setpoint get command overrides the previous two commands so they are never sent.

greens: It can happen, but it's not a situation I'm too concerned about.

greens: Based on current behavior of the device that situation would already be an issue. I think as long as we're getting _any_ setpoint reports, it's an improvement.

cjswedes: If its an improvement, it seems worthwhile.

cjswedes (approved):
Can you comment the code a little bit so its easy to understand why this approach was taken?

greens:
> Can you comment the code a little bit so its easy to understand why this approach was taken?

more than this: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/759/files#diff-1098053ec5c064f9ee13c5285d1e89c8ee7557ebf75ba2db73350edb835d6bf5R127-R129 ?

cjswedes:
On second reading that seems fine. LGTM

PR #760, opened 2023-06-07
Bugfix: Bose devices sometimes marked offline "forever"

cjswedes:
The driver attempted to reconnect a devices websocket at most 10 times, before stopping. This meant that if the device did not come back online within 20 minutes, the device would be offline forever, or at least until the driver is restarted or the device is refreshed while it is reachable on the network.


greens (commented):
Wasn't the "offline forever" part of the original intent of the code? What's changed?

greens:
But this will retry forever, right? I'm just asking what the original rationale was for capping retries.

greens (approved):

PR #763, opened 2023-06-08
BUG-7236 Fixes typo leading to less-sensitive Centralite multi-sensors

greens:
Original DTH was setting this value at 0x02 whereas the driver was configuring it to 20, leading to an order of magnitude-higher acceleration threshold.

### DIFF
--- a/drivers/SmartThings/zigbee-contact/src/multi-sensor/centralite-multi/init.lua
+++ b/drivers/SmartThings/zigbee-contact/src/multi-sensor/centralite-multi/init.lua
@@ -44,7 +44,7 @@ end
 
 local do_configure = function(self, device)
   device:configure()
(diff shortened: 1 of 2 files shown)


greens:
Original line from DTH that was duplicated incorrectly and led to this error: 

https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy#L415

cjswedes:
Is this one of those typos that has been copy pasted from another place? IIRC we have some user reports of acceleration/multi-sensors not reporting any events for a few different models.

greens:
> Is this one of those typos that has been copy pasted from another place? IIRC we have some user reports of acceleration/multi-sensors not reporting any events for a few different models.

I just checked the other models and the configs seem correct. They all have a different config flow.

PR #764, opened 2023-06-12
Forward all Hue logs to hub-core for debugging

dljsjr:


dljsjr:
This shouldn't be a huge impact on most of the fleet because the `hub_logs = true` option respects hubcore's log level configuration when a driver log is received on the hub side, so we won't be filling people's logs unless they've already upped the verbosity on their hub. Which is what we want.

ctowns (approved):
Let there be logs!

drivers/SmartThings/philips-hue/src/disco.lua
```diff
@@ -63,26 +63,26 @@ function HueDiscovery.search_for_bridges(driver, computed_mac_addresses, callbac
...
       goto continue
     end
     if info.service_info.service_type ~= SERVICE_TYPE then -- response for a different service type. Shouldn't happen.
-      log.debug("Unexpected service type response: " .. info.service_info.service_type)
+      log.debug_with({ hub_logs = true }, "Unexpected service type response: " .. info.service_info.service_type)
       goto continue
     end
     if info.service_info.domain ~= DOMAIN then -- response for a different domain. Shouldn't happen.
-      log.debug("Unexpected domain response: " .. info.service_info.domain)
+      log.debug_with({ hub_logs = true }, "Unexpected domain response: " .. info.service_info.domain)
```
cjswedes, line 85: I could see these wanting to be info level to diagnose reasons for devices not joining. Up to you though.

drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -250,7 +254,39 @@ end
...
+  local num_devices_expected = 0
+  for _, device in ipairs(driver:get_devices()) do
+    if not device.parent_device_id then
+      log.warn_with(string.format("Device %s does not have a parent_device_id",
+        (device.label or device.id or "unknown device")))
+    elseif device.parent_device_id == bridge_device.id then
+      num_devices_expected = num_devices_expected + 1
+    end
+  end
+  log.trace_with({ hub_logs = true },
```
cjswedes, line 266: Info level here too? Is this info that you'd want to confirm in most investigations?

drivers/SmartThings/philips-hue/src/hue/api.lua
```diff
@@ -83,13 +104,13 @@ function PhilipsHueApi.new_bridge_manager(base_url, api_key, socket_builder)
...
       if err then
-        log.error("Error receiving on control channel for REST API thread", err)
+        log.error_with({ hub_logs = true }, "Error receiving on control channel for REST API thread", err)
         goto continue
       end
 
       if msg and msg._type then
         if msg._type == ControlMessageTypes.Shutdown then
-          log.trace("REST API Control Thread received shutdown message");
+          log.trace_with({ hub_logs = true }, "REST API Control Thread received shutdown message");
```
cjswedes, line 113: Info level? Maybe other logs happen that indicate this as well, otherwise it seems like a noteworthy event.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -141,7 +142,7 @@ local function migrate_bridge(driver, device)
     }
 
     device:try_update_metadata(new_metadata)
-    log.trace("Bridge Migrated, re-adding")
+    log.trace_with({ hub_logs = true }, "Bridge Migrated, re-adding")
```
cjswedes, line 145: Seems like this could be info, but perhaps other device init logging indicates successful migration as well.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -562,6 +574,7 @@ end
 ---@param driver HueDriver
 ---@param device HueChildDevice
 local function init_light(driver, device)
+  log.trace_with({hub_logs = true}, string.format("Init Light for device %s", (device.label or device.id or "unknown device")))
```
cjswedes, line 577: info level?

cjswedes (approved):
@dljsjr you know best what is needed in hub logs when investigating and which logs are on hot paths that would log too much and not be useful, so feel free to ignore my queries about making more things info level. But with bose/wemo I aired more on the side of log to hubCore than you did here. Particularly the init/discovery path for devices on fresh join and migration.

drivers/SmartThings/philips-hue/src/disco.lua
dljsjr: I wanted to leave these at Debug because they can be potentially chatty and we're likely not debugging an mDNS issue.

drivers/SmartThings/philips-hue/src/hue/api.lua
dljsjr: This is for graceful shutdowns; hard crashes are logged at error.

elbe0046 (commented):
Were you also going to add logging for these cases?  Or where did that end up? https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/604d814596fb0667fce6b2c59ce5ef309a8ea7bd/drivers/SmartThings/philips-hue/src/init.lua#L333-L334

drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -250,7 +254,39 @@ end
 ---@param bridge_device HueBridgeDevice
 local function do_refresh_all_for_bridge(driver, bridge_device)
   local child_devices = bridge_device:get_child_list() --[=[@as HueChildDevice[]]=]
+  local num_devices_expected = 0
+  for _, device in ipairs(driver:get_devices()) do
+    if not device.parent_device_id then
+      log.warn_with(string.format("Device %s does not have a parent_device_id",
```
tpmanley, line 260: Missing `{ hub_logs = true }`

dljsjr:
> Were you also going to add logging for these cases? Or where did that end up?
> 
> https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/604d814596fb0667fce6b2c59ce5ef309a8ea7bd/drivers/SmartThings/philips-hue/src/init.lua#L333-L334

I put that in a stash because it was on another branch and forgot to pop it.

elbe0046 (approved):

PR #766, opened 2023-06-13
Bugfix: Bose devices sometimes marked offline "forever"

cjswedes:
We want this bugfix to go to beta before the release this afternoon so that it can go to prod this afternoon and fix users hub experiencing the issue with bose.

### DIFF
--- a/drivers/SmartThings/bose/src/listener.lua
+++ b/drivers/SmartThings/bose/src/listener.lua
@@ -21,7 +21,6 @@ local CloseCode = require"lustre.frame.close".CloseCode
 local capabilities = require "st.capabilities"
 local utils = require "st.utils"
 local bose_utils = require "utils"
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/bose/src/listener.lua
```diff
@@ -182,11 +181,9 @@ function Listener:try_reconnect()
       return
     end
     retries = retries + 1
-    log.info(string.format("Retry reconnect in %s seconds", RECONNECT_PERIOD))
+    log.info(string.format("Reconnect attempt %s in %s seconds", retries, RECONNECT_PERIOD))
     socket.sleep(RECONNECT_PERIOD)
```
dljsjr, line 185: Possible TODO for a future release: use a backoff here scoped to the loop so it gets GC'd/reset on successful connection.

dljsjr (approved):

PR #774, opened 2023-06-15
fix: Make the Hue Driver resilient to crashes during migration

dljsjr:
Best viewed commit-by-commit. First commit is a revert to the debug logging changes, so it's kind of noisy.

- Be smarter about situations where we create new API connections
- Recover some necessary state information in bridge `init` that can get lost but is needed for lights that still need to migrate
- Wrap a throwing function call in a `pcall`

Fixes: [CHAD-11148](https://smartthings.atlassian.net/browse/CHAD-11148)

### DIFF
--- a/drivers/SmartThings/philips-hue/src/disco.lua
+++ b/drivers/SmartThings/philips-hue/src/disco.lua
@@ -140,8 +140,7 @@ discovered_bridge_callback = function(driver, bridge_ip, bridge_id, known_dni_to
     log.info_with({ hub_logs = true }, string.format("Scanning bridge %s for devices...", bridge_id))
 
     HueDiscovery.disco_api_instances[bridge_id] = HueDiscovery.disco_api_instances[bridge_id]
-        or
-        HueApi.new_bridge_manager("https://" .. bridge_ip, HueDiscovery.api_keys[bridge_id])
+      or HueApi.new_bridge_manager("https://" .. bridge_ip, HueDiscovery.api_keys[bridge_id])
 
     HueDiscovery.search_bridge_for_supported_devices(driver, HueDiscovery.disco_api_instances[bridge_id],
       function(hue_driver, svc_info, device_info)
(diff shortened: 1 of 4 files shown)


varzac:
Given that we haven't validated this fix in situ on an affected user, I think I would prefer to leave the elevated logging so we can target the 3 (now 2) users and follow the elevated logs after releasing this to bot validate the fix and be able to see if any other unexpected behavior happens.  We can follow up with removing logs if we are confident that we have corrected all issues.

varzac:
If we're going to fix the luacheck complaint I would prefer to just remove the `revert` commit as opposed to revert -> revert  especially given we are going to revert again to ultimately remove them.  But I would like to get this moving so if that's going to take a while to rebase without the revert, we can move forward.

dljsjr:
> If we're going to fix the luacheck complaint I would prefer to just remove the `revert` commit as opposed to revert -> revert especially given we are going to revert again to ultimately remove them. But I would like to get this moving so if that's going to take a while to rebase without the revert, we can move forward.

I was planning to squash everything down to a handful of commits to get rid of the double revert. I just didn't mark stuff as `fixup!`.

FreeMasen (approved):

drivers/SmartThings/philips-hue/src/disco.lua
```diff
@@ -166,7 +165,7 @@ discovered_bridge_callback = function(driver, bridge_ip, bridge_id, known_dni_to
 
         local api_key = item.success.username
         local bridge_base_url = "https://" .. bridge_ip
-        local api_instance = HueApi.new_bridge_manager(bridge_base_url, api_key)
+        local api_instance = HueDiscovery.disco_api_instances[bridge_id]
```
varzac, line 168: This is another place where we switched to re-using a connection where we weren't previously.  Are we OK with that?

dljsjr: Nope, reverted. Caught it when I was dealing with `luacheck`.

NoahCornell (approved):

PR #775, opened 2023-06-15
Add missing wemo model name to profile match

cjswedes:
Some users are unable to onboard wemo devices with the LightSwitch model


ctowns (approved):

drivers/SmartThings/wemo/src/init.lua
```diff
@@ -37,6 +37,7 @@ local profiles = {
   ["Dimmer"] = "wemo.dimmer-switch.v1",
   ["Motion"] = "wemo.motion-sensor.v1",
   ["Lightswitch"] = "wemo.light-switch.v1",
```
ctowns, line 39: This is intentionally the same word with two different capitalizations, right?

cjswedes: Yes, I dont think the original was a typo. There certainly could be devices out there that report the lowercase version, so I figured Id leave it in just in case.

greens (approved):

PR #776, opened 2023-06-15
deploy: Add search-parameters to driver package

tandres:
Previous changes that added search-parameters files to some drivers didn't include changes to the deploy script to include those files so they were never included in the driver package. This change updates the deploy script to include them.


greens (approved):

greens (changes requested):
Could you also add this to the CI workflow that creates the channel whenever a PR is created?

greens (approved):

cjswedes (approved):

varzac (approved):

PR #784, opened 2023-06-22
TLJ Door Lock fingerprint

VincentStark:
Add support for TLJ Door Lock.

### DIFF
--- a/drivers/SmartThings/zwave-lock/fingerprints.yml
+++ b/drivers/SmartThings/zwave-lock/fingerprints.yml
@@ -288,6 +288,12 @@ zwaveManufacturer:
     productType: 0x0005
     productId: 0x0001
     deviceProfileName: lock-battery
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/zwave-lock/fingerprints.yml
```diff
@@ -288,6 +288,12 @@ zwaveManufacturer:
     productType: 0x0005
     productId: 0x0001
     deviceProfileName: lock-battery
+  - id: "TLJ"
```
ctowns, line 291: Please move this to `line 248` so that it is grouped with the other `base-lock` devices

add a comment with the manufacturer name above the fingerprint just like it is done elsewhere throughout the file.
```
  # TLJ
  - id: "TLJ"
```

ctowns (approved):
Looks good to me!

PR #791, opened 2023-06-26
Vendor `cosock/ssl.lua` to patch some issues

dljsjr:
1. Fix an issue where a send ready socket was being put in the receive position during dohandshake by adjusting the allowlist for receive errors.
2. Add timeout to the error allowlist for all operations.


dljsjr:
No need to vendor this patch, it was included in the core Lua Libs.

PR #792, opened 2023-06-26
Refactor migration to avoid over-spawning tasks

dljsjr:
We introduced some accidentally quadratic behavior that was leading
to a massive amount of cosock tasks getting spawned.

We had to refactor the data flow a little bit in order to achieve this,
so the behavior may appear to have a changed a bit more than preferred.
However, the savings on CPU/Memory/network overhead are appreciable.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -120,20 +120,24 @@ local function migrate_bridge(driver, device)
   cosock.spawn(
     function()
       local bridge_found = false
-      local timeout_time = cosock.socket.gettime() + 30
-      repeat
-        local time_remaining = math.max(0, timeout_time - cosock.socket.gettime())
-        if time_remaining == 0 then
-          log.error_with({ hub_logs = true }, string.format(
-            "Couldn't successfully discover Hue Bridge on the network for migrated device %s",
-            device.label
-          ))
-          return
-        end
-
+      local backoff_generator = utils.backoff_builder(10, 0.1, 0.1)
+      local sleep_time = 0.1
+      while true do
+        log.info_with({hub_logs = true},
+          string.format(
+            "[MigrateBridge] Scanning for Hue Bridge info for migrated device %s",
+            (device.label or device.id or "unknown device")
+          )
+        )
         Discovery.search_for_bridges(driver, known_macs, function(hue_driver, bridge_ip, bridge_id)
           if bridge_id ~= device_dni then return end
 
+          log.info_with({hub_logs = true},
+            string.format(
+              "[MigrateBridge] Matching Hue Bridge for migrated device %s found, querying configuration values",
+              (device.label or device.id or "unknown device")
+            )
+          )
           local bridge_info, err, _ = HueApi.get_bridge_info(bridge_ip)
           if err ~= nil or not bridge_info then
             log.error_with({ hub_logs = true }, "Error querying bridge info: ", err)
@@ -164,7 +168,20 @@ local function migrate_bridge(driver, device)
           logged_bridge_added(hue_driver, device)
           bridge_found = true
         end)
-      until bridge_found
+        if bridge_found then return end
+
+        if sleep_time < 10 then
+          sleep_time = backoff_generator()
+        end
+        log.info_with({hub_logs = true},
+          string.format(
+            "[MigrateBridge] Failed to find bridge info for device %s, waiting %s seconds then trying again",
+             (device.label or device.id or "unknown device"),
+             sleep_time
+          )
+        )
+        cosock.socket.sleep(sleep_time)
+      end
     end,
     string.format("bridge migration thread for %s", device.label)
   )
@@ -259,14 +276,25 @@ bridge_added = function(driver, device)
     local known_macs = {}
     cosock.spawn(
       function()
-        local duration = 30 -- seconds
-        local timeout_time = cosock.socket.gettime() + duration
-        local time_remaining = duration
-        while time_remaining > 0 do
-          local bridge_found = false
+        local backoff_generator = utils.backoff_builder(10, 0.1, 0.1)
+        local sleep_time = 0.1
+        local bridge_found = false
+
+        while true do
+          log.info_with({hub_logs = true},
+            string.format(
+              "[BridgeAdded] Scanning the network for Hue Bridge matching device %s",
+              (device.label or device.id or "unknown device")
+            )
+          )
           Discovery.search_for_bridges(driver, known_macs, function(driver, bridge_ip, bridge_id)
             if bridge_id ~= device_bridge_id then return end
 
+            log.info_with({hub_logs = true}, string.format(
+                "[BridgeAdded] Found Hue Bridge on the network for %s, querying configuration values",
+                (device.label or device.id or "unknown device")
+              )
+            )
             local bridge_info, err, _ = HueApi.get_bridge_info(bridge_ip)
             if err ~= nil or not bridge_info then
               log.error_with({ hub_logs = true }, "Error querying bridge info: ", err)
@@ -284,9 +312,19 @@ bridge_added = function(driver, device)
             update_bridge_fields_from_info(driver, bridge_info, device)
             bridge_found = true
           end)
-
           if bridge_found then return end
-          time_remaining = math.max(0, timeout_time - cosock.socket.gettime())
+
+          if sleep_time < 10 then
+            sleep_time = backoff_generator()
+          end
+          log.info_with({hub_logs = true},
+            string.format(
+              "[BridgeAdded] Failed to find bridge info for device %s, waiting %s seconds then trying again",
+               (device.label or device.id or "unknown device"),
+               sleep_time
+            )
+          )
+          cosock.socket.sleep(sleep_time)
         end
       end,
       "bridge added re-scan"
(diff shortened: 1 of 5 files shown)


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -377,116 +368,60 @@ local function migrate_light(driver, device, parent_device_id)
...
-    end,
-    "migrate light re-scan"
-  )
+  local new_metadata = {
+    profile = profile_ref,
+    manufacturer = light_resource.hue_device_data.product_data.manufacturer_name,
+    model = light_resource.hue_device_data.product_data.model_id,
+    vendor_provided_label = light_resource.hue_device_data.product_data.product_name,
+  }
+  device:try_update_metadata(new_metadata)
```
varzac, line 411: We should not touch this as a part of this version, but I do wonder if sending all these updates are really necessary on migrate.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -891,65 +836,35 @@ local stray_bulb_tx, stray_bulb_rx = cosock.channel.new()
...
-              light_device.parent_assigned_child_key,
-              light_device.parent_device_id,
-              bridge_device_uuid,
-              Discovery.api_keys[driver:get_device_info(bridge_device_uuid).device_network_id],
-              driver.api_key_to_bridge_id[api_key_extracted]
-            ))
-            table.insert(dnis_to_remove, light_dni)
-            logged_stray_init(hue_driver, light_device, nil, nil, bridge_device_uuid)
-          end
+      for light_dni, light_device in pairs(strays) do
```
varzac, line 848: I'm wondering if we want to keep a list of "found" strays and log any that don't find a match by the end of the for loop.  Just because I don't think we would have great visibility currently if some device never finds a match.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -982,7 +901,89 @@ cosock.spawn(function()
...
+                hue_driver.api_key_to_bridge_id[api_key_extracted]
+              ))
+              stray_dni_to_rid[stray_dni] = svc_info.rid
+            end
+          end
+        end,
+        "[process_strays]"
+      )
+      log.info_with({ hub_logs = true }, "Finished querying bridge for devices from stray light handler")
+      process_strays(thread_local_driver, stray_lights, msg_device.id)
```
varzac, line 986: Is this a recursive call?  Wouldn't the loop handle this?

varzac (commented):
OK, I did a first pass and have a couple of questions, but overall looks good.  I will take another pass tomorrow morning with fresh eyes.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -982,7 +901,89 @@ cosock.spawn(function()
...
+              on = light.on,
+              color = light.color,
+              dimming = light.dimming,
+              color_temp = light.color_temperature,
+              mode = light.mode,
+              parent_device_id = bridge_device_uuid,
+              hue_device_id = light.owner.rid,
+              hue_device_data = device_data
+            }
+            Discovery.light_state_disco_cache[light.id] = light_resource_description
```
kianooshkarami, line 946: Proposing the following for logging purposes:

```
if Discovery.light...[light.id] is nil then
  log (light_resource_description)
end
Discovery.light_state_disco_cache[light.id]
```

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -982,7 +901,89 @@ cosock.spawn(function()
...
+                matching_v1_id,
+                matching_uuid,
+                stray_light.device_network_id,
+                stray_light.parent_assigned_child_key,
+                stray_light.parent_device_id,
+                bridge_device_uuid,
+                Discovery.api_keys[hue_driver:get_device_info(bridge_device_uuid).device_network_id],
+                hue_driver.api_key_to_bridge_id[api_key_extracted]
+              ))
+              stray_dni_to_rid[stray_dni] = svc_info.rid
```
kianooshkarami, line 979: Could we do a break here if already found?

drivers/SmartThings/philips-hue/src/init.lua
kianooshkarami: So if I understand the changes correct, what happens is that in the main coroutine/task, after a Bridge is found, all the devices are iterated and once a stray is matched to a `rid`, the following gets set `              stray_dni_to_rid[stray_dni] = svc_info.rid` which then allows the `process_strays` to then migrate the newly matched strays.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -377,116 +368,60 @@ local function migrate_light(driver, device, parent_device_id)
...
-  )
+  local new_metadata = {
+    profile = profile_ref,
+    manufacturer = light_resource.hue_device_data.product_data.manufacturer_name,
+    model = light_resource.hue_device_data.product_data.model_id,
+    vendor_provided_label = light_resource.hue_device_data.product_data.product_name,
+  }
+  device:try_update_metadata(new_metadata)
+
+  local updated_parent_id = parent_device_id or known_dni_to_device_map[bridge_id].id
```
kianooshkarami, line 413: You already have `bridge_device` from line 332 and validate that it exists. you could just use `bridge_device.id` instead of `updated_parent_id`

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: :+1:

We do technically have that list, since we only remove keys from the strays list if we find a match, and that's probably a better place to print them than at the top.

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: @varzac @kianooshkarami `process_strays` isn't recursive, but it's confusing because the function is declared inside of the coroutine so that it can create a closure over variables on the closure's stack if needed. The need for this has come and gone throughout refactors.

Currently it's closing over the `stray_dni_to_rid` map. If we can get (and keep) it to be referrentially pure then we should definitely move the definition to the module local scope to reduce this confusion cuz I have confused myself with it multiple times.

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: Good callout. This is a leftover from a time when the `parent_device_id` was accidentally getting set to the Hub's ID instead of the Hue Bridge during migration. That bug was fixed a long time ago and I wouldn't expect devices to be experiencing this anymore.

drivers/SmartThings/philips-hue/src/init.lua
varzac: I don't know for certain.  It's very possible that the fields aren't migrated directly to the new ones.  IMO, that could/should be fixed in the migration script.  But this is relatively harmless, and easier than re-working/testing the migration process.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -891,65 +836,35 @@ local stray_bulb_tx, stray_bulb_rx = cosock.channel.new()
...
+        if cached_light_description then
+          table.insert(dnis_to_remove, light_dni)
+          logged_migrate_light(driver, light_device, bridge_device_uuid, cached_light_description)
         end
-      end, "[process_strays]")
-      log.info_with({hub_logs = true}, "Finished querying bridge for devices from stray light handler")
+      end
+
       for _, dni in ipairs(dnis_to_remove) do
         strays[dni] = nil
```
dljsjr, line 858: I'm a little shaky on Lua's pass-by-reference semantics; this *will* mutably update the `stray_lights` map since that's what's beeing passed in, right? Or should this be closing over that stack variable?

drivers/SmartThings/philips-hue/src/init.lua
varzac: Yeah, understanding scope is just pretty hard on github.  Once I pulled this locally, this made sense.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -694,15 +690,8 @@ local function init_bridge(driver, device)
...
-        local timeout_time = cosock.socket.gettime() + duration
-        local time_remaining = duration
         local bridge_info, err
-
-        while time_remaining > 0 do
-          log.info_with({ hub_logs = true }, string.format(
-            "time remaining bridge init for %s" .. time_remaining, (device.label or device.id or "unknown device")
-          ))
+        while true do
           bridge_info, err = HueApi.get_bridge_info(ip)
```
varzac, line 695: What exactly is involved here?  This is a network operation correct? Just want to make sure this isn't a super tight loop on the continue when bridge_info is nil

dljsjr: Yeah this is an un-authed REST call so it doesn't take an API instance, it creates and then tears down a REST client under the hood.

Regardless of that, though, I do believe that we should probably have a (bounded) back-off here.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -259,8 +257,11 @@ bridge_added = function(driver, device)
     local known_macs = {}
     cosock.spawn(
       function()
+        local backoff_generator = utils.backoff_builder(10, 0.1, 0.1)
+        local sleep_time = 0.1
         while true do
           local bridge_found = false
+
```
kianooshkarami, line 264: Please add some form of log here.

kianooshkarami (approved):

PR #799, opened 2023-06-28
Refactor Hue Driver back to non-debug state

dljsjr:
## Overview
This PR preps Hue to get back in to its normal, non-debug state that we've been using for the EU Alpha users.

We guard against one small capability bug, and then we revise logging entirely to be suitable for a production driver once again.

## Commits
- fix: clamp hue/sat attr emit values to [0,100]: ad8490bcd40a1bebf4f6907dcf930584357ac461
- chore: Tidy up Hue driver logging: 3ccd9c36b6ddddb52739c567fa2deb3ed6f7fde7

### DIFF
--- a/drivers/SmartThings/philips-hue/src/hue/api.lua
+++ b/drivers/SmartThings/philips-hue/src/hue/api.lua
@@ -57,17 +57,17 @@ end
 
 local function process_rest_response(response, err, partial, err_callback)
   if err == nil and response == nil then
-    log.info_with({ hub_logs = true },
-    st_utils.stringify_table(
-      {
-        resp = response,
-        maybe_err = err,
-        maybe_partial = partial
-      },
-      "[PhilipsHueApi] Unexpected nil for both response and error processing REST reply",
-      false
+    log.error(
+      st_utils.stringify_table(
+        {
+          resp = response,
+          maybe_err = err,
+          maybe_partial = partial
+        },
+        "[PhilipsHueApi] Unexpected nil for both response and error processing REST reply",
+        false
+      )
     )
-  )
   end
   if err ~= nil then
     if type(err_callback) == "function" then err_callback(err) end
@@ -115,17 +115,16 @@ function PhilipsHueApi.new_bridge_manager(base_url, api_key, socket_builder)
       local msg, err = control_rx:receive()
       if err then
         if err ~= "timeout" then
-          log.error_with({ hub_logs = true }, "[PhilipsHueApi] Error receiving on control channel for REST API thread", err)
+          log.error("[PhilipsHueApi] Error receiving on control channel for REST API thread", err)
         else
-          -- TODO convert this to TRACE logs when debugging is over: dougstephen@smartthings.com
-          log.info_with({ hub_logs = true }, "Timeout on Hue API Control Channel, continuing")
+          log.trace("Timeout on Hue API Control Channel, continuing")
         end
         goto continue
       end
 
       if msg and msg._type then
         if msg._type == ControlMessageTypes.Shutdown then
-          log.info_with({ hub_logs = true }, "[PhilipsHueApi] REST API Control Thread received shutdown message");
+          log.info("[PhilipsHueApi] REST API Control Thread received shutdown message");
           self._running = false
           goto continue
         end
@@ -148,7 +147,7 @@ function PhilipsHueApi.new_bridge_manager(base_url, api_key, socket_builder)
           )
         end
       else
-        log.warn_with({ hub_logs = true },
+        log.warn(
           st_utils.stringify_table(msg, "[PhilipsHueApi] Unexpected Message on REST API Control Channel", false))
       end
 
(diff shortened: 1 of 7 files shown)


drivers/SmartThings/philips-hue/src/hue/api.lua
```diff
@@ -115,17 +115,16 @@ function PhilipsHueApi.new_bridge_manager(base_url, api_key, socket_builder)
       local msg, err = control_rx:receive()
       if err then
         if err ~= "timeout" then
-          log.error_with({ hub_logs = true }, "[PhilipsHueApi] Error receiving on control channel for REST API thread", err)
+          log.error("[PhilipsHueApi] Error receiving on control channel for REST API thread", err)
         else
-          -- TODO convert this to TRACE logs when debugging is over: dougstephen@smartthings.com
-          log.info_with({ hub_logs = true }, "Timeout on Hue API Control Channel, continuing")
+          log.trace("Timeout on Hue API Control Channel, continuing")
```
dljsjr, line 120: This log has ended up being a nice little heartbeat indicator but do we feel that there are problems with printing this to the live logs every 30 seconds?

drivers/SmartThings/philips-hue/src/utils.lua
```diff
@@ -64,10 +64,102 @@ function utils.backoff_builder(max, inc, rand)
...
+end
+
+-- This version of log_func_wrapper has slightly different
+-- characteristics and formatting than the one in the st/utils
+-- files.
+--
+-- Of note, this log_func_wrapper *also* sets hub_logs = true,
+-- so it should be used in chatty paths sparingly if using
+-- info level.
 function utils.log_func_wrapper(func, func_name, log_level)
```
varzac, line 159: is this still used in the driver at all?

drivers/SmartThings/philips-hue/src/utils.lua
```diff
@@ -64,10 +64,102 @@ function utils.backoff_builder(max, inc, rand)
   end
 end
 
+function utils.labeled_socket_builder(label)
```
varzac, line 67: Where was all this happening before?  What was this change meant to handle?

dljsjr: This is a direct copy over from the default socket builder in `rest.lua` (the function call `make_socket`).

@kianooshkarami had mused about being able to have more context in some of the low level rest calls so that we can know e.g. which bridge we're having failures on.

The socket builder itself isn't new functionality, all the REST stuff already uses it, we were just inlining the default one in the rest.lua module and now instead we create the function via this builder so that we can prefix it with a string label and get more context.

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/799/files#diff-7e9e351ebde6cef0a6366cf2bc69bd47ec79864fdf869ba8c8964502046c9f45L237

drivers/SmartThings/philips-hue/src/utils.lua
dljsjr: It is not. I figured it could be useful to preserve in case we want to start using it again.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -949,32 +985,45 @@ cosock.spawn(function()
...
+            utils.labeled_socket_builder((msg_device.label or msg_device.device_network_id or msg_device.id or "unknown bridge"))
+          )
           Discovery.disco_api_instances[msg_device.device_network_id] = api_instance
         end
 
         found_bridges[msg_device.id] = msg.device
         local bridge_device_uuid = msg_device.id
 
-        log.info_with({ hub_logs = true }, "Querying bridge for devices from stray light handler")
+        -- TODO: We can optimize around this by keeping track of whether or not this bridge
```
ctowns, line 999: I know this was mentioned at some point, but for my own understanding - is this where we would implement a backoff for `process_strays` if we decide to do that in the future?

ctowns (approved):
Looks good to me!

FreeMasen (approved):

drivers/SmartThings/philips-hue/src/utils.lua
varzac: Eh, I'd say gut it since the code exists in the main utils we can always pull it back in with the minor changes we use here.

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: Not necessarily; I was thinking we'd most likely do that check a bit further up in the `err == "timeout"` branch that starts on line 964, since that's the only place where we'd be scanning repeatedly in the event that everything else onboards properly.

cjswedes (approved):
This seems fine to me, but why is everything no longer being sent to hub core logs? I worry that without any hubCore logging that will be really hard to do investigation into hue. IMO it is worthwhile to at least send all the error logs to hub core, if not most of the discovery and initialization logging. However, many others here have looked at a bunch of hue logs recently and know better than me if this is useful.

dljsjr:
> This seems fine to me, but why is everything no longer being sent to hub core logs? I worry that without any hubCore logging that will be really hard to do investigation into hue. IMO it is worthwhile to at least send all the error logs to hub core, if not most of the discovery and initialization logging. However, many others here have looked at a bunch of hue logs recently and know better than me if this is useful.

We're mostly considering the investigation "resolved" at this point, in the sense that the version on Alpha right now seems to have achieve everything that's possible to achieve purely in the driver. We're prepping this for beta and prod at this point. We're going to roll out slowly and if we do see some stuff show back up we can re-introduce the logging to hubcore pretty easily.

PR #800, opened 2023-06-28
Release lan thing driver

greens:


drivers/SmartThings/lan-thing/src/init.lua
```diff
@@ -0,0 +1,12 @@
+local capabilities = require "st.capabilities"
+local Driver = require "st.driver"
+
+--should anything beyond refresh be added?
```
lelandblue, line 4: nit - do we want to remove this comment?

greens: We can do it later in a commit to main. This is just the PR to move to beta. It copies the existing commit.

ctowns (approved):

PR #807, opened 2023-06-30
Update Sonos version of Lunchbox with some fixes from Hue.

dljsjr:


kianooshkarami (approved):
LGTM at a high level and a quick cross reference with Hue. Thanks. I'd appreciate it if additional eyes were checked beside me as well.

drivers/SmartThings/sonos/src/lunchbox/rest.lua
```diff
@@ -165,9 +157,18 @@ local function handle_response(sock)
...
+    return nil, "Called `execute request` on a terminated REST Client"
+  end
 
-  local should_retry = retry_fn or function() return false end
+  if client.socket == nil then
+    local success, err = connect(client)
+    if not success then return nil, err end
+  end
+
+  if type(should_retry) ~= "function" then
```
ctowns, line 169: Looks like `should_retry` needs to be initalized somewhere, and I'm guessing this should be checking if `retry_fn()` is a function? I am somewhat confused how this wasn't caught in luacheck unless I'm looking at this wrong

drivers/SmartThings/sonos/src/utils.lua
```diff
@@ -0,0 +1,96 @@
...
+  local log = require "log"
+  local socket = require "cosock.socket"
+  local ssl = require "cosock.ssl"
+
+  label = (label or "")
+  if #label > 0 then
+    label = label .. " "
+  end
+
+  local function make_socket(host, port, wrap_ssl)
```
dljsjr, line 38: The only difference here between the Hue version of `make_socket` is that we don't set the TCP keepalive option. That's because Sonos only ever uses one-shot REST stuff to establish basic details and after that everything is websocket. So we don't want long-lived sockets.

drivers/SmartThings/sonos/src/utils.lua
```diff
@@ -0,0 +1,96 @@
...
+    log.info(
+      string.format(
+        "%sCreating TCP socket for REST Connection", label
```
ctowns, line 41: nit: the Hue equivalent for this has `Hue` in the log message (i.e. `Creating TCP socket for Hue REST Connection`). However, I think the label that's included in the logs is enough to distinguish this is Sonos

ctowns (commented):
LGTM based on my understanding of the changes called out in your write up and the changes be made in hue. Just one little potential change, but looks good!

drivers/SmartThings/sonos/src/utils.lua
dljsjr: Yeah I wanted to make it a little more general and use the label passed in. I wanna start unifying this code.

drivers/SmartThings/sonos/src/lunchbox/rest.lua
dljsjr: Good catch. Accidentally removed the local initializer during the translation.

ctowns (approved):

cjswedes (approved):
LGTM

PR #809, opened 2023-07-03
Release Sonos to Beta with some REST API hardening, and Hue with some hub-side logging

dljsjr:
This release takes some changes we made on the Alpha driver based on potential error surfaces discovered in the Hue driver as a pre-emptive hardening before migration.

Also re-enables some hub-side logging for Hue.


drivers/SmartThings/sonos/src/init.lua
```diff
@@ -29,7 +31,7 @@ local function find_player_for_device(driver, device, should_continue)
     local attempts = 0
     should_continue = function()
       attempts = attempts + 1
-      return attempts <= 10
+      return attempts <= DEFAULT_SSDP_RETRY_ATTEMPTS
```
cjswedes, line 34: Whats the reasoning for limiting the discovery attempts? If the max attempts are hit, a device will be non-functional because it failed to initialize; even if the device came back online, it wouldn't initialize until the driver restarted. IMO an exponential backoff would be preferred, but that is a relatively substantial change to be making directly to beta now.

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -36,7 +38,13 @@ local function find_player_for_device(driver, device, should_continue)
     --- @param ssdp_group_info SonosSSDPInfo
     SSDP.search(SONOS_SSDP_SEARCH_TERM, function(ssdp_group_info)
       driver:handle_ssdp_discovery(ssdp_group_info, function(dni, _, _, _)
+        log.info(string.format(
```
cjswedes, line 43: Logging this to hub logs could really help us during migration should we need to look into issues.

cjswedes (commented):
I think we should consider adding more logging for device init/discovery, and websocket connection maintenance logs to hub core logs.

drivers/SmartThings/sonos/src/init.lua
dljsjr: We override info in this file to use info_with

cjswedes: I missed that log was overridden, so this and many others are going to logs already.

cjswedes (approved):

drivers/SmartThings/sonos/src/init.lua
dljsjr: Agreed, we should make a ticket and revisit this. Don't wanna make that change going in to beta.

cjswedes: @dljsjr @ctowns and I discussed this and decided it is nice to have but because we have yet to see an issue in tests around this that it is not critical. We will re-evaluate after the first set of migrations.

PR #810, opened 2023-07-03
Bugfix/missing ge fingerprint

cjswedes:
This [device](https://products.z-wavealliance.org/products/2939?selectedFrequencyId=-1) joins as a basic switch with the GenericSwitch/1 fingerprint, it should have metering capabilities.

I am not sure if a manufacturer specific fingerprint or the added generic fingerprint is best. Both could be added, but I hope to get feedback on if one is preferred or if both can be added.


greens (commented):
Just add the generic FP. The `0x3032` device is not certified, only `0x3031`.

drivers/SmartThings/zwave-switch/fingerprints.yml
```diff
@@ -943,6 +943,14 @@ zwaveGeneric:
         - 0x60 # multi channel
         - 0x26
     deviceProfileName: multichannel-switch-level
+  - id: "MeteringSwitch"
```
tpmanley, line 946:
```suggestion
  - id: "GenericMeteringSwitch"
```

tpmanley:
Can you update the PR title to better capture what is now being changed?

greens (commented):
Wondering if we should add a metering dimmer FP while we're at it?

tpmanley (approved):

cjswedes:
@greens I added the fingerprint for metering dimmers. Ill wait for your approval before merging.

greens (approved):

PR #811, opened 2023-07-03
Add back some hue logging in hub logs for migration testing

ctowns:
Adds back some choice hue logging statements for better migration debugging.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/disco.lua
+++ b/drivers/SmartThings/philips-hue/src/disco.lua
@@ -31,7 +31,7 @@ process_discovered_light
 ---@param _ table
 ---@param should_continue function
 function HueDiscovery.discover(driver, _, should_continue)
-  log.info("Starting Hue discovery")
+  log.info_with({ hub_logs = true }, "Starting Hue discovery")
 
   while should_continue() do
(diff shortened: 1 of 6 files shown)


ctowns:
Would it still be worth it to find a way to print out some of the stray lights? I know that took up a lot of the logs previously, but I also know that was useful for figuring out if devices were real or not by checking for the id in the `lights/` group

kianooshkarami (approved):

drivers/SmartThings/philips-hue/src/disco.lua
```diff
@@ -31,7 +31,11 @@ process_discovered_light
 ---@param _ table
 ---@param should_continue function
 function HueDiscovery.discover(driver, _, should_continue)
-  log.info("Starting Hue discovery")
+  log.info_with({ hub_logs = true }, "Starting Hue discovery")
+  -- add logging for strays
+  -- err on more logging than not
+  -- cosock timeout 30 second heartbeat - see the PR
```
cjswedes: nit: comments should be removed

dljsjr (approved):

cjswedes (approved):

PR #824, opened 2023-07-06
Sonos Edge Driver networking reliability fixes

dljsjr:
Network conditions can cause some values to come in incomplete,
be missing entirely, or cause other such issues.

This PR adds some resiliency to this:

1. SSDP Scanning can now run indefinitely, to deal with delay and
congestion on some networks
2. We've added some nil checking in a few places where network
conditions can lead to nil values

### DIFF
--- a/drivers/SmartThings/sonos/src/init.lua
+++ b/drivers/SmartThings/sonos/src/init.lua
@@ -27,6 +27,7 @@ end
 
 local Driver = require "st.driver"
 
+local cosock = require "cosock"
 local capabilities = require "st.capabilities"
 local st_utils = require "st.utils"
 
@@ -38,6 +39,7 @@ local SonosDisco = require "disco"
 local SonosRestApi = require "api.rest"
 local SonosState = require "types".SonosState
 local SSDP = require "ssdp"
+local utils = require "utils"
 
 local DEFAULT_SSDP_RETRY_ATTEMPTS = 20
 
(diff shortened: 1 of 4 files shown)


drivers/SmartThings/sonos/src/init.lua
```diff
@@ -86,54 +88,66 @@ end
...
-
-    log.trace("Setting persistent fields")
-    device:set_field(PlayerFields.WSS_URL, fields.wss_url, { persist = true })
-    device:set_field(PlayerFields.HOUSEHOULD_ID, fields.household_id, { persist = true })
-    device:set_field(PlayerFields.PLAYER_ID, fields.player_id, { persist = true })
-
-    device:set_field(PlayerFields._IS_INIT, true)
-  end
+  if not device:get_field(PlayerFields._IS_SCANNING) then
+    device:set_field(PlayerFields._IS_SCANNING, true)
```
varzac, line 92: I don't see this ever marked false

varzac, line 92: If we get past this point, so this is marked true, but we fail to find the player before a driver restart, would we ever recover?

dljsjr: This should default to the transient store since there are no optional args, I would expect that this would re-run after a driver restart.

varzac: Ah!  right 👍

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -86,54 +88,66 @@ end
...
+          ((driver.found_ips and driver.found_ips[device.device_network_id])
+              or driver.sonos:is_player_joined(device.device_network_id)) and
+              driver._field_cache[device.device_network_id]
+
+          if not is_already_found then
+            log.debug("Rescanning for player with DNI " .. device.device_network_id)
+            device:offline()
+            local success = false
+
+            local backoff = utils.backoff_builder(360, 1, 0.5)
```
greens, line 107: Is there a reason to have randomness here? This is all hub local, right, so load shouldn't be a huge issue.

dljsjr: We're introducing this change because we've seen cases where 3 minutes (which was the derived time limit based on other parameters) was insufficient to find a speaker, so we know that it could take a *very* long time.

While it's not gonna be the end of the world if we do this, say, every 30 seconds, we're starting to see cases where power users have a ton of load on their hubs via drivers so every little saving helps IMO.

greens: Sure, but that last parameter adjusts the backoff value randomly. I'm not understanding the advantage of randomly having a shorter or longer backoff has over a deterministic one.

dljsjr: Copy/paste from another location, mostly, we can absolutely omit the randomization parameter here.

varzac (approved):

greens (approved):

drivers/SmartThings/sonos/src/init.lua
dljsjr: @greens This has been adjusted. Thanks for the call-out!

PR #828, opened 2023-07-09
Add fingerprint for GE 40amp embrighten switch with metering.

Felixowns:
I added a new ge enbrighten 40amp switch to smartthings and noticed the driver choices I had were to either provide switch control or only show metering information. My previous switch I had that migrated shows both the switch control and the metering. I looked at the smartthings adavanced web app and noticed a slight difference in product ID that might be the culprit? 

properly configured device:
image

new not properly configured:
image


Felixowns:
someone else already brought up this issue

PR #836, opened 2023-07-11
Bug fix:  avoid indexing a nil object

liuyd96:
It's possible that config is nil.
We should not use it without checking, otherwise we will hit nid index error.


greens (approved):

greens:
Thanks for the fix!

PR #839, opened 2023-07-11
fix: Don't re-detect profiles during migration

dljsjr:
Instead we rely on the static mapping. There have been reported issues with this code in the community and given that we have well established static mappings in the cloud now it's mostly not needed.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/disco.lua
+++ b/drivers/SmartThings/philips-hue/src/disco.lua
@@ -348,11 +348,12 @@ process_discovered_light = function(driver, bridge_id, resource_id, device_info,
     }
 
     HueDiscovery.light_state_disco_cache[light.id] = {
+      hue_provided_name = light.metadata.name,
       id = light.id,
       on = light.on,
       color = light.color,
       dimming = light.dimming,
-      color_temp = light.color_temperature,
+      color_temperature = light.color_temperature,
       mode = light.mode,
       parent_device_id = bridge_device.id,
       hue_device_id = light.owner.rid,
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -93,7 +93,16 @@ local function emit_light_status_events(light_device, light)
       local kelvin = math.floor(
         st_utils.clamp_value(handlers.mirek_to_kelvin(mirek), min, HueApi.MAX_TEMP_KELVIN)
       )
-      light_device:emit_event(capabilities.colorTemperature.colorTemperature(kelvin))
+      if utils.is_nan(kelvin) then
+        light_device.log.warn(
+          string.format(
+            "Non numeric value %s computed for colorTemperature Attribute Event, ignoring.",
+            kelvin
+          )
+        )
+      else
+        light_device:emit_event(capabilities.colorTemperature.colorTemperature(kelvin))
+      end
     end
 
     if light.color then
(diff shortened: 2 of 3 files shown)


greens (commented):
So the profile mapping is all done in the migration script now, so the profile-detecting code can be removed, but you've left in debug info because it's probably useful?

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -445,35 +445,29 @@ local function migrate_light(driver, device, parent_device_id, hue_light_descrip
...
-        , light_resource.id
-      )
-    )
-  end
+  local dbg_table = {
+    bulb_label = (device.label or device.id or "unknown identifier"),
+    profile_supports = {
+      switch = device:supports_capability_by_id(capabilities.switch.ID, nil),
+      switchLevel = device:supports_capability_by_id(capabilities.switchLevel.ID, nil),
+      colorTemperature = device:supports_capability_by_id(capabilities.colorControl.ID, nil),
```
ctowns, line 453: Looks like you need to swap the colorTemperature and colorControl mapping here
```suggestion
      colorTemperature = device:supports_capability_by_id(capabilities.colorTemperature.ID, nil),
```

ctowns (changes requested):

dljsjr:
> So the profile mapping is all done in the migration script now, so the profile-detecting code can be removed, but you've left in debug info because it's probably useful?

I would actually like to keep this profile detecting code around in some form, but right now we're seeing evidence that the Hue Bridge is reflecting capabilities that aren't what is in the profile from the static mapping; we're moving people to the wrong profile and dropping capabilities as a result.

So the idea is that we'll stop questioning the static mapping for now, and we leave the debug in to see if we can figure out what's going on w/ the Hue bridge not reporting all of the functionality for the bulbs all the time.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -97,8 +115,30 @@ local function emit_light_status_events(light_device, light)
...
 
-      light_device:emit_event(capabilities.colorControl.hue(st_utils.clamp_value(st_utils.round(hue * 100), 0, 100)))
-      light_device:emit_event(capabilities.colorControl.saturation(st_utils.clamp_value(st_utils.round(sat * 100), 0, 100)))
+      local adjusted_hue = st_utils.clamp_value(st_utils.round(hue * 100), 0, 100)
+      local adjusted_sat = st_utils.clamp_value(st_utils.round(sat * 100), 0, 100)
+
+      if utils.is_nan(adjusted_hue) then
+        light_device.log.warn(
+          string.format(
+            "Non numeric value %s computed for switchLevel Attribute Event, ignoring.",
```
varzac, line 124: Incorrect attribute name `switchLevel` should be `hue`

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -97,8 +115,30 @@ local function emit_light_status_events(light_device, light)
...
+          )
+        )
+      else
+        light_device:emit_event(capabilities.colorControl.hue(adjusted_hue))
+      end
+
+      if utils.is_nan(adjusted_sat) then
+        light_device.log.warn(
+          string.format(
+            "Non numeric value %s computed for switchLevel Attribute Event, ignoring.",
```
varzac, line 135: Another use of `switchLevel`

varzac (approved):
Approved pending test and log message corrections

drivers/SmartThings/philips-hue/src/disco.lua
```diff
@@ -348,11 +348,12 @@ process_discovered_light = function(driver, bridge_id, resource_id, device_info,
     }
 
     HueDiscovery.light_state_disco_cache[light.id] = {
+      hue_provided_name = light.metadata.name,
       id = light.id,
       on = light.on,
       color = light.color,
       dimming = light.dimming,
-      color_temp = light.color_temperature,
+      color_temperature = light.color_temperature,
```
varzac, line 356: how confident are we that this is correctly referred to everywhere now?

dljsjr: The only place it was being used incorrectly was in the `migrate_light` code path.

varzac (approved):

ctowns (approved):

PR #840, opened 2023-07-12
Reduce Sonos logging but keep important logs in hub logs

ctowns:
This removes all logs below info-level from the hub logs.

Note, a handful of the critical log messages we used for debugging will still be available:
* `Found device for Sonos search query with MAC addr %s, comparing to %s` - for MAC addr.  Will still be in hub logs, but the quantity should be significantly reduced thanks to https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/826
* `Received SSDP response for non-primary Sonos device in a bonded set, skipping; SSDP Response: %s` - for checking number of bonded devices
* `Looking for Sonos Player on network for device (%s:%s)` - for dni
* `Couldn't find Sonos Player [%s] during SSDP scan, trying again shortly` - for finding DNI of devices that are still trying to be matched

### DIFF
--- a/drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua
+++ b/drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua
@@ -1,4 +1,4 @@
-local old_log = require "log"
+local log = require "log"
 local cosock = require "cosock"
 local socket = require "cosock.socket"
 local channel = require "cosock.channel"
@@ -9,15 +9,6 @@ local Message = require "lustre".Message
 local CloseCode = require "lustre.frame.close".CloseCode
 local lb_utils = require "lunchbox.util"
 local st_utils = require "st.utils"
-
-local log = {}
-log.info = old_log.info
-log.warn = old_log.warn
-log.error = old_log.error
-log.debug = old_log.debug
-log.trace = function(...)
-  old_log.info_with({hub_logs = true}, table.concat({"[TRACE->INFO] ", ...}))
-end
 
 --- A "singleton" module that maintains all of the websockets for a
 --- home's Sonos players. A player as modeled in the driver will have
--- a/drivers/SmartThings/sonos/src/init.lua
+++ b/drivers/SmartThings/sonos/src/init.lua
@@ -1,29 +1,26 @@
 SONOS_API_KEY = require 'app_key'
 local old_log = require "log"
 
--- Forward all debug logs to hubcore info logs.
-old_log.debug = function(...)
-  old_log.info_with({hub_logs = true}, table.concat({"[DEBUG->INFO] ", ...}))
-end
-
--- Print all errors to hubcore for now
+-- Print all errors, warnings, and info to hubcore for now
 old_log.error = function(...)
   old_log.error_with({hub_logs = true}, ...)
 end
 
--- Print all warnings to hubcore for now
 old_log.warn = function(...)
   old_log.warn_with({hub_logs = true}, ...)
 end
 
+old_log.info = function(...)
+  old_log.info_with({hub_logs = true}, ...)
+end
+
 local log = {}
+
 log.warn = old_log.warn
 log.error = old_log.error
 log.debug = old_log.debug
 log.trace = old_log.trace
-log.info = function(...)
-  old_log.info_with({hub_logs = true}, ...)
-end
+log.info = old_log.info
 
 local Driver = require "st.driver"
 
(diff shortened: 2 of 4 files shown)


drivers/SmartThings/sonos/src/init.lua
```diff
@@ -1,26 +1,12 @@
...
 local old_log = require "log"
 
--- Forward all debug logs to hubcore info logs.
-old_log.debug = function(...)
-  old_log.info_with({hub_logs = true}, table.concat({"[DEBUG->INFO] ", ...}))
-end
-
--- Print all errors to hubcore for now
-old_log.error = function(...)
-  old_log.error_with({hub_logs = true}, ...)
```
cjswedes, line 11: I would prefer an approach where we can leave all error and warning logs going to hub logs. If we think the error/warning wouldn't actually be informative during an investigation or it would be too chatty, we should remove the log or reduce its level. As we solve crashes that are occuring on errors https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/841, I would like to ensure we still have some way to get insight to if errors are occuring.

dljsjr: I'm actually in support of leaving the driver-global error and warning overrides here in place for Warning and Error as well.

dljsjr (commented):
@ctowns  missed a spot: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/7c84e014a8fe1b47d310645267c43a6e9b38018e/drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua#L19

Also, I think we should leave the Error and Warning global overrides in place like @cjswedes said. The SSDP scanning path has a lot of warning prints, but with #826 merged the SSDP scanning path during migration should now be subject to the backoff every scan instead of every 30 scans.

## Stretch Goals

There are a few more logging changes I'd like to make, if you think it makes sense as well:

1. I wanna log all SSDP responses to the Sonos search term even if they don't have the proper headers, just like the Protocol Handler
2. `find_player_for_device` should use `device.log` (I wish I knew about *that* earlier :rage: )
3. I wanna add more identifying information around the bonded subordinate logging but I'm not sure how much additional context we have available in that scope without attempting a REST call, so that might be a non-starter.

Granted, the end goal here is to be expedient and low risk, so feel free to veto any of those if they don't seem legit.

drivers/SmartThings/sonos/src/init.lua
ctowns: Added driver-global error and warning logging overrides in cb1f83f1520e03b2374e7b07665335d85e6a76fe

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -1,24 +1,14 @@
 SONOS_API_KEY = require 'app_key'
 local old_log = require "log"
 
--- Forward all debug logs to hubcore info logs.
-old_log.debug = function(...)
-  old_log.info_with({hub_logs = true}, table.concat({"[DEBUG->INFO] ", ...}))
+-- Print all errors, warnings, and info to hubcore for now
+local log = {}
+log.warn = function(...)
```
dljsjr, line 6: For Info/Error/Warn overrides to be global, we should override "`old_log`'s" functions after `require` but before re-scoping it to the `local log` identifier.

drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua
```diff
@@ -10,14 +10,14 @@ local CloseCode = require "lustre.frame.close".CloseCode
 local lb_utils = require "lunchbox.util"
 local st_utils = require "st.utils"
 
+-- override module-wide logging to reduce chattiness
```
dljsjr, line 13: nit: Since we're not overriding anything in this module, I'm okay with getting rid of this and directly `require`ing `"log"` as `log` instead of `old_log`. But I'm also okay leaving this in as a hook to adjust later if needed. Your call.

dljsjr:
@ctowns With the global Info logging re-enabled, I think there's one spot where we may want to change from `log.info` to `log.trace`, and that's all the calls in `make_socket` that are tracing the REST client TCP socket bring-up:
 https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/89e35a4ab40cccc0b74a559f0056dcd4b6132a67/drivers/SmartThings/sonos/src/utils.lua#L38

```
🐧 ❯ rg -i '"%s[a-zA-Z]'

src/utils.lua
39-    log.info(
40-      string.format(
41:        "%sCreating TCP socket for REST Connection", label
42-      )
43-    )
--
50-    log.info(
51-      string.format(
52:        "%sSetting TCP socket timeout for REST Connection", label
53-      )
54-    )
--
60-    log.info(
61-      string.format(
62:        "%sConnecting TCP socket for REST Connection", label
63-      )
64-    )
--
70-      log.info(
71-        string.format(
72:          "%sCreating SSL wrapper for for REST Connection", label
73-        )
74-      )
--
80-      log.info(
81-        string.format(
82:          "%sPerforming SSL handshake for for REST Connection", label
83-        )
84-      )
```

drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua
ctowns: Good point - removed in 80f53ff47cdd202d26cdef7adba4a9c75b8b8e84

PR #841, opened 2023-07-12
Bugfix/sonos nil index

cjswedes:
Each crash signature was broken into its own commit for easier reviewing. These crashes represent the majority of the crashes we are seeing right now, but are all relatively low occurrences wrt the number of drivers actually running.

@dljsjr feel free to takeover the PR after any testing/approvals should it need to be merged prior to me doing so.


drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -93,14 +93,21 @@ local function _open_coordinator_socket(sonos_conn, household_id, self_player_id
 
   if coordinator_id ~= self_player_id then
     local coordinator = sonos_conn.driver.sonos:get_household(household_id).players[coordinator_id]
```
dljsjr, line 95: I think the wider issue is that we should be nil-checking `sonos_conn.driver.sonos:get_household(household_id)` or at least forcing the return of an empty table.

dljsjr:
@cjswedes We also need to nil check `sonos_conn` in the refresh handler; that's a source of a one-off nil crash but still one I'd like to see resolved: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/9bceb3368779264c425037a5a9309bd9ce5c2d53/drivers/SmartThings/sonos/src/init.lua#L188

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -92,15 +92,28 @@ local function _open_coordinator_socket(sonos_conn, household_id, self_player_id
   end
 
   if coordinator_id ~= self_player_id then
-    local coordinator = sonos_conn.driver.sonos:get_household(household_id).players[coordinator_id]
-    _, err = Router.open_socket_for_player(coordinator_id, coordinator.websocketUrl)
+    local household = self.driver.sonos:get_household(household_id)
```
dljsjr: `self` isn't available here to access the `sonos` instance, `sonos` is available on `sonos_conn.driver` though.

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -92,15 +92,28 @@ local function _open_coordinator_socket(sonos_conn, household_id, self_player_id
   end
```
dljsjr, line 92: Should we have an early return here?

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -186,6 +199,7 @@ function SonosConnection.new(driver, device)
         Router.cleanup_unused_sockets(self.driver)
 
         if not self:coordinator_running() then
+          --TODO this is not infallible
```
dljsjr, line 202: Is it worth checking this now? My gut is that this should be "eventually consistent" using the automatic reconnect semantics even without it but it's a good call-out.

drivers/SmartThings/sonos/src/api/sonos_connection.lua
cjswedes: Yeah, I updated the others, so I should probably update this one too. I dont think it makes a difference functionally though.

drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua
```diff
@@ -190,9 +195,9 @@ function SonosWebSocketRouter.register_listener_for_socket(listener,
     log.debug("Registering SonosConnection for device %s as listener for player_id's %s websocket", listener.device.label, player_id_for_socket)
   end
   local ws = websockets[player_id_for_socket]
-  ws._player_id = player_id_for_socket
 
   if ws ~= nil then
+    ws._player_id = player_id_for_socket
```
dljsjr, line 200: :facepalm:

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -184,6 +184,9 @@ end
 local function do_refresh(driver, device, cmd)
   log.trace("Refreshing " .. device.label)
   local sonos_conn = device:get_field(PlayerFields.CONNECTION)
+  if sonos_conn == nil then
+    log.error(string.format("Failed to do refresh, no sonos connection for device: [%s]", device.label))
```
dljsjr, line 188: Early return needed.

drivers/SmartThings/sonos/src/types.lua
```diff
@@ -220,6 +220,14 @@ function SonosState.new()
         role = "auxilary"
       end
 
+      if coordinator_id == "table" then
```
dljsjr, line 223: Should this be `type(coordinator_id) == "table"`?

drivers/SmartThings/sonos/src/api/sonos_connection.lua
cjswedes: I felt it was too big of a change functionally for what we are trying to accomplish in this PR.

drivers/SmartThings/sonos/src/types.lua
```diff
@@ -229,7 +237,13 @@ function SonosState.new()
   --- @param player_id PlayerId
   --- @return GroupId
   ret.get_group_for_player = function(self, household_id, player_id)
-    return private.households[household_id].player_to_group_map[player_id]
+    local household = private.households[household_id]
+    if household == nil then
+      local ut = require "st.utils"
+      log.error(ut.stringify_table({household = household}, "Get group for invalid household", false))
+      return
```
dljsjr, line 244: What's the Lua style here on implicit/explicit nil? I feel like this is fine as-is but my IDE is complaining about the empty return.

cjswedes: I usually do implicit nil returns ¯\_(ツ)_/¯

dljsjr (approved):

PR #842, opened 2023-07-12
add jbl driver

cchillerr:


drivers/SmartThings/jbl/profiles/jbl.yml
```diff
@@ -0,0 +1,30 @@
+name: jbl
+components: 
+- id: main
+  capabilities:
+  - id: mediaPlayback
+    version: 1
+    config:
+      values:
+        - key: "playbackStatus.value"
```
dljsjr, line 9: This means that the device will never be in the "Stopped" state, only the Playing or Paused states. Is that true?

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,216 @@
...
+--  either express or implied. See the License for the specific language governing permissions
+--  and limitations under the License.
+--
+--  ===============================================================================================
+
+local log = require "log"
+
+local capabilities = require "st.capabilities"
+local Driver = require "st.driver"
+local json = require "st.json"
```
dljsjr, line 19: Unused require.

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,216 @@
...
+local utils = require "st.utils"
```
dljsjr, line 20: Another unused require.

drivers/SmartThings/jbl/src/discovery_mdns.lua
```diff
@@ -0,0 +1,154 @@
...
+            ip = answer.kind.ARecord.ipv4
+        end
+
+        if ip ~= nil then
+            if dni == nil then
+                dni = driver.discovery_helper.get_dni(driver, ip, filtered_responses)
+            end
+    
+            if dni ~= nil then
+                target_table[dni] = ip
```
dljsjr, line 105: I would not recommend using the IP address as the Device Network Identifier:

1. The IP can change for a variety of reasons, such as in the case that DHCP is being used on the network,
2. The Device Network ID cannot be updated after the device is created.

The Device Network ID should be both unique and stable. Typically, for a LAN device, we use the MAC address if that's available.

If the MAC isn't available, it should be some other stable/unchanging and unique value.

drivers/SmartThings/jbl/src/discovery_mdns.lua
```diff
@@ -0,0 +1,154 @@
+local log = require "log"
+local mdns = require "st.mdns"
+local net_utils = require "st.net_utils"
+
+local discovery_mdns = {}
+
+local function byte_array_to_plain_text(byte_array)
```
dljsjr, line 7: This function will only work if every byte is an ASCII byte. Are we sure that there won't be any UTF-8 payloads getting passed in to this?

drivers/SmartThings/jbl/src/discovery_mdns.lua
```diff
@@ -0,0 +1,154 @@
...
+    insert_dni_ip_from_found(driver, filtered_responses, dni_ip_table)
+
+    return dni_ip_table
+end
+
+function discovery_mdns.find_ip_table_by_mdns(driver)
+    log.info("discovery_mdns.find_device_ips")
+
+    local service_type, domain = driver.discovery_helper.get_service_type_and_domain()
+    local discovery_responses = mdns.discover(service_type, domain) or {answers = {}, additional = {}}
```
dljsjr, line 147: This is not the correct return type structure for the `mdns.discover` API; this was the return behavior in an early version of the API but it change around hubcore 46.

The final version of the API that is available to production drivers can be found here: https://developer.smartthings.com/docs/edge-device-drivers/mdns.html

drivers/SmartThings/jbl/src/discovery.lua
```diff
@@ -0,0 +1,104 @@
...
+local discovery_mdns = require "discovery_mdns"
+
+local socket = require "cosock.socket"
+
+-- mapping from device DNI to info needed at discovery/init time
+local device_discovery_cache = {}
+
+local function set_device_field(driver, device)
+
+  log.info("set_device_field : dni = " .. tostring(device.device_network_id))
```
dljsjr, line 14: Two things here:

1. The DNI is already a string,
2. `string.format` is a bit more idiomatic here, and takes care of handling the `tostring` in the event that the value isn't a string.

drivers/SmartThings/jbl/src/discovery.lua
```diff
@@ -0,0 +1,104 @@
...
+  }
+end
+
+local function try_add_device(driver, device_dni, device_ip)
+  log.trace("try_add_device : dni=" .. tostring(device_dni) .. ", ip=" .. tostring(device_ip))
+
+  local credential  = driver.discovery_helper.get_credential(driver, device_dni, device_ip)
+
+  if not credential then
+    log.error("failed to get credential. dni=" .. device_dni .. ", ip=" .. device_ip)
```
dljsjr, line 39: Either use `tostring` or switch from concat to `string.format` here to avoid crashing if either value is nil.

drivers/SmartThings/jbl/src/discovery.lua
```diff
@@ -0,0 +1,104 @@
...
+  update_device_discovery_cache(driver, device_dni, device_ip, credential)
+  local create_device_msg = driver.discovery_helper.get_device_create_msg(driver, device_dni, device_ip)
+  driver:try_create_device(create_device_msg)
+end
+
+function discovery.device_added(driver, device)
+  log.info("device_added : dni = " .. tostring(device.device_network_id))
+  set_device_field(driver, device)
+  device_discovery_cache[device.device_network_id] = nil
+  driver.lifecycle_handlers.init(driver, device)
```
dljsjr, line 52: `init` is supposed to get called automatically after the added handler completes.

There *is* a bug where sometimes the `init` handler *doesn't* get called after `added`, and we are still root causing this; however, in our other land drivers, we use some extra state to make sure that we aren't calling the `added` and `init` handlers multiple times on a device.

If your `init` handler is idempotent, then it should be okay, but it's still doing extra work, and I would suggest adding similar checks like we have in our other drivers, to make sure you don't call `init` on a new device two times.

drivers/SmartThings/jbl/src/lunchbox/rest.lua
```diff
@@ -0,0 +1,403 @@
...
+    client.socket = nil
+    return false, err
+  end
+
+  client.socket = sock
+  return true
+end
+
+local function reconnect(client)
+  client.socket:close()
```
dljsjr, line 65: This needs a nil check; this was also a bug in the Hue and Sonos versions of Lunchbox that has been fixed there as well.

drivers/SmartThings/jbl/src/lunchbox/rest.lua
```diff
@@ -0,0 +1,403 @@
...
+  client.socket = nil
+  return connect(client)
+end
+
+local function send_request(client, request)
+  local payload = request:serialize()
+
+  local bytes, err, idx = nil, nil, 0
+  if client.socket == nil then
```
dljsjr, line 74: This check was also missing from the other versions of Lunchbox and has since been fixed. I'd recommend moving it to the top of the function so you can return right away without doing the work to serialize the request.

drivers/SmartThings/jbl/src/lunchbox/rest.lua
```diff
@@ -0,0 +1,403 @@
...
+    EXPECTING_CHUNK_LENGTH = "ExpectingChunkLength",
+    EXPECTING_BODY_CHUNK = "ExpectingBodyChunk",
+  }
+
+  local full_response = Response.new(original_response.status, nil)
+
+  for header in original_response.headers:iter() do full_response.headers:append_chunk(header) end
+
+  local original_body, err = original_response:get_body()
+  if not original_body or err ~= nil then
```
dljsjr, line 96: A bug that was fixed in our other versions of Lunchbox was to change this from checking `original_body`'s nil status, to it's type being explicitly a `string`. That might be useful here as well.

So instead of `not original_body` you would do `type(original_body) ~= "string"`, which will also catch `nil`.

drivers/SmartThings/jbl/src/lunchbox/rest.lua
```diff
@@ -0,0 +1,403 @@
...
+      response, recv_err, partial = nil, nil, nil
+      current_state = RestCallStates.SEND
+      socket.sleep(backoff())
+    end
+  until current_state == RestCallStates.COMPLETE
+
+  return ret, err
+end
+
+local function make_socket(host, port, wrap_ssl)
```
dljsjr, line 291: I would look at the updated version of this in Hue or Sonos, as there's a good bit of error handling that was missing that we added recently.

drivers/SmartThings/jbl/src/lunchbox/rest.lua
```diff
@@ -0,0 +1,403 @@
...
+function RestClient.new(base_url, sock_builder)
+  base_url = utils.force_url_table(base_url)
+
+  if type(sock_builder) ~= "function" then sock_builder = make_socket end
+
+  return
+    setmetatable({base_url = base_url, socket_builder = sock_builder, socket = nil}, RestClient)
+end
+
+return RestClient
```
dljsjr, line 403: I would recommend a general code review of this module in Hue and Sonos again, as it has changed quite a bit in the last month or two as we identified issues after migration. Not all of the changes will apply here, but some of the changes that we made such as moving a lot of the work on to a separate cosock task have been beneficial for us and may be beneficial for this driver as well.

drivers/SmartThings/jbl/src/lunchbox/sse/eventsource.lua
```diff
@@ -0,0 +1,500 @@
...
+--- @param source EventSource
+local function connecting_action(source)
+  if not source._sock then
+    if type(source._sock_builder) == "function" then
+      source._sock = source._sock_builder()
+    else
+      source._sock, err = socket.tcp()
+      if err ~= nil then return nil, err end
+
+      _, err = source._sock:connect(source.url.host, source.url.port)
```
dljsjr, line 251: I would recommend setting a timeout on this socket. We made a similar change in our versions of Lunchbox and added logic to the state machine processing loop to not treat timeouts as fatal errors.

drivers/SmartThings/jbl/src/jbl/api.lua
```diff
@@ -0,0 +1,165 @@
...
+    options = "all",
+    cafile="./selfSignedRoot.crt"
+}
+
+local ADDITIONAL_HEADERS = {
+    ["Accept"] = "application/json",
+    ["Content-Type"] = "application/json",
+}
+
+function jbl_api.socket_builder(host, port, wrap_ssl)
```
dljsjr, line 27: Similar to the comment on in the Event Source, it's strongly encouraged that a timeout be set on any socket that can end up in `receive` position for select to avoid yielding a coroutine and never resuming.

drivers/SmartThings/jbl/src/jbl/api.lua
```diff
@@ -0,0 +1,165 @@
...
+
+local function get_base_url(device_ip)
+    return "https://" .. device_ip .. ":4443"
+end
+
+local function process_rest_response(response, err, partial)
+    if err ~= nil then
+        return response, err, nil
+    elseif response ~= nil then
+        return json.decode(response:get_body()), nil, response.status
```
dljsjr, line 60: This was heavily refactored on our branch because `json.decode` can throw a Lua `error` which will crash the driver. I would suggest doing the same here.

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    local conn_info = device:get_field(fields.CONN_INFO)
+    log.info(string.format("media-playback.set_playback_status_handler : dni = %s, status = %s", device.device_network_id, capability_status))
+
+    local jbl_playback_status = st_status_to_jbl_playback_status_table[capability_status]
+
+    local response, err, status = conn_info:post_playback(string.format('{"playback": "%s"}', jbl_playback_status))
+    if not err and status == 200 then
+        device:emit_event(capabilities.mediaPlayback.playbackStatus[capability_status]())
```
dljsjr, line 24: By convention for LAN devices, we try to avoid doing a "synthetic" emit on success for stuff like this. The appropriate design pattern would be to send the command, and then wait for the appropriate state change to come in on the server sent event stream asynchronously.

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
```diff
@@ -0,0 +1,138 @@
...
+    elseif status == 404 then
+        log.error("404 error. delete device. dni = " .. tostring(device.device_network_id))
+        device:offline()
```
dljsjr, line 27: Do we always want to set the device offline in the case of a 404? The device monitoring thread would eventually set it back online, but could there be other reasons for a 404 to come out of the REST API?

drivers/SmartThings/jbl/src/jbl/device_manager.lua
```diff
@@ -0,0 +1,145 @@
...
+        and (audioTrackData.album == latestTrackData.album)
+        and (audioTrackData.albumArtUrl == latestTrackData.albumArtUrl)
+        and (audioTrackData.mediaSource == latestTrackData.mediaSource) then
+        return false
+    end
+
+    return true
+end
+
+local jbl_playback_state_to_smartthings_playbck_status_table = {
```
dljsjr, line 30: Just for the sake of clarity, the profile YAML for the speaker says that it only supports "playing" and "paused" so do we need the other elements in the table here? If the driver tries to emit an event for something other than "playing" or "paused" it will get rejected before it makes it to the cloud platform because of the profile.

drivers/SmartThings/jbl/src/jbl/discovery_helper.lua
```diff
@@ -0,0 +1,97 @@
...
+        profile = "jbl",
+        manufacturer = device_info.manufacturerName,
+        model = device_info.modelName,
+        vendor_provided_label = device_info.label,
+    }
+
+    return create_device_msg
+end
+
+function discovery_helper.get_credential(driver, bridge_dni, bridge_ip)
```
dljsjr, line 66: I know it's mentioned elsewhere but I wanted to reiterate that using the bridge_ip as a lookup mechanism is quite fragile.

dljsjr (changes requested):
There are some changes that need to be addressed.

The biggest issue is that the mDNS usage isn't correct.

There are other issues that are mostly defensive, the code would work on the happy path but will crash if anything is even a little bit malformed.

Luacheck is also not passing. You can run Luacheck locally if you install it from Luarocks and then point it to the config file that's in the repo's `.github` directory.

Also since this PR has been hanging around for a while there are conflicts with `main`, this branch needs to be rebased on `main` before continuing.

drivers/SmartThings/jbl/profiles/jbl.yml
cchillerr: Yes, that's the way, JBL bar only support Play and pause, no "stopped" state.

drivers/SmartThings/jbl/src/init.lua
cchillerr: Done, unused require removed

drivers/SmartThings/jbl/src/init.lua
cchillerr: Done, unused require removed

drivers/SmartThings/jbl/src/init.lua
cchillerr: Done, unused require removed

drivers/SmartThings/jbl/src/discovery.lua
cchillerr: change to

> log.info(string.format("set_device_field : %s", device.device_network_id))

drivers/SmartThings/jbl/src/discovery.lua
cchillerr: Done, by checking device:get_field(fields._INIT)

drivers/SmartThings/jbl/src/discovery.lua
cchillerr: use string.format instead of tostring

drivers/SmartThings/jbl/src/lunchbox/rest.lua
cchillerr: add nil check

drivers/SmartThings/jbl/src/lunchbox/rest.lua
cchillerr: done

>   if client.socket == nil then
    return nil, "no socket available"
  end

drivers/SmartThings/jbl/src/lunchbox/rest.lua
cchillerr: >   if type(original_body) ~= "string" or err ~= nil then
    return original_body, (err or "unexpected nil in error position")
  end

drivers/SmartThings/jbl/src/lunchbox/rest.lua
cchillerr: update this file to latest version in sample

drivers/SmartThings/jbl/src/lunchbox/sse/eventsource.lua
cchillerr: update this file to latest version in sample

drivers/SmartThings/jbl/src/jbl/api.lua
cchillerr: change to

> function jbl_api.labeled_socket_builder(label)
    local socket_builder = utils.labeled_socket_builder(label, SSL_CONFIG)
    return socket_builder
end

drivers/SmartThings/jbl/src/jbl/api.lua
cchillerr: change to        

> local status, decoded_json = pcall(json.decode, response:get_body())

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
cchillerr: device:emit_event removed

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
cchillerr: For the moment, we don't have other case that could produce 404 error.

drivers/SmartThings/jbl/src/lunchbox/rest.lua
cchillerr: Ok thanks.

Seong-Yeol-Park:
@dljsjr ( cc: @cchillerr )
Hi, I'm testing JBL BAR 300 with this Driver.

Onboarding failed when using Invitation URL from this PR:
https://bestow-regional.api.smartthings.com/invite/1PlYJgXeKyje

But When I package driver by myself from https://github.com/cchillerr/SmartThingsEdgeDrivers, 
I succeed in onboarding.

So Could you guess root cause about this happening?

dljsjr:
@Seong-Yeol-Park @cchillerr would you be able to provide edge driver live logs and hubcore logs from discovery using the two different driver sources so we can see if there are any differences? Thanks!

Seong-Yeol-Park:
> @Seong-Yeol-Park @cchillerr would you be able to provide edge driver live logs and hubcore logs from discovery using the two different driver sources so we can see if there are any differences? Thanks!

@dljsjr I'll share logs tomorrow for you investigation. Thank you.

drivers/SmartThings/jbl/src/fields.lua
```diff
@@ -0,0 +1,16 @@
+--- Table of constants used to index in to device store fields
+--- @module "fields"
+--- @class table
+--- @field IPV4 string the ipV4 address of the device
+
+local fields = {
```
dljsjr, line 6: Style nit: `fields` should be capitalized. Won't hold up review for this, but would be nice to address if time permits.

drivers/SmartThings/jbl/src/discovery_mdns.lua
```diff
@@ -0,0 +1,150 @@
+local log = require "log"
+local mdns = require "st.mdns"
+local net_utils = require "st.net_utils"
+
+local discovery_mdns = {}
+
+local function byte_array_to_plain_text(byte_array)
```
dljsjr, line 7: Not as concerned about the ASCII stuff, but I would advocate for simplifying this to use as much Lua built-in functionality as possible. You can replace this with:

```lua
string.char(table.unpack(byte_array))
```

[Lua's `string.char` function is variadic](https://www.lua.org/manual/5.0/manual.html#5.3).

drivers/SmartThings/jbl/src/discovery_mdns.lua
```diff
@@ -0,0 +1,150 @@
...
+    end
+
+    local answer_text = find_text_in_answers_by_ip(ip, discovery_responses)
+    for _, text_item in pairs(answer_text or {}) do
+        table.insert(text_list, text_item)
+    end
+    return text_list
+end
+
+local function filter_response_by_servie_name(service_type, domain, discovery_responses)
```
dljsjr, line 67: typo: `servie`

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,221 @@
...
+  local new_timer = device.thread:call_on_schedule(monitoring_interval, function()
+    check_and_update_connection(driver, device)
+    driver.device_manager.device_monitor(driver, device, device_info)
+    end, "monitor_timer")
+  device:set_field(fields.MONITORING_TIMER, new_timer)
+end
+
+local function refresh(driver, device, cmd)
+  log.info("refresh : dni =  " .. tostring(device.device_network_id))
+  check_and_update_connection(driver, device)
```
dljsjr, line 138: It might be a good idea to use `cosock.spawn` or `device.thread:call_with_delay` to perform this on a one-shot background task since this function performs mDNS operations and we don't want to hang the device thread while handling the refresh.

drivers/SmartThings/jbl/src/discovery_mdns.lua
```diff
@@ -0,0 +1,150 @@
...
+
+    local service_type, domain = driver.discovery_helper.get_service_type_and_domain()
+    local discovery_responses = mdns.discover(service_type, domain) or {found = {}}
+
+    local dni_ip_table = get_dni_ip_table_from_mdns_responses(driver, service_type, domain, discovery_responses)
+
+    return dni_ip_table
+end
+
+return discovery_mdns
```
dljsjr, line 150: nit: no newline at end of file

drivers/SmartThings/jbl/src/discovery.lua
```diff
@@ -0,0 +1,104 @@
+local log = require "log"
+local discovery = {}
```
dljsjr, line 2: nit: move the module declaration below the requires.

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,221 @@
...
+    log.trace("Creating SSE EventSource for " .. device.device_network_id .. ", sse_url = " .. sse_url)
+    local eventsource = EventSource.new(sse_url, {[CREDENTIAL_KEY_HEADER] = credential}, nil)
+
+    eventsource.onmessage = function(msg)
+      if msg then
+        handle_sse_event(driver, device, msg)
+      end
+    end
+
+    eventsource.onerror = function(msg)
```
dljsjr, line 58: nit: This `msg` will always be `nil` (the EventSource spec doesn't propagate an error message on unexpected close)

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,221 @@
...
+
+  if driver.device_manager.is_valid_connection(driver, device, conn_info) then
+    device:set_field(fields.CONN_INFO, conn_info)
+
+    create_sse(driver,device, credential)
+  end
+end
+
+
+local function find_new_connetion(driver, device)
```
dljsjr, line 95: type: `connetion`

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,221 @@
...
+    find_new_connetion(driver, device)
+    conn_info = device:get_field(fields.CONN_INFO)
+  end
+
+  if driver.device_manager.is_valid_connection(driver, device, conn_info) then
+    device:online()
+  end
+end
+
+local function create_monitoring_thread(driver, device, device_info)
```
dljsjr, line 119: You'll see better performance and responsiveness if you can create a single task for monitoring for connection changes and fanning those changes out, instead of creating a thread-per-device. The hub environments have a single physical thread and background tasks are scheduled using cooperative multitasking as opposed to pre-emptive multitasking. So there is non-trivial latency and memory overhead when using a lot of background tasks.

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,221 @@
...
+  device:set_field(fields.MONITORING_TIMER, new_timer)
+end
+
+local function refresh(driver, device, cmd)
+  log.info("refresh : dni =  " .. tostring(device.device_network_id))
+  check_and_update_connection(driver, device)
+  driver.device_manager.refresh(driver, device)
+end
+
+local function device_removed(driver, device)
```
dljsjr, line 142: Since you are already capturing the spawned timers in fields, it would be beneficial to tear those down on removal as well.

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,221 @@
...
+  local credential = device:get_field(fields.CREDENTIAL)
+
+  if not credential then
+    log.error("failed to find credential.")
+    device:offline()
+    return
+  end
+
+  log.trace("Creating device monitoring for " .. device.device_network_id)
+  create_monitoring_thread(driver, device, device_info)
```
dljsjr, line 174: See https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/842#discussion_r1316311114

drivers/SmartThings/jbl/src/jbl/api.lua
```diff
@@ -0,0 +1,138 @@
...
+
+local function get_base_url(device_ip)
+    return "https://" .. device_ip .. ":4443"
+end
+
+local function process_rest_response(response, err, partial)
+    if err ~= nil then
+        return response, err, nil
+    elseif response ~= nil then
+        local _, decoded_json = pcall(json.decode, response:get_body())
```
dljsjr, line 38: The `pcall` result should be checked before using the returned value.

dljsjr (changes requested):
Most of my comments here are style nits.

I would like to have the `pcall` comment addressed before merging, though.

drivers/SmartThings/jbl/src/discovery_mdns.lua
```diff
@@ -0,0 +1,150 @@
...
+local function get_text_by_srvname(srvname, discovery_responses)
+    for _,answer_item in pairs(discovery_responses.answers or {}) do
+        if answer_item.kind.TxtRecord ~= nil and answer_item.name == srvname then
+            return answer_item.kind.TxtRecord.text
+        end
+    end
+end
+
+local function get_srvname_by_hostname(hostname, discovery_responses)
+    for _,answer_item in pairs(discovery_responses.answers or {}) do
```
greens, line 24: nit: we typically use 2 spaces instead of tabs

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,221 @@
+--  Copyright 2022 SmartThings
```
greens, line 1: nit: date (I think? This PR isn't _that_ old is it?)

lelandblue:
Hey @cchillerr - I hope that you are well. 

Can you please review the recommendations to address on this PR and let us know if you have any questions? Looking forward to your contributions and thank you for your efforts.

chillcyj:
Hi @lelandblue 
I have reviewed the new PR, the code is ok except we have just applied a new patch to improve the performance, it's better to include also this change into this PR.

varzac:
This pr still needs to bring in the changes from the main repo as it is quite old and missing the latest CI changes.  Until those changes are included the automatic channel driver uploads won't work correctly.  In addition, the lua check is pointing out 2 unused variables that should be removed.

Because this change is being requested from an outside fork, we do need to manually approve the CI runs, so let us know when changes are made so we can approve the run.

drivers/SmartThings/jbl/src/init.lua
dljsjr: @cchillerr

drivers/SmartThings/jbl/src/init.lua
dljsjr: @cchillerr This should be changed to a function with no arguments. EventSource errors don't have an error message unfortunately (as part of the spec)

drivers/SmartThings/jbl/src/init.lua
dljsjr: @cchillerr This is still a place where there's potential to have the main driver thread yield, I'd like to see this addressed.

drivers/SmartThings/jbl/src/init.lua
dljsjr: @cchillerr

drivers/SmartThings/jbl/src/discovery.lua
```diff
@@ -0,0 +1,104 @@
...
+
+  for dni, ip in pairs(known_discovered_devices) do
+    log.trace(string.format("known dni=%s, ip=%s", dni, ip))
+  end
+
+  for dni, ip in pairs(unknown_discovered_devices) do
+    log.trace(string.format("unknown dni=%s, ip=%s", dni, ip))
+    if not device_discovery_cache[dni] then
+      try_add_device(driver, dni, ip)
+    end
```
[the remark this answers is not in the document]
dljsjr: @cchillerr

lelandblue:
Hello - @cchillerr 👋  Can you please review the open mentions on the comments above?  We would like to see a resolution to this before the PR is merged. 

Thank you

drivers/SmartThings/jbl/src/fields.lua
cchillerr: Done

drivers/SmartThings/jbl/src/discovery_mdns.lua
cchillerr: Already fixed

drivers/SmartThings/jbl/src/init.lua
cchillerr: Done by using device.thread:call_with_delay

drivers/SmartThings/jbl/src/discovery_mdns.lua
cchillerr: Done

drivers/SmartThings/jbl/src/discovery.lua
cchillerr: Done

drivers/SmartThings/jbl/src/init.lua
cchillerr: Fixed

drivers/SmartThings/jbl/src/init.lua
cchillerr: fixed

drivers/SmartThings/jbl/src/init.lua
cchillerr: function removed

drivers/SmartThings/jbl/src/init.lua
cchillerr: changed to 2023

drivers/SmartThings/jbl/src/discovery_mdns.lua
cchillerr: done

drivers/SmartThings/jbl/src/jbl/api.lua
cchillerr: done

drivers/SmartThings/jbl/src/discovery.lua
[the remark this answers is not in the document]
cchillerr: Simplified

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
```diff
@@ -0,0 +1,140 @@
...
+    end
+end
+
+function capability_handler.set_volume_handler(driver, device, args)
+    local volume = args.args.volume
+    local conn_info = device:get_field(fields.CONN_INFO)
+    log.info(string.format("audio_volume.set_volume : dni = %s, volume = %d", device.device_network_id, volume))
+
+
+    local _, err, status = conn_info:post_volume(string.format('{"volume": %d}', volume))
```
FreeMasen, line 69:
```suggestion
    local _, err, status = conn_info:post_volume(json.encode( { volume = volume }))
```

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
```diff
@@ -0,0 +1,140 @@
...
+    end
+end
+
+function capability_handler.audioNotification_handler(driver, device, args)
+    local uri = args.args.uri
+    local level = args.args.level
+    local conn_info = device:get_field(fields.CONN_INFO)
+
+    log.info(string.format("%s, %s : level = %s", args.capability, args.command, device.device_network_id, level))
+    log.info(string.format("URI: %s", uri))
```
FreeMasen, line 84: nit: these logs are a bit complicated and using %s generates the same as just providing the value so

```
log.info(args.capability, device.device_network_id, "level:", level)
log.info("URI:", uri)
```
is going to give you very similar output w/o having these super long format calls. An alternative would be to use the `%q` format argument which will quote strings and also print `nil` instead `""` which can be helpful sometimes

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
```diff
@@ -0,0 +1,140 @@
...
+
+    local payload_table = {
+        ["uri"] = uri,
```
FreeMasen, line 87: nit: No reason not to use `uri = uri` here

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    if level then
+        payload_table["volume"] = tostring(level)
```
FreeMasen, line 91: I was assuming that `level` was a number, does the API require the parameter to be a string? it might be good to have a comment here about the requirement that it get converted to a string if not `nil`

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
```diff
@@ -0,0 +1,140 @@
...
+    local conn_info = device:get_field(fields.CONN_INFO)
+    local st_state_to_jbl_muted_state_table = {
+        muted = "mute",
+        unmuted = "unmute",
+    }
+    log.info(string.format("smartthings_audioMute_capability_handler dni = %s, status = %s", device.device_network_id, mute_state))
+
+    local jbl_mute_state = st_state_to_jbl_muted_state_table[mute_state]
+
+    local _, err, status = conn_info:post_playback(string.format('{"mute": "%s"}', jbl_mute_state))
```
FreeMasen, line 118:
```suggestion
    local _, err, status = conn_info:post_playback(json.encode({ mute = jbl_mute_state }))
```

drivers/SmartThings/jbl/src/jbl/capability_handler.lua
```diff
@@ -0,0 +1,140 @@
...
+
+function capability_handler.mute_handler(driver, device, args)
+    smartthings_audioMute_capability_handler(driver, device, "muted")
+end
+
+function capability_handler.unmute_handler(driver, device, args)
+    smartthings_audioMute_capability_handler(driver, device, "unmuted")
+end
+
+return capability_handler
```
FreeMasen, line 140: nit: new line

drivers/SmartThings/jbl/src/jbl/device_manager.lua
```diff
@@ -0,0 +1,144 @@
...
+local json = require "st.json"
+local fields = require "fields"
+
+local capabilities = require "st.capabilities"
+
+local device_manager = {}
+device_manager.__index = device_manager
+
+local function is_new_audioTrackData(device, audioTrackData)
+    local latestTrackData = device:get_latest_state("main", capabilities.audioTrackData.ID, capabilities.audioTrackData.audioTrackData.NAME, "")
```
FreeMasen, line 11: Since this is expected to be an object if populated, I think `{}` or `nil` is more appropriate for the default value

drivers/SmartThings/jbl/src/jbl/device_manager.lua
```diff
@@ -0,0 +1,144 @@
...
+    end
+
+    return true
+end
+
+local jbl_playback_state_to_smartthings_playback_status_table = {
+    paused = "paused",
+    playing = "playing",
+    stopped = "stopped",
+}
```
FreeMasen, line 32: This doesn't appear to be mapping between anything?

drivers/SmartThings/jbl/src/jbl/device_manager.lua
```diff
@@ -0,0 +1,144 @@
...
+
+function device_manager.handle_status(driver, device, status)
+    if not status then
+        log.error("device_manager.handle_status : status is nil")
+        return
+    end
+
+    local playback_status = jbl_playback_state_to_smartthings_playback_status_table[status.playback]
+    if playback_status and playback_status ~= device:get_latest_state("main", capabilities.mediaPlayback.ID, capabilities.mediaPlayback.playbackStatus.NAME, "") then
```
FreeMasen, line 41: nit: line length here is a big gnarly (throughout this function)

drivers/SmartThings/jbl/src/jbl/device_manager.lua
FreeMasen: Looking further on, it seems like a "set" table here would be enough.

```suggestion
local valid_playback_status = {
    paused = true,
    playing = true,
    stopped = true
}
```

```lua
if not valid_playback_status[status.playback] then
  return
end
```

drivers/SmartThings/jbl/src/jbl/device_manager.lua
```diff
@@ -0,0 +1,144 @@
...
+        audioTrackData.artist = status.track.artist or ""
+        audioTrackData.album = status.track.album or ""
+        if status.track.mediaSource then
+            audioTrackData.mediaSource = status.track.mediaSource.name
+        else
+            audioTrackData.mediaSource = ""
+        end
+
+        if is_new_audioTrackData(device, audioTrackData) then
+            log.info("device_manager.handle_status : update audioTrackData = " .. tostring(json.encode(audioTrackData)) .. ", dni = " .. tostring(device.device_network_id) .. tostring(json.encode(status)))
```
FreeMasen, line 69: This is a lot of work just for a log message, is all this information needed?

drivers/SmartThings/jbl/src/jbl/device_manager.lua
```diff
@@ -0,0 +1,144 @@
...
+    if err or status ~= 200 then
+        log.error(" device_manager.is_valid_connection : failed to connect to device, dni = " .. tostring(device.device_network_id) .. ", err = " .. tostring(err) .. ", status = " .. tostring(status))
+        return false
+    end
+
+    return true
+end
+
+function device_manager.device_monitor(driver, device, device_info)
+    --TODO: add device monitoring logic (ip change, online/offline, etc ..)
```
FreeMasen, line 136: This seems like a pretty big todo, I just want to confirm it is still undone?

drivers/SmartThings/jbl/src/jbl/device_manager.lua
```diff
@@ -0,0 +1,144 @@
...
+    log.info("device_monitor = " .. tostring(device.device_network_id))
+    device_manager.refresh(driver, device)
+end
+
+function device_manager.get_sse_url(driver, device, conn_info)
+    return conn_info:get_sse_url()
+end
```
FreeMasen, line 143: nit: new line between statements

drivers/SmartThings/jbl/src/jbl/discovery_helper.lua
```diff
@@ -0,0 +1,97 @@
...
+
+local jbl_api = require "jbl.api"
+local discovery_mdns = require "discovery_mdns"
+
+function discovery_helper.get_dni(driver, ip, discovery_responses)
+    local text_list = discovery_mdns.find_text_list_in_mdns_response(driver, ip, discovery_responses)
+    for _, text in ipairs(text_list) do
+        for key, value in string.gmatch(text, "(%S+)=(%S+)") do
+            if key == "mac" then
+                log.info("discovery_helper.get_dni : use mac as dni, dni = " .. value)
```
FreeMasen, line 31: I don't think we want to log the device's mac address?

drivers/SmartThings/jbl/src/jbl/discovery_helper.lua
```diff
@@ -0,0 +1,97 @@
...
+
+function discovery_helper.get_service_type_and_domain()
+    return SERVICE_TYPE, DOMAIN
+end
+
+function discovery_helper.get_device_create_msg(driver, device_dni, device_ip)
+    local device_info = jbl_api.get_info(device_ip, jbl_api.labeled_socket_builder(device_dni))
+
+    if not device_info then
+        log.error("failed to create device create msg. device_info is nil. dni = " .. device_dni)
```
FreeMasen, line 49: no need to concat here
```suggestion
        log.error("failed to create device create msg. device_info is nil. dni = ", device_dni)
```

drivers/SmartThings/jbl/src/log.lua
```diff
@@ -0,0 +1,56 @@
...
+  original_log.info_with({hub_logs = true}, ...)
+end
+
+hubcore_enabled_log.warn = function(...)
+  original_log.warn_with({hub_logs = true}, ...)
+end
+
+hubcore_enabled_log.error = function(...)
+  original_log.error_with({hub_logs = true}, ...)
+end
```
FreeMasen, line 36: This seems like it would always be pushing data into the hub-core logs which I don't think we want to do.

drivers/SmartThings/jbl/src/discovery_mdns.lua
cjswedes: not done, still looks like 4 spaces

drivers/SmartThings/jbl/src/init.lua
cjswedes: was this change missed in the pr update?

drivers/SmartThings/jbl/src/init.lua
cjswedes: I dont see any timers being cancelled.

drivers/SmartThings/jbl/src/log.lua
cjswedes: I suggest limiting to only warn/error going to hub logs, and only if we think that is needed.

cjswedes (commented):
Overall LGTM with the assumption that there has been decent testing with the device. I'll defer to my approval to @dljsjr since he has been the most involved in the review. Looking over the previous comments there seem to be some that were resolved but dont have the changes so I commented on those.

drivers/SmartThings/jbl/src/init.lua
cchillerr: image
https://github.com/cchillerr/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/jbl/src/init.lua 
It seems the code is not up to date here?

drivers/SmartThings/jbl/src/discovery_mdns.lua
cchillerr: change tab from 4 spaces to 2 spaces

drivers/SmartThings/jbl/config.yml
```diff
@@ -0,0 +1,7 @@
+name: 'JBL'
+packageKey: 'jbl'
+permissions:
+  lan: {}
+  discovery: {}
+description: "SmartThings driver for JBL devices"
+vendorSupportInformation: "https://support.smartthings.com"
```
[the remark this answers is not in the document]
cchillerr: done

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,220 @@
...
+    device_manager = jbl_device_manager,
+    controlled_devices = {},
+  }
+)
+
+lan_driver:call_on_schedule(CONNECTION_MONITORING_INTERVAL, connection_monitoring, "JBL Connection monitoring thread")
+
+log.info("Starting lan driver")
+lan_driver:run()
+log.warn("lan driver exiting")
```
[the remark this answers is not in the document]
cchillerr: done

drivers/SmartThings/jbl/src/jbl/api.lua
```diff
@@ -0,0 +1,143 @@
...
+jbl_api.__index = jbl_api
+
+local CREDENTIAL_TIME_OUT_SECONDS = 30
+
+local SSL_CONFIG = {
+    mode = "client",
+    protocol = "any",
+    verify = "peer",
+    options = "all",
+    cafile="./selfSignedRoot.crt"
```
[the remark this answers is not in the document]
cchillerr: done

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,220 @@
...
+
+  if not driver.device_manager.is_valid_connection(driver, device, conn_info) then
+    log.error("create_sse : invalid connection")
+    return
+  end
+
+  local sse_url = driver.device_manager.get_sse_url(driver, device, conn_info)
+  if not sse_url then
+    log.error("failed to get sse_url")
+  else
```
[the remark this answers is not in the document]
cchillerr: done

drivers/SmartThings/jbl/src/jbl/device_manager.lua
cchillerr: Todo removed.

drivers/SmartThings/jbl/src/jbl/device_manager.lua
cchillerr: Done

drivers/SmartThings/jbl/src/jbl/discovery_helper.lua
cchillerr: change to log.debug

drivers/SmartThings/jbl/src/init.lua
```diff
@@ -0,0 +1,220 @@
...
+  local conn_info = device:get_field(fields.CONN_INFO)
+
+  if not driver.device_manager.is_valid_connection(driver, device, conn_info) then
+    log.error("create_sse : invalid connection")
+    return
+  end
+
+  local sse_url = driver.device_manager.get_sse_url(driver, device, conn_info)
+  if not sse_url then
+    log.error("failed to get sse_url")
```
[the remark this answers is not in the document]
cchillerr: change to     log.error("failed to get sse_url, dni = " .. tostring(device.device_network_id))

drivers/SmartThings/jbl/src/jbl/discovery_helper.lua
cchillerr: done

drivers/SmartThings/jbl/src/log.lua
cchillerr: remove other log levels except warn and error

drivers/SmartThings/jbl/src/init.lua
cchillerr: @cjswedes Sorry, I'm confused, which spawned timers we are talking about here?

drivers/SmartThings/jbl/src/init.lua
cchillerr: Now it's available in the PR, please check.

dljsjr:
I'm back from being OOO and reviewing latest changes.

@cchillerr @chill-cai Please rebase on top of latest `main`, that is needed before we can merge the PR.

drivers/SmartThings/jbl/src/init.lua
dljsjr: @cjswedes @cchillerr I believe that with some of the refactors that were made, this is no longer needed.

dljsjr (approved):
Once this is rebased I think it's ready for merging.

PR #847, opened 2023-07-12
Refresh matter devices on device added event

ctowns:
This change adds a device refresh to matter drivers that have a non-default added handler. There is another PR to change the default behavior for matter devices so that a refresh is done by default on device add.


ctowns:
Closing this - we believe the real issue is a result of https://smartthings.atlassian.net/browse/CHAD-11243, which will be fixed separately .

PR #854, opened 2023-07-17
Remove supportedTrackControlCommands when playing radio source

ctowns:
This un-reverts commit 5f5947952c1b32711a6d8406a8a923d618461797 for Bose and adds the same functionality to Sonos.


cjswedes (commented):
The driver events for bose seem correct from my testing just now. However, I am concerned that this may break scenes for bose devices that have been migrated; I dont have a way to verify since my bose devices were not migrated.

lelandblue:
Closing this out, if we need it later we can revisit this.

PR #855, opened 2023-07-18
Send WakeUpIntervalGet once when device wakes up

cjswedes:
This helps the hubs device health implementation work better for migrated devices by ensuring a WakeUpIntervalReport is received post migration. Note this will be done by default in the lua libs post 0.49.x hub FW release.


cjswedes:
Open question, should we forgo these changes and just let 0.49.x release fix it? I dont think many devices are affected.

ctowns (approved):

ctowns:
> Open question, should we forgo these changes and just let 0.49.x release fix it? I dont think many devices are affected.
My vote is that we should still make these changes and then just remove them once 49.x is widely available. It seems like a safe enough change even though it might only affect a few devices.

greens (approved):

greens:
I'm with Cooper

PR #861, opened 2023-07-19
Add sinope water sensor wl4210

aguilbault-Sinope:


drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
```diff
@@ -0,0 +1,70 @@
...
+
+local SINOPE_TECHNOLOGIES_MFR_STRING = "Sinope Technologies"
+
+local generate_event_from_zone_status = function(driver, device, zone_status, zb_rx)
+  local wet_probe = 0x0031
+  local wet_probe_disconnected = 0x0033
+  local event
+
+  if zone_status.value == wet_probe or zone_status.value == wet_probe_disconnected then
+      event = capabilities.waterSensor.water.wet()
```
greens, line 27: nit: we use 2-spaces for tabs

drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
```diff
@@ -0,0 +1,70 @@
...
+          [IASZone.ID] = {
+              [IASZone.attributes.ZoneStatus.ID] = ias_zone_status_attr_handler
+          }
+      },
+      cluster = {
+          [IASZone.ID] = {
+              [IASZone.client.commands.ZoneStatusChangeNotification.ID] = ias_zone_status_change_handler
+          }
+      }
+  },
```
greens, line 66: same comment about spacing

drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
```diff
@@ -0,0 +1,70 @@
...
+local IASZone = zcl_clusters.IASZone
+
+local SINOPE_TECHNOLOGIES_MFR_STRING = "Sinope Technologies"
+
+local generate_event_from_zone_status = function(driver, device, zone_status, zb_rx)
+  local wet_probe = 0x0031
+  local wet_probe_disconnected = 0x0033
+  local event
+
+  if zone_status.value == wet_probe or zone_status.value == wet_probe_disconnected then
```
greens, line 26: Is there a reason you _only_ emit wet if supervision and restore reports are enabled? Just from a quick look, the defaults, which check for whether alarm 1 or alarm 2 are set would also both trigger for ias zone status values of 0x0031 and 0x0033.

It doesn't seem like you need a sub-driver at all.

greens (changes requested):
If, after you've done some inspection, you still think you need a sub-driver, please write some tests.

I think you could just add fingerprints and the defaults would work fine with what you have here.

drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
greens: You'll want to check this out: https://developer.smartthings.com/docs/devices/hub-connected/set-up-dev-env

greens: So the default driver checks for bits 0 and 1 to be set in the IAS Zone Status to emit a water leak event.

Your behavior here is to check bits 0, 1, 4, and 5. Bits 4 and 5 simply indicate the reporting behavior of the device, namely, whether the device sends periodic updates about this zone and whether it emits a separate event when the issue is cleared.

What was the bit value emitted by the device for which you were seeing a false positive? It doesn't seem like bits 4 and 5 should affect anything.

drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
aguilbault-Sinope: fixed

drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
aguilbault-Sinope: fixed

drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
```diff
@@ -0,0 +1,70 @@
+-- Copyright 2022 SmartThings
```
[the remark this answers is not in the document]
aguilbault-Sinope: fixed

drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
aguilbault-Sinope: I've made a mistake, I meant whenever the probe is disconnected. When the probe is disconnected, the device report this:
(Sinope Water Leak Sensor)> received Zigbee message: < ZigbeeMessageRx || type: 0x00, < AddressHeader || src_addr: 0x3397, src_endpoint: 0x01, dest_addr: 0x0000, dest_endpoint: 0x01, profile: 0x0104, cluster: IASZone >, lqi: 0xFF, rssi: -20, body_length: 0x0009, < ZCLMessageBody || < ZCLHeader || frame_ctrl: 0x19, seqno: 0x1D, ZCLCommandId: 0x00 >, < ZoneStatusChangeNotification || zone_status: 0x0032, extended_status: 0x00, zone_id: 0x01, delay: 0x0000 > > >

This kind of change is considered a wet accord to the main driver.

I did the change with a bitwise operator instead so only the first bit is considered.

drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
```diff
@@ -0,0 +1,67 @@
...
+  if zone_status.value & 1 == 1 then
+    event = capabilities.waterSensor.water.wet()
+  else
+    event = capabilities.waterSensor.water.dry()
+  end
+
+  if event ~= nil then
+    device:emit_event(event)
+  end
+end
```
greens, line 32: so the distinction here is that you're only checking `zone_status:is_alarm1_set()` as opposed to the default driver, which checks for `alarm2` as well, because it looks like the device sends an event with the alarm2 field set when the sensor is disconnected?

Sounds good, but can you use the `is_alarm1_set()` function of the zone status object, for clarity's sake?

drivers/SmartThings/zigbee-water-leak-sensor/src/test/test_sinope_zigbee_water.lua
```diff
@@ -0,0 +1,314 @@
...
+        direction = "receive",
+        message = { mock_device.id, ZoneStatusAttribute:build_test_attr_report(mock_device, 0x0000) }
+      },
+      {
+        channel = "capability",
+        direction = "send",
+        message = mock_device:generate_test_message("main", capabilities.waterSensor.water.dry())
+      }
+    }
+)
```
greens, line 78: I'd include a test for the value of `0x0002` sending a "dry" event since that's the primary difference between your driver and the base driver

drivers/SmartThings/zigbee-water-leak-sensor/src/sinope/init.lua
aguilbault-Sinope: Done

drivers/SmartThings/zigbee-water-leak-sensor/src/test/test_sinope_zigbee_water.lua
aguilbault-Sinope: Done

greens (approved):

PR #862, opened 2023-07-19
BUG-7996 Re-bind yoolax window shades if no update received

greens:
Some users reported window shades that were not generating final closed/open events after being issued an open/close command. Upon investigation, the window shade was not sending attribute updates despite supposedly having attribute reporting configured. With this change, if no update is received within 30s, the driver will re-bind to the cluster and re-configure the attribute, and then issue an attribute read

### DIFF
--- a/drivers/SmartThings/zigbee-window-treatment/src/test/test_zigbee_window_shade_battery_yoolax.lua
+++ b/drivers/SmartThings/zigbee-window-treatment/src/test/test_zigbee_window_shade_battery_yoolax.lua
@@ -159,4 +159,129 @@ test.register_coroutine_test(
   end
 )
 
+test.register_coroutine_test(
+  "an attribute read should be sent after 30s of no response",
+  function ()
+    test.timer.__create_and_queue_test_time_advance_timer(30, "oneshot")
+    test.socket.capability:__queue_receive(
+      {
+        mock_device.id,
+        { capability = "windowShade", component = "main", command = "open", args = {} }
+      }
+    )
+    test.socket.zigbee:__expect_send({
+      mock_device.id,
+      WindowCovering.server.commands.GoToLiftPercentage(mock_device, 0)
+    })
+    test.wait_for_events()
+    test.mock_time.advance_time(30)
+    test.socket.zigbee:__expect_send({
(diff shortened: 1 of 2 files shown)


cjswedes (approved):

ctowns (approved):

cjswedes (approved):
This behavior might be something worth considering to add as default behavior for all attribute reporting situations for zigbee, and/or maybe the Matter subscription interaction.

drivers/SmartThings/zigbee-window-treatment/src/test/test_zigbee_window_shade_battery_yoolax.lua
```diff
@@ -198,8 +198,7 @@ test.register_coroutine_test(
 test.register_coroutine_test(
   "an attribute read should not be sent after 30s if there is a response",
   function ()
-    test.timer.__create_and_queue_test_time_advance_timer(25, "oneshot")
-    test.timer.__create_and_queue_test_time_advance_timer(5, "oneshot")
+    test.timer.__create_and_queue_test_time_advance_timer(10, "oneshot")
```
ctowns, line 201: Removing this line seems to make the test pass successfully every time for me.

drivers/SmartThings/zigbee-window-treatment/src/yoolax/init.lua
```diff
@@ -31,26 +35,35 @@ local function is_yoolax_window_shade(opts, driver, device)
...
 local function set_window_shade_level(level)
   return function(driver, device, cmd)
-    set_shade_level(device, level, cmd)
+    set_shade_level(driver, device, level, cmd)
   end
 end
 
 local function current_position_attr_handler(driver, device, value, zb_rx)
+  local timer = device:get_field(LEVEL_UPDATE_TIMEOUT)
+  if timer then device.thread:cancel_timer(timer) end
```
cjswedes, line 66: I think you need to set timer to nil on the device object.

greens: that's a good catch

PR #863, opened 2023-07-19
Switch generic locks that support lock codes

greens:
Because our fingerprinting system cannot distinguish between zigbee locks that do and do not support lock codes, users with non-WWST-certified devices that do support lock codes are unable to use that functionality on our platform. Previously they could easily switch the DTH to enable this, but with drivers this is not possible without a custom driver installed.

This logic attempts to determine whether a lock that was fingerprinted without user codes actually supports user codes or not, and switches the profile to one which does. It will also attempt to recover and lock codes carried over from migration.

### DIFF
--- a/drivers/SmartThings/zigbee-lock/src/init.lua
+++ b/drivers/SmartThings/zigbee-lock/src/init.lua
@@ -57,6 +57,14 @@ local refresh = function(driver, device, cmd)
   device:refresh()
   device:send(LockCluster.attributes.LockState:read(device))
   device:send(Alarm.attributes.AlarmCount:read(device))
+  -- we can't determine from fingerprints if devices support lock codes, so
+  -- here in the driver we'll do a check once to see if the device responds here
+  -- and if it does, we'll switch it to a profile with lock codes
+  if not device:supports_capability_by_id(LockCodes.ID) and not device:get_field(lock_utils.CHECKED_CODE_SUPPORT) then
+    device:send(LockCluster.attributes.NumberOfPINUsersSupported:read(device))
+    -- we won't make this value persist because it's not that important
+    device:set_field(lock_utils.CHECKED_CODE_SUPPORT, true)
+  end
 end
 
 local do_configure = function(self, device)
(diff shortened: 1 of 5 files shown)


drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -57,6 +57,14 @@ local refresh = function(driver, device, cmd)
   device:refresh()
   device:send(LockCluster.attributes.LockState:read(device))
   device:send(Alarm.attributes.AlarmCount:read(device))
+  -- we can't determine from fingerprints if devices support lock codes, so
```
ctowns, line 60: Why have this in `refresh` instead of a less frequent handler like `configure`? Is the idea that this would this just make it easier for recovering since a refresh would be all that's needed?

ctowns (approved):

drivers/SmartThings/zigbee-lock/src/init.lua
greens: Right. Users should only need to pull down to recover.

cjswedes (approved):

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -57,6 +57,14 @@ local refresh = function(driver, device, cmd)
...
+  -- here in the driver we'll do a check once to see if the device responds here
+  -- and if it does, we'll switch it to a profile with lock codes
+  if not device:supports_capability_by_id(LockCodes.ID) and not device:get_field(lock_utils.CHECKED_CODE_SUPPORT) then
+    device:send(LockCluster.attributes.NumberOfPINUsersSupported:read(device))
+    -- we won't make this value persist because it's not that important
+    device:set_field(lock_utils.CHECKED_CODE_SUPPORT, true)
```
varzac, line 66: This does mean that for devices that truly don't support lock codes, this will be re-checked every time the driver restarts.  I don't think that's a big deal, but wanted to call it out.

greens: I was _attempting_ to point that out with my comment, but I guess it wasn't that clear.

PR #864, opened 2023-07-20
BUG-8069 Z-Wave Switch: fix generic fingerprint ranking

greens:
Because we had two fingerprints with only one command class each, a device that supported both command classes would match to both with equal 'match score'. I've added a third fingerprint such that devices that support both binary and multilevel switch ccs will match to a multilevel profile


ctowns (approved):
Do we know what happens when we have a case of "equal match score"? Is it a coin toss?

cjswedes (approved):

PR #868, opened 2023-07-25
Nuki SmartLock Pull Request

Brianj94:
Adding this profile and fingerprint for Nuki Matter Smart Lock WWST Certification.

### DIFF
--- a/drivers/SmartThings/matter-lock/fingerprints.yml
+++ b/drivers/SmartThings/matter-lock/fingerprints.yml
@@ -1,3 +1,11 @@
+matterManufacturer:
+#Nuki
+  - id: "Nuki Smart Lock"
+    deviceLabel: Nuki Smart Lock
+    vendorId: 0x135d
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-lock/profiles/lock-without-codes-or-tamper.yml
```diff
@@ -0,0 +1,21 @@
+name: lock-without-codes-or-tamper
```
ctowns, line 1: I would suggest changing this to `lock-nocodes-notamper` to follow the naming precedent of excluding capabilities that is used in the `matter-thermostat` profiles: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/tree/main/drivers/SmartThings/matter-thermostat/profiles

When changing the name, remember to change the `name:` field, filename, and the profile name in the fingerprint above.

ctowns (approved):

PR #870, opened 2023-07-25
Fix incorrect fingerprint

greens:


drivers/SmartThings/zwave-switch/fingerprints.yml
```diff
@@ -923,13 +923,6 @@ zwaveGeneric:
     deviceLabel: Dimmer Switch
     genericType: 0x11 # GENERIC_TYPE_SWITCH_MULTILEVEL
     deviceProfileName: switch-level
-  - id: "GenericDimmerSwitch/3"
```
ctowns, line 926: Why was this generic fingerprint removed?

greens: that was not intentional!

ctowns (approved):

PR #873, opened 2023-07-26
Feature/wemo power energy

cjswedes:
Im not convinced in the energy meter reporting of this device at all, in fact it may be worthwhile to only report powerMeter capabilities for the device since it is very inconsistent in its values for energy reports (total energy fluctuates a lot including negative values, and it seems very high)

This implementation is based off the way device report payloads are interpreted [here](https://github.com/tigoe/WeMoExamples/tree/master) and [here](https://github.com/SmartThingsUle/Wemo/blob/4f5236c3fb1aa14834cd11d3173495edfff6d1a4/WeMo%20Insight%20Switch.groovy#L110)

### DIFF
--- a/drivers/SmartThings/wemo/src/parser.lua
+++ b/drivers/SmartThings/wemo/src/parser.lua
@@ -74,10 +74,11 @@ local function handle_binary_state(device, value)
   if result.current_power_mW ~= nil and result.energy_today_Wh ~= nil and
     device:supports_capability_by_id("powerMeter") and device:supports_capability_by_id("energyMeter") then
     device:emit_event(capabilities.powerMeter.power(result.current_power_mW / 1000)) --ST uses watts by default
-    --Report energy
-    if device:get_field("DAILY_ENERGY_REPORT") then
+    --Sometimes total energy reported is way off, in that case use the daily energy reported
+    if result.energy_today_Wh > result.energy_total_Wh then
       device:emit_event(capabilities.energyMeter.energy({value = result.energy_today_Wh, unit = "Wh"}))
-      device:set_field("DAILY_ENERGY_REPORT", false)
+    else
+      device:emit_event(capabilities.energyMeter.energy({value = result.energy_total_Wh, unit = "Wh"}))
     end
   end
 end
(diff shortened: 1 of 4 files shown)


cjswedes:
Due to the weird energy reports I have gotten from the device I see a few options to move this PR forward:
1. Remove energyMeter cap from the insight devices
2. Leave it in there as it is now, but implement a profile switch based on a preference value, so that users can opt into the energy reporting, which is somewhat suspect
3. Implement energy reporting in software. The `energy_today_Wh` field that is reported seems to be accurate and could be used to keep a running total for the device and periodically emit that total. This would also allow for resetting of the energyMeter since we control the total rather than the device
4. Implement the powerConsumptionReport capability instead. This sort of seems like option 3 but the capability is setup for it. Although, I am not sure how accurate all the fields are that would be needed from the device, and if the device is actually resetting the energy totals daily.

drivers/SmartThings/wemo/src/parser.lua
```diff
@@ -55,7 +70,17 @@ local function handle_binary_state(device, value)
       log.warn("parse| BinaryState event on device that supports neither `switch` nor `motionSensor`")
     end
   end
-  -- TODO: there's a bunch more values from insight plugs, what do they mean?
+
+  if result.current_power_mW ~= nil and result.energy_today_Wh ~= nil and
+    device:supports_capability_by_id("powerMeter") and device:supports_capability_by_id("energyMeter") then
+    device:emit_event(capabilities.powerMeter.power(result.current_power_mW / 1000)) --ST uses watts by default
```
greens, line 76: I thought you felt safe always emitting the power events?

greens:
Since I suggested it, I'm most partial to option 2.

lelandblue:
@cjswedes Should we close this PR?

cjswedes:
I would prefer to leave it available for any developer out there to pick up, since there was non-trivial engineering time put into figuring out the message format for energy from these devices. If it needs to be closed, I will move it to a personal repo to avoid losing the work.

hcarter-775:
@cjswedes could we close this PR?

cjswedes:
😢

PR #876, opened 2023-07-27
initial commit: add profiles, fingerprints, init.lua

not-tacos:
ready for review!


drivers/SmartThings/zwave-switch/src/leviton-smart-switch/init.lua
```diff
@@ -0,0 +1,114 @@
+-- Copyright 2022 SmartThings
```
greens, line 1: year

drivers/SmartThings/zwave-switch/src/leviton-smart-switch/init.lua
```diff
@@ -0,0 +1,114 @@
...
+      if new_parameter_value ~= old_parameter_value then
+        device:send(Configuration:Set({
+          parameter_number = pref.parameter_number,
+          size = pref.size,
+          configuration_value = new_parameter_value
+        }))
+      end
+    end
+  end
+end
```
greens, line 97: on quick glance, this looks remarkably similar to the existing `infoChanged` handler from the base driver: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zwave-switch/src/init.lua#L72 I don't think you need to replace it.

drivers/SmartThings/zwave-switch/src/leviton-smart-switch/init.lua
```diff
@@ -0,0 +1,114 @@
...
+
+  if target_parameter_name == nil then
+    return
+  end
+
+  device:set_field(target_parameter_name, reported_value)
+  device:set_field(target_parameter_name, reported_value, {persist = true})
+
+  -- FIXME: SmartThings Mobile App UI does not sync with the updated value
+  -- https://community.smartthings.com/t/st-edge-driver-for-leviton-z-wave-plus-dimmers-dz6hd-dz1kd/257938/4
```
greens, line 58: these values you're setting are never read

I believe what you're alluding to are two-way preferences. Right now there's no way for a driver to change its own preference values, that must be done by the user in the UI.

drivers/SmartThings/zwave-switch/src/leviton-smart-switch/init.lua
```diff
@@ -0,0 +1,114 @@
...
+  local parameters = preferences.get_device_parameters(device)
+
+  if parameters then
+    for id, pref in pairs(parameters) do
+      device:send(Configuration:Get({
+        parameter_number = pref.parameter_number
+      }))
+    end
+  end
+end
```
greens, line 73: We have had users report a race condition where the hub might not be ready for z-wave transmission during the `init` lifecycle event. We haven't been able to reproduce this locally, but I'd be careful about it. Since the values you read are never used, it does not seem particularly crucial in any event.

drivers/SmartThings/zwave-switch/src/leviton-smart-switch/init.lua
```diff
@@ -0,0 +1,114 @@
...
+    {LEVITON_PRODUCT_ID_ZW6HD, LEVITON_PRODUCT_ID_ZW15S},
+    LEVITON_PRODUCT_TYPE_ZWXXX
+  );
+end
+
+local function handle_configuration_update(self, device, cmd)
+
+  local parameters = preferences.get_device_parameters(device)
+  local target_parameter_number = cmd.args.parameter_number
+  local reported_value = cmd.args.configuration_value
```
greens, line 36: you've indicated this function as a handler of CONFIGURATION_GET commands, but those commands do not include the `configuration_value` field.

greens (commented):
In general, I think this driver could just be the fingerprints, profiles, and `preferences.lua` changes.

If you are going to write a separate sub-driver, though, we'd like to see some tests coverage.

drivers/SmartThings/zwave-switch/src/leviton-smart-switch/init.lua
jnoel-leviton: Thanks for clarifying - I was confused as to why the UI would not show the freshly read-in preference values. The main reason we were trying to do this was to avoid having the UI display incorrect information that does not the device's true configuration state.

Are there any plans to have this information read automatically by the hub in the future, so the UI will show correct preference values post-inclusion?

jnoel-leviton:
> In general, I think this driver could just be the fingerprints, profiles, and `preferences.lua` changes.
> 
> If you are going to write a separate sub-driver, though, we'd like to see some tests coverage.

Agreed, the main reason for the custom driver functions in the first place was to synchronize the preferences UI with the device's previously set values. But as you said, that's not currently possible - so will remove all but the profiles and `preferences.lua`changes as you suggested.

drivers/SmartThings/zwave-switch/src/leviton-smart-switch/init.lua
greens: Yes, but I don't have a timeframe for that.

drivers/SmartThings/zwave-switch/src/preferences.lua
```diff
@@ -318,6 +318,34 @@ local devices = {
...
+    }
+  },
+  LEVITON_SWITCH = {
+    MATCHING_MATRIX = {
+      mfrs = 0x001D,
+      product_types = {0x0042},
+      product_ids = 0x0002
+    },
+    PARAMETERS = {
+    statusLEDmode = { parameter_number = 7, size = 4 }
```
greens, line 347: nit: whitespace

greens (approved):

greens (approved):

jnoel-leviton:
Are there any other issues blocking this from getting merged into the main repo?

lelandblue:
Closing this PR due to communication from the WWST Certification team. If you have questions, please reach out to partners@smartthings.com

Thank you.

lelandblue:
This PR is now closed as it is a duplicate of https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1425. Please see that PR for more information.

PR #877, opened 2023-07-31
[Eve] Improve Eve Energy support

Timac:
These changes improve the Eve Energy support

- A subdriver has been implemented
- It implements the switch, powerMeter, energyMeter and powerConsumptionReport capabilities
- It supports the refresh capability
- It reads with a timer each minute the current Watt and Watt Accumulated values from custom attributes
- The powerMeter and energyMeter values are updated each values or on a refresh
- A powerConsumptionReport is sent each 10 minutes by comparing the previous Watt Accumulated value with the current Watt Accumulated value


greens (commented):
I think you accidentally brought in several unintended changesets. I think you are only intending to be touching `matter-switch` files. I think you probably wanted to `rebase` your fork.

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
```diff
@@ -0,0 +1,231 @@
...
+local TIMER_REPEAT = (1 * 60)    -- Run the timer each minute
+local REPORT_TIMEOUT = (10 * 60) -- Report the value each 10 minutes
+
+
+-------------------------------------------------------------------------------------
+-- Eve specifics
+-------------------------------------------------------------------------------------
+
+local function is_eve_energy_products(opts, driver, device)
+	if device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID then
```
greens, line 27: we use 2-width spaces instead of tabs

drivers/SmartThings/matter-switch/profiles/eve-energy.yml
```diff
@@ -0,0 +1,18 @@
+name: eve-energy
```
greens, line 1: I would just name this profile `power-energy-powerConsumption`, since it's generic enough that other devices may end up using it.

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
```diff
@@ -0,0 +1,231 @@
...
+end
+
+local timer = nil
+local function create_poll_schedule(device)
+	-- The powerConsumption report needs to be updated at least every 15 minutes in order to be included in SmartThings Energy
+	-- Eve Energy generally report changes every 10 or 17 minutes
+	timer = device.thread:call_on_schedule(TIMER_REPEAT, function()
+		requestData(device)
+	end, "polling_schedule_timer")
+end
```
greens, line 98: can you not use zigbee reporting configurations here instead of a timer?

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
```diff
@@ -0,0 +1,231 @@
...
+	device:send(clusters.OnOff.attributes.OnOff:read(device))
+
+	-- Update the Watt usage
+	device:send(cluster_base.read(device, 0x01, PRIVATE_CLUSTER_ID, PRIVATE_ATTR_ID_WATT, nil))
+
+	-- Update the energy consumption
+	device:send(cluster_base.read(device, 0x01, PRIVATE_CLUSTER_ID, PRIVATE_ATTR_ID_WATT_ACCUMULATED, nil))
+end
+
+local timer = nil
```
greens, line 91: note that this is a driver-level variable and this will have unintended results if more than one device is using this driver

you'll want to use the device storage for this (`device:set_field`)

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
```diff
@@ -0,0 +1,231 @@
...
+	log.info(string.format("Fallback handler for %s", response_block))
+end
+
+local function on_off_attr_handler(driver, device, ib, response)
+	if ib.data.value then
+		device:emit_event_for_endpoint(ib.endpoint_id, capabilities.switch.switch.on())
+	else
+		device:emit_event_for_endpoint(ib.endpoint_id, capabilities.switch.switch.off())
+	end
+end
```
greens, line 181: you don't need to overwrite defaults here I don't think, this method looks equivalent to the one in the base driver

greens:
Please include some tests, as there is a significant amount happening here.

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
Timac: Because of some unintended changesets in this PR, I created a new PR at https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/940

In this new PR, we use 2-width spaces instead of tabs

drivers/SmartThings/matter-switch/profiles/eve-energy.yml
Timac: Because of some unintended changesets in this PR, I created a new PR at #940

In this new PR, we renamed the profile to `power-energy-powerConsumption`

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
Timac: This plugin is dedicated to Matter devices, so I am not sure if it would be appropriate to use zigbee features in such a plugin. That’s the reason I use a timer.

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
Timac: Because of some unintended changesets in this PR, I created a new PR at #940

In the new PR, the timer has been changed from a driver-level variable to a device storage

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
Timac: Because of some unintended changesets in this PR, I created a new PR at #940

In the new PR, this code has been removed

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
greens: excellent point

PR #880, opened 2023-08-01
fix: Fixes a nil crash in the keyless bridge task.

dljsjr:
If the bridge is missing both the application key + the bridge info, there's not much we can do until we decide how we want to handle the remedy for this situation from a UX perspective.

This addresses the crash signature around CHAD-11420 but does not yet fix that bug itself.


greens (approved):
Should we mark the device offline, or will it already be marked that way?

ctowns (approved):

dljsjr:
@greens It'll already be marked that way, as it won't be able to open its SSE connection.

PR #882, opened 2023-08-01
Add periodic mDNS Scanning to Philips Hue Edge Driver

dljsjr:
The main change to the functionality of the driver in this PR is that *all* mDNS operations have been consolidated in to a function that does the requisite REST calls for discovered bridges and updates state on the driver's persistent store by accessing the `driver.datastore` table proxy directly.

I'm not normally in favor of using side-effect oriented flows like this, but in this specific case it actually solves a lot of problems for us:

1. It reduces the number of tasks trying to use the mDNS socket at the same time by removing mDNS operations from `search_for_bridges`
2. It reduces the number of tasks attempting REST network operations by having the various tasks that run during init and migration simply query the datastore cylically instead of querying the network cyclically (in the case of bridge info; we still need to scan known bridges for their lights)
3. The new entrypoint can be called manually/deliberately to force a scan that updates the state, which can be seen in the Discovery flow: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/ca534245610eaa6aaba4f832229b974dbe53b7e9/drivers/SmartThings/philips-hue/src/disco.lua#L51
4. By using the persistent store, it speeds up the driver's orienting of itself and connecting to bridges during start-up without needing a full re-scan if nothing has changed

This was tested in a variety of combinations, including injecting a driver restart in to the system and having the IP address change during restart, a full disconnect/reconnect, and IP changes while the driver is live.

**_Edit:_**
There is one thing I want to call out, which is that we currently *don't* trigger an immediate state update on disconnect. There's no really good way to disambiguate between a disconnect because re-scanning is needed vs, say, the Hue Bridge getting unplugged temporarily. I figured it was probably more beneficial and less risky to just user a relatively "short" poll interval of 60 seconds instead of coming up with heuristics to try and figure out how to re-scan on disconnect without hammering TCP connection attempts, especially since we don't currently have the ability to subscribe to mDNS events that tell us if something is actually offline or not.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/hue/api.lua
+++ b/drivers/SmartThings/philips-hue/src/hue/api.lua
@@ -86,7 +86,7 @@ local function process_rest_response(response, err, partial, err_callback)
       )
     end
 
-    return table.unpack(json_result)
+    return table.unpack(json_result, 1, json_result.n)
   else
     return nil, "no response or error received"
   end
@@ -185,7 +185,7 @@ local function do_get(instance, path)
     instance.client:close_socket()
     return nil, "cosock error: " .. err
   end
-  return table.unpack(recv)
+  return table.unpack(recv, 1, recv.n)
 end
 
 local function do_put(instance, path, payload)
@@ -198,7 +198,7 @@ local function do_put(instance, path, payload)
     instance.client:close_socket()
     return nil, "cosock error: " .. err
   end
-  return table.unpack(recv)
+  return table.unpack(recv, 1, recv.n)
 end
 
 ---@param bridge_ip string
@@ -220,7 +220,7 @@ function PhilipsHueApi.get_bridge_info(bridge_ip, socket_builder)
   if err ~= nil then
     return nil, "cosock error: " .. err
   end
-  return table.unpack(recv)
+  return table.unpack(recv, 1, recv.n)
 end
 
 ---@param bridge_ip string
@@ -243,7 +243,7 @@ function PhilipsHueApi.request_api_key(bridge_ip, socket_builder)
   if err ~= nil then
     return nil, "cosock error: " .. err
   end
-  return table.unpack(recv)
+  return table.unpack(recv, 1, recv.n)
 end
 
 function PhilipsHueApi:get_lights() return do_get(self, "/clip/v2/resource/light") end
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -726,7 +726,7 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
       if msg and msg.data then
         local json_result = table.pack(pcall(json.decode, msg.data))
         local success = table.remove(json_result, 1)
-        local events, err = table.unpack(json_result)
+        local events, err = table.unpack(json_result, 1, json_result.n)
 
         if not success then
           log.error_with({ hub_logs = true },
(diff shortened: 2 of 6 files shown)


drivers/SmartThings/philips-hue/src/hue/api.lua
```diff
@@ -86,7 +86,7 @@ local function process_rest_response(response, err, partial, err_callback)
       )
     end
 
-    return table.unpack(json_result)
+    return table.unpack(json_result, 1, json_result.n)
```
greens, line 89: would you mind enlightening me about this call? I wasn't even aware `table.unpack` took multiple arguments

dljsjr: It turns out that if you don't pass in a start and end point, `table.unpack` uses `#list` to determine the endpoint instead of using the `n` key that `table.pack` creates… which kinda defeats the point of using `table.pack` because you'll end up stopping at the first `nil` 😵‍💫

dljsjr: See also: https://www.lua.org/manual/5.3/manual.html#pdf-table.unpack

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -1322,10 +1309,64 @@ local hue = Driver("hue",
...
+              maybe_event_source:close()
+              bridge_device:set_field(Fields.EVENT_SOURCE, nil)
+              do_bridge_network_init(self, bridge_device, bridge_url, maybe_api_key)
+            end
+          end
+        end
+      end
+    end,
+    get_device_by_dni = function(self, dni)
+      local device_id = self.datastore.dni_to_device_id[dni]
```
greens, line 1348: Have you considered using the term `uuid` instead of `device_id`? I find `device_id` and `device_network_id` difficult to keep separate in my head.

dljsjr: I thought about it but it "seems" to me that internally we have a convention that has overloaded the acronym "UUID" to mean "User UUID". So I wanted to keep with our conversational and organizational conventions.

I'm not against changing that though.

drivers/SmartThings/philips-hue/src/utils.lua
```diff
@@ -170,4 +170,40 @@ function utils.labeled_socket_builder(label)
...
+				return false
+			end
+		end
+
+		return true
+	end
+
+	return false
+end
+
```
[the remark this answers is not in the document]
dljsjr: That's what I get for copy/pasting a gist.

drivers/SmartThings/philips-hue/src/init.lua
greens: the arg to `get_device_info` is named `uuid`, at least

drivers/SmartThings/philips-hue/src/hue/api.lua
greens: wouldn't it be better to only return non-`nil` values? or is it useful to keep them around?

greens (approved):

drivers/SmartThings/philips-hue/src/hue/api.lua
dljsjr: Yes, but, returning `nil, "error message string"` is a Lua convention for communicating failure, so we don't always control if the first argument is nil or not in the error case.

PR #885, opened 2023-08-02
Initial push of Harman Luxury Driver

alon-tchelet:
Initial push of harman-luxury SmartThings edge driver.
Current version supports only future released device L42ms


drivers/SmartThings/harman-luxury/profiles/stream-sdk.yaml
```diff
@@ -0,0 +1,36 @@
...
+            - 'stopped'
+        - key: "{{enumCommands}}"
+          enabledValues:
+            - 'play'
+            - 'pause'
+            - 'stop'
+  - id: mediaTrackControl
+    version: 1
+  # - id: keypadInput
+  #   version: 1
```
FreeMasen, line 32: nit: commented code

drivers/SmartThings/harman-luxury/profiles/stream-sdk.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - id: refresh
+    version: 1
+  categories:
+  - name: Speaker
```
FreeMasen, line 36: nit: missing trailing new line

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
```diff
@@ -0,0 +1,246 @@
...
+    else
+        local roles = ""
+        for i, role in rolesList do
+            if roles == "all" then
+                return "@all"
+            else
+                roles = roles .. role .. ","
+            end
+        end
+        return roles:sub(1, -2)
```
FreeMasen, line 41:
```suggestion
        local t = {}
        for i, role in rolesList do
            if roles == "all" then
                return "@all"
            else
                table.insert(t, role)
            end
        end
        return table.concat(t, ",")
```

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
```diff
@@ -0,0 +1,246 @@
...
+    }
+    u = u:build()
+    local sink = {}
+    local result, code, headers = http.request {
+        url = u,
+        method = "GET",
+        sink = ltn12.sink.table(sink)
+    }
+    if code == 200 then     -- OK
+        return json.decode(sink[1])[1]
```
FreeMasen, line 66:
```suggestion
        return json.decode(table.concat(sink, ""))[1]
```

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
```diff
@@ -0,0 +1,246 @@
...
+    }
+    u = u:build()
+    local sink = {}
+    local result, code, headers = http.request {
+        url = u,
+        method = "GET",
+        sink = ltn12.sink.table(sink)
+    }
+    if code == 200 then     -- OK
+        return json.decode(sink[1])[1]["rows"]
```
FreeMasen, line 99: again, `sink` should be concatenated

FreeMasen (commented):
Quick look, there is some inconsistency that should probably be addressed but mostly nits.

- All uses of the `sink` in `http.request` should be updated to use `table.concat(sink, "")` to avoid missing data if it comes across as multiple chunks.
- Indentation should be consistent, some files are 8 spaces, some are 4 (I think more are 4 than 8)

drivers/SmartThings/harman-luxury/src/api/apis.lua
```diff
@@ -0,0 +1,238 @@
...
+
+--- get Member ID from Harman Luxury on ip
+---@param ip string
+function APIs.GetMemberId(ip)
+    return get.String(ip, MEMBER_ID_PATH)
+end
+
+--- get device manufacture name from Harman Luxury on ip
+---@param ip string
+function APIs.GetManufatureName(ip)
```
greens, line 52: typo: `Manufacturer`

drivers/SmartThings/harman-luxury/src/disco_helper.lua
```diff
@@ -0,0 +1,185 @@
...
+                params["dni"] = dni
+            end
+            if key == "mnid" then
+                log.info("get_parameters : mnid = " .. value)
+                params["mnid"] = value
+            end
+            if key == "setupid" then
+                log.info("get_parameters : setupid = " .. value)
+                params["setupid"] = value
+            end
```
greens, line 88: these should be `elseif`s I think

drivers/SmartThings/harman-luxury/src/handlers.lua
```diff
@@ -0,0 +1,245 @@
...
+    if ret == "playing" then
+        device:emit_event(capabilities.mediaPlayback.playbackStatus.playing())
+    elseif ret == "paused" then
+        device:emit_event(capabilities.mediaPlayback.playbackStatus.paused())
+    elseif ret == "stopped" then
+        device:emit_event(capabilities.mediaPlayback.playbackStatus.stopped())
+    end
+end
+
+function Handler.handle_set_playback_status(driver, device, cmd)
```
greens, line 191: this should either just call `handle_stop` and `handle_playPause` or they should call it, to avoid unnecessary code duplication

drivers/SmartThings/harman-luxury/src/handlers.lua
```diff
@@ -0,0 +1,245 @@
...
+    -- verify change and update app
+    local ret = api.GetMute(ip)
+    if ret then
+        device:emit_event(capabilities.audioMute.mute.muted())
+    else
+        device:emit_event(capabilities.audioMute.mute.unmuted())
+    end
+end
+
+function Handler.handle_set_mute(driver, device, cmd)
```
greens, line 83: again, since you've already defined mute and unmute methods, this method seems reduplicative

greens (commented):
Please fix all of the linter errors.

Additionally, a lot of the handlers have significant amount of duplicated code. A better pattern would be worthwhile, I think. Either combining methods like on/off and mute/unmute or using a factory would be better I think.

There are also a significant number of magic strings (like `"device_ipv4"`) that I think would be good to define somewhere, maybe in a constants file.

drivers/SmartThings/harman-luxury/search-parameters.yml
```diff
@@ -0,0 +1,2 @@
+mdns:
+  - service: "_sue-st._tcp"
```
dljsjr, line 2: Is this the correct service name? It's not the same one we saw in test tools, so just wanted to make sure this was accurate.

drivers/SmartThings/harman-luxury/src/disco_helper.lua
```diff
@@ -0,0 +1,185 @@
...
+                    ip = ip,
+                    mnid = mnid,
+                    setupid = setupid
+                }
+            end
+        end
+    end
+end
+
+local function filter_response_by_servie_name(service_type, discovery_responses)
```
dljsjr, line 126: type: `servie` should be `service`

drivers/SmartThings/harman-luxury/src/disco_helper.lua
```diff
@@ -0,0 +1,185 @@
...
+
+    for _, answer in pairs(discovery_responses.answers or {}) do
+        table.insert(filtered_responses.answers, answer)
+    end
+
+    for _, additional in pairs(discovery_responses.additional or {}) do
+        table.insert(filtered_responses.additional, additional)
+    end
+
+    for _, found in pairs(discovery_responses.found or {}) do
```
dljsjr, line 140: Just making a note: This is the correct structure for the mDNS API; released versions of the API will always use the `.found` field.

The rest of the code should treat the `{ answers, additional }` form as the fallback, not the `found` form.

drivers/SmartThings/harman-luxury/src/devices.lua
```diff
@@ -0,0 +1,142 @@
+local log = require "log"
+local st_utils = require "st.utils"
+
+local api = require "api.apis"
+
+Devices = {}
+
+local suppourted_devices = { "L75ms", "L42ms", "AVR5" }
```
dljsjr, line 8: typ: `suppourted` -> `supported`

drivers/SmartThings/harman-luxury/src/devices.lua
```diff
@@ -0,0 +1,142 @@
...
+    L42ms = "603",
+    AVR5 = "604"
+}
+
+function Devices.GetSupportedDevices()
+    log.info("GetSupportedDevices: supported models: " .. st_utils.stringify_table(suppourted_devices))
+    return suppourted_devices
+end
+
+local function GetDefaultDeviceInfo(dni, ip)
```
dljsjr, line 27: Since we want the values here to not only be non-nil, but also explicitly `strings`, it would be more defensive and less bug-prone to use `if type(var) ~= "string"` instead of `if not var`.

So for example, `if type(label) ~= "string"` instead of `if not label`. This check will catch both `nil` and cases where the value passed in is not a string.

drivers/SmartThings/harman-luxury/src/devices.lua
```diff
@@ -0,0 +1,142 @@
...
+        profile = "harman-luxury",
+        manufacturer = "ARCAM",
+        model = "AVR5",
+        vendor_provided_label = label
+    }
+
+    return device_info
+end
+
+function Devices.get_device_info(dni, params)
```
dljsjr, line 122: Similar to the comment that @greens made elsewhere, there's a lot of duplicated code here. One option would be to have Lua tables that take `params.setupid` as the key, and return the MFR and Model information for the various speakers, as those are the only things that differ in the various functions defined in this file.

drivers/SmartThings/harman-luxury/src/disco.lua
```diff
@@ -0,0 +1,130 @@
...
+local socket = require "cosock.socket"
+local log = require "log"
+
+local disco_helper = require "disco_helper"
+local devices = require "devices"
+
+local SERVICE_TYPE = "_sue-st._tcp"
+local DOMAIN = "local"
+
+local Discovery = {}
```
dljsjr, line 11: This one is a nit:

This file (and a few others) use a lot of string concatenation (the `..` operator).

`..` is fragile; it will crash on `nil` values if `tostring` isn't called first. Most of the usage here is well protected, so that's good.

However, the overall readability when there are multiple interleaved/concatenated variables in the middle of a string suffers, and the use of `tostring` to elide `nil` isn't very obvious.

In cases where one needs to insert multiple variables in to a string, it can often lead to much more readable and maintainble logging if `string.format()` is used instead of multiple concatenation operations.

drivers/SmartThings/harman-luxury/src/disco.lua
```diff
@@ -0,0 +1,130 @@
...
+
+    driver:try_create_device(device_info)
+end
+
+function Discovery.set_device_field(driver, device)
+    log.info("set_device_field : dni = " .. tostring(device.device_network_id))
+    local device_cache_value = driver.registered_devices[device.device_network_id]
+
+    -- persistent fields
+    device:set_field("device_ipv4", device_cache_value.ip, {
```
dljsjr, line 42: Already mentioned elsewhere but it would be nice if this string (`"device_ipv4`) was captured in a constant somewhere.

drivers/SmartThings/harman-luxury/src/disco_helper.lua
```diff
@@ -0,0 +1,185 @@
...
+        if found.service_info.service_type == service_type then
+            table.insert(filtered_responses.found, found)
+        end
+    end
+
+    return filtered_responses
+end
+
+local function insert_dni_ip_from_found(driver, filtered_responses, target_table)
```
dljsjr, line 149: This is correct.

I think it's good that there is code here that catches both the `{answers, additional}` form and the `found` form, but style-wise, the code should prefer the `found` form.

drivers/SmartThings/harman-luxury/src/handlers.lua
```diff
@@ -0,0 +1,245 @@
+local capabilities = require "st.capabilities"
+local api = require "api.apis"
+local log = require "log"
+
+local Handler = {}
```
dljsjr, line 5: Seconding @greens comment that there's a lot of duplicated code in here.

drivers/SmartThings/harman-luxury/src/api/gets.lua
```diff
@@ -0,0 +1,94 @@
...
+--- get value from path in ip
+---@param ip string
+---@param path string
+---@param type string
+local function GetValue(ip, path, type)
+	local ret = nsdk.GetData {
+		ip = ip,
+		path = path
+	}
+	if ret then
```
dljsjr, line 19: I'd recommend `if type(ret) == "table"` instead.

drivers/SmartThings/harman-luxury/src/api/gets.lua
```diff
@@ -0,0 +1,94 @@
...
+		return ret[type]
+	else
+		return false
+	end
+end
+
+--- get value's type from path in ip
+---@param ip string
+---@param path string
+local function GetType(ip, path)
```
dljsjr, line 29: Unused function.

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
```diff
@@ -0,0 +1,246 @@
+local cosock = require "cosock"
+local http = cosock.asyncify "socket.http"
+local ltn12 = require "ltn12"
+local url = require "net.url"
+local json = require "st.json"
```
dljsjr, line 5: `json.decode` can throw an `error` on malformed input, which can crash the whole driver.

This API needs to be either documented to note this, or cleaned up to avoid this. In the meantime, I would strongly suggest using `pcall` everywhere that you call `json.decode`

dljsjr (changes requested):
I made a lot of comments, but they can be boiled down to a handful of things that need to be addressed before we can merge.

1. Luacheck lints are failing. Luacheck can be installed from Luarocks and can be run with the config file vendored in this repo's .github directory for validating locally without needing to wait for the PR actions to complete.
2. Lots of code duplication that should be cleaned up,
3. The discovery code that consumes the `mDNS` API is technically valid but has a preference for a deprecated API, and it treats the stable API as the "fallback". For style, readability, and maintainability reasons that should be switched.
4. Weak error handling, that can be broken down even further

### Error Handling Improvement Requests

#### JSON Error Handling

`json.decode` needs to be wrapped in `pcall` to avoid accidentally crashing the entire driver if there's malformed input. You can find examples of this in the up to date Hue and Sonos drivers.

#### Function Input and Return Value Validation

There are a lot of places where the return values from network I/O operations are checked for "truthiness" to see if the values are `nil`. Network operations are fragile, and various forms of socket interruptions can lead to values that might not be what one would expect.

When checking the type of a value that has come directly from a socket I/O operation, or when checking that a value is safe to forward to another function, it's safest to use `type(val) == "string"` or whatever the expected type string is. Because `nil` is its own type, this satisfies both nil checking and rudimentary argument type checking needs.

#### API Error Handling

The `apis` directory and all of its modules have very incosistent error handling patterns. In many places they return internal API calls directly, such as `return json.decode(…)`, but in their local error checking branches they will often swallow errors entirely or log them in-flight then return `false`.

This is both diverging from the return semantics of the internal APIs that it passes through, as well as not being idiomatic Lua. It also puts undue burden on the caller to know that the return semantics of those modules can be very polymorphic, and make the code hard to reason about.

A more reliable error handling pattern would be to return `nil, "error message string"` when a function fails instead of logging and returning only `false`. This does a few things:

1. Less code specialization in the function bodies so that there can be more opportunities for code reuse
2. Makes it easier to write functions using the early return pattern, which we prefer
3. Makes the code more predictable, as this is a common pattern in most of the Lua libraries that we use so it makes the code behave more consistently withour dependencies and platform-provided libraries
5. Using `nil` instead of `false` removes any ambiguity from a function that could potentially have a `bool` return value that can be false in the non-error case; even if that type of function isn't in this codebase, it eliminates the question of whether or not it could be from a reader/maintainer.

drivers/SmartThings/harman-luxury/search-parameters.yml
alon-tchelet: Yes. We changed it one of our latest firmware releases. When you test on the L42ms, please ensure you are testing on latest StreamSDK software. Latest release currently is 1793, but we will make a new release once I covered the changes.

Reason for the change was to avoid discovery of unsupported devices

alon-tchelet:
@dljsjr Hi, Thanks for all the comments. Please have a look at latest master after I addressed all the previous comments

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
greens, line 1: you are repeating code a _lot_ in this file

greens (commented):
never mind, I see it

greens (changes requested):
We use 2-width spaces in most of our drivers. You've got a mix of tabs and some 4-width spaces here. Could you update that for consistency?

drivers/SmartThings/harman-luxury/src/handlers.lua
```diff
@@ -0,0 +1,249 @@
...
+    log.info("Starting handle_mute")
+    -- send API mute on message
+    set_mute(device, true, "handle_mute")
+end
+
+--- handler of audioMute.unmute
+function Handler.handle_unmute(_, device, _)
+    log.info("Starting handle_unmute")
+    -- send API mute off message
+    set_mute(device, true, "handle_unmute")
```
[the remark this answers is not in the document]
alon-tchelet: Absolutely, correct. Was just looking at this after some tests results

alon-tchelet:
Hi again, I applied the changes you requested (fixed indents and further minimized repeated code inside nsdk.lua). 
Once again ready for review

drivers/SmartThings/harman-luxury/config.yaml
```diff
@@ -0,0 +1,5 @@
+name: "Harman-Luxury"
```
dljsjr, line 1: This key should be a "human friendly" name, and as such doesn't need the hyphen.

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
```diff
@@ -0,0 +1,286 @@
...
+---@param func_name string
+---@param u string
+---@param sink table
+---@param result boolean
+---@param code integer
+---@param valLocationFunc function
+---@return boolean, boolean|number|string|table|ErrMsg
+local function handleReply(func_name, u, sink, result, code, valLocationFunc)
+  if code == 200 then -- OK
+    local ret, val = pcall(json.decode, table.concat(sink, ""))
```
dljsjr, line 82: Returning the pcall's `ret` value is basically the same thing as re-throwing an exception. I would prefer if the pcall-specific error handling was internal to the function, and I would also prefer to stick with the Lua convention of returning the value or `nil` in the first return spot, and `nil` or an error message in the 2nd return spot.

drivers/SmartThings/harman-luxury/src/init.lua
```diff
@@ -0,0 +1,390 @@
...
+  end
+end
+
+local function do_refresh(driver, device, _)
+  log.info(string.format("Starting do_refresh: %s", device.label))
+
+  -- check and update device IP
+  local dni = device.device_network_id
+  local ip = device:get_field(const.IP)
+  update_connection(device, dni, ip)
```
dljsjr, line 317: See https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/885#discussion_r1317431712

drivers/SmartThings/harman-luxury/src/init.lua
```diff
@@ -0,0 +1,390 @@
...
+  end
+
+  log.info(string.format("create_check_for_updates_thread: dni=%s", device.device_network_id))
+  local new_timer = device.thread:call_on_schedule(const.UPDATE_INTERVAL, function()
+    check_for_updates(device)
+  end, "value_updates_timer")
+  device:set_field(const.UPDATE_TIMER, new_timer)
+end
+
+local function create_check_health_thread(device)
```
dljsjr, line 238: There is non-trivial overhead to creating many tasks; in the case where a user has many speakers in their home, there could be a lot of overhead with having a task per device. I would recommend creating a single background task that monitors for IP address changes and then fans out to the appropriate device.

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
```diff
@@ -0,0 +1,286 @@
...
+
+--- send HTTP request
+---@param u string
+---@return table, boolean, integer
+local function sendRequest(u)
+  local sink = {}
+  local result, code, _ = http.request {
+    url = u,
+    method = "GET",
+    sink = ltn12.sink.table(sink)
```
dljsjr, line 67: Please see @FreeMasen's comment here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/885#pullrequestreview-1565972963

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
```diff
@@ -0,0 +1,286 @@
...
+---@param valLocationFunc function
+---@return boolean, boolean|number|string|table|ErrMsg
+local function handleReply(func_name, u, sink, result, code, valLocationFunc)
+  if code == 200 then -- OK
+    local ret, val = pcall(json.decode, table.concat(sink, ""))
+    val = valLocationFunc(val)
+    log.debug(string.format("Nsdk %s: received value:%s", func_name, st_utils.stringify_table(val)))
+    return ret, val
+  elseif code == 500 then -- ERROR
+    local ret, err = pcall(json.decode, sink[result])
```
dljsjr, line 87: Same here, and everywhere else `pcall` is used; I would prefer to see `pcall` "catches" handled inline instead of being bubbled up.

Just to reiterate: I think it's okay that these functions are fallible. But I would like to see internal error handling with a relevant error based on the `pcall` failure reported in the returns, instead of just bubbling up the boolean transparently.

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
dljsjr: Second this, lots of opportunities for "DRY" refactors in this file I'd like to see seized.

alon-tchelet:
Covered all comments again. Hope it's better now

drivers/SmartThings/harman-luxury/src/api/apis.lua
dljsjr: Typo remains; `Manufature` should be `Manufacturer`

dljsjr (approved):
There are a few remaining style nits but this looks pretty good to me at this point, have there been tests with real devices and real hubs?

dljsjr (approved):
@alon-tchelet Looks ready to go, but this needs to be rebased on the usptream `main` before we can merge.

greens (approved):

PR #886, opened 2023-08-02
Sonos: Add Periodic SSDP scan to detect IP changes

dljsjr:
Similarly to Hue, this adds a periodic scan for detecting changes in IP address
at runtime after a device has onboarded.

This refactor isn't as surgical as the Hue one because the Sonos topology is
simpler, so we didn't do a deep gutting of the discovery logic and left it
mostly untouched, so we still actively perform SSDP scans during discovery.

Relaxing this for Sonos is likely best left to a refactor after we move the
SSDP APIs in to hubcore and expose them via the lua libs.

Fixes: [CHAD-11472](https://smartthings.atlassian.net/browse/CHAD-11472)

cc @TheMegamind

### DIFF
--- a/drivers/SmartThings/sonos/src/init.lua
+++ b/drivers/SmartThings/sonos/src/init.lua
@@ -3,15 +3,15 @@ local old_log = require "log"
 
 -- Print all errors, warnings, and info to hubcore for now
 old_log.error = function(...)
-  old_log.error_with({hub_logs = true}, ...)
+  old_log.error_with({ hub_logs = true }, ...)
 end
 
 old_log.warn = function(...)
-  old_log.warn_with({hub_logs = true}, ...)
+  old_log.warn_with({ hub_logs = true }, ...)
 end
 
 old_log.info = function(...)
-  old_log.info_with({hub_logs = true}, ...)
+  old_log.info_with({ hub_logs = true }, ...)
 end
 
 local log = {}
@@ -56,7 +56,7 @@ local function find_player_for_device(driver, device, should_continue)
   repeat
     --- @param ssdp_group_info SonosSSDPInfo
     SSDP.search(SONOS_SSDP_SEARCH_TERM, function(ssdp_group_info)
-      driver:handle_ssdp_discovery(ssdp_group_info, function(dni, _, _, _)
+      driver:handle_ssdp_discovery(ssdp_group_info, function(dni, inner_ssdp_group_info, player_info, group_info)
         device.log.info(string.format(
             "Found device for Sonos search query with MAC addr %s, comparing to %s",
             dni, device.device_network_id
@@ -64,6 +64,16 @@ local function find_player_for_device(driver, device, should_continue)
         )
         if dni_equal(dni, device.device_network_id) then
           device.log.info(string.format("Found Sonos Player match for device"))
+
+          local field_cache = {
+            household_id = inner_ssdp_group_info.household_id,
+            player_id = player_info.playerId,
+            wss_url = player_info.websocketUrl
+          }
+
+          driver._field_cache[dni] = field_cache
+
+          driver.sonos:update_household_info(player_info.householdId, group_info)
           player_found = true
         end
       end)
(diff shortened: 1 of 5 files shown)


TheMegamind:
To the limited extent I understand what you're trying to accomplish here, I'm curious: All of my Sonos devices have static IPs, so how might a periodic scan have an impact on the problem myself and many other users are having with the speakers becoming unresponsive to the app/API and failing to report events (such as track changes) as they occur? I assume that's why I was cc'd on this PR?

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -73,21 +83,94 @@ local function find_player_for_device(driver, device, should_continue)
...
+  if current_url ~= fields.wss_url then
+    if current_url ~= nil and sonos_conn ~= nil then
+      sonos_conn:stop()
+      device:set_field(PlayerFields.CONNECTION, nil)
+      refresh = true
+    end
+    device:set_field(PlayerFields.WSS_URL, fields.wss_url, { persist = true })
+  end
+
+  if refresh and already_connected then
```
greens, line 122: is the expectation that the existing `sonos_conn` has already been stopped?

dljsjr: Yeah, from #L113-120 above.

The general approach I was going for is:
1. Don't mess with the connection unless the websocket URL has changed
2. If the URL has changed (which is true if the existing URL is either nil or has a different value than the one in the scan), and if there's an existing sonos connection, then we stop it and nil it out.

However, typing it out, I see the *potential* bug. If we need to do a refresh but the connection wasn't changed, then we'll leak a new connection without stopping it.

The reason I missed this is because player_id and household_id are *supposed* to be stable/unchanging so in theory you should never be in a situation where you need to do a refresh unless the websocket URL changes. But I wanted to code very defensively here, and I think in that spirit we should probably also be defensive about that. Probably a `nil` check on the connection is sufficient.

TheMegamind:
> When you say that yours have static IP's, does that mean you did MAC-based IP allocation in your router? 

Yes, I've manually assigned each speaker a static IP address in the router.  My system (6 pairs + 1 sub) has always been that way.  

> I'd be curious if you could provide driver livelogs that have been redirected to a file. 

I assume via the CLI?   I would, but right now there is no driver output because the driver stopped reporting events about an hour ago. 😏

dljsjr:
> > When you say that yours have static IP's, does that mean you did MAC-based IP allocation in your router?
> 
> Yes, I've manually assigned each speaker a static IP address in the router. My system (6 pairs + 1 sub) has always been that way.
> 
> > I'd be curious if you could provide driver livelogs that have been redirected to a file.
> 
> I assume via the CLI? I would, but right now there is no driver output because the driver stopped reporting events about an hour ago. 😏

Yeah, the `edge:drivers:logcat` command from the CLI. My personal preference is to use `smartthings edge:drivers:logcat | tee path/to/file.txt` instead of trying to copy/paste from the scrollback, that way you get file redirection while still being able to use the interactive prompts.

That said, if the livelogs aren't updating *at all*, then that's a sign that something else on your hub is hung and not just the driver. The livelogs should still populate for things like incoming events.

An easy way to check is to do a pull-to-refresh on one of the speakers in the mobile app, or send a refresh to the speaker via the CLI. There should be some activity in the live logs coming from the general edge driver infrastructure regardless of the state of the driver itself, and if you don't see that, then the issue is upstream of the driver in this case.

TheMegamind:
> That said, if the livelogs aren't updating _at all_, then that's a sign that something else on your hub is hung and not just the driver. The livelogs should still populate for things like incoming events.**

_But they don't,_ and I'm stunned you aren't aware of this. This is a huge part of the problem with this integration and has been discussed and reported repeatedly by numerous users from the very beginning. It's not just that the speakers stop responding to commands from the app and API.  Frequently, the driver doesn't report events contemporaneously.  That's why people are complaining that the app doesn't show the correct Status (playing, paused) or Track Name or Album Art or Volume or whatever. 

Sometimes (or perhaps all the time; I can't pour through logs 24-7!) the missing reporting floods into the driver log at some point later, when the speakers "reconnect" or whatever to the integration. I previously sent @nayzlez a log of some two hours of backlogged events that flashed through the log in about 30 seconds when everything "woke up."  I still have that file somewhere.

And, to be clear, this has _only_ been seen with Sonos, and only since the Edge migration. I have no issues with any of my other devices, rules, or scenes.  

> An easy way to check is to do a pull-to-refresh on one of the speakers in the mobile app, or send a refresh to the speaker via the CLI. 

Been there, done that. Did the demo and even posted the log, as I recall. The driver acknowledges the refresh command ... and then does nothing.  The usual return on a refresh just doesn't happen—at least not when the driver-device connection is frozen or stale or otherwise broken. 

There's something bigger at play here, and I'm doubtful that IP changes is all that is—especially since mine never change. 🤷🏻‍♂️

dljsjr:
@TheMegamind Do you mind if I bug you for some clarification on a few statements to help me get to the bottom of this?

You are reporting that the driver isn't servicing any capability commands or attribute events, which is definitely a big problem. For a *large* number of our users (but definitely not all of them), this is because of an issue that impacts *all* of our LAN Edge Drivers related to IP address change detection. It's extremely unfortunate that this fix won't recover your system like I had hoped it would but it will recover it for a lot of users so it's still important.

Obviously there remain users who have integrations that are exhibiting defects that aren't related to this, and I'd like to increase our understanding of this using your experience if possible since you've been so openly communicative with us in the past.

In your comment you've mentioned several symptoms of defective behavior that are extremely frustrating (ones that I myself encounter with my Sonos speakers as well, for what it's worth 😞). It should be noted that the bigger issue is that many of the ones you've called out are overlapping presentations of extremely different root causes, so our best triage strategy is for us to be as data-driven as possible to make sure we're addressing the correct issues even though the symptoms are sometimes very similar.

> Frequently, the driver doesn't report events contemporaneously. That's why people are complaining that the app doesn't show the correct Status (playing, paused) or Track Name or Album Art or Volume or whatever.

We're still isolating situations where the driver or the hub are reponsible for this, but be aware that there is also a mobile app side bug that can cause the app state to fail to update. We've noticed that this is more prevalent on Android than it is on iOS and we're hopeful for many users that upcoming mobile app releases will improve this. There will remain a different set of folks, though, where we're going to need to identify the root cause in the driver still. For now, because the majority of users with defective integrations are here because of connectivity loss, we're hoping to see some improvement with this fix.

> > That said, if the livelogs aren't updating at all, then that's a sign that something else on your hub is hung and not just the driver. The livelogs should still populate for things like incoming events.
> 
> But they don't, and I'm stunned you aren't aware of this.

Clarifying question here: What I was inquiring about is whether or not the livelogs were *completely* silent, or simply silent about events.

In other words, if you fire up the livelog client and you see "connected" but no other activity at all, ever, then that is an issue that is upstream of the driver code itself. While it may be true that folks have only encountered this with Sonos and none of the other integrations, Sonos is also one of the most unique integrations on the platform so the idea that it could be impacted by an upstream bug that doesn't yet have a good test case is extraordinarily likely.

If you are in a situation where there is no live log activity whatsoever, then we'll need hubcore logs to diagnose further. The issue won't be deducable via the edge driver logs.

> Sometimes (or perhaps all the time; I can't pour through logs 24-7!) the missing reporting floods into the driver log at some point later, when the speakers "reconnect" or whatever to the integration. I previously sent @nayzlez a log of some two hours of backlogged events that flashed through the log in about 30 seconds when everything "woke up." I still have that file somewhere.

This further indicates to me that this is likely an issue upstream of the driver itself, and will again require hubcore logs to diagnose further.

> The driver acknowledges the refresh command ... and then does nothing. The usual return on a refresh just doesn't happen—at least not when the driver-device connection is frozen or stale or otherwise broken.

Again seeking clarity: If the driver live logs show activity for the refresh command coming in, but then then there's no **_device_** activity, that definitely sounds like a driver issue. But if the there is no activity in the live logs whatsoever, that is still indicative to me of an upstream/firmware issue.

> There's something bigger at play here, and I'm doubtful that IP changes is all that it is—especially since mine never change.

I very much agree, there are absolutely additional issues to sort out. We're simply trying to take a measured approach and help as many users as we can, as quickly as we can, with the resources we have.

dljsjr:
@TheMegamind if you're comfortable with it, would you mind sending Nayely or Alejandro on the support forums a DM so they can associate the information you're providing here with the information you've provided them in the forums and logs you've sent them as well?

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -304,9 +380,31 @@ local driver = Driver("Sonos", {
...
+  driver.datastore["dni_to_device_id"] = {}
+end
+
+driver._field_cache = driver.datastore._field_cache
+
+-- Kick off a scan right away to attempt to populate some information
+driver:call_with_delay(3, scan_for_ssdp_updates, "Sonos SSDP Initial Scan")
+
+-- re-scan every minute
+local SSDP_SCAN_INTERVAL_SECONDS = 60
```
cjswedes, line 405: Any reason for this value? It seems aggressive to me. I know for most network setups, IPs are not changing that much if at all, so I would think it would be more like once or twice a day rather than every minute.

ctowns (approved):

drivers/SmartThings/sonos/src/init.lua
dljsjr: I had actually intended to set this to 10 minutes, I was using 1 minute in testing.

cjswedes (approved):

greens (approved):

PR #889, opened 2023-08-03
Add fingerprinters and profiles for certification

chenpeixiang:


drivers/SmartThings/matter-switch/profiles/light-color-level-1800-6500K.yml
```diff
@@ -0,0 +1,22 @@
+name: light-color-level-1800-6500K
```
ctowns, line 1: This should be `light-color-level-1800K-6500K` (notice the extra K after 1800) to match naming scheme of other similar fingerprints. Please change the `name:` and the filename, then update the `deviceProfileName` in the fingerprints to match ``light-color-level-1800K-6500K`

lelandblue:
Hello @chenpeixiang I just want to confirm, are you still interested in persuing this device for WWST Certification? We noticed a change on your application. If this was not intentional please reach out to partners@smartthings.com

PR #894, opened 2023-08-04
Add WALLHERO Button

liuyd96:


greens (approved):

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/wallhero/init.lua
```diff
@@ -0,0 +1,103 @@
...
+  local additional_fields = {
+    state_change = true
+  }
+
+  local ep = zb_rx.address_header.src_endpoint.value
+  local button_name = "button" .. ep
+  local event = capabilities.button.button.pushed(additional_fields)
+  local comp = device.profile.components[button_name]
+  if comp ~= nil then
+    device:emit_component_event(comp, event)
```
greens, line 45: this is fine, but in the future you can save yourself a line and just use `emit_event_for_endpoint` which goes and calls `endpoint_to_component`

drivers/SmartThings/zigbee-button/src/zigbee-multi-button/wallhero/init.lua
```diff
@@ -0,0 +1,103 @@
...
+    end
+    -- Without this time delay, the state of some buttons cannot be updated
+    socket.sleep(1)
+  end
+end
+
+local function component_to_endpoint(device, component_id)
+  local ep_num = component_id:match("button(%d)")
+  return ep_num and tonumber(ep_num) or device.fingerprinted_endpoint_id
+end
```
greens, line 71: you technically don't need to define this since we don't send any capability commands to the device

drivers/SmartThings/zigbee-button/src/test/test_wallhero_button.lua
```diff
@@ -0,0 +1,360 @@
...
+
+    test.socket.zigbee:__queue_receive({ mock_device.id, zigbee_test_utils.build_custom_command_id(mock_device, Scenes.ID, Scenes.server.commands.RecallScene.ID, 0x0000, "\x05\x00\x00\x00\x05\x00", 0x1E) })
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message("button30", button_attr.pushed({ state_change = true }))
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message("main", button_attr.pushed({ state_change = true }))
+    )
+    test.wait_for_events()
+
```
greens, line 318: ouch

for future reference, you could have used a loop here. e.g.: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-button/src/test/test_heiman_button.lua#L161

PR #895, opened 2023-08-04
Add WALLHERO Switch

liuyd96:


drivers/SmartThings/zigbee-switch/profiles/wallhero-scene-switch.yml
```diff
@@ -0,0 +1,12 @@
+name: wallhero-scene-switch
```
greens, line 1: I'd just call this something like `button`

drivers/SmartThings/zigbee-switch/src/wallhero/init.lua
```diff
@@ -0,0 +1,125 @@
...
+  end
+  device:refresh()
+end
+
+local function device_added(driver, device)
+  if device.network_type ~= stDevice.NETWORK_TYPE_CHILD then
+    create_child_devices(driver, device)
+  end
+  device:emit_event(capabilities.button.numberOfButtons({ value = 4 }, { visibility = { displayed = false } }))
+  device:emit_event(capabilities.button.supportedButtonValues({ "pushed" }, {visibility = {displayed = false } }))
```
greens, line 86: this will cause an error in the logs, since the switch children do not support the button capability

drivers/SmartThings/zigbee-switch/src/wallhero/init.lua
```diff
@@ -0,0 +1,125 @@
...
+local function scenes_cluster_handler(driver, device, zb_rx)
+  log.info("Enter scenes_cluster_handler")
+  local additional_fields = {
+    state_change = true
+  }
+  local ep = zb_rx.address_header.src_endpoint.value
+  local child_device = device:get_child_by_parent_assigned_key(string.format("%02X", ep))
+  if child_device ~= nil then
+    local event = capabilities.button.button.pushed(additional_fields)
+    child_device:emit_event(event)
```
greens, line 102: you can just call `emit_event_for_endpoint` which will call the `find_child` function you set in `init`

greens (commented):
General question: 

Should the buttons be components of the main device, while the switches are children?

drivers/SmartThings/zigbee-switch/profiles/wallhero-scene-switch.yml
liuyd96: > I'd just call this something like `button`

Got it. I will rename wallhero-scene-switch to button. (Both the filename and name component)

liuyd96:
> General question:
> 
> Should the buttons be components of the main device, while the switches are children?

Did you mean change the deviceProfileName, use the button profile which contains the button capability instead of basic-switch ?

liuyd96:
> > General question:
> > Should the buttons be components of the main device, while the switches are children?
> 
> Did you mean change the deviceProfileName, use the button profile which contains the button capability instead of basic-switch ?

I

drivers/SmartThings/zigbee-switch/src/wallhero/init.lua
```diff
@@ -0,0 +1,118 @@
...
+  end
+  device:refresh()
+end
+
+local function device_added(driver, device)
+  if device.network_type ~= stDevice.NETWORK_TYPE_CHILD then
+    create_child_devices(driver, device)
+  end
+  -- Set Button Capabilities for scene switches
+  if device.profile.components.main.capabilities.button ~= nil then
```
greens, line 86: we would typically use `device:supports_capability_by_id` here

greens:
@liuyd96 that's exactly what I meant. A main profile that has a component for each button, and then the switches are added as children.

greens (approved):

greens:
@liuyd96 I just wanted to say that your work on these drivers has been exceptional. You've picked all of this up very quickly and have been very pleasant to work with. Great work.

PR #897, opened 2023-08-07
[Aqara] Smart Pet Feeder C1

seojune79:
Fix initialization to prevent incorrect behavior


tpmanley:
@seojune79 Is there something that needs to be done to fix devices that are already joined?

seojune79:
> @seojune79 Is there something that needs to be done to fix devices that are already joined?

There's nothing else to do.

seojune79:
> @seojune79 Is there something that needs to be done to fix devices that are already joined?

I'm sorry I misinformed you.
It is necessary the process of onboarding again after removing.

PR #899, opened 2023-08-07
Add support for Aeotec Water Sensor 7

ctowns:
Aeotec Water Sensor 7 device was joining as a generic Zwave Sensor due to a missing fingerprint.


drivers/SmartThings/zwave-sensor/src/zwave-water-temp-humidity-sensor/init.lua
```diff
@@ -22,6 +22,7 @@ local ZWAVE_WATER_TEMP_HUMIDITY_FINGERPRINTS = {
   { manufacturerId = 0x0371, productType = 0x0002, productId = 0x0013 }, -- Aeotec Water Sensor 7 Pro EU
   { manufacturerId = 0x0371, productType = 0x0102, productId = 0x0013 }, -- Aeotec Water Sensor 7 Pro US
   { manufacturerId = 0x0371, productType = 0x0202, productId = 0x0013 }, -- Aeotec Water Sensor 7 Pro AU
+  { manufacturerId = 0x0371, productId = 0x0012 } -- Aeotect Water Sensor 7
```
ctowns, line 25: The non-pro version of the aeotec water sensor 7 does not support temperature or humidity while the pro-version does. However, the only special handling in this sub driver is for the notification report which maps to the water sensor and tamper alert capabilities, which both the pro and non-pro versions of this device support. It seems like it would be useful to have both of these device types use the same sub driver since the sub driver only affects their shared capabilities. However, would it be a good idea to then change the name of the sub driver since one of the devices technically would not support temp and humidity?

greens: yeah, I think this is just copying the name of the DTH these devices were on, but you can change the name of the sub-driver itself no problem

greens (approved):

ctowns:
Just fixed a spelling mistake in a comment

PR #906, opened 2023-08-09
Yeelight Matter Request

jacobvoegele:
Pull request for Yeelight Cubes.


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -33,6 +33,13 @@ matterManufacturer:
     productId: 0xAC
     deviceProfileName: plug-binary
 
+#Yeelight
+  - id: "Yeelight Smart Lamp"
+    deviceLabel: Yeelight Smart Lamp
+    vendorId: 0x1312
+    productId: 0x1
```
ctowns, line 40: nit: I would add a leading zero to make this `0x01` to for consistency with the other fingerprints above

drivers/SmartThings/matter-switch/profiles/light-level-colorTemperature-2700k-6500k.yml
```diff
@@ -0,0 +1,26 @@
+name: light-level-colorTemperature-2700k-6500k
```
ctowns, line 1: This profile is a duplicate of the profile `light-color-level-2700K-6500K` that was introduced here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/872/files

Please delete this file and then update your repo to include the latest changes, then point your fingerprint to the duplicate profile that already exists. Then this should be good to go!

drivers/SmartThings/matter-switch/fingerprints.yml
jacobvoegele: Fixed

ctowns (approved):
Looks great, thanks Jacob!

PR #921, opened 2023-08-17
Lazy load zwave switch sub drivers

ctowns:
This utilizes a new method of loading subdrivers that allows the sub drivers to be dynamically loaded when needed rather than having every subdriver loaded at initialization.

### DIFF
--- a/drivers/SmartThings/zwave-switch/src/init.lua
+++ b/drivers/SmartThings/zwave-switch/src/init.lua
@@ -19,6 +19,7 @@ local defaults = require "st.zwave.defaults"
 local st_device = require "st.device"
 --- @type st.zwave.Driver
 local ZwaveDriver = require "st.zwave.driver"
+local Driver = require "st.driver"
 --- @type st.zwave.CommandClass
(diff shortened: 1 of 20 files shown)


drivers/SmartThings/zwave-switch/src/init.lua
```diff
@@ -130,25 +131,25 @@ local driver_template = {
...
-    require("qubino-switches"),
-    require("fibaro-double-switch"),
-    require("fibaro-single-switch"),
-    require("eaton-5-scene-keypad"),
-    require("ecolink-switch"),
-    require("multi-metering-switch"),
-    require("zooz-zen-30-dimmer-relay"),
-    require("multichannel-device"),
-    require("aeotec-smart-switch")
+    Driver.lazy_load_subdriver(require("eaton-accessory-dimmer")),
```
greens, line 134: won't this be in the `ZwaveDriver` object? I don't think you have to additionally include `Driver` here, do you?

greens (approved):

PR #922, opened 2023-08-21
Hue bugfix bridge offline every driver restart

cjswedes:
Whenever a driver is restarted, a bridge will temporarily be marked offline while the event source connection is created. This seems to be affecting the online/offline state of children incorrectly post 0.49.x FW.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -701,6 +701,8 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
       log.info_with({ hub_logs = true },
         string.format("Event Source Connection for Hue Bridge \"%s\" established, marking online", device.label))
       device:online()
+      --TODO should we do a refresh of all child devices when we first connect?
+      --handlers.refresh_handlers(driver, device)
     end
 
     eventsource.onerror = function()
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -701,7 +701,9 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
       log.info_with({ hub_logs = true },
         string.format("Event Source Connection for Hue Bridge \"%s\" established, marking online", device.label))
       device:online()
-      --We do a refresh of all child devices when we first connect, to determine which are still online.
+      --We do a refresh of all child devices when we first connect to emit events for all child devices
+      -- after the bridge is online, this will cause ST to mark the children online.
+      --TODO use the actual zigbee service to get the actual child device connectivity
       handlers.refresh_handler(driver, device)
     end
 
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -701,8 +701,8 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
       log.info_with({ hub_logs = true },
         string.format("Event Source Connection for Hue Bridge \"%s\" established, marking online", device.label))
       device:online()
-      --TODO should we do a refresh of all child devices when we first connect?
-      --handlers.refresh_handlers(driver, device)
+      --We do a refresh of all child devices when we first connect, to determine which are still online.
+      handlers.refresh_handler(driver, device)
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -700,6 +701,8 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
       log.info_with({ hub_logs = true },
         string.format("Event Source Connection for Hue Bridge \"%s\" established, marking online", device.label))
       device:online()
+      --TODO should we do a refresh of all child devices when we first connect?
+      --handlers.refresh_handlers(driver, device)
```
cjswedes, line 705: Child devices are "refreshed" on the first join, but not on driver restart. Its not clear to me if we should make the change since that would be a relatively large change to driver restart flow if we are trying to get the fix out in time for 0.49.8 prod release. However, such a change may be required to fix this issue for all cases due to the way the lan child device health feature works with 0.49.8.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -688,7 +690,6 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
   if not device:get_field(Fields.EVENT_SOURCE) then
     log.info_with({ hub_logs = true }, "Creating SSE EventSource for bridge " ..
       (device.label or device.device_network_id or device.id or "unknown bridge"))
-    device:offline()
```
varzac, line 691: Will the bridge ever be marked offline if we can't establish a connection?

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -54,8 +54,10 @@ local function emit_light_status_events(light_device, light)
   if light_device ~= nil then
     if light.status then
       if light.status == "connected" then
+        light_device.log.info_with({hub_logs=true}, "Light status event, marking device online")
```
varzac, line 57: Do we want this logging elevated to hub logs?

drivers/SmartThings/philips-hue/src/init.lua
greens: if we don't do that fix, doesn't this not fix the problem (i.e. child devices remaining offline)?

drivers/SmartThings/philips-hue/src/init.lua
cjswedes: There are surprisingly few logs related to when these apis are called, so I added it to detect a situation in logs where the hubs Device watch state is not updated (no logging at all) but the api was called.

drivers/SmartThings/philips-hue/src/init.lua
cjswedes: Yes, if the connection was connected and then errored out, or there was an error when connecting. I will manually test this case again though to be sure.

drivers/SmartThings/philips-hue/src/init.lua
cjswedes: Yup, you're right. Without the change any time a bridge comes online after being offline the children would erroneously remain offline rather than being queried to see if they are online.

drivers/SmartThings/philips-hue/src/init.lua
ctowns: I tested this by disconnecting the bridge and restarting the driver. As the driver restarted, the bridge (and children) were correctly marked as offline.

ctowns: Now using the correct test setup - I still that the hue bridge is correctly marked offline at startup when it is disconnected.

greens (approved):

cjswedes:
> Firstly, the zigbee connectivity status isn't guaranteed to come in over a refresh. The only way this would be "fixing" things is by emitting events for lights that are connected to the bridge and so forcing an online event in that way, and the Hue bridge will gladly give you the last known state for a freshly disconnected device to get emitted. So it could be a false positive.

Thanks for the clarification on how the bridge actually reports this. After discussing offline we should make a change that will poll the zigbee connectivity service for each device when the bridge is initialized (avoiding an http request flood). However, we should leave the refresh call in place because it seems that the attribute events emitted to ST cause the child devices to come online.

PR #926, opened 2023-08-22
Poll bulbs for their Zigbee Connectivity status explicitly when refreshing

dljsjr:
Adds explicit checking for zigbee connectivity status as part of refresh, instead of always waiting for it to come in on the SSE stream.

Note that the zigbee connectivity status itself within the Hue network can be delayed.

Because we're now doing more network calls during refresh, I also adjusted the "refresh all" handler to use the bulk endpoints.

To keep the changeset on this small, I did it by overloading the functionality of the existing refresh for single lights. It's noted in comments that this isn't the most ideal, however, I'm in the process of planning out a rearch of this driver including with a test coverage plan, so I didn't think deep surgery was warranted given that we're going to be doing a complete redo soon.

Pardon the Mess™

### DIFF
--- a/drivers/SmartThings/philips-hue/src/handlers.lua
+++ b/drivers/SmartThings/philips-hue/src/handlers.lua
@@ -2,7 +2,9 @@ local Fields = require "hue.fields"
 local HueApi = require "hue.api"
 local HueColorUtils = require "hue.cie_utils"
 local log = require "log"
+local utils = require "utils"
 
+local cosock = require "cosock"
 local capabilities = require "st.capabilities"
 local st_utils = require "st.utils"
 
@@ -245,8 +247,45 @@ end
 
 ---@param driver HueDriver
 ---@param light_device HueChildDevice
-local function do_refresh_light(driver, light_device)
+---@param conn_status_cache table|nil
+---@param light_status_cache table|nil
+local function do_refresh_light(driver, light_device, conn_status_cache, light_status_cache)
   local light_resource_id = light_device:get_field(Fields.RESOURCE_ID)
+  local hue_device_id = light_device:get_field(Fields.HUE_DEVICE_ID)
+
+  local do_zigbee_request = true
+  local do_light_request = true
+  local light_online = true
+
+  if type(conn_status_cache) == "table" then
+    local zigbee_status = conn_status_cache[hue_device_id]
+    if zigbee_status ~= nil then
+      if zigbee_status.status and zigbee_status.status == "connected" then
+        light_device.log.debug(string.format("Zigbee Status for %s is connected", light_device.label))
+        light_device:online()
+        do_zigbee_request = false
+      else
+        light_device.log.debug(string.format("Zigbee Status for %s is not connected", light_device.label))
+        light_device:offline()
+        do_zigbee_request = false
+        light_online = false
+      end
+    end
+  end
+
+  if type(light_status_cache) == "table" then
+    local light_info = light_status_cache[hue_device_id]
+    if light_info ~= nil then
+      if light_info.id == light_resource_id then
+        if light_info.color ~= nil and light_info.color.gamut then
+          light_device:set_field(Fields.GAMUT, light_info.color.gamut_type, { persist = true })
+        end
+        driver.emit_light_status_events(light_device, light_info)
+        do_light_request = false
+      end
+    end
+  end
+
   local bridge_id = light_device.parent_device_id or light_device:get_field(Fields.PARENT_DEVICE_ID)
   local bridge_device = driver:get_device_info(bridge_id)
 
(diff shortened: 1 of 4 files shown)


drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -247,6 +248,7 @@ end
 ---@param light_device HueChildDevice
 local function do_refresh_light(driver, light_device)
```
dljsjr, line 249: This function has become exceedingly smelly, but I didn't think that now was the right time to be dealing with that. One of the major issues is that the whole driver is oriented around light devices on our platform representing a light resource in the Hue representation. We should probably revise that to orient around the Hue device representation instead of the light resource representation, but that's a deep surgery on the entire driver.

drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -323,7 +389,9 @@ end
 ---@param device HueDevice
 function handlers.refresh_handler(driver, device, cmd)
   if device:get_field(Fields.DEVICE_TYPE) == "bridge" then
-    do_refresh_all_for_bridge(driver, device --[[@as HueBridgeDevice]])
+    cosock.spawn(function()
+      do_refresh_all_for_bridge(driver, device --[[@as HueBridgeDevice]])
```
dljsjr, line 393: I mistakenly assumed that Past Doug was smart enough to write this using the bulk REST call and filtering the returned values, but that is not the case; in the name of DRY-ness, this makes 3 API calls per device.

This definitely needs to be fixed soon.

drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -265,28 +267,92 @@ local function do_refresh_light(driver, light_device)
...
+  --- the information for the light that we expect to getting the info for.
   repeat
-    local light_resp, err = hue_api:get_light_by_id(light_resource_id)
     count = count + 1
-    if err ~= nil then
-      log.error_with({ hub_logs = true }, err)
-    elseif light_resp ~= nil then
-      if #light_resp.errors > 0 then
-        for _, err in ipairs(light_resp.errors) do
+    rest_resp, rest_err = hue_api:get_device_by_id(hue_device_id)
```
varzac, line 280: So is this call to the device record required because we need this to get the zigbee resource, which is needed to get that connectivity state?  After seeing these changes I'm a little more concerned about this approach of adding all this to refresh.  Would we be better off explicitly checking all the children connectivity status via a separate method only on the transition from hub offline -> online.

drivers/SmartThings/philips-hue/src/handlers.lua
```diff
@@ -265,28 +267,92 @@ local function do_refresh_light(driver, light_device)
...
   local count = 0
   local num_attempts = 3
+  local rest_resp, rest_err = nil, nil
+  local zigbee_resource_id = nil
+  --- this loop is a rate-limit dodge.
+  ---
+  --- One of the various symptoms of hitting the Hue Bridge's rate limit is that you'll get a silent
+  --- failure that takes the form of the bridge returning the last valid response it replied with.
+  --- So we hit the bridge 2-3 times and check the IDs in the responses to verify that we're getting
+  --- the information for the light that we expect to getting the info for.
```
greens, line 277: would it make sense to do a backoff/sleep/whatever if we notice that this has happened?

drivers/SmartThings/philips-hue/src/handlers.lua
dljsjr: I'm of two minds about it.

On the one hand, yes, that's a lot of REST calls in a single event handler. We could pare it down to 2 instead of 3 if we stored the zigbee ID on the ST device record, but we're already storing so many strings in the driver that there's a real memory usage concern.

On the other hand, I do think it makes sense to check the connectivity status when you do a pull-to-refresh for a variety of reasons, not least of which being that Hue bridge will sometimes return last-known-state for an offline bulb.

drivers/SmartThings/philips-hue/src/handlers.lua
dljsjr: Probably but I didn't want to make changes outside of the scope of what was needed to unblock the firmware release.

drivers/SmartThings/philips-hue/src/handlers.lua
dljsjr: I think the third option could be not reusing the refresh code in the "refresh all" handler on the bridge, and having the "refresh all" code use the bulk endpoints, but that didn't seem like a change that was straight forward enough to have ready on the timelines we were targeting.

greens (approved):

tandres (approved):
LGTM

PR #930, opened 2023-08-23
[Aqara] Cube T1 Pro

seojune79:
add Aqara Cube T1 Pro

### DIFF
--- a/drivers/SmartThings/zigbee-button/src/aqara/init.lua
+++ b/drivers/SmartThings/zigbee-button/src/aqara/init.lua
@@ -1,0 +1,128 @@
+local capabilities = require "st.capabilities"
+local data_types = require "st.zigbee.data_types"
+local cluster_base = require "st.zigbee.cluster_base"
+local battery_defaults = require "st.zigbee.defaults.battery_defaults"
+local clusters = require "st.zigbee.zcl.clusters"
+local PowerConfiguration = clusters.PowerConfiguration
+
+local PRI_CLU = 0xFCC0
+local PRI_ATTR = 0x0009
+local MFG_CODE = 0x115F
+
+local ROTATE_CLU = 0x000C
+local EVENT_CLU = 0x0012
+local FACE_ATTR = 0x0149
+local ACTION_ATTR = 0x0055
+local CUBE_MODE = 0x0148
+
+local cubeAction = capabilities["stse.cubeAction"]
+local cubeFace = capabilities["stse.cubeFace"]
+local cubeFaceVal = { "face1Up", "face2Up", "face3Up", "face4Up", "face5Up", "face6Up" }
+local cubeFlipToSideVal = { "flipToSide1", "flipToSide2", "flipToSide3", "flipToSide4", "flipToSide5", "flipToSide6" }
+
+local CUBEACTION_TIMER = "cubeAction_timer"
+local CUBEACTION_TIME = 3
+
+local FINGERPRINTS = { mfr = "LUMI", model = "lumi.remote.cagl02" }
+
+local is_aqara_products = function(opts, driver, device, ...)
+  return device:get_manufacturer() == FINGERPRINTS.mfr and device:get_model() == FINGERPRINTS.model
+end
+
+local callback_timer = function(driver, device, cmd)
+  return function()
+    device:emit_event(cubeAction.cubeAction("noAction"))
+  end
+end
+
+local function delete_timer(device)
+  local timer = device:get_field(CUBEACTION_TIMER)
+  if timer then
+    device.thread:cancel_timer(timer)
+    device:set_field(CUBEACTION_TIMER, nil, { persist = true })
+  end
+end
+
+local function data_handler(driver, device, value, zb_rx)
+  local val = value.value
+  local timer = nil
+  if val == 0x0000 then -- Shake
+    device:emit_event(cubeAction.cubeAction("shake"))
+    delete_timer(device)
+    device:set_field(CUBEACTION_TIMER, device.thread:call_with_delay(CUBEACTION_TIME, callback_timer(driver, device)))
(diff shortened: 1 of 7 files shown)


drivers/SmartThings/zigbee-button/src/aqara/init.lua
```diff
@@ -0,0 +1,128 @@
...
+  end
+end
+
+local function data_handler(driver, device, value, zb_rx)
+  local val = value.value
+  local timer = nil
+  if val == 0x0000 then -- Shake
+    device:emit_event(cubeAction.cubeAction("shake"))
+    delete_timer(device)
+    device:set_field(CUBEACTION_TIMER, device.thread:call_with_delay(CUBEACTION_TIME, callback_timer(driver, device)))
```
greens, line 52: These two lines always appear together. I would suggest you pull all of this logic out into one function rather than defining two functions that are always called together in the same order.

drivers/SmartThings/zigbee-button/src/test/test_aqara_cube_t1_pro.lua
```diff
@@ -0,0 +1,165 @@
+local test = require "integration_test"
+local cluster_base = require "st.zigbee.cluster_base"
+local t_utils = require "integration_test.utils"
+local zigbee_test_utils = require "integration_test.zigbee_test_utils"
+local capabilities = require "st.capabilities"
+local data_types = require "st.zigbee.data_types"
+local clusters = require "st.zigbee.zcl.clusters"
+local PowerConfiguration = clusters.PowerConfiguration
```
greens, line 8: unused: this is the luacheck error

greens: maybe you should write a configuration test

drivers/SmartThings/zigbee-button/src/aqara/init.lua
seojune79: I fixed it.

drivers/SmartThings/zigbee-button/src/test/test_aqara_cube_t1_pro.lua
seojune79: I removed them.

drivers/SmartThings/zigbee-button/src/aqara/init.lua
```diff
@@ -0,0 +1,124 @@
...
+  return function()
+    device:emit_event(cubeAction.cubeAction("noAction"))
+  end
+end
+
+local function reset_thread(device)
+  local timer = device:get_field(CUBEACTION_TIMER)
+  if timer then
+    device.thread:cancel_timer(timer)
+    device:set_field(CUBEACTION_TIMER, nil, { persist = true })
```
greens, line 42: this doesn't need to persist

it takes more than 3s to reboot the hub anyway

drivers/SmartThings/zigbee-button/src/aqara/init.lua
```diff
@@ -0,0 +1,124 @@
...
+  end
+  device:set_field(CUBEACTION_TIMER, device.thread:call_with_delay(CUBEACTION_TIME, callback_timer(driver, device)))
```
greens, line 44: `driver` is undefined here

greens: but isn't used in `callback_timer`, so it can be removed there too

drivers/SmartThings/zigbee-button/src/aqara/init.lua
seojune79: Okay, I see it.

drivers/SmartThings/zigbee-button/src/aqara/init.lua
seojune79: I applied the changes you requested.

lelandblue:
Should the firmwareUpdate capability also be included?

greens (approved):

tpmanley:
> So this is another case where it doesn't really seem right to have this device under one of our generic drivers. It has no shared functionality with zigbee-button at all. @tpmanley and @ctowns what do you think?

Apologies for the delay. In addition to no shared functionality with zigbee-button, it also doesn't use the button capability (although it does use the RemoteController category). I wouldn't be opposed to putting this in its own driver, similar to the Pet Feeder. Is that what you were thinking too?

greens:
@tpmanley that's correct

lelandblue:
Hey @seojune79 We would like to see this device use its own driver due to the uniqueness of the device. Can you please refactor this PR to be an independent driver in the Aqara folder please? 

Also just so your aware I attempted this driver with a T1 Pro device on my hub and it was not responsive to the rotation of the cube.

seojune79:
> Also just so your aware I attempted this driver with a T1 Pro device on my hub and it was not responsive to the rotation of the cube

I have changed the folder of the cube. 
Please refer to the video for the cube rotation.
https://drive.google.com/file/d/1NEebgX0n_pPR8gFul2hyD9drBITsuoM2/view

drivers/Aqara/aqara-cube/src/init.lua
```diff
@@ -0,0 +1,127 @@
...
+      [PRI_CLU] = {
+        [FACE_ATTR] = face_handler
+      }
+    }
+  },
+  lifecycle_handlers = {
+    init = device_init,
+    added = device_added
+  },
+  can_handle = is_aqara_products
```
greens, line 122: as a base driver, you do not need to supply a `can_handle` function

seojune79: It was removed can_handle.

greens (approved):

PR #931, opened 2023-08-23
fix: Don't emit events for bulbs known offline

dljsjr:
Requires tracking some non-persistent state for devices.


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -751,8 +758,10 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
               if status.status == "connected" then
                 child_device.log.trace("Marking Online after SSE Reconnect")
                 child_device:online()
+                child_device:set_field(Fields.IS_ONLINE, true)
```
dljsjr, line 761: This runs every time the SSE stream opens, so will run at driver start

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -55,12 +55,18 @@ local function emit_light_status_events(light_device, light)
     if light.status then
       if light.status == "connected" then
         light_device:online()
+        light_device:set_field(Fields.IS_ONLINE, true)
```
dljsjr, line 58: This is where the SSE event gets processed.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -55,12 +55,18 @@ local function emit_light_status_events(light_device, light)
...
       elseif light.status == "connectivity_issue" then
         light_device:offline()
+        light_device:set_field(Fields.IS_ONLINE, false)
         return
       end
     end
 
+    if not light_device:get_field(Fields.IS_ONLINE) then
```
varzac, line 66: I would like to understand the consequences of the difference between 

```
if not is_online
```
and
```
if is_online is false
```

dljsjr: I was actually debating on making this an explicit boolean check, so I'm glad you asked that question, cuz it's forcing me to think about it harder.

`if not <val>` will also capture the `nil` case, e.g., the value hasn't been set yet, e.g., we haven't had a status update for that bulb in any capacity. I figured it was good to treat that the same as the false case and so treat that bulb as offline. So `if not <val>` will only happen if value is `false` or `nil`; `true` or any other value will evaluate truthy.

`if <val> == false` explicitly looks for that value being assigned false, meaning in the case that we don't know what's going on, we'd treat this as `true`.

If anything, what we may want to do is make an explicit check against the value of `true`, which is what I was going back and forth on.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -681,6 +687,7 @@ light_added = function(driver, device, parent_device_id, resource_id)
 
   driver.light_id_to_device[device_light_resource_id] = device
 
+  device:set_field(Fields.IS_ONLINE, true)
```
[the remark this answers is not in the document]
cjswedes: Bulbs still are added if they are offline, correct? I think we should not emit any events until we know a device is online. In this case we will do the refresh for the device and emit initial state making the ST platform think the device is online when its not.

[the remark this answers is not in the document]
dljsjr: I thought it would be better to have the field follow the synthetic state we establish in `added` to populate DW, and then let the SSE onopen code handle rectifying this.

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -705,8 +710,14 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
 
       local bridge_api = device:get_field(Fields.BRIDGE_API)
       cosock.spawn(function()
+        -- Wait bit before doing querying to allow initial state for added devices to be reported
+        -- Children will all remain offline for this amount of time when a bridge comes back online.
+        -- Children will all be online for this amount of time when a bridge id first joined.
+        cosock.socket.sleep(15) --TODO is this long enough for large hue setup initial onboarding to complete for all devices?
```
cjswedes, line 716: A delay is needed for the onboarding case where we do emit events for offline devices.  Perhaps there is an event based solution to do this scan at the opportune time, but I struggled to find an event for "all devices added" during the onboarding of a hue setup.

cjswedes (commented):
My commit fixed an onboarding issue I was having where a bulb was onboarded as online despite being offline. Another option we could do is scan connectivity every time after a bridge comes online, and then also do a scan 1 min or so after the bridge added event. This way we have responsive device health for children when bridge is naturally coming online/offline and also attempt to correct any offline devices that are marked online immediately post join due to the synthetic capability event emission.

dljsjr:
@cjswedes after merging #926 and rebasing this on top of it, then cleaning up the conflicts, I was able to remove the hard-coded cosock delay and I believe this is now behaving the way we want it to. Could use another review, though.

NoahCornell:
Does this need its field set or does it not matter because its the bridge and not a light? https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/931/files#diff-01ac799356794c771516edfa70fbd5b602c9fa11a15603b8566c2bc245fd22d5R711

NoahCornell (approved):
Looks good to me, the only thing I can think of is that it would be nice if you could easily override the device table's `online` and `offline` methods so that they could do this without having to add it everywhere we set the device offline and online but I know that is difficult/not possible with the way device tables are setup.

dljsjr:
> Does this need its field set or does it not matter because its the bridge and not a light? [#931 (files)](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/931/files#diff-01ac799356794c771516edfa70fbd5b602c9fa11a15603b8566c2bc245fd22d5R711)

Right now it only needs to be set on lights.

dljsjr:
> Looks good to me, the only thing I can think of is that it would be nice if you could easily override the device table's `online` and `offline` methods so that they could do this without having to add it everywhere we set the device offline and online but I know that is difficult/not possible with the way device tables are setup.

Yeah I discovered this when I did the tracing to validate the changes in #926. I had to replicate the entire `device.lua` file in-tree.

cjswedes (approved):

PR #935, opened 2023-08-28
Fix up unit tests and refresh behavior for subdrivers

cjswedes:
I think we should hotfix this to our beta channel asap to make the prod driver release this week. It seems there were a few cases that were not fully addressed by the lua libs in 0.49 wrt sending wake up interval get commands once when a driver starts up. These changes address those.

### DIFF
--- a/drivers/SmartThings/zwave-sensor/src/everspring-motion-light-sensor/init.lua
+++ b/drivers/SmartThings/zwave-sensor/src/everspring-motion-light-sensor/init.lua
@@ -15,7 +15,8 @@
 local capabilities = require "st.capabilities"
 local SwitchBinary = (require "st.zwave.CommandClass.SwitchBinary")({version=2,strict=true})
 local SensorBinary = (require "st.zwave.CommandClass.SensorBinary")({version=2})
-
+local WakeUp = (require "st.zwave.CommandClass.WakeUp")({ version = 2 })
+local cc = require "st.zwave.CommandClass"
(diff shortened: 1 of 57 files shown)


drivers/SmartThings/zwave-sensor/src/everspring-motion-light-sensor/init.lua
```diff
@@ -32,11 +33,23 @@ local function device_added(driver, device)
   device:send(SensorBinary:Get({ sensor_type = SensorBinary.sensor_type.MOTION }))
 end
 
+local wakeup_notification = nil
+local version = require "version"
+if version.api == 6 then
+  --TODO remove once this happens properly for subdrivers that dont override the default
```
cjswedes, line 39: There are more subdrivers that rely on the battery default for WakeUpNotifications to issue `device:refresh()`. That commit is incoming and will affect several more sub drivers with a similar change as is here.

cjswedes: They were not fixed in the initial commit because there weren't any test that verify the refresh for several of these devices.

greens (commented):
What would happen if we overrode the defaults in each base driver, or had a subdriver whose can_handle was `version.api == 6`?

cjswedes:
> I think this can work...Im gonna give it a try.

I put up #936 which seems like a better option than the changes here.

cjswedes:
Im closing this in favor of #936

PR #936, opened 2023-08-29
Fix/default wakeup override bug 49 release

cjswedes:
This is an alternative solution suggfested by @greens to #935. It seems better because it achieves the same coverage of tcases where a battery powered device is using a subdriver, but it will be easier to revert.

### DIFF
--- a/drivers/SmartThings/zwave-button/src/apiv6_bugfix/init.lua
+++ b/drivers/SmartThings/zwave-button/src/apiv6_bugfix/init.lua
@@ -1,0 +1,26 @@
+local cc = require "st.zwave.CommandClass"
+local WakeUp = (require "st.zwave.CommandClass.WakeUp")({ version = 1 })
+
+
+local function can_handle(opts, driver, device, cmd, ...)
+  local version = require "version"
+  return version.api == 6 and
+    cmd.cmd_class == cc.WAKE_UP and
+    cmd.cmd_id == WakeUp.NOTIFICATION
+end
+
+local function wakeup_notification(driver, device, cmd)
+  device:refresh()
+end
+
+local apiv6_bugfix = {
+  zwave_handlers = {
+    [cc.WAKE_UP] = {
+      [WakeUp.NOTIFICATION] = wakeup_notification
+    }
+  },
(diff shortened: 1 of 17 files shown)


drivers/SmartThings/zwave-sensor/src/aeotec-multisensor/multisensor-6/init.lua
```diff
@@ -34,7 +34,8 @@ local function wakeup_notification(driver, device, cmd)
     device:set_field("__wakeup_interval_get_sent", true)
   end
   device:send(Configuration:Get({parameter_number = PREFERENCE_NUM}))
-  device:refresh()
+  local version = require "version"
+  if version.api ~= 6 then device:refresh() end
```
cjswedes, line 38: I ended up doing it this way along the way, but I could instead exclude the multisensor 6 and 7 in the `apiv6_bugfix` can_handle function like I did for a few other devices that have their own wakeup_notification handler. Looking at it this morning it seems better to do this to keep the bugfix local to its own subdriver, so I will go forward with that.

drivers/SmartThings/zwave-sensor/src/test/test_enerwave_motion_sensor.lua
```diff
@@ -81,6 +81,12 @@ test.register_coroutine_test(
           mock_sensor,
           Battery:Get({})
       ))
+      --This extra battery get is because of apiv6 default handler override sub driver
+      --It was left in so when we remove that subdriver, a new bug isn't created
```
cjswedes, line 85: We could specifically avoid these if we wanted to, but I thought that it opened us up to extra bugs. Like the multisensor case, I will try to exclude this device from the bugfix with fingerprints.

greens (approved):

greens:
should probably squash these btw

greens:
```
Checking drivers/SmartThings/zwave-sensor/src/apiv6_bugfix/init.lua 1 warning

    drivers/SmartThings/zwave-sensor/src/apiv6_bugfix/init.lua:11:7: (W211) unused variable SENSATIVE_STRIP_FP
```

PR #945, opened 2023-09-07
Minor fix: Fix compatibility issues

liuyd96:
Windows's path is different with Linux's path. Therefore, some codes for processing file paths cannot be used across platforms.


greens (approved):
Nice. Thanks @liuyd96 .

PR #947, opened 2023-09-08
BUG-8470 Add support for SmartWings window covering

greens:


cjswedes (approved):

drivers/SmartThings/zigbee-window-treatment/src/IKEA/init.lua
```diff
@@ -94,7 +94,8 @@ local ikea_window_treatment = {
     }
   },
   can_handle = function(opts, driver, device, ...)
-    return device:get_manufacturer() == "IKEA of Sweden"
+    return device:get_manufacturer() == "IKEA of Sweden" or
+      device:get_manufacturer() == "Smartwings"
```
tpmanley, line 98: If we're putting other manufacturers in here could we rename the subdriver too to better describe what this subdriver covers?

tpmanley (approved):

PR #948, opened 2023-09-08
CHAD-11494: Add localizations for Chinese join names

greens:


cjswedes (approved):
Is this just a one of case of doing this because the devices are only in china, or are we going to have to chose which language to use by default for everything (I thought that it would be English if we did this).

ctowns (approved):

drivers/SmartThings/zigbee-button/fingerprints.yml
```diff
@@ -1,6 +1,6 @@
 zigbeeManufacturer:
   - id: "HEIMAN/SOS-EM"
-    deviceLabel: HEIMAN Button
+    deviceLabel: 海曼智能紧急按钮
```
ctowns, line 3: Just to confirm - these device label changes are going to be reverted back to the english names, right?

greens:
@cjswedes this is only possible because the CN environment is segregated from the rest of the global environments. If we do end up with a join name localization strategy in the future, I do not expect it will look like this.

PR #949, opened 2023-09-11
Fix color temperature handling when attribute is outside capability range

ctowns:
Some devices could report a value of 0 mireds, which would result
in a divide by zero error when converting to kelvin. This
introduces a check to make sure that 0 and other mired values
that are outside the range of the colorTemperature capability are
not handled since they are likely not real values.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -23,6 +23,11 @@ local RECEIVED_X = "receivedX"
 local RECEIVED_Y = "receivedY"
 local HUESAT_SUPPORT = "huesatSupport"
 local CONVERSION_CONSTANT = 1000000
+-- These values are taken from the min/max definined in the colorTemperature capability
+local COLOR_TEMPERATURE_KELVIN_MAX = 30000
+local COLOR_TEMPERATURE_KELVIN_MIN = 1
+local COLOR_TEMPERATURE_MIRED_MAX = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN
+local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX
 
 local function convert_huesat_st_to_matter(val)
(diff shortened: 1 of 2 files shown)


ctowns:
As it stands right now, this PR will clamp the bounds that the color temperature attribute reported in mireds to fit within the range provided by our capability (1-30,000 Kelvin, which is 1,000,000 - 33 mired, respectively). Most bulbs seem to fall in the 1800K - 8000K range, and I don't think there is any practical application that would go beyond 30,000K, so I think this range checking might be overkill and missing the real point of the problem (the zero case, addressed below). I suppose it would still be good to check and provide a log message for, but anything outside this range is certainly strange - any thoughts on this?

Right now this PR also treats 0 mired as being beyond the max allowable value so it treats it as 30,000K. This seems like a bad idea since zero is actually not a valid value to begin with (kelvin = 1,000,000 / mired). What would be the best way to handle the zero case beyond just leaving a message in the logs? Should be default it to some kind of sane value (i.e. the default value for the cluster definition, 4000K)? The spec does not seem to provide a satisfying answer for how to handle an impossible value like this.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -217,7 +222,12 @@ end
 
 local function temp_attr_handler(driver, device, ib, response)
   if ib.data.value ~= nil then
-    local temp = utils.round(CONVERSION_CONSTANT/ib.data.value)
+    local temp_in_mired = ib.data.value
+    if (temp_in_mired < COLOR_TEMPERATURE_MIRED_MIN or temp_in_mired > COLOR_TEMPERATURE_MIRED_MAX) then
+      device.log.warn_with({hub_logs = true}, "Device reported color temperature outside of supported capability range")
+      temp_in_mired = utils.clamp_value(temp_in_mired, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX)
```
greens, line 228: I think we probably just want to drop this on the floor, since it's probably not a _real_ value.

greens: i.e. just `return` after the log message

greens (approved):

PR #959, opened 2023-09-15
Add a nil check for matter window covering level in status handler

ctowns:
The CurrentPositionLiftPercent100ths attribute can report a value of nil when the current position is unknown (e.g. calibration is needed). Add a check to make sure we do not try to use position if it is nil.


greens:
not sure from where/how, but you've got a failing test

greens (approved):

PR #963, opened 2023-09-18
[Aqara] Smart Wall Switch(With Neutral, No Neutral)

seojune79:
fix routine condition error on child button of light switch

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
@@ -8,9 +8,9 @@ local PRIVATE_ATTRIBUTE_ID = 0x0009
 local MFG_CODE = 0x115F
 
 local FINGERPRINTS = {
-  { mfr = "LUMI", model = "lumi.switch.n1acn1", children = 1, child_profile = "" },
-  { mfr = "LUMI", model = "lumi.switch.n2acn1", children = 2, child_profile = "aqara-switch-child" },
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
```diff
@@ -81,6 +81,12 @@ local function device_added(driver, device)
 
     device:send(cluster_base.write_manufacturer_specific_attribute(device,
       PRIVATE_CLUSTER_ID, PRIVATE_ATTRIBUTE_ID, MFG_CODE, data_types.Uint8, 0x01)) -- private
+  elseif device.network_type == "DEVICE_EDGE_CHILD" then
+    device:emit_event(capabilities.button.supportedButtonValues({ "pushed" },
+      { visibility = { displayed = false } }))
+    device:emit_event(capabilities.button.numberOfButtons({ value = 1 },
+      { visibility = { displayed = false } }))
+    device:emit_event(capabilities.button.button.pushed({ state_change = false }))
```
greens, line 89: just move the identical code from above outside of the `if` block

also write some tests to capture this behavior.

seojune79: I applied the changes you requested.

greens (approved):

greens (changes requested):
Actually, can you capture this behavior in a test, please?

greens (approved):

PR #967, opened 2023-09-19
Matter-sensor add support for mains powered sensors

ctowns:
The `matter-sensor` generic fingerprints all join to profiles with `battery` support by default. This change will attempt to check for battery support when the device is configured and then change to the appropriate version of the profile (battery or no battery).

### DIFF
--- a/drivers/SmartThings/matter-sensor/profiles/motion-illuminance-battery.yml
+++ b/drivers/SmartThings/matter-sensor/profiles/motion-illuminance-battery.yml
@@ -1,10 +1,12 @@
-name: motion-battery
+name: motion-illuminance-battery
 components:
 - id: main
   capabilities:
   - id: motionSensor
     version: 1
   - id: battery
+    version: 1
+  - id: illuminanceMeasurement
     version: 1
   - id: firmwareUpdate
     version: 1
--- a/drivers/SmartThings/matter-sensor/src/init.lua
+++ b/drivers/SmartThings/matter-sensor/src/init.lua
@@ -23,6 +23,46 @@ local function device_init(driver, device)
   device:subscribe()
 end
 
+local function do_configure(driver, device)
+  local battery_eps = device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
+  local profile_name = ""
+
+  if device:supports_capability(capabilities.motionSensor) then
+    profile_name = profile_name .. "-motion"
+  end
+
+  if device:supports_capability(capabilities.contactSensor) then
+    profile_name = profile_name .. "-contact"
+  end
+
+  if device:supports_capability(capabilities.illuminanceMeasurement) then
+    profile_name = profile_name .. "-illuminance"
+  end
+
+  if device:supports_capability(capabilities.temperatureMeasurement) then
+    profile_name = profile_name .. "-temperature"
+  end
+
+  if device:supports_capability(capabilities.relativeHumidityMeasurement) then
+    profile_name = profile_name .. "-humidity"
+  end
+
+  if #battery_eps > 0 then
+    profile_name = profile_name .. "-battery"
+  end
+
+  -- remove leading "-"
+  profile_name = string.sub(profile_name, 2)
+
+  device:try_update_metadata({profile = profile_name})
+end
+
+local function info_changed(driver, device, event, args)
+  if device.profile.id ~= args.old_st_store.profile.id then
+    device:subscribe()
+  end
+end
+
 local function illuminance_attr_handler(driver, device, ib, response)
   local lux = math.floor(10 ^ ((ib.data.value - 1) / 10000))
   device:emit_event_for_endpoint(ib.endpoint_id, capabilities.illuminanceMeasurement.illuminance(lux))
@@ -60,7 +100,9 @@ end
 
 local matter_driver_template = {
   lifecycle_handlers = {
-    init = device_init
+    init = device_init,
+    doConfigure = do_configure,
+    infoChanged = info_changed
   },
   matter_handlers = {
     attr = {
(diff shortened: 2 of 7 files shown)


ctowns:
The unit tests are currently misbehaving. Although I have been able to pass 3/3, I cannot get them to pass consistently when running locally.

drivers/SmartThings/matter-sensor/profiles/motion-battery.yml
```diff
@@ -1,4 +1,4 @@
-name: matter-motion-battery
+name: motion-battery
```
ctowns, line 1: I know that deleting profiles can cause some trouble, but is renaming them okay or would this lead to similar issues? I renamed this so that it would follow the naming scheme used to construct the profile name in `do_configure`

drivers/SmartThings/matter-sensor/profiles/motion-illuminance-battery.yml
```diff
@@ -0,0 +1,16 @@
+name: motion-illuminance-battery
```
ctowns, line 1: This profile is a copy of `matter-motion-battery-illuminance` but I duplicated it so that it would follow the naming scheme that is used in `do_configure` for updating the profile

drivers/SmartThings/matter-sensor/profiles/motion-battery.yml
greens: I believe it leaves users who have already installed the device with this profile in an "orphaned" state just the same as a delete.

greens (commented):
Won't we need to re-subscribe after a profile change?

drivers/SmartThings/matter-sensor/src/test/test_matter_sensor_featuremap.lua
```diff
@@ -0,0 +1,180 @@
...
+      subscribe_request_temp_humidity:merge(cluster:subscribe(mock_device_temp_humidity))
+    end
+  end
+
+  test.socket.matter:__expect_send({mock_device_humidity_battery.id, subscribe_request_humidity_battery})
+  test.socket.matter:__expect_send({mock_device_humidity_no_battery.id, subscribe_request_humidity_no_battery})
+  test.socket.matter:__expect_send({mock_device_temp_humidity.id, subscribe_request_temp_humidity})
+  test.mock_device.add_test_device(mock_device_humidity_battery)
+  test.mock_device.add_test_device(mock_device_humidity_no_battery)
+  test.mock_device.add_test_device(mock_device_temp_humidity)
```
greens, line 149: maybe stagger these adds and `expect_send`s?

drivers/SmartThings/matter-sensor/profiles/motion-illuminance-battery.yml
tpmanley: Is the plan to make copies of all the existing profiles that don't follow the naming scheme, and then add code to switch from the old profile to the new one, and finally we remove the old profiles from the driver once we know that no devices are still using them?

ctowns: This is what I was imagining we would do. At the moment, only two profiles don't fit the naming scheme: `matter-motion-battery-illuminance`, and `motion-battery` which has an incorrect `name:` field despite the filename being correct. The plan would be to make a duplicates of both and remove the old profiles once they are no longer used, just like you said.

drivers/SmartThings/matter-sensor/profiles/motion-battery.yml
cjswedes: I dont think this actually causes problems though. Those devices should still function.

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -23,6 +23,41 @@ local function device_init(driver, device)
   device:subscribe()
 end
 
+local function do_configure(driver, device)
+  print("do_configure")
```
cjswedes, line 27: nit:print statement

cjswedes (approved):

drivers/SmartThings/matter-sensor/src/init.lua
ctowns: Removed in 9473139b5f9e4b8c2b2466c509e3cd90ef9b251e

drivers/SmartThings/matter-sensor/src/test/test_matter_sensor_featuremap.lua
ctowns: Fixed by separating each device into its own `test_init` function

ctowns:
> Won't we need to re-subscribe after a profile change?
Yes, I added it in the latest

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -23,6 +23,44 @@ local function device_init(driver, device)
...
+  end
+
+  -- remove leading "-"
+  profile_name = string.sub(profile_name, 2)
+
+  device:try_update_metadata({profile = profile_name})
+end
+
+local function info_changed(driver, device, event, args)
+  device:subscribe()
```
greens, line 61: probably want to restrict this to the profile changing:

`if device.profile.id ~= args.old_st_store.profile.id`

ctowns: Added in the latest change

ctowns:
I have made sure we have battery and non-battery versions of every sensor combination that we currently support in profiles. There are also the `matter-motion-battery-illuminance` and `matter-motion-battery` profiles that I have created duplicates of because they do not follow the naming scheme of the other profiles. I think the plan would be to eventually just delete these profiles once we can confirm they are no longer used.

greens (approved):
duplicate profile check is going to be obnoxious for a while

PR #976, opened 2023-09-20
aqara-security-demo

tandres:
Change adds a demo for aqara security features.


lelandblue:
@tandres Can we close this?

PR #982, opened 2023-09-22
wallhero add model ACL-401S1I

thinkaName:
Manufacturers wallhero add new model ACL-401S1I zigbee-switch.
Please review it.


greens (approved):

greens:
if it's not a multi-switch then maybe it shouldn't have a "1" at the end of the join name

greens:
also the branch needs to be rebased

thinkaName:
> Is this a multi-switch? I'm guessing it is from the "1" at the end of the join name.
> 
> You might need to add the model info here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-switch/src/wallhero/init.lua#L23

This is not a multi-switch

greens (approved):

PR #984, opened 2023-09-22
edge driver support for aeotec tri sensor 8 added

ronie-z-wave:


ronie-z-wave:
Hi team,

unfortunately I don't understand why the **Run driver tests / run-driver-tests (pull_request)** and the **Luacheck / Test Results (pull_request)** fail. I can see that it has problems with **WakeUp:IntervalGet({})** but in my test/driver this Command is not implemented. Does it com from the parent driver and needs to be handled somehow?

PR #986, opened 2023-09-25
Matter Switch: fix endpoint to component mapping

ctowns:
The matter switch driver was initially written with the assumption that for multi-endpoint devices the endpoints would be sequentially numbered starting from 1. However, this is not a requirement in the matter spec, and devices like composed bridged devices often do not follow this assumption. To address this, endpoints are now queried on device add and then mapped to the corresponding component.


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -71,6 +57,24 @@ local function device_init(driver, device)
...
+    if current_component_number == 1 then
+      endpoint_map[ep] = "main"
+    else
+      endpoint_map[ep] = string.format("switch%d", current_component_number)
+    end
+    current_component_number = current_component_number + 1
+  end
+
+  device:set_field(ENDPOINT_TO_COMPONENT_MAP, endpoint_map, {persist = true})
+end
```
Con-Tejus, line 76: Is there any concern with race conditions on creating and populating the map within `device_added` compared to `do_configure`? Or is the general flow that `device_added` will be called before any other operation can take place for the driver?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -29,39 +29,25 @@ local COLOR_TEMPERATURE_KELVIN_MIN = 1
 local COLOR_TEMPERATURE_MIRED_MAX = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN
 local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX
 
+local ENDPOINT_TO_COMPONENT_MAP = "__endpoint_to_component_map"
```
tpmanley, line 32: Thoughts on checking for the existence of this field on each device on driver init and populating the map if it doesn't exist? The goal would be to fix devices that are already joined without requiring them to be removed and re-commissioned.

drivers/SmartThings/matter-switch/src/init.lua
hubTab: It looks like the lifecycle handlers do invoke the added first. @ctowns please correct if this is not correct.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -29,39 +29,25 @@ local COLOR_TEMPERATURE_KELVIN_MIN = 1
...
+    if component == component_name then return ep end
   end
-  return res
-end
-
-local function component_to_endpoint(device, component_id)
-  -- Assumes matter endpoint layout is sequentional starting at 1.
-  local ep_num = component_id:match("switch(%d)")
-  return ep_num and tonumber(ep_num) or find_default_endpoint(device, component_id)
 end
```
tpmanley, line 43: Do we need a default return value here?

drivers/SmartThings/matter-switch/src/init.lua
hubTab: That would be a great way of dealing with existing devices, and both, device_init and device_add could call into a commonly share function, if the map is null.

hubTab: Also, if this could be implemented, it would be helpful to log the creation of the map, especially for existing devices.

drivers/SmartThings/matter-switch/src/init.lua
ctowns: Good question, Tejus. Eugen is correct, device_add is queued and handled before the doConfigure event in this case.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -82,18 +86,14 @@ local function do_configure(driver, device)
...
-      device:try_update_metadata({profile = string.format("switch-%d", math.min(num_switch_eps, MAX_MULTI_SWITCH_EPS))})
-    end
-    if num_switch_eps > MAX_MULTI_SWITCH_EPS then
-      error(string.format(
-        "Matter multi switch device will not function. Profile doesn't exist with %d components",
-        num_switch_eps
-      ))
-    end
+  if num_switch_eps > 1 then
+    device:try_update_metadata({profile = string.format("switch-%d", math.min(num_switch_eps, MAX_MULTI_SWITCH_EPS))})
```
hubTab, line 90: As it was used as such in the past,  it could have been useful to define `switch-%d` as a constant, to be shared across all drivers, to ensure the format would be consistent across.

drivers/SmartThings/matter-switch/src/test/test_multi_switch.lua
```diff
@@ -152,4 +211,26 @@ test.register_message_test(
...
+        mock_3switch_non_sequential.id,
+        { capability = "switch", component = "switch3", command = "on", args = { } }
+      }
+    },
+    {
+      channel = "matter",
+      direction = "send",
+      message = {
+        mock_3switch_non_sequential.id,
+        clusters.OnOff.server.commands.On(mock_3switch_non_sequential, 15) -- switch 3 is on endpoint 15
```
hubTab, line 230: This is correct, but it got a little thinking to realize that we sequentially increase the switch numbers, internally, as opposed to using the endpoint_id number.

drivers/SmartThings/matter-switch/src/init.lua
ctowns: I think this is a great idea, Tom! Although on a side note, I am wondering if there is any precedent in the past for reconfiguring already joined devices? Were we just supposed to remove and rejoin the devices? @greens  do you know if there is any precedent?

cjswedes: I know we have done fixes in device init for devices that aren't functioning properly due to bugs in the way we add them. The risk of the change is heightened though, because the change then impacts all matter devices using the switch driver rather than just newly joined devices.

drivers/SmartThings/matter-switch/src/test/test_multi_switch.lua
ctowns: In our profiles, the switch components are numbered sequentially (except the first switch component is called "main", otherwise it continues on as "switch2", "switch3", etc.). 

The purpose of these changes is that the endpoint number does not need to match the component number. It can be a little confusing, but the idea is that when mapping a component to an endpoint, the component will map to the next lowest endpoint that hasn't been mapped yet. So for example, in a case where we have endpoints 10, 12, and 15, the mapping would be:
"main" -> 10
"switch2" -> 12
"switch3" -> 15

I will add some additional documentation to make this clear. Thanks for pointing it out!

hubTab: Thank you @ctowns !

drivers/SmartThings/matter-switch/src/init.lua
cjswedes: I think so. I think this will break existing devices that dont have an`ENDPOINT_TO_COMPONENT_MAP` setup on the device object. Previously it would have returned the lowest non-0 endpoint, but now it will return nil

drivers/SmartThings/matter-switch/src/init.lua
cjswedes: In an attempt to reduce complexity of dealing with devices in the field that have a ep to comp map and ones that dont, we should probably do reprocessing on init of the ep/comp map. It will help us ensure correctness with unit tests.

ctowns: Moved component mapping to `init` and removed `added` in a9e0c3ac0e37798c0fd3f4927ce6b4ad80d88fc0

drivers/SmartThings/matter-switch/src/init.lua
ctowns: Added a default return of the lowest non-0 endpoint in a9e0c3ac0e37798c0fd3f4927ce6b4ad80d88fc0

drivers/SmartThings/matter-switch/src/test/test_multi_switch.lua
ctowns: Added a comment in the `create_endpoint_to_component_map()` function to help explain the reasoning behind the component numbering in a9e0c3ac0e37798c0fd3f4927ce6b4ad80d88fc0

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -49,23 +51,47 @@ local function find_default_endpoint(device, component)
   return res
 end
 
-local function component_to_endpoint(device, component_id)
-  -- Assumes matter endpoint layout is sequentional starting at 1.
-  local ep_num = component_id:match("switch(%d)")
-  return ep_num and tonumber(ep_num) or find_default_endpoint(device, component_id)
+local function create_endpoint_to_component_map(device)
```
tpmanley, line 54: I think we may want to also do the profile switching behavior in this function rather than in `do_configure` so that the endpoint map and the profile are in sync. Otherwise we could be fixing the endpoint map of an existing device but the profile would be wrong because the device was already configured.

I also think we probably want to do this stuff only once and it should be skipped if the `ENDPOINT_TO_COMPONENT_MAP` already exists.

greens:
@Con-Tejus the CI generates a channel invite automatically for all PRs: https://bestow-regional.api.smartthings.com/invite/oDM8dkQVkYML

drivers/SmartThings/matter-switch/src/init.lua
greens: the matter device's default behavior is to return the value of `find_default_endpoint` if the `component_to_endpoint` function doesn't return anything, so it's not technically necessary

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -82,18 +108,14 @@ local function do_configure(driver, device)
...
-        num_switch_eps
-      ))
-    end
+  if num_switch_eps > 1 then
+    device:try_update_metadata({profile = string.format("switch-%d", math.min(num_switch_eps, MAX_MULTI_SWITCH_EPS))})
+  end
+  if num_switch_eps > MAX_MULTI_SWITCH_EPS then
+    error(string.format(
+      "Matter multi switch device will have limited function. Profile doesn't exist with %d components",
+      num_switch_eps
```
greens, line 117: is it worth setting it to the highest available value?

ctowns: If a device has more than the max number of supported switches, it will switch to the highest available profile on line 112 above. This would mean that only those endpoints that could be mapped would actually be functional, and the excess endpoints would be mapped to the "main" component

drivers/SmartThings/matter-switch/src/init.lua
ctowns: > the matter device's default behavior is to return the value of find_default_endpoint if the component_to_endpoint function doesn't return anything, so it's not technically necessary

I think it only defaults to `find_default_endpoint` if the `component_to_endpoint` function itself is nil (i.e. has not been set by the driver), not if the `component_to_endpoint` function returns nil

drivers/SmartThings/matter-switch/src/init.lua
ctowns: In d70a3099c1abcd571de9b96b22ad399797ffd37a I moved both the component-to-endpoint mapping and the profile switching into one `initialize_switch` function that will be called on `device_init` if there is no `ENDPOINT_TO_COMPONENT_MAP` on the device

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -49,23 +51,65 @@ local function find_default_endpoint(device, component)
...
-  local switch_comp = string.format("switch%d", ep)
-  if device.profile.components[switch_comp] ~= nil then
-    return switch_comp
-  else
-    return "main"
+  local map = device:get_field(ENDPOINT_TO_COMPONENT_MAP) or {}
+  if map[ep] and device.profile.components[map[ep]] then
+    return map[ep]
   end
+  return "main"
```
greens, line 104: I think what we probably want to do here is, if we have _any_ endpoint to component mapping done at all, to ignore messages from a non-mapped endpoint, rather than emit them from main

I'm thinking of a situation where you have more switches than we have profiles for. Anything from endpoint > X will change the state of "main" which is also a _real_ component that we have mapped to.

drivers/SmartThings/matter-switch/src/test/test_multi_switch.lua
```diff
@@ -205,9 +199,35 @@ test.register_message_test(
...
       }
     }
-  }
+  },
+  { test_init = test_init_mock_3switch }
 )
 
+-- The custom "test_init" function also checks that the appropriate profile is switched on init
+test.register_message_test(
+  "On command to component switch should send the appropriate commands",
```
cjswedes, line 208: nit: name of test is same as previous test

cjswedes (approved):

drivers/SmartThings/matter-switch/src/init.lua
ctowns: The `emit_event_for_endpoint` function expects the `endpoint_to_component` function to return something, or else we would run into an issue in `emit_component_event` when we try to use the nil value for the component. So, I think there needs to be some kind of default value, even if it is not ideal.

> I'm thinking of a situation where you have more switches than we have profiles for. Anything from endpoint > X will change the state of "main" which is also a real component that we have mapped to.

I think this is a valid problem that would be caused by this. The solution would be to add another profile to account for the number of switches. Ideally we'd be able to quickly see the error in the logs since it is logged if a device has more than our supported number of switches

hubTab (approved):

drivers/SmartThings/matter-switch/src/init.lua
greens: if you feel like digging into it, what I did with buttons was use parent-child in the case that I didn't have a profile configured for that number of components

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -49,23 +51,65 @@ local function find_default_endpoint(device, component)
...
+  -- For switch devices, the profile components follow the naming convention "switch%d",
+  -- with the exception of "main" being the first component. Each component will then map
+  -- to the next lowest endpoint that hasn't been mapped yet.
+  for _, ep in ipairs(switch_eps) do
+    if current_component_number == 1 then
+      endpoint_map[ep] = "main"
+    else
+      endpoint_map[ep] = string.format("switch%d", current_component_number)
+    end
+    current_component_number = current_component_number + 1
```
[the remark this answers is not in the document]
ctowns: Simplified loop in 2ec17138768b2e55bf7fa10f7979eae59ed034db

greens (approved):

PR #990, opened 2023-09-26
Matter Thermostat: fix default endpoint mapping

ctowns:
Our "default" endpoint logic current returns the lowest endpoint. For matter composed bridged devices, the current "default" endpoint would be the BRIDGED_NODE_DEVICE_TYPE endpoint because it is the lowest endpoint for the bridged device. However, that endpoint does not support the desired clusters, so we want to find the first endpoint that supports our desired cluster and map that to the corresponding component.

This should not affect existing behavior because for single endpoint non-bridged devices, the default endpoint would also support the given cluster.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -73,8 +73,27 @@ local subscribed_attributes = {
   }
 }
 
+local function find_default_endpoint(device, cluster)
+  local res = device.MATTER_DEFAULT_ENDPOINT
+  local eps = device:get_endpoints(cluster)
+  table.sort(eps)
+  for _, v in ipairs(eps) do
+    if v ~= 0 then --0 is the matter RootNode endpoint
+      return v
+    end
+  end
+  return res
+end
+
+local function component_to_endpoint(device, component_name)
+  -- Use the find_default_endpoint function to return the first endpoint that
+  -- supports a given cluster.
+  return find_default_endpoint(device, clusters.Thermostat.ID)
+end
+
 local function device_init(driver, device)
   device:subscribe()
+  device:set_component_to_endpoint_fn(component_to_endpoint)
 end
 
 local function info_changed(driver, device, event, args)
--- a/drivers/SmartThings/matter-thermostat/src/test/test_matter_thermostat_composed_bridged.lua
+++ b/drivers/SmartThings/matter-thermostat/src/test/test_matter_thermostat_composed_bridged.lua
@@ -1,0 +1,673 @@
+-- Copyright 2023 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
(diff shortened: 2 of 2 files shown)


cjswedes (approved):
It'd be good to add a unit test for the case where the first non-0 endpoint is skipped over because it doesn't support the Thermostat cluster

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -73,8 +73,28 @@ local subscribed_attributes = {
   }
 }
 
+local function find_default_endpoint(device, component, cluster)
```
greens, line 76: `component` is unused

ctowns: removed

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -73,8 +73,28 @@ local subscribed_attributes = {
...
 }
 
+local function find_default_endpoint(device, cluster)
+  local res = device.MATTER_DEFAULT_ENDPOINT
+  local eps = device:get_endpoints(cluster)
+  table.sort(eps)
+  for _, v in ipairs(eps) do
+    if v ~= 0 then --0 is the matter RootNode endpoint
+      res = v
+      break
```
hubTab, line 83: Nit: this could probably return, instead of breaking and returning.

ctowns:
Added a unit test for a mock "composed bridge thermostat" device that shows that the `find_default_endpoint()` will skip non-zero endpoints and return the first endpoint that supports the given cluster.

drivers/SmartThings/matter-thermostat/src/test/test_matter_thermostat_composed_bridged.lua
```diff
@@ -0,0 +1,654 @@
+-- Copyright 2022 SmartThings
```
greens, line 1: date

drivers/SmartThings/matter-thermostat/src/test/test_matter_thermostat_composed_bridged.lua
```diff
@@ -0,0 +1,654 @@
...
+		},
+		{
+			channel = "matter",
+			direction = "send",
+			message = {
+				mock_device.id,
+				refresh_request
+			}
+		}
+	}
```
greens, line 651: whitespace

drivers/SmartThings/matter-thermostat/src/test/test_matter_thermostat_composed_bridged.lua
```diff
@@ -0,0 +1,673 @@
...
+        capabilities.thermostatMode.supportedThermostatModes({ "off", "cool" }))
+    },
+  }
+)
+
+test.register_message_test(
+  "Additional mode reports should extend the supported modes",
+  {
+    {
+      channel = "matter",
```
hubTab, line 343: Is the `ControlSequenceOfOperation` channel required by this test?

hubTab (approved):

samadDotDev (approved):

PR #994, opened 2023-09-26
matter-button: Fix multi-button devices that don't start at endpoint 1

tpmanley:
This fixes two problems with the endpoint/component mapping:
- The Lua table serializer cannot handle numerical keys that don't start at 1 (aka an array) because it serializing to JSON and JSON objects require keys to be strings.
- Using `#` to check the size of a Lua table only works if the keys start at 1

Due to the second issue, the driver wasn't even trying to write the map to persistance, but even if that was fixed the first issue caused the driver to crash.

To fix the first issue we switch to using the components as the key which are strings so there is no issue converting to/from JSON.

To fix the second issue we use another variable to determine if we should write the map to persistance.

With these changes and the 0.50 hub firmware the Hue multi-button devices are now working.

https://smartthings.atlassian.net/browse/MTR-453

### DIFF
--- a/drivers/SmartThings/matter-button/src/test/test_matter_multi_button.lua
+++ b/drivers/SmartThings/matter-button/src/test/test_matter_multi_button.lua
@@ -7,14 +7,18 @@ local dkjson = require "dkjson"
 local clusters = require "st.matter.generated.zap_clusters"
 local button_attr = capabilities.button.button
 
---mock the actual device
+-- Mock a 4-button device using endpoints non-consecutive endpoints
 local mock_device = test.mock_device.build_test_matter_device(
   {
     profile = t_utils.get_profile_definition("4-button-battery.yml"), -- on a real device we would switch to this, rather than fingerprint to it
     manufacturer_info = {vendor_id = 0x0000, product_id = 0x0000},
     endpoints = {
     {
-      endpoint_id = 1,
+      endpoint_id = 0,
+      clusters = {},
+    },
+    {
+      endpoint_id = 10,
       clusters = {
         {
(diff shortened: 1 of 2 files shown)


greens (commented):
@ctowns did you repeat my mistake in your drivers?

ctowns:
Yes that's my bad :( but I am going to get up a fix shortly

drivers/SmartThings/matter-button/src/test/test_matter_multi_button.lua
```diff
@@ -14,7 +14,7 @@ local mock_device = test.mock_device.build_test_matter_device(
     manufacturer_info = {vendor_id = 0x0000, product_id = 0x0000},
     endpoints = {
     {
-      endpoint_id = 1,
+      endpoint_id = 10,
```
cjswedes, line 17: Can you add an endpoint id of 0 to this test device as well. Just want to make sure we get coverage of the common case where there is a root node endpoint without the switch cluster.

cjswedes (approved):

drivers/SmartThings/matter-button/src/test/test_matter_multi_button.lua
tpmanley: Added in [a263d34](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/994/commits/a263d34d1dece766e7b46330ca1ab1cc939438ef)

ctowns (approved):

tpmanley:
The fix in this PR applies to newly joined devices. I'll do a follow up PR to address already joined devices.

PR #995, opened 2023-09-27
Matter Switch: use component as key for endpoint map

ctowns:
This changes the endpoint-component mapping to use the component as the key for the map. This is the same as was done for https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/994

This change is required because our lua libs cannot handle tables with numerical keys that don't start at 1. See here for more detail: https://smartthings.atlassian.net/browse/MTR-453

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -29,7 +29,7 @@ local COLOR_TEMPERATURE_KELVIN_MIN = 1
 local COLOR_TEMPERATURE_MIRED_MAX = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN
 local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX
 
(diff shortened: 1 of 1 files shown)


ctowns:
I have added a unit test that isn't passing, and I believe it is because `device.profile.components` only seems to be returning the `main` component. Therefore, when we check if the returned component is actually a component in the profile, any component other than `main` will be false. I will investigate further tomorrow.

greens (approved):

tpmanley (approved):

ctowns:
Tested with matter device test app to confirm functionality for: 

1. a standard non-bridged non-composed light bulb
2. multiple bridged color bulbs
3. multiple bridged composed matter devices

NOTE: matter test app currently supports only a composed device with two bulbs, but we do not have multi-light device support in our driver/profiles. However, in this case the composed multi-light device changes to a switch profile (since it has multiple endpoints with OnOff support), so I was able to confirm that the OnOff commands are handled correctly by the composed device as if it was a composed switch device.

If we want add support for multi-light composed devices I can do that in a separate PR, but I not sure we would expect to see a lot of devices like that right now.

PR #996, opened 2023-09-28
matter-switch: Fix devices (ex: Hue White) that did not have levelControl

tpmanley:
In Matter it is valid for an endpoint that supports a specific device type to also support additional device types that are subsets of that device type. The Hue White is an example of a device that does this. Since we didn't have a fingerprint for an endpoint with both OnOff Light and Dimmable Light device types the fingerprinting logic happened to be picking the fingerprint with just the OnOff device type which uses a profile without levelControl.

The fix is to add additional fingerprints for each of the device types that are a superset of other device type(s).

**Note: This requires devices to be removed and re-added since it only changes the fingerprinting and does not change the profile of already joined devices.**

https://smartthings.atlassian.net/browse/MTR-473

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -121,21 +121,42 @@ matterGeneric:
     deviceTypes:
       - id: 0x0101 # Dimmable Light
     deviceProfileName: light-level
+  - id: "matter/dimmable/light/2"
+    deviceLabel: Matter Dimmable Light
+    deviceTypes:
+      - id: 0x0100 # OnOff Light
+      - id: 0x0101 # Dimmable Light
+    deviceProfileName: light-level
   - id: "matter/colorTemperature/light"
     deviceLabel: Matter Color Temperature Light
     deviceTypes:
       - id: 0x010C # Color Temperature Light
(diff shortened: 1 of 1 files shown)


ctowns (approved):

kianooshkarami (approved):

drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -134,8 +147,16 @@ matterGeneric:
   - id: "matter/color/temp/light"
     deviceLabel: Matter Color Temperature Light
     deviceTypes:
+      - id: 0x010C # Color Temperature Light
       - id: 0x010D # Extended Color Light
+    deviceProfileName: light-color-level
+  - id: "matter/color/temp/light/2"
+    deviceLabel: Matter Color Temperature Light
```
ctowns, line 154: nit: should this be `Matter Extended Color Light`? Same for the fingerprint above

samadDotDev (approved):

hubTab (approved):

drivers/SmartThings/matter-switch/fingerprints.yml
tpmanley: I went with "Matter Color Light" which seems more user friendly to me.

PR #1000, opened 2023-09-29
matter-switch: Add firmwareUpdate capability to the profile used by bridges

tpmanley:
Matter bridges do expose their firmware version but it wasn't visible to users in the SmartThings app because the profile did not include the necessary firmwareUpdate capability. This just adds that capability to the profile.

https://smartthings.atlassian.net/browse/MTR-475


ctowns (approved):
WOOO PR 1000! 🎉  💯0

Con-Tejus (approved):
1000! 🎉

PR #1001, opened 2023-10-02
motion-sensor: (Hue) Expose temperature and remove non-functional battery

tpmanley:
The Hue motion sensor supports motion, temperature and illuminance sensors so added a fingerprint for those device types. Like the other battery powered Hue accessories, the motion does not support reporting battery percentage remaining so the battery capability is not used for Hue sensors.

**Note**: This fix only applies to newly joined devices so existing devices will need to be removed and re-added.

https://smartthings.atlassian.net/browse/MTR-476

### DIFF
--- a/drivers/SmartThings/matter-sensor/src/init.lua
+++ b/drivers/SmartThings/matter-sensor/src/init.lua
@@ -20,8 +20,19 @@ local utils = require "st.utils"
 
 local BATTERY_CHECKED = "__battery_checked"
 
+local HUE_MANUFACTURER_ID = 0x100B
+
+local function supports_battery_percentage_remaining(device)
+  local battery_eps = device:get_endpoints(clusters.PowerSource.ID,
+          {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
+  -- Hue devices support the PowerSource cluster but don't support reporting battery percentage remaining
+  if #battery_eps > 0 and device.manufacturer_info.vendor_id ~= HUE_MANUFACTURER_ID then
+    return true
+  end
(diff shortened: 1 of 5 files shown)


ctowns (approved):

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -20,8 +20,19 @@ local utils = require "st.utils"
 
 local BATTERY_CHECKED = "__battery_checked"
 
+local HUE_MANUFACTURER_ID = 0x100B
+
+local function supports_battery_percentage_remaining(device)
+  local battery_eps = device:get_endpoints(clusters.PowerSource.ID,
+          {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
+  -- Hue devices support the PowerSource cluster but don't support reporting battery percentage remaining
```
kianooshkarami, line 28: Do you have any thoughts about checking this through checking the feature map or supported attributes, if possible, as opposed to checks against manufacturers?

tpmanley: Thanks for bringing this up. We should do exactly as you suggested and then we can remove the manufacturer specific test. This was a short term solution to get things fixed for Hue since we saw the issues there and will follow up with the better long term solution.

greens (approved):

drivers/SmartThings/matter-sensor/src/init.lua
z-michel: Created https://smartthings.atlassian.net/browse/CHAD-11842

PR #1002, opened 2023-10-03
Find default endpoint by cluster for remaining matter drivers

ctowns:
The default endpoint used in matter drivers is currently the first non-zero endpoint. However, this can cause problems in cases like composed matter devices or aggregated bridged devices where the first non-zero endpoint is not where we would expect to find the relevant clusters for each driver. These changes return the first endpoint that supports a relevant cluster for each driver in order to prevent this issue.

These changes were already ported to matter-switch, matter-button, and matter-thermostat.

[CHAD-11752](https://smartthings.atlassian.net/browse/CHAD-11752)

### DIFF
--- a/drivers/SmartThings/matter-button/src/init.lua
+++ b/drivers/SmartThings/matter-button/src/init.lua
@@ -79,6 +79,7 @@ local function find_default_endpoint(device, component)
       break
     end
   end
+  device.log.warn(string.format("Did not find default endpoint, will use endpoint %d instead", device.MATTER_DEFAULT_ENDPOINT))
   return res
 end
 
--- a/drivers/SmartThings/matter-lock/src/init.lua
+++ b/drivers/SmartThings/matter-lock/src/init.lua
@@ -364,30 +364,33 @@ local function lock_user_change_event_handler(driver, device, ib, response)
 end
 
 local function handle_refresh(driver, device, command)
-  local req = DoorLock.attributes.LockState:read(device, device.MATTER_DEFAULT_ENDPOINT)
-  req:merge(PowerSource.attributes.BatPercentRemaining:read(device, device.MATTER_DEFAULT_ENDPOINT))
+  -- Note: no endpoint specified indicates a wildcard endpoint
+  local req = DoorLock.attributes.LockState:read(device)
+  req:merge(PowerSource.attributes.BatPercentRemaining:read(device))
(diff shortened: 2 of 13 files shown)


drivers/SmartThings/matter-button/src/init.lua
```diff
@@ -69,6 +69,7 @@ end
 
 --end of helper functions
 --------------------------------------------------------------------------
+-- TODO: remove this and use device:find_default_endpoint once those lua libs changes are widely available
```
cjswedes, line 72: nit: I think you should remove this comment, because Im not sure this function will ever be in the lua libs.

cjswedes (approved):

hubTab (approved):

drivers/SmartThings/matter-button/src/init.lua
ctowns: I agree with your feedback and have closed the associated lua libs PR and removed the comments here and in the other drivers 👍

greens (approved):

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -476,7 +479,26 @@ local function handle_name_slot(driver, device, command)
...
+local function find_default_endpoint(device, cluster)
+  local res = device.MATTER_DEFAULT_ENDPOINT
+  local eps = device:get_endpoints(cluster)
+  table.sort(eps)
+  for _, v in ipairs(eps) do
+    if v ~= 0 then --0 is the matter RootNode endpoint
+      return v
+    end
+  end
+  return res
```
tpmanley, line 491: Thoughts on adding a warning message here? I think we don't expect to ever hit this so it'd be helpful to log if we are

ctowns: Added log message in this function for each driver 👍

tpmanley: Looks like this one might have gotten missed

tpmanley (approved):
Thanks @ctowns !

drivers/SmartThings/matter-lock/src/init.lua
ctowns: welp, forgot to save the file 🤦, should be good now!

PR #1011, opened 2023-10-10
Matter Switch: use parent-child for multi-switch devices

ctowns:
https://smartthings.atlassian.net/browse/CHAD-11858

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -17,6 +17,7 @@ local log = require "log"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
 local utils = require "st.utils"
+local device_lib = require "st.device"
 
 local MOST_RECENT_TEMP = "mostRecentTemp"
 local RECEIVED_X = "receivedX"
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -48,74 +49,56 @@ local function find_default_endpoint(device, component)
...
-    else
-      component_map[string.format("switch%d", i)] = ep
+    if ep ~= main_endpoint then -- don't create a child device that maps to the main endpoint
+      local name = string.format("%s %d", device.label, current_component_number)
+      device.log.info(string.format("creating child device %s", name))
+      driver:try_create_device(
+        {
+          type = "EDGE_CHILD",
+          label = name,
+          profile = "switch-binary",
```
hubTab, line 73: Is it possible to refer to this string by leveraging the fingerprint::deviceProfileName?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -48,74 +49,56 @@ local function find_default_endpoint(device, component)
...
+          parent_device_id = device.id,
+          parent_assigned_child_key = string.format("%02X", ep),
```
hubTab, line 75: Is there a utility function available that, provided an ep, returns its formatted value, as It is being used throughout?

ctowns:
Running into some issues with testing and need to do some more clean up, but the plan is now to not get this in for the beta deploy today and instead target end of the week.

drivers/SmartThings/matter-switch/src/init.lua
greens: not available in the driver afaik

drivers/SmartThings/matter-switch/src/test/test_matter_switch_parent_child.lua
```diff
@@ -0,0 +1,161 @@
...
+  {
+    {
+      channel = "matter",
+      direction = "receive",
+      message = {
+        mock_device.id,
+        clusters.OnOff.attributes.OnOff:build_test_report_data(mock_device, 3, true)
+      }
+    },
+		{
```
greens, line 152: whitespace

lelandblue:
What is the status of this PR is it still needed?

ctowns:
This was merged here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1094

PR #1015, opened 2023-10-11
DRAFT: Driver tests for lazy subdriver loading

ctowns:
This utilizes a new method of loading subdrivers that allows the sub drivers to be dynamically loaded when needed rather than having every subdriver loaded at initialization.


drivers/SmartThings/matter-switch/src/eve-energy/init.lua
```diff
@@ -217,13 +217,15 @@ end
 -------------------------------------------------------------------------------------
 
 local function watt_attr_handler(driver, device, ib, zb_rx)
+  print("watt_attr_handler")
```
tpmanley, line 220: Did you mean to include this? If so would be better to switch to using logging

ctowns: Did not mean to include this! Still cleaning up this PR for testing purposes.

PR #1023, opened 2023-10-19
fix: Only allow a single discovery handler

dljsjr:
There are some situations where many discovery handlers can be started
at the same time, and in some cases this can lead to issues like duped
devices being created because the event create requests happen close
enough together that they skirt through checks for things like DNI
duplication.

The proper fix for this is to disallow this in firmware, but this is
impacting users currently so a stop-gap fix in the LAN Edge Drivers
at a minimum -- where this is more likely to happen -- may not be a bad
idea.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/disco.lua
+++ b/drivers/SmartThings/philips-hue/src/disco.lua
@@ -18,6 +18,7 @@ local HueDiscovery = {
   light_state_disco_cache = {},
   ServiceType = SERVICE_TYPE,
   Domain = DOMAIN,
+  discovery_active = false
 }
 
 local supported_resource_types = {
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/philips-hue/src/disco.lua
```diff
@@ -33,7 +34,12 @@ process_discovered_light
 ---@param _ table
 ---@param should_continue function
 function HueDiscovery.discover(driver, _, should_continue)
+  if HueDiscovery.discovery_active then
+    log.info("Hue discovery already in progress, ignoring new discovery request")
```
NoahCornell, line 38: should there be an early return here?

dljsjr: Oof yes. I had one, then I reset the entire branch cuz I accidentally based it off #931, and it didn't get brought over.

NoahCornell (approved):

tandres (approved):

cjswedes (approved):

PR #1024, opened 2023-10-20
bugfix: Matter Switch Device Type (0x010C) profile

kianooshkarami:
Per Matter Specification, 0x010C has color control as well as temperate, switching the profile to "light-color-level" to add the Color control support.

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -138,7 +138,7 @@ matterGeneric:
     deviceLabel: Matter Color Temperature Light
     deviceTypes:
       - id: 0x010C # Color Temperature Light
-    deviceProfileName: light-level-colorTemperature
+    deviceProfileName: light-color-level
   - id: "matter/colorTemperature/light/2"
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -138,7 +138,7 @@ matterGeneric:
     deviceLabel: Matter Color Temperature Light
     deviceTypes:
       - id: 0x010C # Color Temperature Light
-    deviceProfileName: light-level-colorTemperature
+    deviceProfileName: light-color-level
```
Con-Tejus, line 141: We would probably want to do this for all fingerprints involving device type `0x010C`

tpmanley: I'm not sure if this change should be needed. The Color Control cluster is used for both setting the color and the color temperature. 0x010C is the Color Temperature Light device type so I would expect it to only use the Color Control cluster for setting the color temperature. A device that supports setting the color and color temperature would likely use device type 0x010D which is the Extended Color Light.

kianooshkarami:
Closing due to misreading of the device type

PR #1027, opened 2023-10-20
MTR-497 don't switch profiles if the lock doesn't support tamper

greens:


drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -506,7 +506,7 @@ local function device_added(driver, device)
   --Note: May want to write RequirePINForRemoteOperation, to avoid cota cases if possible.
   device:emit_event(capabilities.tamperAlert.tamper.clear())
   local eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
-  if #eps == 0 then
+  if #eps == 0 and device:supports_capability_by_id(capabilities.tamperAlert.ID) then
```
tpmanley, line 509: This would hit the else case if the device doesn't support PIN creds or tamper

cjswedes (approved):

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -507,8 +507,10 @@ local function device_added(driver, device)
   device:emit_event(capabilities.tamperAlert.tamper.clear())
   local eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
   if #eps == 0 then
-    device.log.debug("Device does not support lockCodes")
-    device:try_update_metadata({profile = "lock-without-codes"})
+    if device:supports_capability_by_id(capabilities.tamperAlert.ID) then
+      device.log.debug("Device does not support lockCodes")
+      device:try_update_metadata({profile = "lock-without-codes"})
+    end
```
tpmanley, line 513: What do you think about adding an else with a warning that we don't have a profile for a device that doesn't support codes or tamper?

tpmanley (approved):

drivers/SmartThings/matter-lock/src/init.lua
greens: I think this is only temporary until the incoming profile gets merged. Then we can actually do switching in that `else`

greens: updated anyway, though

PR #1029, opened 2023-10-24
Matter Thermostat: Fix thermostatMode capability

yhoyoon:
The matter-thermostat driver is using the "fanOnly" value of the thermostatMode capability. But since this capability requires using the "fanonly" value, events using the "fanOnly" value are ignored. This change renames "fanOnly" to "fanonly".

https://developer.smartthings.com/docs/devices/capabilities/capabilities-reference/#thermostatMode


chulspro (approved):
Looks good to me.

greens (approved):
based on https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/main/datatypes/ThermostatMode.yml

this capitalization is correct

PR #1026, opened 2023-10-20
add Heiman matter Motion and door sensor

greens:

PR #1032, opened 2023-10-24
WWSTCERT-987 Fingerprint Add

lelandblue:
Fingerprint add for the Door Sensor from Heiman through the Edge Builder Tool.


hubTab (approved):

z-michel:
Seems like this device already covered in: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1026?

Leo2442926161:
> Seems like this device already covered in: #1026?

Yes, we have two sensors are covered in that PR.

lelandblue:
I'd recommend that we edit that PR to remove the duplicate device and have one device per PR. Each device will be tested individually and could be published on different schedules.

Leo2442926161:
> I'd recommend that we edit that PR to remove the duplicate device and have one device per PR. Each device will be tested individually and could be published on different schedules.

Thanks for your quick response, we actually submitted two matter sensors for WWST, would it make sense if the PR contains two sensors?

Leo2442926161:
Hi @lelandblue  Thank you very much for your clarifying.
Yes, that is correct.

Q:Remove the Heiman Door and Window D1-M from PR 1026, and it will remain as part of this PR 1032. Each pull request should contain one device per pull request.
A: Do I need to remove the Door sensor and resubmit it in PR1026?

Question for the D1-M Device in this PR, Do you want the device label to be "Heiman Door and Window D1-M" OR "Door sensor". The device label is the default name given to a device when it joins our platform. Users will see the device in their home named what is defined in the deviceLabel:. There were differences in how this was submitted so I just want to confirm with you what your preference is. Users can change this after the device joins the platform.
A: we would like the deviceLabel to be "Heiman Door and Window D1-M", how to change it in this PR?

Thanks again for your help.

lelandblue:
Hello @Leo2442926161 

I want to help make this easy for you. I have edited this PR, to fix the join name to "Heiman Door and Window D1-M". If you are ok with how this PR is and do not need any more changes, please let me know. Then we will proceed with this PR and there will be no need for you to open another.

Leo2442926161:
> Hello @Leo2442926161
> 
> I want to help make this easy for you. I have edited this PR, to fix the join name to "Heiman Door and Window D1-M". If you are ok with how this PR is and do not need any more changes, please let me know.
> 
> I think we can proceed with this PR, and there is no further action needed from you regarding Pull Requests.
> 
> Please confirm.

Hi @lelandblue,
That would be great, we confirm that join name is okay for us, thank you very much.

drivers/SmartThings/matter-sensor/fingerprints.yml
```diff
@@ -33,6 +33,12 @@ matterManufacturer:
     vendorId: 0x1021
     productId: 0x0001
     deviceProfileName: motion-contact-battery
+# HEIMAN
```
ctowns, line 36: The `deviceLabel` and `deviceProfileName` should not be in quotes based on the precedent set by the other fingerprints

ctowns (approved):

PR #1040, opened 2023-10-30
Samsung audio fallback to http when https fails

cjswedes:
Sometimes old devices are failing to successfully use https to download samsung hosted audio notification files, in that case we should re-attempt with http before completely failing.

### DIFF
--- a/drivers/SmartThings/samsung-audio/src/command.lua
+++ b/drivers/SmartThings/samsung-audio/src/command.lua
@@ -348,14 +348,40 @@ function Command.getPlayStatus(ip)
   return response_map
 end
 
+--- Return err code from xml handler or nil if it doesn't exist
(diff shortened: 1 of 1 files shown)


cjswedes:
@mikesip or @drewpeterson I think one of you has the device. Could you check that this is working as expected? Repro steps:
1. Install the driver via the install link
2. Join the device
3. Setup a routine/scene to use the audionotification
4. Start capturing the samsung audio driver log using the cli `smartthings edge:drivers:logcat`
5. Ensure the audionotification works and validate with logs whether the http fallback was used or not (if it wasn't used, you may need the audio notification to be a text to speech audionotification)

cjswedes:
This has been tested by @mikesip and it seems to work.

mikesip (approved):

PR #1041, opened 2023-10-31
CHAD-11987: Fix potential infinite loops in lunchbox

NoahCornell:
Previously, it was possible to get stuck in an infinite loop while parsing a chunked response if the socket was closed and didn't return any partial data. Now, we will return in that case and propagate the "closed" error so that try to reconnect after the error.

I have tested this with hue and can with sonos but I don't have a JBL to test with.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/lunchbox/rest.lua
+++ b/drivers/SmartThings/philips-hue/src/lunchbox/rest.lua
@@ -92,6 +92,8 @@ local function parse_chunked_response(original_response, sock)
         if partial ~= nil and #partial >= 1 then
           full_response:append_body(partial)
           next_chunk_bytes = 0
+        else
+          return nil, next_err
         end
       else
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/philips-hue/src/lunchbox/rest.lua
```diff
@@ -92,6 +92,8 @@ local function parse_chunked_response(original_response, sock)
         if partial ~= nil and #partial >= 1 then
           full_response:append_body(partial)
```
NoahCornell, line 93: My first thoughts with this were that this was potentially incorrect and that if the socket was closed, then we would always want to return that as an error but that caused some issues with commands while testing on hub.

dljsjr (approved):
Tested on my hub and working well.

The luacheck and unit test failures are related to #1043. They should go away if you rebase.

PR #1042, opened 2023-10-31
fix: Remove some lookups from persistent store

dljsjr:
Subtle edge cases with the persistent store's eventual consistency mean that it's possible for these stores to get in to a bad state, and some of the APIs on the hub will hard crash when given stale data on older hub firmware versions that are still actively supported. Moving these stores to be ephemeral could slow down detection of IP address changes but shouldn't hamper functionality.


tandres (approved):

dljsjr:
Blocked on #1043

PR #1043, opened 2023-10-31
 fix: finish restoring popp_danfoss -> danfoss

dljsjr:
Some changes in another PR modified an existing profile in an breaking
way, and we asked for those changes to be reverted in the PR. Some stuff
got missed, though. This addresses the stragglers.


dljsjr:
Failing tests are repro'd locally as well.

greens (approved):

PR #1046, opened 2023-11-02
Eve Energy: Improve the performance by removing unnecessary read

Timac:
Remove an unnecessary switch state read. The switch state should be already reported by the device itself so this frequent read is not necessary here.


ctowns (changes requested):
Just remove the now unused "clusters" from the top of the file, then this looks good to me!

tpmanley (approved):

ctowns (approved):

Issue #911, opened 2023-08-15
BUG FIX: Bose Speaker Audio Notification - set Default 30 ONLY if error getting current volume of device


ketchumj1:
Right now, default is always set to 30.   Having a new feature to automation is nice, but why not just keep the device at same volume it was instead of change to 30?   the handler for volume up shows can get current volume, as seen below

  local ip = device:get_field("ip")
  local vol, _ = command.volume(ip)
  if vol then
    log.info(string.format("[%s](%s) BoseCmd: volume up +5 {%s}", bose_utils.get_serial_number(device),
                           device.label, ip))
    local err = command.set_volume(ip, vol.actual + 5)

cjswedes:
@ketchumj1 Feel free to test with the invitation link in the PR prior to merge. I will also test later and merge. (If the PR has merged the invite link becomes invalid FYI)

PR #1047, opened 2023-11-02
Bose use current volume as default audio notification

cjswedes:
Rather than defaulting to 30 if the audio notification doesn't provide a volume level, default to the current volume with a default of 30 only if the current volume is unknown.

addresses: #911


cjswedes:
I tested this locally and it plays at the existing level rather than 30. @lelandblue feel free to approve and merge if this is the behavior you think we should be using.

dljsjr (approved):

PR #1049, opened 2023-11-03
BUG2-532 Setpoint values could have been sent as floats instead of wh…

greens:
…ole numbers

### DIFF
--- a/drivers/SmartThings/zigbee-thermostat/src/leviton/init.lua
+++ b/drivers/SmartThings/zigbee-thermostat/src/leviton/init.lua
@@ -16,6 +16,7 @@ local clusters = require "st.zigbee.zcl.clusters"
 local Thermostat = clusters.Thermostat
 local FanControl = clusters.FanControl
 local capabilities = require "st.capabilities"
+local utils = require "st.utils"
 
 local ENDPOINT = 10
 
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/zigbee-thermostat/src/leviton/init.lua
```diff
@@ -89,12 +90,12 @@ end
...
 local function set_cooling_setpoint(driver, device, command)
   device:emit_event(capabilities.thermostatCoolingSetpoint.coolingSetpoint({value = command.args.setpoint*1.0, unit = "C"}))
-  device:send_to_component(command.component, Thermostat.attributes.OccupiedCoolingSetpoint:write(device, command.args.setpoint*100))
+  device:send_to_component(command.component, Thermostat.attributes.OccupiedCoolingSetpoint:write(device, utils.round(command.args.setpoint*100)))
 end
 
 local function set_heating_setpoint(driver, device, command)
   device:emit_event(capabilities.thermostatHeatingSetpoint.heatingSetpoint({value = command.args.setpoint*1.0, unit = "C"}))
-  device:send_to_component(command.component, Thermostat.attributes.OccupiedHeatingSetpoint:write(device, command.args.setpoint*100))
+  device:send_to_component(command.component, Thermostat.attributes.OccupiedHeatingSetpoint:write(device, utils.round(command.args.setpoint*100)))
```
cjswedes, line 98: Are these events being rejected because of this? I assume thats the reason for the change, but if not then it seems like we are reducing the precision for users, which I know has been frustrating in the past.

pegor-karoglanian (approved):

drivers/SmartThings/zigbee-thermostat/src/leviton/init.lua
greens: 1. Yes the events were being rejected, because this `write` expects an Int, and we were sending a float
2. There's no loss of precision here, since zigbee expects tenths of a degree here (we're rounding after multiplying by 100).

cjswedes (approved):

mtilchen (approved):

PR #1050, opened 2023-11-03
Matter lock subscribe request unit test expectation fix

cjswedes:
A bug in the unit test framework has allowed these tests to get past. The integration test framework bugfix will be released with the 51 lua libs.

- [x] Investigate if the workarounds for multiple test init functions (seen [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/967/files#diff-9302cdbb223bfda1db868ce1766871b4e8faa95fa4da3ed0e162f64556d5247dR120)) are still needed


cjswedes:
> Investigate if the workarounds for multiple test init functions (seen [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/967/files#diff-9302cdbb223bfda1db868ce1766871b4e8faa95fa4da3ed0e162f64556d5247dR120)) are still needed

My investigation lead to the conclusion there is no existing way to allow for a case where multiple devices are initialized and have metadata updates expectations set that would be fulfilled by calls in the device init handler. This is because of the way the integration test framework adds devices to a drivers device cache prior to running the test function, and because the mock_devices_api, always checks device update/create expectations in order.

greens (approved):

PR #1055, opened 2023-11-06
feature/Kaadas-259-1104

lelandblue:
WWST CERT Request for Kaadas Z-Wave Lock 259


lelandblue:
Closing, we need to validate some information.

PR #1059, opened 2023-11-07
CHAD-12031 Fixes incorrectly-ported logic for Schlage user codes

greens:


greens:
I think there may be some more subtle differences here in the code report handling for Schlage but trying to pick this code apart is challenging.

cjswedes (approved):

PR #1065, opened 2023-11-08
matter-button: Missing button press after a long press

tpmanley:
Some Switch devices will send a MultiPressComplete event at the end of a long press sequence and others will not. Normally the driver creates button capability events on receipt of the MultiPressComplete but in the case of a long press it creates the "held" event on receipt of the LongPress event. If it also receives a MultiPressComplete as part of that long press sequence then it should ignore it so as not to generate an incorrect extra event.

The problem was that if a new button press was started (as determined by the receipt of a InitialPress) the driver was still ignoring the next MultiPressComplete event, and so that button press sequence would be ignored.

The fix is to only ignore MultiPressComplete that occurs as part of a long press sequence.

https://smartthings.atlassian.net/browse/MTR-509

### DIFF
--- a/drivers/SmartThings/matter-button/src/init.lua
+++ b/drivers/SmartThings/matter-button/src/init.lua
@@ -13,17 +13,22 @@ local STATIC_PROFILE_SUPPORTED = {2, 4, 8}
 
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 local DEFERRED_CONFIGURE = "__DEFERRED_CONFIGURE"
+
+-- Some switches will send a MultiPressComplete event as part of a long press sequence. Normally the driver will create a
+-- button capability event on receipt of MultiPressComplete, but in this case that would result in an extra event because
+-- the "held" capability event is generated when the LongPress event is received. The IGNORE_NEXT_MPC flag is used
+-- to tell the driver to ignore MultiPressComplete if it is received after a long press to avoid this extra event.
 local IGNORE_NEXT_MPC = "__ignore_next_mpc"
 
 -- These are essentially storing the supported features of a given endpoint
 -- TODO: add an is_feature_supported_for_endpoint function to matter.device that takes an endpoint
-local EMULATE_HELD = "__emulate_held" -- for MSR devices we can emulate this on the software side
-local MULTI_BUTTON = "__multi_button" -- for multi-press devices, only trigger an event on a multi-button complete
-local INITIAL_PRESS_ONLY = "__initial_press_only" -- for devices that support MS, but not MSR
+local EMULATE_HELD = "__emulate_held" -- for non-MSR (MomentarySwitchRelease) devices we can emulate this on the software side
+local SUPPORTS_MULTI_PRESS = "__multi_button" -- for MSM devices (MomentarySwitchMultiPress), create an event on receipt of MultiPressComplete
+local INITIAL_PRESS_ONLY = "__initial_press_only" -- for devices that support MS (MomentarySwitch), but not MSR (MomentarySwitchRelease)
 
 local HUE_MANUFACTURER_ID = 0x100B
 
---helper function to create liste of multi press values
+--helper function to create list of multi press values
 local function create_multi_list(size, supportsHeld)
   local list = {"pushed", "double"}
   if supportsHeld then table.insert(list, "held") end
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-button/src/init.lua
```diff
@@ -281,19 +290,19 @@ local function initial_event_handler(driver, device, ib, response)
...
 end
 
--- if the devce distinguishes a long press event, it will always be a "held"
+-- if the device distinguishes a long press event, it will always be a "held"
 -- there's also a "long release" event, but this event is required to come first
 local function long_press_event_handler(driver, device, ib, response)
   device:emit_event_for_endpoint(ib.endpoint_id, capabilities.button.button.held({state_change = true}))
-  local MSM = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH_MULTI_PRESS})
-  if contains(MSM, ib.endpoint_id) then
+  if get_field_for_endpoint(device, SUPPORTS_MULTI_PRESS, ib.endpoint_id) then
```
tpmanley, line 297: I didn't see a reason it would recheck the feature flag rather than using the `SUPPORTS_MULTI_PRESS` field but let me know if I'm missing something.

greens: probably because initial drafts didn't set a field here, and I missed this one when that was changed

drivers/SmartThings/matter-button/src/init.lua
```diff
@@ -13,17 +13,22 @@ local STATIC_PROFILE_SUPPORTED = {2, 4, 8}
 
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 local DEFERRED_CONFIGURE = "__DEFERRED_CONFIGURE"
+
+-- Some switches will send a MultiPressComplete event as part a long press sequence. Normally the driver will create a
+-- button capability event on receipt of MultiPressComplete, but in this case that would result in an extra event because
+-- the "held" capability event is generated when the LongPress event is received. The IGNORE_NEXT_MPC flag is used
+-- to tell the driver to ignore MultiPressComplete if it is received after a long press to avoid this extra event.
 local IGNORE_NEXT_MPC = "__ignore_next_mpc"
```
greens, line 21: So this flag was here but never used? Hmm...

greens (approved):

drivers/SmartThings/matter-button/src/init.lua
tpmanley: It was being used, but it was ignore the next MultiPressComplete whether it came following a long press (InitialPress, LongPress, LongRelease, MultiPressComplete) or it was part of a separate press event following a long press (long press: InitialPress, LongPress, LongRelease, single press: InitialPress, ShortRelease, MultiPressComplete). The spec does seem to require MultiPressComplete to be sent as part of a long press event sequence, but not all devices are doing that and there has been some discussion that the spec needs to be more clear in this area.

drivers/SmartThings/matter-button/src/init.lua
```diff
@@ -13,17 +13,22 @@ local STATIC_PROFILE_SUPPORTED = {2, 4, 8}
 
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 local DEFERRED_CONFIGURE = "__DEFERRED_CONFIGURE"
+
+-- Some switches will send a MultiPressComplete event as part a long press sequence. Normally the driver will create a
```
ctowns, line 17: nit:
```suggestion
-- Some switches will send a MultiPressComplete event as part of a long press sequence. Normally the driver will create a
```

drivers/SmartThings/matter-button/src/init.lua
```diff
@@ -270,8 +275,12 @@ end
 ----------------------------------------------------------------------------
 
 -- initial press
-local function initial_event_handler(driver, device, ib, response)
-  if not get_field_for_endpoint(device, MULTI_BUTTON, ib.endpoint_id) then
+local function initial_press_event_handler(driver, device, ib, response)
+  if get_field_for_endpoint(device, SUPPORTS_MULTI_PRESS, ib.endpoint_id) then
+    -- Receipt of an InitialPress event means we do not want to ignore the next MultiPressComplete event
+    -- or else we would potentially not create the expected button capability event
+    set_field_for_endpoint(device, IGNORE_NEXT_MPC, ib.endpoint_id, nil)
```
ctowns, line 282: nit: is there a reason to set this as nil instead of `false`? nil would make me think it hadn't been intentionally set yet.

ctowns (approved):
Adding the definitions for all the M** acronyms in the comments was very helpful, good thinking!

drivers/SmartThings/matter-button/src/init.lua
greens: just saves memory. truth value would be the same, though

PR #1067, opened 2023-11-09
CHAD-11469: Detect Bose IP Changes

NoahCornell:
Tested with my bose and changing its IP.

### DIFF
--- a/drivers/SmartThings/bose/src/init.lua
+++ b/drivers/SmartThings/bose/src/init.lua
@@ -300,6 +300,28 @@ local bose = Driver("bose", {
   },
 })
 
+local function ip_change_check()
+  local id_to_device = {}
+  local device_list = bose:get_devices()
+  for _, device in ipairs(device_list) do
+    local id = bose_utils.get_serial_number(device)
+    id_to_device[id] = device
+  end
+  discovery.find(nil, function(found)
+    local known = id_to_device[found.id]
+    if known ~= nil then
+      local known_ip = known:get_field("ip")
+      if known_ip == nil or known_ip ~= found.ip then
+        log.info_with({hub_logs = true}, "Updating device ip:", found.id, found.ip)
+        known:set_field("ip", found.ip, { persist = true })
+      end
+    end
+  end)
+end
+
+local IP_CHANGE_CHECK_INTERVAL_S = 600
+bose:call_on_schedule(IP_CHANGE_CHECK_INTERVAL_S, ip_change_check, "IP Change Check Task")
+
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/bose/src/init.lua
```diff
@@ -300,6 +300,28 @@ local bose = Driver("bose", {
...
+      local known_ip = known:get_field("ip")
+      if known_ip == nil or known_ip ~= found.ip then
+        log.trace("Updating device ip:", found.id, found.ip)
+        known:set_field("ip", found.ip, { persist = true })
+      end
+    end
+  end)
+end
+
+local IP_CHANGE_CHECK_INTERVAL_S = 60
```
dljsjr, line 322: I think for Hue and Sonos we defaulted to 10m instead of 1m, I'm not sure which makes more sense here tbh though. I feel like there's value to being very responsive but there's also value in reducing LAN traffic and hub load. Not a nit or a blocker from my perspective, but wanted to make a comment to have a discussion; no pressure to actually incorporate this change.

```suggestion
local IP_CHANGE_CHECK_INTERVAL_S = 600
```

dljsjr (approved):

drivers/SmartThings/bose/src/init.lua
```diff
@@ -300,6 +300,28 @@ local bose = Driver("bose", {
...
+  for _, device in ipairs(device_list) do
+    local id = bose_utils.get_serial_number(device)
+    id_to_device[id] = device
+  end
+  discovery.find(nil, function(found)
+    local known = id_to_device[found.id]
+    if known ~= nil then
+      local known_ip = known:get_field("ip")
+      if known_ip == nil or known_ip ~= found.ip then
+        log.trace("Updating device ip:", found.id, found.ip)
```
cjswedes, line 315: I think this should be info level and in hub logs, since it represents a moment when we essentially re-discover the device on the network.

cjswedes (approved):

drivers/SmartThings/bose/src/init.lua
NoahCornell: I think hue is also at a minute right now https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/philips-hue/src/init.lua#L1448 

I wasn't 100% on this being a minute so I have no problem changing this to be 10 minutes.

dljsjr: Womp, I should fix that. I thought we did that already.

PR #1068, opened 2023-11-09
Fixes bad merge. Re-adds missing profiles

greens:


lelandblue:
Thanks Steven, this will address the merge of 980, 982, 983 where I resolved comflicts improperly. All of those devices use the same profile basic switch.

lelandblue (approved):
I have verified the model numbers and profiles, looks good.

PR #1072, opened 2023-11-13
new device (matter-window-covering): Zemismart MT01

lelandblue:
WWST Certification request to add the Zemismart MT01 window covering device.


drivers/SmartThings/matter-window-covering/fingerprints.yml
```diff
@@ -5,7 +5,12 @@ matterManufacturer:
     vendorId: 0x130A
     productId: 0x55
     deviceProfileName: window-covering-profile
-
+#Zemismart
+- id: 5020/65534
```
ctowns, line 9: Indent this line to follow the formatting in the other fingerprints. Also this should be a string (should be surrounded with quotes "").

Where did this ID come from? I am wondering if we can change it to something more human readable? `5020/65534` doesn't really tell us much about what this device is. See above where the id is "Eve MotionBlinds", which makes it clear what we are looking at.

lelandblue: The partner provided that information, I will need to follow up with the certification tool team regarding that. Thanks Cooper. I made another commit.

ctowns (approved):

drivers/SmartThings/matter-window-covering/fingerprints.yml
```diff
@@ -5,7 +5,12 @@ matterManufacturer:
     vendorId: 0x130A
     productId: 0x55
     deviceProfileName: window-covering-profile
-
+#Zemismart
+  - id: "Zemismart MT01 Slide Curtain"
+    deviceLabel: Zemismart MT01 Slide Curtain
+    vendorId: 0x139C
+    productId: 0xFFFE
+    deviceProfileName: window-covering-battery
```
ctowns, line 13: I believe this should just be `window-covering` (no battery) as per discussion in the latest two comments here: https://smartthings.atlassian.net/browse/MTR-605?focusedCommentId=1416175

ctowns (changes requested):

drivers/SmartThings/matter-window-covering/fingerprints.yml
lelandblue: Updated, thanks @ctowns

ctowns (approved):

PR #1076, opened 2023-11-15
Add back matter switch fingerprints

ctowns:
Generic fingerprints for the OnOff Switch, Dimmer Switch, and Color Dimmer Switch device types were removed because they require bindings to work correctly, and we do not currently support bindings (removed here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1003). These generic fingerprints were removed in favor of having these devices join as `matter-thing` until binding support is added. However, the Matter Test app does currently have support for OnOff Switch, Dimmer Switch, and Color Dimmer Switch devices, but with the removal of these generic fingerprints these devices just show up as `matter-thing`. We want to add back these fingerprints so that these test app devices show up correctly in the SmartThings app.

https://smartthings.atlassian.net/browse/BUG2-624

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -1,4 +1,4 @@
-matterManufacturer:
+matterManufacturer:    
 #Eve
   - id: "Eve/Energy/US"
     deviceLabel: Eve Energy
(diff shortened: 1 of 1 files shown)


ctowns:
@tpmanley what are your thoughts on adding these fingerprints back for the sake of testing with the Matter Test app?

PR #1082, opened 2023-11-21
Support switching from matter-thing to switch-binary

ctowns:
The On/Off Light switch device type requires bindings in order to work correctly because the On/Off cluster on this device type is client type as specified in the matter spec. Because we do not support bindings, we have removed the generic fingerprint for this device type so that it will join as a matter-things so we do not expose broken funcitonality to users. However, we have found some devices that do not follow the spec and report an On/Off Light Switch device type and yet have a server type OnOff cluster. These out-of-spec devices would technically have worked before we decided to remove the fingerprints due to lack of binding support, so these devices will be broken now that they will join as matter-thing. This adds a workaround that allows us to check for a special case where a matter-thing supports the OnOff cluster as a server, and then we switch to the switch-binary profile.

Additionally, this improves our multi-switch profile switching logic to only count server type On/Off cluster endpoints towards the switch total for the same reasons as described above.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -32,6 +32,7 @@ local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 -- New profiles need to be added for devices that have more switch endpoints
 local MAX_MULTI_SWITCH_EPS = 7
+local detect_matter_thing
 
 local function convert_huesat_st_to_matter(val)
   return math.floor((val * 0xFE) / 100.0 + 0.5)
@@ -69,7 +70,7 @@ local function initialize_switch(device)
   -- support for bindings.
   local num_server_eps = 0
   for _, ep in ipairs(switch_eps) do
-    if(device:supports_server_cluster(clusters.OnOff.ID, ep)) then
+    if device:supports_server_cluster(clusters.OnOff.ID, ep) then
       num_server_eps = num_server_eps + 1;
       if num_server_eps == 1 then
         component_map["main"] = ep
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -64,25 +64,40 @@ local function initialize_switch(device)
...
-  if num_switch_eps > 1 then
-    device:try_update_metadata({profile = string.format("switch-%d", math.min(num_switch_eps, MAX_MULTI_SWITCH_EPS))})
+  -- Also, the case where num_server_eps == 1 is a workaround for devices that have the On/Off
+  -- Light Switch device type but implement the On Off cluster as server (which is against the spec
+  -- for this device type). By default, we do not support On/Off Light Switch because by spec these
+  -- devices need bindings to work correctly (On/Off cluster is client in this case), so this device type
+  -- does not have a generic fingerprint and will join as a matter-thing. However, we have
+  -- seen some devices claim to be On/Off Light Switch device type and still implement On/Off server, so this
+  -- is a workaround for those devices.
+  if num_server_eps == 1 and device.label == "Matter Thing" then
```
ctowns, line 92: I am not sure if using the `device.label` is the best way to determine if the device was joined as a `matter-thing`. The device will join with this label so I do think it is safe to use for this case where we are checking if the device is currently joined as a matter thing so we can attempt to profile switch as needed.

However, after the switch, the device label would still be "Matter Thing", which could be somewhat confusing to a user. Is there a way to change that or is it something we have to accept with this workaround?

ctowns: > I think instead you could try to detect the shape of the matter thing profile (only refresh capability).

This is a great idea! I have implemented this with the `detect_matter_thing` function in the latest commit. Based on my testing with the matter test app as well as my nanoleaf light strip, this is working as well.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -64,25 +65,40 @@ local function initialize_switch(device)
...
-    if i == 1 then
-      component_map["main"] = ep
-    else
-      component_map[string.format("switch%d", i)] = ep
+  -- Additionally, since we do not support bindings at the moment, we only want to count
+  -- On/Off clusters that have been implemented as server. This can be removed when we have
+  -- support for bindings.
+  local num_server_eps = 0
+  for _, ep in ipairs(switch_eps) do
+    if(device:supports_server_cluster(clusters.OnOff.ID, ep)) then
```
tpmanley, line 73: nit:

```suggestion
    if device:supports_server_cluster(clusters.OnOff.ID, ep) then
```

ctowns: Done

tpmanley (approved):

greens (approved):

cjswedes (approved):

PR #1086, opened 2023-11-24
wallhero s3i s2i s1i add backlight control

thinkaName:


greens (approved):
Do you also want to go ahead and make the change for the ACL-401ON, or does it not support this?

vangoran:
> Do you also want to go ahead and make the change for the ACL-401ON, or does it not support this?

Hi, @greens 
This device(model: ACL-401ON) is a smart plug, so it doesn't support the backlight control

greens (approved):

PR #1090, opened 2023-11-28
Reverse switch operation

Detailzc:
Because the APP page displays "Turn off indicator light" , so when "true", it should be closed


greens (commented):
Could you write a test to encapsulate this behavior? It would have been nice if there were one already.

greens (approved):

PR #1091, opened 2023-11-28
Matter Window Covering: battery profile switching support

ctowns:
Adds ability to profile switch to a profile with battery support on init.


ctowns:
> Duplicate profile check: Warning - duplicate profiles detected. window-covering-battery.yml == window-covering-profile.yml

This is going to happen because the `window-covering-profile` has been intentionally duplicated to be renamed as `window-covering-battery`. The old profile must linger until it is not longer used, but we can delete it once all devices in the field have been updated.

samadDotDev (approved):
Thanks for putting this together so quickly and for accommodating the profile switch on info change, @ctowns

PR #1094, opened 2023-11-30
Matter Switch: use parent-child for multi switch devices

ctowns:
Updates matter switch driver to use parent-child for multi switch devices rather than multi-component devices. Along with these changes, we will check for matter bridge devices before attempting a profile switch to prevent bridge devices from incorrectly switching to MCD switch profiles.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/eve-energy/init.lua
+++ b/drivers/SmartThings/matter-switch/src/eve-energy/init.lua
@@ -22,6 +22,7 @@ local clusters = require "st.matter.clusters"
 local cluster_base = require "st.matter.cluster_base"
 local utils = require "st.utils"
 local data_types = require "st.matter.data_types"
+local device_lib = require "st.device"
 
 local EVE_MANUFACTURER_ID = 0x130A
 local PRIVATE_CLUSTER_ID = 0x130AFC01
@@ -41,7 +42,9 @@ local REPORT_TIMEOUT = (15 * 60) -- Report the value each 15 minutes
 -------------------------------------------------------------------------------------
 
 local function is_eve_energy_products(opts, driver, device)
-  if device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID then
+  -- this sub driver does not support child devices
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
+     device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID then
     return true
   end
 
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -118,12 +118,26 @@ local function find_child(parent, ep_id)
   return parent:get_child_by_parent_assigned_key(string.format("%d", ep_id))
 end
 
+local function detect_bridge(device)
+  for _, ep in ipairs(device.endpoints) do
+    for _, dt in ipairs(ep.device_types) do
+      if dt.device_type_id == 0x000E then -- look for aggregator type
+        return true
+      end
+    end
+  end
+  return false
+end
+
 local function device_init(driver, device)
   if device.network_type == device_lib.NETWORK_TYPE_MATTER then
     -- initialize_switch will create parent-child devices as needed for multi-switch devices.
     -- However, we want to maintain support for existing MCD devices, so do not initialize
     -- device if it has already been previously initialized as an MCD device.
-    if not device:get_field(COMPONENT_TO_ENDPOINT_MAP) and not device:get_field(SWITCH_INITIALIZED) then
+    -- Also, do not attempt a profile switch for a bridge device.
+    if not device:get_field(COMPONENT_TO_ENDPOINT_MAP) and
+       not device:get_field(SWITCH_INITIALIZED) and
+       not detect_bridge(device) then
       -- create child devices as needed for multi-switch devices
       initialize_switch(driver, device)
     end
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -32,6 +32,7 @@ local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN
 
 local SWITCH_INITIALIZED = "__switch_intialized"
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
+local AGGREGATOR_DEVICE_TYPE_ID = 0x000E
 local detect_matter_thing
 
 local function convert_huesat_st_to_matter(val)
@@ -121,7 +122,7 @@ end
 local function detect_bridge(device)
   for _, ep in ipairs(device.endpoints) do
     for _, dt in ipairs(ep.device_types) do
-      if dt.device_type_id == 0x000E then -- look for aggregator type
+      if dt.device_type_id == AGGREGATOR_DEVICE_TYPE_ID then
         return true
       end
     end
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -17,6 +17,7 @@ local log = require "log"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
 local utils = require "st.utils"
+local device_lib = require "st.device"
 
 local MOST_RECENT_TEMP = "mostRecentTemp"
 local RECEIVED_X = "receivedX"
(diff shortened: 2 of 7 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -44,46 +44,47 @@ end
...
-      else
-        component_map[string.format("switch%d", num_server_eps)] = ep
+      num_server_eps = num_server_eps + 1
+      if ep ~= main_endpoint then -- don't create a child device that maps to the main endpoint
+        local name = string.format("%s %d", device.label, num_server_eps)
+        driver:try_create_device(
+          {
+            type = "EDGE_CHILD",
+            label = name,
+            profile = "switch-binary",
```
ctowns, line 76: At the moment, this only supports multi-switch devices (just as we had before with MCD). However, I think this could be expand on in a future PR by checking the clusters/device type on each endpoint and mapping to the correct profile now that we have the parent-child paradigm in place.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -44,46 +44,47 @@ end
...
+            parent_device_id = device.id,
+            parent_assigned_child_key = string.format("%02d", ep),
```
ctowns, line 78: I have seen elsewhere in our drivers that the `parent_assigned_child_key` is converted to hex with something like `parent_assigned_child_key = string.format("%02X", i)`, however this will not work here due to an issue in the lua libs where the `MatterChildDevice:get_endpoint()` function does not convert the child key as base 16. In this case, any child key for any endpoint above 10 would not be converted correctly back to a number.

This can be address with a simple change in the lua libs: https://github.ecodesamsung.com/iot-hub/scripting-engine/pull/1608, however, to get this working without the lua libs change I have just used base 10 for the parent key for now.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -29,9 +30,8 @@ local COLOR_TEMPERATURE_KELVIN_MIN = 1
 local COLOR_TEMPERATURE_MIRED_MAX = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN
 local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX
 
-local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
+local SWITCH_INITILIZED = "__switch_intialized"
```
greens, line 33: typo: INITIALIZED

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -264,7 +246,7 @@ end
 local function temp_attr_handler(driver, device, ib, response)
   if ib.data.value ~= nil then
     if (ib.data.value < COLOR_TEMPERATURE_MIRED_MIN or ib.data.value > COLOR_TEMPERATURE_MIRED_MAX) then
-      device.log.warn_with({hub_logs = true}, "Device reported color temperature %d mired outside of supported capability range", ib.data.value)
+      device.log.warn_with({hub_logs = true}, string.format("Device reported color temperature %d mired outside of supported capability range", ib.data.value))
```
greens, line 249: doesn't log do the `string.format` call internally?

ctowns: Apparently not :/

drivers/SmartThings/matter-switch/src/init.lua
ctowns: Done in ecc24c0b70e61b0b1ea56a8550c315aaf23e9158

drivers/SmartThings/matter-switch/src/test/test_multi_switch_parent_child.lua
```diff
@@ -0,0 +1,222 @@
+-- Copyright 2022 SmartThings
```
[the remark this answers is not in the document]
ctowns: Done in ecc24c0b70e61b0b1ea56a8550c315aaf23e9158

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -44,46 +44,47 @@ end
...
   for _, ep in ipairs(switch_eps) do
     if device:supports_server_cluster(clusters.OnOff.ID, ep) then
-      num_server_eps = num_server_eps + 1;
-      if num_server_eps == 1 then
-        component_map["main"] = ep
-      else
-        component_map[string.format("switch%d", num_server_eps)] = ep
+      num_server_eps = num_server_eps + 1
+      if ep ~= main_endpoint then -- don't create a child device that maps to the main endpoint
+        local name = string.format("%s %d", device.label, num_server_eps)
```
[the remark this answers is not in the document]
ctowns: Just making sure - I imagine that we only want to update the child device name/location when we onboard, right? So the child devices are updated to the parent label/room that is set in the initial onboarding flow, but not subsequent changes, right? That way if sometime later a user sets the name of a child device and then changes the name of the parent, it doesn't overwrite the child label.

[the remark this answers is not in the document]
greens: Presumably, but probably challenging to do.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -92,46 +93,27 @@ local function initialize_switch(device)
...
-    error(string.format(
-      "Matter multi switch device will have limited function. Profile doesn't exist with %d components, max is %d",
-      num_server_eps,
-      MAX_MULTI_SWITCH_EPS
-    ))
   end
 end
 
 local function component_to_endpoint(device, component)
-  local map = device:get_field(COMPONENT_TO_ENDPOINT_MAP) or {}
```
tpmanley, line 108: We talked about this offline too, but I think we need to keep the support for MCD devices in the driver since there may be existing devices relying on that functionality.

ctowns:
Here is the changes we have discussed spread across a few commits:
* 1st commit: changes the matter-switch driver to use parent-child moving forward for new devices
* 2nd commit: adds logic to recover matter-bridge devices
* 3rd commit: fixes up the mock devices in the unit tests so that the device_types are used correctly in the bridge detection

drivers/SmartThings/matter-switch/src/init.lua
tpmanley: I think that's definitely worth doing, but agreed it can be follow up PR

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -118,12 +118,26 @@ local function find_child(parent, ep_id)
   return parent:get_child_by_parent_assigned_key(string.format("%02d", ep_id))
 end
 
+local function detect_bridge(device)
+  for _, ep in ipairs(device.endpoints) do
+    for _, dt in ipairs(ep.device_types) do
+      if dt.device_type_id == 0x000E then -- look for aggregator type
```
tpmanley, line 124: Do we have the device types available in a lib already? If not, I'd suggest adding a local var to define the device type so it's less of a magic number

tpmanley (approved):

drivers/SmartThings/matter-switch/src/init.lua
ctowns: Done!

greens (approved):

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
```diff
@@ -41,7 +42,9 @@ local REPORT_TIMEOUT = (15 * 60) -- Report the value each 15 minutes
 -------------------------------------------------------------------------------------
 
 local function is_eve_energy_products(opts, driver, device)
-  if device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID then
+  -- this sub driver does not support child devices
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
+     device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID then
```
cjswedes, line 47: This will often require 2 operations, whereas if the eve mfr check was first it will often only need 1.

drivers/SmartThings/matter-switch/src/init.lua
cjswedes: I dont think we should be aiming to switch the way we store these values, since it would be an api breaking change. ~Limiting the endpoint id to two digits is not spec compliant since an endpoint id is a u16 which can be > 99~ Looks like `"%02X` isn't actually limiting, it is just 0 padding, but I still think my suggested change is valid, since the 0 padding is not typically done for base 10 numbers.
```suggestion
            parent_assigned_child_key = string.format("%d", ep),
```
Also, I dont think we should link to private repos here.

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
ctowns: `if device.network_type == device_lib.NETWORK_TYPE_MATTER` is put first because child devices do not have `manufacturer_info`, so calling `can_handle` for the child devices results in a warning. Therefore, I added the check for the `network_type` first to prevent the warning.

drivers/SmartThings/matter-switch/src/init.lua
ctowns: Done!

ctowns:
This is now working and tested with the composed, bridged multi-switch device I was testing with in the matter device test app.

samadDotDev (approved):

Con-Tejus (approved):

cjswedes (approved):

PR #1095, opened 2023-11-30
WASH-1872 Join names in CN not being localized.

greens:


tools/deploy.py
```diff
@@ -50,7 +50,12 @@
       reader = csv.reader(csvfile)
       for row in reader:
         print("en: "+row[0]+" "+LOCALE+": "+row[1])
-        subprocess.run("find . -name 'fingerprints.yml' | xargs sed -i '' 's/deviceLabel: "+row[0].translate(slash_escape)+"/deviceLabel: "+row[1].translate(slash_escape)+"/g'", shell=True)
+        subprocess.run(
+          "find . -name 'fingerprints.yml' | xargs sed -i '' 's/deviceLabel: "+row[0].translate(slash_escape)+"/deviceLabel: "+row[1].translate(slash_escape)+"/g'",
+          shell=True
+          cwd=os.path.dirname(current_path)
```
greens, line 56: explicitly setting the working directory is the only real thing that changed here

tools/deploy.py
```diff
@@ -50,7 +50,12 @@
...
+        )
+    subprocess.run("git status")
```
greens, line 58: this is mostly so we can confirm there were file changes when this runs

cjswedes (approved):

ctowns (approved):

PR #1108, opened 2023-12-08
Fix parsing patterns for SSDP responses in Sonos

dljsjr:
The current way we parse the SSDP response headers can cause failures in discovering Sonos speakers with Unicode characters that require more than one byte for their representation, such as CJK characters.

The current parsing patterns use the '`%g`' character class, which does not match characters that don't fit in a single byte. While Lua itself is utf-8 compatible, and it is safe to use multi-code-point graphemes or clusters in Lua strings, the actual string library is byte-oriented for many operations. The pattern used currently tries to match each byte against the provided class, which causes it to fail to match and capture many headers.

We now use a parsing pass for header key/value pairs derived from [the Luncheon library with bug fixes](https://github.com/cosock/luncheon/pull/9) that is aware of the proper structure for HTTP Header key/values, which is the format that an SSDP response uses.

### DIFF
--- a/drivers/SmartThings/sonos/src/ssdp.lua
+++ b/drivers/SmartThings/sonos/src/ssdp.lua
@@ -22,7 +22,15 @@ local function process_response(val)
     if l == nil or l == "" then
       break
     end
-    local k, v = string.match(l, "(.-):%s*(.*)$")
+    -- SSDP Messages use the HTTP/1.1 Header Field rules described in RFC 2616, 4.2: https://datatracker.ietf.org/doc/html/rfc2616#section-4.2
+    -- This pattern extracts the Key/Value pairs in to a Lua table via the two capture groups.
+    -- The key capture group is composed entirely of a negating matcher to exclude illegal characters, ending at the `:`.
+    -- The RFC states that after the colon there may be any arbitrary amount of leading space between the colon
+    -- and the value, and that the value shouldn't have any trailing whitespace, so we exclude those as well.
+    -- The original Luncheon implementation of this Lua Pattern used iteration and detected the `;` separator
+    -- that indicates key/value parameters, however, we don't make that distinction here and instead leave parsing
+    -- values with parameters to the consumers of the output of this function.
+    local k, v = string.match(l, '([^%c()<>@,;:\\"/%[%]?={} \t]+):%s*(.-)%s*$')
     if k == nil or k == "" then
       return nil, string.format("Couldn't parse header/value pair for line %q", l)
     end
--- a/drivers/SmartThings/sonos/src/ssdp.lua
+++ b/drivers/SmartThings/sonos/src/ssdp.lua
@@ -92,12 +92,28 @@ function SSDP.search(search_term, callback)
     local val, rip, _ = s:receivefrom()
 
     if val then
-      local headers = process_response(val)
+      local headers, err = process_response(val)
 
-      -- log all SSDP responses, even if they don't have proper headers
+      if err ~= nil then
+        log.error(err or "Unknown error while parsing SSDP response headers")
+        goto continue
+      end
+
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/sonos/src/ssdp.lua
```diff
@@ -10,8 +10,8 @@ local SSDP = {}
 local function process_response(val)
   local info = {}
   val = string.gsub(val, "HTTP/1.1 200 OK\r\n", "", 1)
-  for k, v in string.gmatch(val, "([%g]+): ([%g ]*)\r\n") do
-    info[string.lower(k)] = v
+  for k, v in string.gmatch(val, "(.-):(.-)\r\n") do
```
varzac, line 13: This looks like it will now capture the space after the colon in the second group, where previously the leading space would be omitted.

dljsjr: Correct, but it's trimmed [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1108/files#diff-20c5b6b201061a964f9e0ffcf0e8bee4b2a5ad1eaa2579e391c566abc87b7672R14)

tandres (approved):

drivers/SmartThings/sonos/src/ssdp.lua
dljsjr: It took me a second to get a pattern that worked well with the oddly formed `EXT:\r\n` header, but I've updated the code here to exclude leading whitespace without using a hard-coded single space.

dljsjr:
Waiting for https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1107 to fix tests.

drivers/SmartThings/sonos/src/ssdp.lua
```diff
@@ -8,9 +8,24 @@ SONOS_SSDP_SEARCH_TERM = "urn:smartspeaker-audio:service:SpeakerGroup:1"
...
   val = string.gsub(val, "HTTP/1.1 200 OK\r\n", "", 1)
-  for k, v in string.gmatch(val, "([%g]+): ([%g ]*)\r\n") do
+
+  local info = {}
+  -- iterate line-by-line by splitting on `\r\n`
+  for l in string.gmatch(val, "([^\r\n]*)\r\n") do
+    if l == nil or l == "" then
+      break
+    end
+    local k, v = string.match(l, '([^%c()<>@,;:\\"/%[%]?={} \t]+):%s*(.-);?$')
```
varzac, line 25: This is a very opaque regex, I think a comment could be helpful here.

drivers/SmartThings/sonos/src/ssdp.lua
```diff
@@ -8,9 +8,24 @@ SONOS_SSDP_SEARCH_TERM = "urn:smartspeaker-audio:service:SpeakerGroup:1"
 local SSDP = {}
 
 local function process_response(val)
-  local info = {}
+  -- check first line assuming it's the HTTP Status Line, which if not is invalid
+  local status_line = string.match(val, "([^\r\n]*)\r\n")
+  if not (status_line and string.match(status_line, "HTTP/1.1 200 OK"))  then
```
cjswedes, line 13: Does it need to be `HTTP/1.1` here or could it be a different version and still work?

cjswedes (approved):

drivers/SmartThings/sonos/src/ssdp.lua
dljsjr: UPnP/SSDP is always HTTP/1.1: https://openconnectivity.org/upnp-specs/UPnP-arch-DeviceArchitecture-v2.0-20200417.pdf

Section 1.1

PR #1111, opened 2023-12-11
BUG2-659 Zigbee locks: fix out-of-order commands

greens:
Commands can arrive from a zigbee lock after an unlock event in any order. This can be a problem if the door lock operation event arrives _after_ the attribute report about the door lock state. Because the door lock attribute update doesn't include any useful metadata, but it arrives first, the second command with the useful metadata is marked as a duplicate and debounced, dropping all of that useful metadata on the floor. This same issue can also happen with Z-Wave, but because of the overloading of the notification defaults, it will have to be fixed via updates to the lua libraries. This fix is also going into the lua libraries, but it's easier to also deploy it quicker here.

### DIFF
--- a/drivers/SmartThings/zigbee-lock/src/init.lua
+++ b/drivers/SmartThings/zigbee-lock/src/init.lua
@@ -34,7 +34,11 @@ local UserStatusEnum            = LockCluster.types.DrlkUserStatus
 local UserTypeEnum              = LockCluster.types.DrlkUserType
 local ProgrammingEventCodeEnum  = LockCluster.types.ProgramEventCode
 
+local socket = require "cosock.socket"
 local lock_utils = require "lock_utils"
+
+local DELAY_LOCK_EVENT = "_delay_lock_event"
+local MAX_DELAY = 10
 
 local reload_all_codes = function(driver, device, command)
   -- starts at first user code index then iterates through all lock codes as they come in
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -300,6 +303,88 @@ local function init(driver, device)
...
 end
 
+-- The following two functions are from the lock defaults
+-- They are in the base driver temporarily because of BUG2-659
+local lock_state_handler = function(driver, device, value, zb_rx)
+  local attr = capabilities.lock.lock
+  local LOCK_STATE = {
+    [value.NOT_FULLY_LOCKED]     = attr.unknown(),
+    [value.LOCKED]               = attr.locked(),
+    [value.UNLOCKED]             = attr.unlocked()
```
tpmanley, line 313: ZCL8 also includes 0xFF as `Undefined` which we can map to `unknown`

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -300,6 +303,88 @@ local function init(driver, device)
...
+  -- this is where we decide whether or not we need to delay our lock event because we've
+  -- observed it coming before the event (or we're starting to compute the timer)
+  local delay = device:get_field(DELAY_LOCK_EVENT) or 100
+  if (delay < MAX_DELAY) then
+    device.thread:call_with_delay(delay+.5, function ()
+      device:emit_event_for_endpoint(zb_rx.address_header.src_endpoint.value, LOCK_STATE[value.value])
+    end)
+  else
+    local socket = require "socket"
+    device:set_field(DELAY_LOCK_EVENT, socket.gettime())
```
tpmanley, line 325: Is the goal just to set `DELAY_LOCK_EVENT` to something bigger than `MAX_DELAY` or does it really need to be the time? Wondering if we could just set it to `MAX_DELAY` and avoid having to require `socket`

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -300,6 +303,88 @@ local function init(driver, device)
...
+local lock_operation_event_handler = function(driver, device, zb_rx)
+  local event_code = zb_rx.body.zcl_body.operation_event_code.value
+  local source = zb_rx.body.zcl_body.operation_event_source.value
+  local OperationEventCode = require "st.zigbee.generated.zcl_clusters.DoorLock.types.OperationEventCode"
+  local METHOD = {
+    [0] = "keypad",
+    [1] = "command",
+    [2] = "manual",
+    [3] = "rfid",
+    [4] = "fingerprint",
```
tpmanley, line 339: `4` and `5` aren't defined in ZCL8. Did they get added in something later?

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -300,6 +303,88 @@ local function init(driver, device)
...
+  local STATUS = {
+    [OperationEventCode.LOCK]           = capabilities.lock.lock.locked(),
+    [OperationEventCode.UNLOCK]         = capabilities.lock.lock.unlocked(),
+    [OperationEventCode.ONE_TOUCH_LOCK] = capabilities.lock.lock.locked(),
+    [OperationEventCode.KEY_LOCK]       = capabilities.lock.lock.locked(),
+    [OperationEventCode.KEY_UNLOCK]     = capabilities.lock.lock.unlocked(),
+    [OperationEventCode.AUTO_LOCK]      = capabilities.lock.lock.locked(),
+    [OperationEventCode.MANUAL_LOCK]    = capabilities.lock.lock.locked(),
+    [OperationEventCode.MANUAL_UNLOCK]  = capabilities.lock.lock.unlocked()
+  }
```
tpmanley, line 351: Should probably add `SCHEDULE_LOCK` and `SCHEDULE_UNLOCK` too

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -300,6 +303,88 @@ local function init(driver, device)
...
+  local event = STATUS[event_code]
+  if (event ~= nil) then
+    event["data"] = {}
+    if (event_code == OperationEventCode.AUTO_LOCK) then
```
tpmanley, line 355: `SCHEDULE_LOCK` and `SCHEDULE_UNLOCK` are probably closest to "auto"

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -300,6 +303,88 @@ local function init(driver, device)
...
+      event.data.method = "auto"
+    else
+      event.data.method = METHOD[source]
+    end
+    if (source == 0) then
```
tpmanley, line 360: Could use a variable to at least a comment that 0 is Keypad

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -300,6 +303,88 @@ local function init(driver, device)
...
+
+    -- if this is an event corresponding to a recently-received attribute report, we
+    -- want to set our delay timer for future lock attribute report events
+    if device:get_latest_state(
+        device:get_component_id_for_endpoint(zb_rx.address_header.src_endpoint.value),
+        capabilities.lock.ID,
+        capabilities.lock.lock.ID) == event.value.value then
+      local preceding_event_time = device:get_field(DELAY_LOCK_EVENT) or 0
+      local socket = require "socket"
+      local time_diff = socket.gettime() - preceding_event_time
```
tpmanley, line 379: I see here that the actual time is used. Maybe just require `socket` once at the top rather than at separate places in the code?

greens:
@tpmanley other than the timing code this was all pulled directly from lua libs defaults. If you see some errors or enhancements we need to make there, can you open a ticket?

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -34,8 +34,12 @@ local UserStatusEnum            = LockCluster.types.DrlkUserStatus
 local UserTypeEnum              = LockCluster.types.DrlkUserType
 local ProgrammingEventCodeEnum  = LockCluster.types.ProgramEventCode
 
+local socket = require "socket"
```
cjswedes, line 37: nit: `require "cosock.socket"` instead. It ends up being the same function, but we want to push everyone to use cosock wrapped socket rather than directly accessing the native socket.

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -300,6 +304,92 @@ local function init(driver, device)
   lock_utils.populate_state_from_data(device)
 end
 
+-- The following two functions are from the lock defaults
+-- They are in the base driver temporarily because of BUG2-659
```
cjswedes, line 308: nit: only internal devs know what BUG2-659 is, the comment could be more useful for the community

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -34,8 +34,12 @@ local UserStatusEnum            = LockCluster.types.DrlkUserStatus
 local UserTypeEnum              = LockCluster.types.DrlkUserType
 local ProgrammingEventCodeEnum  = LockCluster.types.ProgramEventCode
 
+local socket = require "socket"
 local lock_utils = require "lock_utils"
 
+local DELAY_LOCK_EVENT = "_delay_lock_event"
+local MAX_DELAY = 10
```
cjswedes, line 41: So if there is a lock that doesn't send the LockOperationEvent, we will be delaying lock state capability events for 10 seconds. Is that right? If so, why 10, and should it instead be like 1-3 seconds?

cjswedes (approved):

drivers/SmartThings/zigbee-lock/src/init.lua
greens: This _should_ be written in a way such that if a lock _never_ sends a LockOperationEvent, we just send the event immediately.

greens: i.e. only a device that sends _both_ can have its messages delayed, and then by a max of 10 seconds

drivers/SmartThings/zigbee-lock/src/init.lua
tpmanley: OK I'll write a separate ticket about this and the other issues

tpmanley (approved):

wkhenon (approved):

PR #1115, opened 2023-12-14
Create to Zigbee fan for SLED

seungkwan-choi:
This is a new project for Samsung LED.
It's a Zigbee fan light and the model no. is SAMSUNG-ITM-Z-003.
For the SLED Fan light, we creat a folder " Zigbee Fan" in the smartThings directory.
This includes profiles and so on.

It makes two device cards : fan light and light.

In the Fan light card, you can fan control on/off, fan speed and light control on/off, brightness level.
In the Light card, you can light cotrol on/off, brightness level

Please review and approve this request.


drivers/SmartThings/zigbee-fan/src/configurations.lua
```diff
@@ -0,0 +1,66 @@
+-- Copyright 2022 SmartThings
```
cbaumler, line 1:
```suggestion
-- Copyright 2023 SmartThings
```

drivers/SmartThings/zigbee-fan/README.md
greens, line 1: this file should be deleted

drivers/SmartThings/zigbee-fan/config.yml
```diff
@@ -0,0 +1,6 @@
+name: 'ITM Fan Light 231201'
+packageKey: 'itm-fan-light'
+description: 'EDGE Driver, ITM Fan Light'
+vendorSupportInformation:
+permissions:
+  zigbee: {}
```
greens, line 6: since you're creating a new base driver category, I would be as generic in this config as possible (this is basically a copy/paste of z-wave fan's config):
```suggestion
name: 'Zigbee Fan'
packageKey: 'zigbee-fan'
permissions:
  zigbee: {}
description: "SmartThings driver for zigbee fan devices"
vendorSupportInformation: "https://support.smartthings.com"
```

drivers/SmartThings/zigbee-fan/src/init.lua
```diff
@@ -0,0 +1,46 @@
...
+  },
+  sub_drivers = {
+    require("itm-fan-light")
+  },
+  lifecycle_handlers = {
+    init = device_init
+  }
+}
+
+zigbee_switch:run()
```
greens, line 46:
```suggestion
local zigbee_fan_driver = {
  supported_capabilities = {
    capabilities.switch,
    capabilities.switchLevel,
    capabilities.colorControl,
    capabilities.colorTemperature,
    capabilities.fanspeed
  },
  sub_drivers = {
    require("itm-fan-light")
  },
  lifecycle_handlers = {
    init = device_init
  }
}

defaults.register_for_default_handlers(zigbee_fan_driver, zigbee_fan_driver.supported_capabilities)
local fan = ZigbeeDriver("zigbee-fan", zigbee_fan_driver)
fan:run()
```

drivers/SmartThings/zigbee-fan/src/init.lua
```diff
@@ -0,0 +1,46 @@
+-- Copyright 2022 SmartThings
```
greens, line 1:
```suggestion
-- Copyright 2023 SmartThings
```

drivers/SmartThings/zigbee-fan/src/itm-fan-light/init.lua
```diff
@@ -0,0 +1,168 @@
...
+        profile = profile,
+        parent_device_id = parent.id,
+        parent_assigned_child_key = child_type,
+        vendor_provided_label = string.format("%s %s", parent.label, child_type)
+    }
+    driver:try_create_device(child_metadata)
+end
+
+local function info_changed(driver, device, event, args)
+  if (device.preferences or {}).childLight2 then
```
greens, line 50: if the user sets this preference to `false`, should it not delete the device?

Also, as written, on join the device will be a multi-component device with a fan component and a light component, when the preference is changed, there will still be the multi-component device but also a separate child device that is just a light. Is the intent that the component and the child device are the same and will mirror each other? there's no endpoint to component mapping or anything like that. as written, the switch component is non-functional

greens (changes requested):
At the very minimum you should write some tests here so that at least it's clear what you're trying to do and you can demonstrate that the driver runs. As it is right now this driver will not run.

drivers/SmartThings/zigbee-fan/src/configurations.lua
choiseungkwan: It has been modified.

drivers/SmartThings/zigbee-fan/README.md
choiseungkwan: I deleted it.

drivers/SmartThings/zigbee-fan/config.yml
choiseungkwan: It has been modified.

drivers/SmartThings/zigbee-fan/src/init.lua
choiseungkwan: It has been modified.

drivers/SmartThings/zigbee-fan/src/init.lua
choiseungkwan: It has been modified.

drivers/SmartThings/zigbee-fan/src/itm-fan-light/init.lua
choiseungkwan: Yes, I tried to use it like that.
Is it a problem if I use it like this?
If it is a problem, should it be automatically deleted from the main card (Light) during child device false?

greens: Having a component and a child device that control the same functionality is a problem, yes. I believe you should just stick with the multi-component implementation and get rid of the preference (and the separate presentation file). Since this is for one specific device, we should know whether it has a light or not and can make a decision when we integrate it how we want to present it.

choiseungkwan: Is it okay to delete the preference in profiles and make the lighting and fan automatically main card at the same time when onboarding?
Also, is the presentation file the jason file of CANVAS? Do I have to modify this file as well as the profile? Or can I just modify the profile?

greens: So, specifically, I would remove 

```
preferences:
  - title: "Child Device"
    name: childLight2
    description: "Create Child Device : Light"
    required: true
    preferenceType: boolean
    definition:
      default: false
```

from `drivers/SmartThings/zigbee-fan/profiles/itm-fan-light.yml` and then remove the preference handling code from the driver.

Then I would `git rm drivers/SmartThings/zigbee-fan/profiles/switch-level.yml`.

choiseungkwan: I tried to modify the init file or config file in seungkwan-choi:main according to your request, but I can not modify it because the edit file is not activated. Please guide me on how to modify it.

seungkwan-choi: Files have been modified including init. Please review it.

drivers/SmartThings/zigbee-fan/src/configurations.lua
```diff
@@ -0,0 +1,66 @@
+-- Copyright 2023 SmartThings
```
wkhenon, line 1:
```suggestion
-- Copyright 2024 SmartThings
```

drivers/SmartThings/zigbee-fan/src/init.lua
```diff
@@ -0,0 +1,49 @@
+-- Copyright 2023 SmartThings
```
wkhenon, line 1:
```suggestion
-- Copyright 2024 SmartThings
```

drivers/SmartThings/zigbee-fan/src/itm-fan-light/init.lua
```diff
@@ -0,0 +1,201 @@
+-- Copyright 2023 philh30
```
wkhenon, line 1:
```suggestion
-- Copyright 2024 SmartThings
```

lelandblue:
Hey @seungkwan-choi Please take a look at the above when possible. Thank you.

drivers/SmartThings/zigbee-fan/src/itm-fan-light/init.lua
seungkwan-choi: The suggested content han been revised.
Also, we are setting up and checking the test folder commented by greens.
We are going to add test code to src folder.

drivers/SmartThings/zigbee-fan/src/configurations.lua
seungkwan-choi: The suggested content han been revised.

drivers/SmartThings/zigbee-fan/src/init.lua
seungkwan-choi: The suggested content han been revised.

lelandblue:
Hey @seungkwan-choi - Thank you for your efforts. We will be able to complete this request after additional test code is added. Please review at your earliest opportunity.

seungkwan-choi:
I have written a test code for the fan light and modified the init code accordingly to match the test code.

seungkwan-choi:
We have discovered an issue in the code and need to re-edit it.

drivers/SmartThings/zigbee-fan/src/init.lua
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+end
+
+local zigbee_fan_driver = {
+  supported_capabilities = {
+    capabilities.switch,
+    capabilities.switchLevel,
+    capabilities.colorControl,
+    capabilities.colorTemperature,
```
greens, line 35: these two capabilities seem to be unused

greens (commented):
You need to decide whether you're using multi-component or parent-child. The current implementation where you do both is not what we or most users would prefer.

There shouldn't be two different places to control the same functionality (the light).

seungkwan-choi:
Device

When onboarding, it is separated into fan and fan light, with the fan plug-in having only fan control.
And the light plug-in having only light control.
Two devices do not have the same functions.
Please refer to the attached photo.
If this becomes an issue, please explain how to make the change.

drivers/SmartThings/zigbee-fan/profiles/itm-fan-light.yml
greens, line 1: The reason that the second `light` component is not shown is because of the linked metadata.

Because this is present, it overwrites what is present in the actual profile. However, the included profile should match as near as possible the actual presentation. 

The `light` component should be removed from this file. The `switchLevel` capability should also be removed from the `main` component.

Additionally, this file should be renamed to `fan-3speed.yml` as it is in zwave, since there is nothing device-specific about this profile.

One more thing I'd like you to check is the "Information" menu for the device when it's added. I'm worried that using a linked presentation that does not include the `firmwareUpdate` capability will prevent the firmware info from being displayed.

drivers/SmartThings/zigbee-fan/src/itm-fan-light/init.lua
```diff
@@ -0,0 +1,182 @@
...
+
+local levels_for_speed = {
+  [0] = 0,
+  [1] = 25,
+  [2] = 50,
+  [3] = 100,
+}
+
+local function level_to_speed(level)
+  local speed = 0
```
greens, line 42: the luacheck error you are getting is because this `0` value is not used

drivers/SmartThings/zigbee-fan/src/itm-fan-light/init.lua
```diff
@@ -0,0 +1,182 @@
...
+    create_child_devices(driver, device, 'switch-level', 'light')
+  end
+end
+
+local function device_init(driver, device)
+  local dev = device:get_parent_device()
+    if dev == nil then
+    dev = device
+  end
+     dev:refresh(driver, dev)
```
greens, line 81: nit:spacing

additionally, there are race conditions involved when you call `refresh` from `init`, as the zigbee stack may not be fully initialized.

drivers/SmartThings/zigbee-fan/src/itm-fan-light/init.lua
```diff
@@ -0,0 +1,182 @@
...
+ end
+
+ local function info_changed(driver, device, event, args)
+  local dev = device:get_parent_device()
+  if dev == nil then
+    dev = device
+  end
+  dev:send(FanControl.attributes.FanMode:read(dev))
```
greens, line 89: is this FanControl read not part of the `refresh`?

drivers/SmartThings/zigbee-fan/src/itm-fan-light/init.lua
```diff
@@ -0,0 +1,182 @@
...
+
+-- CAPABILITY HANDLERS
+
+local function on_handler(driver, device, command)
+  local dev = device
+  if device.network_type == stDevice.NETWORK_TYPE_CHILD then
+    command.component = 'light'
+    dev = device:get_parent_device()
+  end
+  if command.component == 'light' then
```
greens, line 101: unused

drivers/SmartThings/zigbee-fan/src/itm-fan-light/init.lua
```diff
@@ -0,0 +1,182 @@
...
+  device:emit_component_event(device.profile.components.main,evt)
+  device:emit_component_event(device.profile.components.main,capabilities.switchLevel.level(levels_for_speed[value.value], { visibility = { displayed = false } }))
+  if value.value > 0 then
+    device:set_field('LAST_FAN_SPD', value.value, {persist = true})
+  end
+end
+
+local function zb_level_handler(driver, device, value, zb_rx)
+  local evt = capabilities.switchLevel.level(math.floor((value.value / 254.0 * 100) + 0.5))
+  device:emit_component_event(device.profile.components.light,evt)
```
greens, line 160: so this `light` component is not visible to the user as you have this set up, but you're still emitting events here then duplicating them to the child

greens (commented):
Please see this documentation on how to properly implement parent/child devices:

https://developer.smartthings.com/docs/edge-device-drivers/zigbee/device.html#parent-child-devices

In several places you're directly calling `device:get_parent_device()` or `get_child_by_assigned_parent_key()` when you should be using the `set_find_child` in your `init` handler.

seungkwan-choi:
We have modified the code to use multi-component.
Please review.

drivers/SmartThings/zigbee-fan/profiles/fan-light.yml
```diff
@@ -0,0 +1,27 @@
+name: fan-light
+components:
+  - id: main
+    label: Fan 
+    capabilities:
+      - id: switch
+        version: 1
+      - id: fanSpeed
+        version: 1
```
greens, line 9: if you want to restrict this to 3 speeds, you could set the range like this:

```
    config:
      values:
        - key: "fanSpeed.value"
          range: [0, 3]
```

We also use this for most zigbee dimmers as well: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-switch/profiles/on-off-level.yml#L10

and I would suggest you also use it in the light component below.

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
```diff
@@ -0,0 +1,161 @@
...
+  [2] = 50,
+  [3] = 100,
+}
+
+local function level_to_speed(level)
+  local speed = level
+  if speed ~= nil then
+    if level == 0 then
+      speed  = 0
+    elseif level  > 0 and level <= 25 then
```
greens, line 46: nit:
```suggestion
    elseif level > 0 and level <= 25 then
```

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
```diff
@@ -0,0 +1,161 @@
...
+  end
+end
+
+-- ZIGBEE HANDLERS
+
+local function zb_fan_control_handler(driver, device, value, zb_rx)
+  device:emit_event(capabilities.fanSpeed.fanSpeed(value.value))
+  local evt = capabilities.switch.switch(value.value > 0 and 'on' or 'off', { visibility = { displayed = false } })
+  device:emit_component_event(device.profile.components.main,evt)
+  device:emit_component_event(device.profile.components.main,capabilities.switchLevel.level(levels_for_speed[value.value], { visibility = { displayed = false } }))
```
greens, line 112: why do you have these set as `displayed = false`? I expect the user would want to see these values in history.

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
```diff
@@ -0,0 +1,161 @@
...
+local function on_handler(driver, device, command)
+  local dev = device
+  if command.component == 'light' then
+    dev:send(OnOff.server.commands.On(dev))
+  else
+    local speed = dev:get_field('LAST_FAN_SPD') or 1
+    dev:send(FanControl.attributes.FanMode:write(dev,speed))
+    dev:send(FanControl.attributes.FanMode:read(dev))
+  end
+  dev:send(FanControl.attributes.FanMode:read(dev))
```
greens, line 70: you're reading `FanMode` twice in a row if the fan is turned on

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
```diff
@@ -0,0 +1,161 @@
...
+end
+
+local function off_handler(driver, device, command)
+  local dev = device
+  if command.component == 'light' then
+    dev:send(OnOff.server.commands.Off(dev))
+  else
+    dev:send(FanControl.attributes.FanMode:write(dev,FanControl.attributes.FanMode.OFF))
+  end
+  dev:send(FanControl.attributes.FanMode:read(dev,FanControl.attributes.FanMode.OFF))
```
greens, line 80: I don't think you meant to include an argument here

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
```diff
@@ -0,0 +1,161 @@
...
+  else
+    speed = 0
+  end
+  return speed
+end
+
+-- CAPABILITY HANDLERS
+
+local function on_handler(driver, device, command)
+  local dev = device
```
greens, line 62: this local variable is not needed

same elsewhere

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
```diff
@@ -0,0 +1,161 @@
...
+
+-- ZIGBEE HANDLERS
+
+local function zb_fan_control_handler(driver, device, value, zb_rx)
+  device:emit_event(capabilities.fanSpeed.fanSpeed(value.value))
+  local evt = capabilities.switch.switch(value.value > 0 and 'on' or 'off', { visibility = { displayed = false } })
+  device:emit_component_event(device.profile.components.main,evt)
+  device:emit_component_event(device.profile.components.main,capabilities.switchLevel.level(levels_for_speed[value.value], { visibility = { displayed = false } }))
+  if value.value > 0 then
+    device:set_field('LAST_FAN_SPD', value.value, {persist = true})
```
greens, line 114: this value is available to you already using `device:get_latest_state("main", capabilities.fanSpeed.ID, capabilities.fanSpeed.fanSpeed.NAME)`

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
```diff
@@ -0,0 +1,161 @@
...
+
+local function switch_level_handler(driver, device, command)
+  local dev = device
+  if command.component == 'light' then
+    local level = math.floor(command.args.level/100.0 * 254)
+    dev:send(Level.server.commands.MoveToLevelWithOnOff(dev, level, command.args.rate or 0xFFFF))
+  else
+    local speed = level_to_speed(command.args.level)
+    dev:send(FanControl.attributes.FanMode:write(dev,speed))
+    dev:send(FanControl.attributes.FanMode:read(dev))
```
greens, line 91: the fan component should no longer be getting `setLevel` commands.

drivers/SmartThings/zigbee-fan/profiles/fan-light.yml
seungkwan-choi: I will change it.

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
seungkwan-choi: I will change it.
But even if you change it to true, the behavior of the history doesn't seem to change.

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
seungkwan-choi: I will change it.

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
seungkwan-choi: I will change it.

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
seungkwan-choi: I will change it.

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
seungkwan-choi: When I apply the code that was guided to me, if I turn on the fan after turning it off, I can't retrieve the last value (fan speed level) and instead retrieve the off state. Can I still use set field & get_field as before?

drivers/SmartThings/zigbee-fan/src/fan-light/init.lua
seungkwan-choi: I didn't understand well. Please specify how you want me to change it in detail.

greens: The profile definition for the `light` component does not include the `Switch Level` capability, so it should be impossible for the `light` component to receive a `setLevel` command.

seungkwan-choi: Do you mean "fan control" instead of "light"? This code actually does not use the switch level, but uses the switch level for compatibility with external devices and wants to use it

greens: Yes. Sorry I meant the "main" fan component:
```
components:
  - id: main
    label: Fan 
    capabilities:
      - id: switch
        version: 1
      - id: fanSpeed
        version: 1
      - id: firmwareUpdate
        version: 1
      - id: refresh
        version: 1
    categories:
      - name: Fan
```
it does not have the Switch Level capability so will never receive these commands.

seungkwan-choi: As mentioned earlier, the fan does not use the switch level, but uses it for compatibility with external devices.

greens: Unless the capability is listed in the device's presentation, the `Switch Level` functionality should not be available for the component. It would not be visible as an option for external devices.

greens (approved):

PR #1119, opened 2023-12-15
add Resideo Korea DT300ST-M0000 Thermostat

aqwerf:
Resideo Korea Thermostat for WWST


wkhenon (approved):

aqwerf:
> Do you think you could include a test that works with one of the child devices? Here's an example of how this can be done in zigbee: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-switch/src/test/test_wallhero_switch.lua

I've updated the pull request to include unit tests for each child device as you suggested. These tests should provide better coverage and ensure the functionality of the child devices. Please let me know if there are any further changes or improvements needed.
Thanks for your guidance!

greens (approved):
Looks great. Thank you.

PR #1126, opened 2023-12-20
Automatically add/remove Hue lights when they are added/removed to Hue Bridge

dljsjr:
Three commits here:

1. Automatically add or remove devices when SSE events indicating topology changes come in,
2. Check for changes in the devices joined to a Hue Bridge when we establish connectivity with a bridge. This handles the case where the Hue configuration changes while the hub is offline or the driver isn't running for some reason.
3. Wrap this new functionality so we have fallbacks on hubs that don't have the `try_delete_device` API in their Lua Libs.

Note that this won't be fully supported in production until the next release of hubcore that adds the delete API, by which time we also hope that the relevant cloud side changes will be ready as well. Hence the inclusion of the fallback if the `try_delete_device` function isn't available on the `Driver` table.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/disco.lua
+++ b/drivers/SmartThings/philips-hue/src/disco.lua
@@ -83,6 +83,31 @@ function HueDiscovery.search_for_bridges(driver, computed_mac_addresses, callbac
   end
 end
 
+function HueDiscovery.scan_bridge_and_update_devices(driver, bridge_id)
+  if driver.ignored_bridges[bridge_id] then return end
+
+  local known_identifier_to_device_map = {}
+  for _, device in ipairs(driver:get_devices()) do
+    -- the bridge won't have a parent assigned key so we give that boolean short circuit preference
+    local dni = device.parent_assigned_child_key or device.device_network_id
+    known_identifier_to_device_map[dni] = device
+  end
+
+  local known_bridge_device = known_identifier_to_device_map[bridge_id]
+  if known_bridge_device and known_bridge_device:get_field(Fields.API_KEY) then
+    HueDiscovery.api_keys[bridge_id] = known_bridge_device:get_field(Fields.API_KEY)
+  end
+
+  HueDiscovery.search_bridge_for_supported_devices(driver, bridge_id, HueDiscovery.disco_api_instances[bridge_id],
+    function(hue_driver, svc_info, device_info)
+      discovered_device_callback(hue_driver, bridge_id, svc_info, device_info, known_identifier_to_device_map)
+    end,
+    "[Discovery: " ..
+    (known_bridge_device.label or bridge_id or known_bridge_device.id or "unknown bridge") .. " bridge re-scan]",
+    true
+  )
+end
+
 ---@param driver HueDriver
 ---@param bridge_ip string
 ---@param bridge_id string
@@ -109,7 +134,7 @@ discovered_bridge_callback = function(driver, bridge_ip, bridge_id, known_identi
           utils.labeled_socket_builder((known_bridge_device.label or bridge_id or known_bridge_device.id or "unknown bridge"))
         )
 
-    HueDiscovery.search_bridge_for_supported_devices(driver, HueDiscovery.disco_api_instances[bridge_id],
+    HueDiscovery.search_bridge_for_supported_devices(driver, bridge_id, HueDiscovery.disco_api_instances[bridge_id],
       function(hue_driver, svc_info, device_info)
         discovered_device_callback(hue_driver, bridge_id, svc_info, device_info, known_identifier_to_device_map)
       end,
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -830,38 +831,137 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
...
+                    color = add_data.color,
+                    dimming = add_data.dimming,
+                    color_temperature = add_data.color_temperature,
+                    mode = add_data.mode,
+                    parent_device_id = bridge_device.id,
+                    hue_device_id = add_data.owner.rid,
+                    hue_device_data = new_device_info
+                  }
+
+                  driver:try_create_device(create_device_msg)
```
tandres, line 949: Mostly curious: What happens to the device if the add fails? Like this event is an sse that a new device has been added are we just back at the original behavior?

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -830,38 +831,137 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
...
+                  end
+
+                  if new_device_info == nil then
+                    log.warn("Couldn't get all device info for new light, unable to join. Try using Scan Nearby to find new Hue lights.")
+                    return
+                  end
+
+                  log.info(
+                    string.format(
+                      "Adding light \"%s\" to Philips Hue Edge Driver",
```
tpmanley, line 897: nit: The driver name will already be logged so probably not necessary to include it in the message

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -830,38 +831,137 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
...
                   string.format(
-                    "New light added to Hue Bridge \"%s\", light properties: \"%s\", " ..
-                    "re-run discovery to join new lights to SmartThings",
-                    device.label, st_utils.stringify_table(add_data, nil, false)
+                    "New light added to Hue Bridge \"%s\", light properties: \"%s\"",
+                    bridge_device.label, json.encode(add_data)
                   )
                 )
+
+                cosock.spawn(function()
```
tpmanley, line 852: This code seems similar to what exists in `disco.lua`. Could it be combined?

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -830,38 +831,137 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
...
+                  local hue_api = bridge_device:get_field(Fields.BRIDGE_API)
+                  if hue_api == nil then
+                    bridge_device.log.warn("No Hue API instance available for new light event.")
+                    return
+                  end
+
+                  local hue_device_rid = add_data.owner.rid
+                  local rest_resp, rest_err = hue_api:get_device_by_id(hue_device_rid)
```
tpmanley, line 860: Could/should this be using `get_light_by_id` instead of `get_device_by_id`?

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -802,9 +803,9 @@ local function do_bridge_network_init(driver, device, bridge_url, api_key)
         end
 
         for _, event in ipairs(events) do
-          log.trace(
-            string.format("Bridge %s processing event from SSE stream: %s",
-              (device.label or device.id or "unknown device"), st_utils.stringify_table(event)))
+          -- log.trace(
```
tpmanley, line 806: nit: remove?

drivers/SmartThings/philips-hue/src/disco.lua
```diff
@@ -222,6 +250,20 @@ function HueDiscovery.search_bridge_for_supported_devices(driver, api_instance,
...
   end
+
+  if do_delete then
+    for _, device in ipairs(driver:get_devices()) do ---@cast device HueDevice
+      local device_type = device:get_field(Fields.DEVICE_TYPE)
+      local not_known_to_bridge = device_is_joined_to_bridge[device:get_field(Fields.HUE_DEVICE_ID) or ""]
+      local parent_device_id = device.parent_device_id or device:get_field(Fields.PARENT_DEVICE_ID) or ""
+      local parent_bridge_device = driver:get_device_info(parent_device_id)
+      local is_child_of_bridge = parent_bridge_device and (parent_bridge_device:get_field(Fields.BRIDGE_ID) == bridge_id)
+      if device_type == "light" and is_child_of_bridge and not not_known_to_bridge then
```
tpmanley, line 261: Just curious, why is it necessary to check that the device_type is a light? I know we only support lights now, but if in the future we supported other devices this seems like one more place that would need to get updated. Is it not enough to know the device is a child of the bridge and it's been deleted?

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: If this fails, there's no way for the driver to know. So it would either get picked up at the next driver restart or the next "Scan Nearby".

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: No; `get_light_by_id` provides information about the light resource service for the device, which we already have as part of the SSE event. We need the device resource service information as well, in order to populate all of the metadata in the `create_device_msg`. So this call is to obtain additional data.

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: Yup; in fact, there's a *ton* of code that's not very `DRY` in this driver.

I'd like to make a separate PR that is a purely hygeine refactor without any behavior changes to catch some of this stuff.

drivers/SmartThings/philips-hue/src/disco.lua
dljsjr: I think you're right. I can adjust this.

varzac (approved):

cjswedes (approved):

tpmanley (approved):

PR #1130, opened 2023-12-25
Added usage of datastore to save discovery cache

cchillerr:
Samsung told us that in an edge case in which the hub or the device crashes during discovery stages the device could be initialised improperly. They informed us about how to try to reduce the damage


dljsjr (approved):

greens (changes requested):
it's unclear to me in which circumstances you're using `.datastore.discovery_cache` and in which you're using `joined_device`.

lelandblue:
> it's unclear to me in which circumstances you're using `.datastore.discovery_cache` and in which you're using `joined_device`.

Hello @chill-cai or @cchillerr Could you please review this question at your nearest opportunity? 

Thank you.

chill-cai:
Hi @greens ,
We use both of them, as you can see joined device is only a key-bool dictionary.
``` joined_device[dni] = true ```
In other hand, driver.datastore.discovery_cache will store the detail of the device joined:
```
driver.datastore.discovery_cache[dni] = {
  ip = ip,
  device_info = device_info,
  credential = credential,
  device_info = device_info,
}
```
We implement this structure to have a better performance to discover device, joined_device doesn't need retrieve data from device. You can considerer that the discovery of device is separated into two steps:
1. Mark down the device is found (joined_device)
2. If the credential of device is provided, continue to retrieve device info and save it into cache (driver.datastore.discovery_cache). This step will take several seconds.
```
      if not joined_device[dni] then
        try_add_device(driver, dni, ip)
        joined_device[dni] = true
      end
```
```
local function try_add_device(driver, device_dni, device_ip)
  log.trace(string.format("try_add_device : dni=%s, ip=%s", device_dni, device_ip))

  local credential  = driver.discovery_helper.get_credential(driver, device_dni, device_ip)

  if not credential then
    log.error(string.format("failed to get credential. dni=%s, ip=%s", device_dni, device_ip))
    joined_device[device_dni] = nil
    return
  end

  update_device_discovery_cache(driver, device_dni, device_ip, credential)
  local create_device_msg = driver.discovery_helper.get_device_create_msg(driver, device_dni, device_ip)
  driver:try_create_device(create_device_msg)
end
```
Hope that answers your question.

dljsjr:
@chill-cai Seems reasonable. However, we need the `luacheck` failures cleaned up before we can merge. Thanks!

chill-cai:
@dljsjr Sure, I just fix the issue, please have a check. Thanks.

lelandblue:
> it's unclear to me in which circumstances you're using `.datastore.discovery_cache` and in which you're using `joined_device`.

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1130#issuecomment-1925520319

PR #1140, opened 2024-01-02
Add fingerprints for WiZ devices

ctowns:


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -93,6 +93,1796 @@ matterManufacturer:
     vendorId: 0x100b
```
ctowns, line 93: Note: this `0x2168` fingerprint existed already but it also appeared in the device spreadsheet, so I removed the new fingerprint to keep this original one.

greens (approved):

PR #1142, opened 2024-01-03
CHAD-12259 zigbee motion: high battery usage

greens:
Users reported higher battery consumption with edge drivers and a disparity was noted between DTH and driver cluster configurations. A forthcoming change in the lua libraries will change the defaults, but this change should significantly improve the situation for most users in the meantime with a simple refresh. Changes the IAS Zone reporting configuration to use the device defaults and changes temperature reporting to a max interval of 10 minutes with a 1 degree celsius reporting threshold.


greens:
I think I'll also be expanding this to `zigbee-contact`. The temperature issue isn't present, but the IAS zone one is.

Additionally, I think the monitored attribute health check might also still be a culprit even if the reporting configuration is updated. In that sense, the simple refresh from this PR should be an _improvement_ but a rejoin will be required for a full fix, since I don't feel comfortable fully removing the monitoring of the IAS Zone attribute...

drivers/SmartThings/zigbee-contact/src/init.lua
```diff
@@ -45,7 +45,6 @@ local zigbee_contact_driver_template = {
   },
   sub_drivers = {
     require("aqara"),
-    require("battery-overrides"),
```
cjswedes, line 48: Is this now released in 51?

cjswedes (approved):

drivers/SmartThings/zigbee-contact/src/init.lua
greens: It wasn't for 51. It was just a sub-driver for one specific Visonic device. Behavior could have been captured by simply adding the model to drivers/SmartThings/zigbee-contact/src/contact-temperature-sensor/init.lua, so that's what I did.

wkhenon (approved):

PR #1144, opened 2024-01-03
Disable hue light controls when in sync mode

bflorian:
Update Philips Hue light device profiles to reference device configurations that disable the light controls when it is in sync mode. This PR should not be merged until the device configurations have been deployed.

### DIFF
--- a/drivers/SmartThings/philips-hue/profiles/legacy-color.yml
+++ b/drivers/SmartThings/philips-hue/profiles/legacy-color.yml
@@ -6,10 +6,6 @@ components:
     version: 1
   - id: switchLevel
     version: 1
-    config:
-      values:
-        - key: "level.value"
-          range: [ 1, 100 ]
   - id: colorControl
     version: 1
   - id: samsungim.hueSyncMode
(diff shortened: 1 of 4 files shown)


drivers/SmartThings/philips-hue/profiles/legacy-color.yml
```diff
@@ -6,10 +6,6 @@ components:
     version: 1
   - id: switchLevel
     version: 1
-    config:
```
dljsjr, line 9: Since the custom presentation isn't visible in this changeset, I wanted to make a visibility comment here to call out that these embedded device configs are removed in this PR because they're rolled in to the custom presentation upon which this PR depends.

greens (approved):

dljsjr:
Added `do not merge`, it was requested that we wait for some mobile app and infrastructure work before merging this.

cjswedes (approved):

mantiger (commented):
This code is the last code to finalize hue music sync.
It should be included in R2 timeline so please merge it as soon as possible.

PR #1145, opened 2024-01-04
Aqara Wireless Mini Switch T1

Hongming6:


drivers/SmartThings/zigbee-button/src/aqara/init.lua
```diff
@@ -0,0 +1,94 @@
+local battery_defaults = require "st.zigbee.defaults.battery_defaults"
```
liuyd96, line 1: Missing CopyRight Profile.

drivers/SmartThings/zigbee-button/src/test/test_aqara_button.lua
```diff
@@ -0,0 +1,135 @@
+---@diagnostic disable: param-type-mismatch, missing-parameter
```
liuyd96, line 1: Please remove this.

Hongming6: It has been modified

greens (approved):

drivers/SmartThings/zigbee-button/src/aqara/init.lua
```diff
@@ -0,0 +1,108 @@
...
+    end
+end
+
+  local function added_handler(self, device)
+    device:send(cluster_base.write_manufacturer_specific_attribute(device,
+                PRIVATE_CLUSTER_ID, PRIVATE_ATTRIBUTE_ID, MFG_CODE, data_types.Uint8, 1))
+    device:emit_event(capabilities.button.supportedButtonValues({"pushed","held","double"}, {visibility = { displayed = false }}))
+    device:emit_event(capabilities.button.numberOfButtons({value = 1}))
+    device:emit_event(capabilities.button.button.pushed({state_change = false}))
+    device:emit_event(capabilities.battery.battery(100))
```
wkhenon, line 88: I could be missing something, but is the battery value held at 100%? Is there a specific function that is updating the value? Or is it an automatic reportable change?

greens: This is the "added" lifecycle event handler that fires when the device is added for the first time. In general we do this so that the value is populated on add so there are no UI hiccups.

After the fact, the value will be read using the standard zigbee attributes and handled by the defaults.

PR #1148, opened 2024-01-04
add branch for new luncheon version

FreeMasen:
### TODO

- [ ] Test on target


drivers/SmartThings/philips-hue/src/lunchbox/rest.lua
```diff
@@ -129,6 +129,9 @@ local function parse_chunked_response(original_response, sock)
 end
 
 local function handle_response(sock)
+  if require("version").api > 7 then
```
cjswedes, line 132: I think this, but need to double check:
```suggestion
  if require("version").api > 8 then
```

dljsjr:
Superceded by https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1152

PR #1155, opened 2024-01-10
Add line numbers to failure outputs for non-ci test script

dljsjr:
This change adds line number output to failure reports from the non-ci `run_driver_tests.py` script. This script doesn't run in CI/CD but can be useful for local development since it can take the `--filter` arg, which lets developers run groups of tests at once with JUnit reports.

Adding the line numbers to the `stdout` report is a QoL improvement for devs to jump to tests quickly; I've even been able to create a problem matcher for the output in VS Code now.

This also adds the line number of the test to the JUnit XML.

The `run_driver_tests_p.py` script that runs in CI/CD is unmodified.

Previous output example:

```console
Unit test failures in /home/dstephen/Developer/Workspaces/SmartThings/git/SmartThingsEdgeDrivers/drivers/SmartThings/zigbee-water-leak-sensor/src/test/test_centralite_water_leak_sensor.lua:
    Configure should configure all necessary attributes
```

Output with these changes example:
```console
Unit test failures in /home/dstephen/Developer/Workspaces/SmartThings/git/SmartThingsEdgeDrivers/drivers/SmartThings/zigbee-water-leak-sensor/src/test/test_centralite_water_leak_sensor.lua:
    Configure should configure all necessary attributes [line 63]
```

### DIFF
--- a/tools/run_driver_tests.py
+++ b/tools/run_driver_tests.py
@@ -73,7 +73,14 @@ def run_tests(verbosity_level, filter, junit, coverage_files):
             if m is not None:
                 test_title = line
                 in_progress_test_name = m.group(1)
-                test_case = junit_xml.TestCase(in_progress_test_name)
+                test_name_regex = re.compile(in_progress_test_name)
+                line_number = None
+                with open(test_file, 'r') as search_file:
+                    for idx, line in enumerate(search_file, 1):
+                        if test_name_regex.search(line) :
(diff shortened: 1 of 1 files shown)


cjswedes (approved):
Seems somewhat useful. It is slightly confusing that the line is the test case not the line that had the expectation that failed, but that would be much harder.

greens (approved):

PR #1156, opened 2024-01-10
Update Zigbee tests for next Lua Libs release

dljsjr:
The next Lua Libs release has a few changes on timing and tolerance for some clusters to ease battery drain, so these test updates will need to be merged in at that time.


cjswedes:
FYI, I hope to release a beta artifact here when we do the hub 0.52.x release to customer beta. We have done this once before, but haven't been super consistent about it.

greens:
> FYI, I hope to release a beta artifact here when we do the hub 0.52.x release to customer beta. We have done this once before, but haven't been super consistent about it.

@dljsjr if that happens we should be able to add a tag to this PR and it'll pull in the artifact to run the tests. Haven't used this feature in a while but it should still work. The tag would be `release-0.52`.

cjswedes:
I just tested this with the 0.52 lua libs we intend to release today, and all tests are passing.

cjswedes (approved):

greens (approved):

PR #1157, opened 2024-01-10
Add fingerprint for Aqara light

ctowns:
Add a fingerprint in order to provide custom embedded profile configs


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -1884,6 +1884,13 @@ matterManufacturer:
     productId: 0x228F
     deviceProfileName: light-color-level-2200K-6500K
 
+#Aqara
+  - id: "Aqara LED Light Bulb T1"
+    deviceLabel: Aqara LED Light Bulb T1
+    vendorId: 0x115F
+    productLabel: Aqara LED Light Bulb T1
```
greens, line 1891: is this field supported? shouldn't this be `productId`?

ctowns: This is a fingerprint for a bridged device, and the `productId` is not required to be provided for bridged devices, so the `productLabel` is used instead. This is the first bridged fingerprint we have. I don't have this particular device, but I have tested with creating a bridged fingerprint for my hue bulbs and these fields do work

greens (approved):

PR #1163, opened 2024-01-15
matter-button: Change category for icon

yhoyoon:
The icon for the RemoteController category is not a button icon. This commit changes the category for the single button profile among the matter-button profiles from RemoteController to Button.

### DIFF
--- a/drivers/SmartThings/matter-button/profiles/button-battery.yml
+++ b/drivers/SmartThings/matter-button/profiles/button-battery.yml
@@ -11,4 +11,4 @@ components:
       - id: refresh
         version: 1
     categories:
-    - name: RemoteController
+    - name: Button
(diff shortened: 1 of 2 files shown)


lelandblue:
Nearly identical profiles for Z-Wave and Zigbee devices, such as [https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/fb87ed9130a12fb687efbbd50095b299657a2381/drivers/SmartThings/zigbee-button/profiles/one-button.yml#L12](url) and [https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/production/drivers/SmartThings/zwave-button/profiles/button-generic.yml](url) with "single button" have a category of  `name: RemoteController`. 

I would recommend that we keep consistency with  all protocols. If this change is accepted then we should update all profiles that would be considered to be "one button".

lelandblue (approved):

tpmanley (approved):
Not all Generic Switches are buttons (they could be light switches or other form factors) but I think the Button category (and therefore icon) is an improvement over the remote controller so I'm fine with making the change.

yhoyoon:
IOTE-3345
IOTE-3474

PR #1168, opened 2024-01-17
Z-Wave Switch: lazy load sub drivers

ctowns:
Adds the ability to lazy load sub drivers. Sub driver that are not currently used by the driver will be "lazy loaded", meaning just a smaller version of the sub driver will be loaded until a device joins that requires the entire sub driver, at which point all the sub driver handlers will be loaded. Given the high number of sub drivers in the zwave-switch driver, this should provide a good amount of memory savings in cases where many of the sub drivers go unused.

Note, this will only work with the latest lua libs changes, so the lazy loading is only activated after an API version check.

### DIFF
--- a/drivers/SmartThings/zwave-switch/src/aeon-smart-strip/init.lua
+++ b/drivers/SmartThings/zwave-switch/src/aeon-smart-strip/init.lua
@@ -40,7 +40,8 @@ local POWER_UNIT_WATT = "W"
 local function can_handle_aeon_smart_strip(opts, driver, device, ...)
   for _, fingerprint in ipairs(AEON_SMART_STRIP_FINGERPRINTS) do
     if device:id_match(fingerprint.mfr, fingerprint.prod, fingerprint.model) then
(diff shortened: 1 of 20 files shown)


greens (approved):

cjswedes (approved):
Approving with the condition that the driver tests were run against the api v9 lua libs we have internally.

ctowns:
Just a testing recap of these changes before merging:

* Completed regression testing with help from QA to test multiple devices to confirm onboarding is successful and devices are functional. Multiple devices were onboarded at once to confirm that multiple sub drivers can be loaded successfully.
* Driver switching was tested and passed successfully
* Unit tests for these changes have been successfully run against lua libs for FW version 50, 51, and 52
* Device is onboarded correctly to a hub running 51 FW and Lua Libs. The lazy loading is correctly not "activated" unless lua libs api version is >= 9, which will only be true for 52 FW for now.

Thank you everyone for your help with this!

PR #1183, opened 2024-01-25
interfree-switch-driver

857831494:
interfree switch driver


857831494:
interfree switch driver

Mariano-Github:
> interfree switch driver

Hi,

They are using a code from a driver from my public repository, to which I added this devices, as a request from community users, for your use and that you can use freely, but using it to certify some commercial products without asking permission seems a bit inappropriate to me.

On the other hand, my driver has custom capabilities and cannot be certified

PR #1195, opened 2024-01-30
[Eve]: Support for Eve Energy Outlet

Timac:
- Add Eve Energy Outlet fingerprint
- Only create the polls (create_poll_schedule and create_poll_report_schedule) if the device supports capabilities.powerConsumptionReport
- When the device is initialized, create child devices with the profile plug-binary for each additional outlet
- Update component_to_endpoint, endpoint_to_component and find_child to support multiple children
- Each time an oulet is turned on or off, store the value in a dictionary
- If one of the outlet is on, we should create the poll to monitor the power consumption
- If all the outlet are off, we should delete the poll

### DIFF
--- a/drivers/SmartThings/matter-switch/src/eve-energy/init.lua
+++ b/drivers/SmartThings/matter-switch/src/eve-energy/init.lua
@@ -251,18 +251,20 @@ end
 -------------------------------------------------------------------------------------
 
 local function device_init(driver, device)
-  if not device:get_field(COMPONENT_TO_ENDPOINT_MAP) and
-      not device:get_field(SWITCH_INITIALIZED) then
-    -- create child devices as needed for multi-switch devices
-    initialize_switch(driver, device)
-  end
-  device:set_component_to_endpoint_fn(component_to_endpoint)
-  device:set_endpoint_to_component_fn(endpoint_to_component)
-  device:set_find_child(find_child)
-  device:subscribe()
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/src/eve-energy/init.lua
```diff
@@ -186,9 +251,14 @@ end
 -------------------------------------------------------------------------------------
 
 local function device_init(driver, device)
-  log.info_with({ hub_logs = true }, "device init")
+  if not device:get_field(COMPONENT_TO_ENDPOINT_MAP) and
```
ctowns, line 254:
```suggestion
if device.network_type == device_lib.NETWORK_TYPE_MATTER then
  if not device:get_field(COMPONENT_TO_ENDPOINT_MAP) and
```

We want this check so that we only run this function for the parent device. You can have this check around the entirety of the init function.

ctowns (changes requested):
Just a small change, but this looks good! The device should arrive for testing shortly and we can confirm everything is working as expected. 

Thank you for your help @Timac !

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
Timac: Change implemented, thanks for the review!

ctowns (approved):

PR #1198, opened 2024-01-30
fix: Add some hardening to the Hue discovery logic.

dljsjr:
The first thing we do is we store the API keys in the datastore right away instead of waiting for a device record to exist. This hardens us against the situation where the driver restarts for any reason while the Hue Bridge device's `added` handler is still executing and is early enough in its execution that we haven't stored the key on the device object yet.

We also address a regression from the automatic add/delete changes, which can lead to the creation of duplicate device records for bulbs in some specific circumstances.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -715,7 +715,15 @@ local function do_bridge_network_init(driver, bridge_device, bridge_url, api_key
 
       local bridge_api = bridge_device:get_field(Fields.BRIDGE_API)
       cosock.spawn(function()
-        Discovery.scan_bridge_and_update_devices(driver, bridge_device:get_field(Fields.BRIDGE_ID))
+        -- We don't want to do a scan if we're already in a discovery loop,
+        -- because the event source connection will open if a bridge is discovered
+        -- and we'll effectively be scanning twice.
+        -- Two scans that find the same device close together can emit events close enough
+        -- together that the dedupe logic at the cloud layer will get bypassed and lead to
+        -- duplicate device records.
+        if not Discovery.discovery_active then
+          Discovery.scan_bridge_and_update_devices(driver, bridge_device:get_field(Fields.BRIDGE_ID))
+        end
         local child_device_map = {}
         local children = bridge_device:get_child_list()
         bridge_device.log.debug(string.format("Scanning connectivity of %s child devices", #children))
@@ -1417,6 +1425,8 @@ local function remove(driver, device)
       event_source:close()
       device:set_field(Fields.EVENT_SOURCE, nil)
     end
+
+    Discovery.api_keys[device.device_network_id] = nil
   end
 end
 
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -712,7 +712,15 @@ local function do_bridge_network_init(driver, bridge_device, bridge_url, api_key
 
       local bridge_api = bridge_device:get_field(Fields.BRIDGE_API)
       cosock.spawn(function()
-        Discovery.scan_bridge_and_update_devices(driver, bridge_device:get_field(Fields.BRIDGE_ID))
+        -- We don't want to do a scan if we're already in a discovery loop,
+        -- because the even tsource connection will open if a bridge is discovered
```
cjswedes, line 716: nit:
```suggestion
        -- because the event source connection will open if a bridge is discovered
```

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -1554,6 +1564,39 @@ if hue.datastore["dni_to_device_id"] == nil then
...
+  local datastore_dirty = false
+  for k, v in pairs(Discovery.api_keys) do
+    hue.datastore.api_keys[k] = v
+    datastore_dirty = true
+  end
+
+  if datastore_dirty then
+    hue.datastore:save()
+  end
+end
```
cjswedes, line 1582: What is the point of the `datastore_dirty` flag? It doesn't seem to be doing anything here. A follow up is that I dont fully understand what is trying to be accomplished by this section that runs prior to the driver being run, and prior to the metatable being set on the Discovery.api_keys table. Is it the goal to copy any previously Discovered api keys into the datastore? Will `Discovery.api_keys` ever have an entry when this code is run?

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -1554,6 +1564,39 @@ if hue.datastore["dni_to_device_id"] == nil then
...
+
+Discovery.api_keys = setmetatable({}, {
```
cjswedes, line 1584: Metatables used like this is what makes people hate lua...It is counterintuitive to reach into a different component's fields, and modify its functionality from outside the component. I see why it was done here, and you're comment in disco.lua helps sort things out, but its pretty hacky. Ill approve with this as is though, since i dont think there is a way to do it  without many changes to disco.lua.

dljsjr: Oh I hate this with a passion. But it's a minimum-entropy fix so that I can get back to finishing the integration tests so that I can have regression coverage so that I can refactor the driver.

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: It's overly-defensive coding after all the odd things I've seen happen.

cjswedes: > Will Discovery.api_keys ever have an entry when this code is run?

Prior to the run loop of the driver, this is impossible, so I would suggest removing it all tbh.

cjswedes (approved):

drivers/SmartThings/philips-hue/src/init.lua
dljsjr: I was thinking about instances in the past where the Discovery loop could be loaded multiple times simultaneously, but thinking about it harder, the package path cache would prevent that from being problematic here anyway. I've removed it.

PR #1199, opened 2024-01-31
Fix run_driver_tests.py UnicodeDecodeError

liuyd96:
I accidentally discovered that smartsense test run failed due to decode error.
The root cause is that our script use wrong encoding to open file.

This test script is "ISO-8859" encoding which includes "0xBD" which UTF-8 doesn't support.
```
❯ file test_smartsense_motion_sensor.lua
test_smartsense_motion_sensor.lua: ISO-8859 text
```

By default python3 uses UTF-8 to open file. That is we our script can't run smartsense test.

Import chardet lib to detect encoding format automatically to fix this issue.

### DIFF
--- a/drivers/SmartThings/zigbee-motion-sensor/src/test/test_smartsense_motion_sensor.lua
+++ b/drivers/SmartThings/zigbee-motion-sensor/src/test/test_smartsense_motion_sensor.lua
@@ -71,7 +71,7 @@ test.register_coroutine_test(
   function()
     test.socket.zigbee:__queue_receive({
       mock_device.id,
-      build_motion_status_message(mock_device, "|")
+      build_motion_status_message(mock_device, "\x7C")
     })
     test.socket.capability:__expect_send(mock_device:generate_test_message("main", capabilities.battery.battery(100)))
     test.socket.capability:__expect_send(mock_device:generate_test_message("main", capabilities.motionSensor.motion.inactive()))
@@ -85,7 +85,7 @@ test.register_coroutine_test(
   function()
     test.socket.zigbee:__queue_receive({
       mock_device.id,
-      build_motion_status_message(mock_device, "~")
+      build_motion_status_message(mock_device, "\x7E")
     })
     test.socket.capability:__expect_send(mock_device:generate_test_message("main", capabilities.battery.battery(100)))
     test.socket.capability:__expect_send(mock_device:generate_test_message("main", capabilities.motionSensor.motion.active()))
@@ -99,7 +99,7 @@ test.register_coroutine_test(
   function()
     test.socket.zigbee:__queue_receive({
       mock_device.id,
-      build_motion_status_message(mock_device, "X")
+      build_motion_status_message(mock_device, "\x58")
     })
     test.socket.capability:__expect_send(mock_device:generate_test_message("main", capabilities.battery.battery(70)))
     test.socket.capability:__expect_send(mock_device:generate_test_message("main", capabilities.motionSensor.motion.inactive()))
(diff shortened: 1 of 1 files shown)


liuyd96:
Before
```
❯ ./run_driver_tests.py -f "smartsense"
############################################################################################################################################
## Running tests from /home/liuyd/SmartThingsEdgeDrivers/drivers/SmartThings/zigbee-motion-sensor/src/test/test_smartsense_motion_sensor.lua
Traceback (most recent call last):
  File "./run_driver_tests.py", line 176, in <module>
    run_tests(verbosity_level, args.filter, args.junit, args.coverage)
  File "./run_driver_tests.py", line 79, in run_tests
    for idx, line in enumerate(search_file, 1):
  File "/usr/lib/python3.8/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbd in position 5391: invalid start byte
```

Now (Ubuntu):
```

❯ ./run_driver_tests.py -f "smartsense"
############################################################################################################################################
## Running tests from /home/liuyd/SmartThingsEdgeDrivers/drivers/SmartThings/zigbee-motion-sensor/src/test/test_smartsense_motion_sensor.lua
Passed 8 of 8 tests
############################################################################################################################################
##############################################################################################################################
## Running tests from /home/liuyd/SmartThingsEdgeDrivers/drivers/SmartThings/zigbee-contact/src/test/test_smartsense_multi.lua
Passed 14 of 14 tests
##############################################################################################################################
##############################
Total unit tests passes: 22/22
##############################
```

liuyd96:
> Perhaps we should just define the values using their hex codes? As in, use `"\xBD"` and `"\xBF"`. I just don't think we need the extra complexity in the test driver.

Hi, @gharveymn 

Sorry for the late reply.

Yes. I agree with you ~
The reason I'm editing `run_driver_tests.py` instead of test case file is due to this commit
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/1d67522ed9fec37c43c3f0bf38b623355fe922a3

This commit convert "\xBD" and "\xBF" into real char. 
I don't know the background so I choose to modify `run_driver_tests.py`.

Thanks.

gharveymn:
@liuyd96 Ah, I see. It looks like those accidentally got converted back to unicode as part of a larger reversion.

@greens Is that correct? Can we switch all of those constants back to hex form?

ctowns:
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/1d67522ed9fec37c43c3f0bf38b623355fe922a3 reverts https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/1d4a18e5e2eba73c5ca1907ae81ac91468c5bfed. 1d4a18e5e2 made changes to synthetic events and looks like it updated some tests as well, namely it updated the smartsense test to use the hex form of those values. The hex changes were then reverted when the original commit was reverted in 1d67522. You can see the original conversion to hex values here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/1d4a18e5e2eba73c5ca1907ae81ac91468c5bfed#diff-6bc32bf963b2e92af3b2f0943cf3c459d8fb6630853f5d8b953ab5fd9a4e03d5R74

So, I think we can fix this by just changing those payloads to using the hex values rather than the ascii characters because I think that was originally intended but got accidentally removed in the revert commit.

Thanks for looking into this @liuyd96  and @gharveymn !

gharveymn (approved):

PR #1201, opened 2024-02-01
Fix the battery check in matter-thermostat, look for feature map

samadDotDev:
This was just looking for the presence of the power source cluster whereas the correct check would be to look for the battery feature map bit in that cluster.


ctowns (approved):
Great find! Looks like this was missing in the matter-button driver too, I will put up a PR for that

PR #1205, opened 2024-02-07
Matter Switch: add handling for multi-On/Off Light Switch device types

ctowns:
The On/Off Light Switch type supports the On/Off cluster as client, however some deviecs still report the On/Off light switch but implement On/Off cluster as server or both. There was previously a check to handle this case, but it did not account for multi-devices, so this expands the check to handle multi-devices that fall in this category as well.


lelandblue (approved):
This PR was tested with SRPOL, the device parent device joined to the switch-binary profile and so did the child. 

Note: The child's name will join as "Matter Thing 2" but will be functional.

greens (approved):

PR #1206, opened 2024-02-07
Hotfix: Revert auto-add/auto-delete on 0.47 hubs

dljsjr:
This change puts an API Version guard around the call that performs the auto-add/auto-delete scan at startup, which is causing a crash on 0.47 TV's due to an issue with looking up non-existent UUID's.

We believe the root of the issue is that the non-existent UUID lookup is related to trying to look up the Hub UUID as the parent of the bridge. But because this is a pretty gnarly issue we're going to roll back completely while we're confirming our suspicion.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -37,6 +37,8 @@ local utils = require "utils"
 local syncCapabilityId = "samsungim.hueSyncMode"
 local hueSyncMode = capabilities[syncCapabilityId]
 
+local api_version = require("version").api
+
 local StrayDeviceMessageTypes = {
   FoundBridge = "FOUND_BRIDGE",
   NewStrayLight = "NEW_STRAY_LIGHT",
@@ -712,7 +714,12 @@ local function do_bridge_network_init(driver, bridge_device, bridge_url, api_key
 
       local bridge_api = bridge_device:get_field(Fields.BRIDGE_API)
       cosock.spawn(function()
-        Discovery.scan_bridge_and_update_devices(driver, bridge_device:get_field(Fields.BRIDGE_ID))
+        -- auto-add/auto-delete is only supported with newer lua libs. We do a conditional
+        -- check in the delete routine, but we try to avoid doing it at all here with an
+        -- additional defensive check against the Lua Libs API version.
+        if api_version >= 9 then
+          Discovery.scan_bridge_and_update_devices(driver, bridge_device:get_field(Fields.BRIDGE_ID))
+        end
         local child_device_map = {}
         local children = bridge_device:get_child_list()
         bridge_device.log.debug(string.format("Scanning connectivity of %s child devices", #children))


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -712,7 +714,12 @@ local function do_bridge_network_init(driver, bridge_device, bridge_url, api_key
 
       local bridge_api = bridge_device:get_field(Fields.BRIDGE_API)
       cosock.spawn(function()
-        Discovery.scan_bridge_and_update_devices(driver, bridge_device:get_field(Fields.BRIDGE_ID))
+        -- auto-add/auto-delete is only supported with newer lua libs. We do a conditional
+        -- check in the delete routine, but we try to avoid doing it at all here with an
+        -- additional defensive check against the Lua Libs API version.
+        if api_version >= 9 then
```
dljsjr, line 720: This is an optimistic mitigation that tries to preserve the feature for 0.52 hubs. This condition *will* be false on 0.47 Tizen hubs.

However, if we're concerned that this is dodgy for any reason, I'm fully in favor of just nuking this function call entirely.

cjswedes (approved):

ctowns (approved):

FreeMasen (approved):

drivers/SmartThings/philips-hue/src/init.lua
elbe0046: Just posting this as me trying to follow along & confirm things, nothing actionable here.

I see this version coming from here: https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/c0a00cd57b3e5c4f6c6b4333ba1d20dd38df9ee2/lua_libs/version.lua#L3

If I look back at 0.47.x, here is what we'd have for v2 / v3: https://github.com/PhysicalGraph/hub-os/blob/3.27.11/meta-smartthings/recipes-smartthings/edge-driver-lua-libs/edge-driver-lua-libs_0.47.bb

And we see that archive here: https://smartthings.jfrog.io/ui/native/edge-driver-libs/release-0.47/

Which contains `version.lua`:
```lua
local version = {      
  rpc = rpc_version(),      
  api = 4,      
}      
      
return version 
```

elbe0046: @dljsjr I know that CELL & TV are on lower API versions since @cjswedes just bumped this recently: https://github.ecodesamsung.com/iot-hub/scripting-engine/commit/49eb1132bea4bb102201786029a0cb84fa735ed2

But how can I tell for sure which archive artifact those platforms have?

elbe0046:
@dljsjr what about https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/4ddaa3a4d65ba87b68691e3d368c7640afef90b8/drivers/SmartThings/philips-hue/src/disco.lua#L137, do we not have concerns with that usage?  I'm still trying to get my bearings on much of this

elbe0046 (approved):
I see the `true` `do_delete` arg now, all makes sense to me and LGTM

dljsjr:
> @dljsjr what about
> 
> https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/4ddaa3a4d65ba87b68691e3d368c7640afef90b8/drivers/SmartThings/philips-hue/src/disco.lua#L137
> 
> , do we not have concerns with that usage? I'm still trying to get my bearings on much of this
> edit: e.g. is an alternative version-gating this particular block?
> 
> https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/4ddaa3a4d65ba87b68691e3d368c7640afef90b8/drivers/SmartThings/philips-hue/src/disco.lua#L254-L266
> 
> Even if 'no', looking to understand

That is an alternative, yes. Or rather I would qualify it as additional defensive coding. The call you linked:

```
    HueDiscovery.search_bridge_for_supported_devices(driver, bridge_id, HueDiscovery.disco_api_instances[bridge_id],
      function(hue_driver, svc_info, device_info)
        discovered_device_callback(hue_driver, bridge_id, svc_info, device_info, known_identifier_to_device_map)
      end,
      "[Discovery: " ..
      (known_bridge_device.label or bridge_id or known_bridge_device.id or "unknown bridge") .. " bridge scan]"
    )
```

Does not pass an argument in the `do_delete` position, and never will. But it doesn't mean we can't be extra defensive.

In the current version of the PR, the call we're gating is the only call where that results in a truthy value in the `do_delete` position

elbe0046:
> Does not pass an argument in the do_delete position, and never will. But it doesn't mean we can't be extra defensive.

> In the current version of the PR, the call we're gating is the only call where that results in a truthy value in the do_delete position

Yeah I see it now, thanks.  I would keep things as you have them.

kianooshkarami (approved):

PR #1210, opened 2024-02-07
New Device (Matter Switch) Zemismart ZM606-6

lelandblue:
This is a WWST Certification submission for the Zemismart ZM606-6. 

This will be a parent child Matter Device.

The switch-binary profile was selected to support this device.

We know that this device according to the DCL is 0x0103. Previous devices from the MFG have supported the On/Off Cluster BOTH as a Server and as a Client, so we are assuming that the switch-binary profile is best for this device. Please review.


ctowns (approved):

lelandblue:
Partner has closed the WWST Certification submission, closing this PR

PR #1211, opened 2024-02-07
Skip processing Hue Bridges as deletion candidates + some static analysis cleanup

dljsjr:
This removes the API version guard on the auto add/delete scan during
event stream connection and instead skips to the next element in the
iterator if it detects that the current iteration target is a device
record that represents a bridge.

We do this because we *do* want the auto-add/auto-delete routine to
run on older firmware, as the delete function does feature detection
on the lua libs and uses `device:offline()` as a fallback. Plus, auto
add will work on all firmware versions, so this is the better UX all
around.

We also address some static analysis lints from the Lua Language Server
VS Code plugin.

Most of these are just refining/augmenting gradual typing to quiet the
IDE warnings, but there were a few places that we were missing nil
checks or otherwise had other potential bugs that got addressed here.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/disco.lua
+++ b/drivers/SmartThings/philips-hue/src/disco.lua
@@ -253,14 +253,17 @@ function HueDiscovery.search_bridge_for_supported_devices(driver, bridge_id, api
 
   if do_delete then
     for _, device in ipairs(driver:get_devices()) do ---@cast device HueDevice
+      -- We're only interested in processing child/non-bridge devices here.
+      if utils.is_bridge(driver, device) then goto continue end
       local not_known_to_bridge = device_is_joined_to_bridge[device:get_field(Fields.HUE_DEVICE_ID) or ""]
       local parent_device_id = device.parent_device_id or device:get_field(Fields.PARENT_DEVICE_ID) or ""
       local parent_bridge_device = driver:get_device_info(parent_device_id)
       local is_child_of_bridge = parent_bridge_device and (parent_bridge_device:get_field(Fields.BRIDGE_ID) == bridge_id)
-      if is_child_of_bridge and not not_known_to_bridge then
+      if parent_bridge_device and is_child_of_bridge and not not_known_to_bridge then
         device.log.info(string.format("Device is no longer joined to Hue Bridge %q, deleting", parent_bridge_device.label))
         driver:do_hue_light_delete(device)
       end
+      ::continue::
     end
   end
 end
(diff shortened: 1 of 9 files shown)


dljsjr:
I tested this on a hub running the older firmware that could have problems with the delete logic and this seems to be a sufficient mitigation. Merging this PR would let us restore the auto-add functionality on all hubs, and the marking offline of deleted lights for hubs that don't support auto-delete.

FreeMasen (approved):

drivers/SmartThings/philips-hue/src/disco.lua
```diff
@@ -253,14 +253,17 @@ function HueDiscovery.search_bridge_for_supported_devices(driver, bridge_id, api
...
       local not_known_to_bridge = device_is_joined_to_bridge[device:get_field(Fields.HUE_DEVICE_ID) or ""]
       local parent_device_id = device.parent_device_id or device:get_field(Fields.PARENT_DEVICE_ID) or ""
       local parent_bridge_device = driver:get_device_info(parent_device_id)
       local is_child_of_bridge = parent_bridge_device and (parent_bridge_device:get_field(Fields.BRIDGE_ID) == bridge_id)
-      if is_child_of_bridge and not not_known_to_bridge then
+      if parent_bridge_device and is_child_of_bridge and not not_known_to_bridge then
         device.log.info(string.format("Device is no longer joined to Hue Bridge %q, deleting", parent_bridge_device.label))
         driver:do_hue_light_delete(device)
       end
+      ::continue::
```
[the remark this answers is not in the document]
dljsjr: Line 257

drivers/SmartThings/philips-hue/src/lunchbox/sse/eventsource.lua
```diff
@@ -156,8 +156,8 @@ local valid_fields = util.read_only {
 -- h/t to github.com/FreeMasen for the suggestions on the efficient implementation of this
 local function find_line_endings(chunk)
   local r_idx, n_idx = string.find(chunk, "[\r\n]+")
-  if r_idx == n_idx then
-    -- 1 character
+  if r_idx == nil or r_idx == n_idx then
+    -- 1 character or no match
```
cjswedes, line 160: When is the no match happening?

cjswedes (approved):

drivers/SmartThings/philips-hue/src/lunchbox/sse/eventsource.lua
dljsjr: For a well-formed message, it shouldn't, but for any kind of adversarial input, someone could hand off a string that doesn't use CRLF.

For the specific case of the EventSource it *shouldn't* happen unless the payload gets corrupted.

But from the perspective of type checking, it's possible for `find` to return `nil` on no match, so we handle it.

PR #1212, opened 2024-02-07
Fix the repair task for unlinked bridges

dljsjr:
Sometimes a data loss event can occur that prevents the Bridge
Application Key from being persisted in the Edge Driver Datastore.

This coroutine broke at some point during a refactor, unsure when, but
this commit restores the functionality that would let a user re-link
their bridge in the event that they're aware that this is what has
happened.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -246,6 +246,26 @@ local function migrate_bridge(driver, device)
   )
 end
 
+local function update_bridge_fields_from_info(driver, bridge_info, bridge_device)
+  local bridge_ip = bridge_info.ip
+  local device_bridge_id = bridge_device.device_network_id
+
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -1003,6 +1004,26 @@ init_bridge = function(driver, device)
   local api_key = device:get_field(HueApi.APPLICATION_KEY_HEADER)
   local bridge_url = "https://" .. ip
 
+  if not api_key then
+    if not Discovery.api_keys[device_bridge_id] then
+      log.error_with({ hub_logs = true },
```
greens, line 1009: these go to _both_ the hub and the driver logs, right?

cjswedes: yes.

lelandblue:
@dljsjr  No Rush - Do we want to close this PR?

dljsjr:
Yeah, we should. This is no longer needed with the refactors done for the new device types.

PR #1216, opened 2024-02-13
BUG2-1182-fix: hide non functional attribute update 

hdlee27:
Some blind drivers have exposed unnecessary information to history. (bug2-1182)
modify the aqara and hansem blind drivers not to show some events
(for supportedWindowShadeCommands, supportedInitializedState)

### DIFF
--- a/drivers/SmartThings/matter-window-covering/src/init.lua
+++ b/drivers/SmartThings/matter-window-covering/src/init.lua
@@ -74,7 +74,7 @@ end
 
 local function device_added(driver, device)
   device:emit_event(
-    capabilities.windowShade.supportedWindowShadeCommands({"open", "close", "pause"})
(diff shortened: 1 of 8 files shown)


tpmanley:
Could you fix https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/8639f88e40e9245cbca67a8e21a82cff0d7488f7/drivers/SmartThings/matter-window-covering/src/init.lua#L77 as well which has the same issue?

hdlee27:
> Could you fix
> 
> https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/8639f88e40e9245cbca67a8e21a82cff0d7488f7/drivers/SmartThings/matter-window-covering/src/init.lua#L77
> 
> as well which has the same issue?

@tpmanley I added an additional commit for matter window covering. Please review it again

tpmanley (approved):

greens (approved):

PR #1220, opened 2024-02-13
Matter Switch: Read color temp and level limits on add

nickolas-deboom:
When device is added, read the color temperature and level cluster min/max and set capability constraint based on the limits.

### DIFF
--- a/drivers/SmartThings/matter-switch/profiles/light-color-level.yml
+++ b/drivers/SmartThings/matter-switch/profiles/light-color-level.yml
@@ -20,6 +20,3 @@ components:
     version: 1
   categories:
   - name: Light
-metadata:
-  mnmn: SmartThingsEdge
-  vid: generic-rgbw-color-bulb-2580K-7050K
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -13,6 +13,7 @@
 -- limitations under the License.
 
 local capabilities = require "st.capabilities"
+local im = require "st.matter.interaction_model"
 local log = require "log"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
@@ -29,6 +30,9 @@ local COLOR_TEMPERATURE_KELVIN_MAX = 30000
 local COLOR_TEMPERATURE_KELVIN_MIN = 1
 local COLOR_TEMPERATURE_MIRED_MAX = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN
 local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX
+local SWITCH_LEVEL_MAX = 100
+local SWITCH_LEVEL_LIGHTING_MIN = 1
+local SWITCH_LEVEL_MIN = 0
 
 local SWITCH_INITIALIZED = "__switch_intialized"
 -- COMPONENT_TO_ENDPOINT_MAP is here only to perserve the endpoint mapping for
@@ -37,6 +41,13 @@ local SWITCH_INITIALIZED = "__switch_intialized"
 -- in the device table for devices that joined prior to this transition, and it
 -- will not be set for new devices.
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
+local BOUNDS_CHECKED = "__bounds_checked"
+local COLOR_TEMP_BOUND_RECEIVED = "__colorTemp_bound_received"
+local COLOR_TEMP_MIN = "__color_temp_min"
+local COLOR_TEMP_MAX = "__color_temp_max"
+local LEVEL_BOUND_RECEIVED = "__level_bound_received"
+local LEVEL_MIN = "__level_min"
+local LEVEL_MAX = "__level_max"
 local AGGREGATOR_DEVICE_TYPE_ID = 0x000E
 local ON_OFF_LIGHT_DEVICE_TYPE_ID = 0x0100
 local DIMMABLE_LIGHT_DEVICE_TYPE_ID = 0x0101
@@ -54,8 +65,27 @@ local device_type_profile_map = {
 }
 local detect_matter_thing
 
+local function get_field_for_endpoint(device, field, endpoint)
+  return device:get_field(string.format("%s_%d", field, endpoint))
+end
+
+local function set_field_for_endpoint(device, field, endpoint, value, additional_params)
+  device:set_field(string.format("%s_%d", field, endpoint), value, additional_params)
+end
+
 local function convert_huesat_st_to_matter(val)
   return math.floor((val * 0xFE) / 100.0 + 0.5)
+end
+
+local function mired_to_kelvin(value)
+  local CONVERSION_CONSTANT = 1000000
+  if value == 0 then -- shouldn't happen, but has
+    value = 1
+    log.warn(string.format("Received a color temperature of 0 mireds. Using a color temperature of 1 mired to avoid divide by zero"))
+  end
+  -- we divide inside the rounding and multiply outside of it because we expect these
+  -- bounds to be multiples of 100
+  return utils.round((CONVERSION_CONSTANT / value) / 100) * 100
 end
 
 --- component_to_endpoint helper function to handle situations where
@@ -186,6 +216,22 @@ local function device_init(driver, device)
     device:set_endpoint_to_component_fn(endpoint_to_component)
     device:set_find_child(find_child)
     device:subscribe()
+
+    if not device:get_field(BOUNDS_CHECKED) then
+      local limit_read = im.InteractionRequest(im.InteractionRequest.RequestType.READ, {})
+      if device:supports_capability(capabilities.colorTemperature) then
+        limit_read:merge(clusters.ColorControl.attributes.ColorTempPhysicalMinMireds:read())
+        limit_read:merge(clusters.ColorControl.attributes.ColorTempPhysicalMaxMireds:read())
+      end
+      if device:supports_capability(capabilities.switchLevel) then
+        limit_read:merge(clusters.LevelControl.attributes.MinLevel:read())
+        limit_read:merge(clusters.LevelControl.attributes.MaxLevel:read())
+      end
+      if #limit_read.info_blocks ~= 0 then
+        device:send(limit_read)
+      end
+      device:set_field(BOUNDS_CHECKED, true)
+    end
   end
 end
 
@@ -335,6 +381,72 @@ local function temp_attr_handler(driver, device, ib, response)
   end
 end
 
+local mired_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value == nil then
+      return
+    end
+    local color_temp_mireds = ib.data.value
+    if (color_temp_mireds < COLOR_TEMPERATURE_MIRED_MIN or color_temp_mireds > COLOR_TEMPERATURE_MIRED_MAX) then
+      device.log.warn_with({hub_logs = true}, string.format("Device reported a min or max color temperature %d mired outside of supported capability range", color_temp_mireds))
+      if (minOrMax == COLOR_TEMP_MIN) then
+        color_temp_mireds = COLOR_TEMPERATURE_MIRED_MIN
+      else
+        color_temp_mireds = COLOR_TEMPERATURE_MIRED_MAX
+      end
+    end
+    local temp_in_kelvin = mired_to_kelvin(color_temp_mireds)
+    set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp_in_kelvin)
+    local min = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MIN, ib.endpoint_id)
+    local max = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MAX, ib.endpoint_id)
+    if min ~= nil and max ~= nil then
+      if min < max then
+        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = min, maximum = max} }))
+        set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MAX, ib.endpoint_id, nil)
+        set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MIN, ib.endpoint_id, nil)
+      else
+        device.log.warn_with({hub_logs = true}, string.format("Device reported a min color temperature %d mired that is not lower than the reported max color temperature %d mired", min, max))
+      end
+    end
+  end
+end
+
+local level_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value == nil then
+      return
+    end
+    local current_level = ib.data.value
+    local lighting_endpoints = device:get_endpoints(clusters.LevelControl.ID, {feature_bitmap = clusters.LevelControl.FeatureMap.LIGHTING})
+    if ((#lighting_endpoints ~= 0 and current_level < SWITCH_LEVEL_LIGHTING_MIN) or current_level < SWITCH_LEVEL_MIN or current_level > SWITCH_LEVEL_MAX) then
+      device.log.warn_with({hub_logs = true}, string.format("Device reported a min or max switch level %d mired outside of supported capability range", ib.data.value))
+      if (minOrMax == LEVEL_MIN) then
+        if (#lighting_endpoints ~= 0) then
+          current_level = SWITCH_LEVEL_LIGHTING_MIN
+        else
+          current_level = SWITCH_LEVEL_MIN
+        end
+      else
+        current_level = SWITCH_LEVEL_MAX
+      end
+    end
+    -- Convert level from given range of 0-254 to range of 0-100.
+    local level = utils.round(current_level / 254.0 * 100)
+    set_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..minOrMax, ib.endpoint_id, level)
+    local min = get_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..LEVEL_MIN, ib.endpoint_id)
+    local max = get_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..LEVEL_MAX, ib.endpoint_id)
+    if min ~= nil and max ~= nil then
+      if min < max then
+        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.switchLevel.levelRange({ value = {minimum = min, maximum = max} }))
+        set_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..LEVEL_MAX, ib.endpoint_id, nil)
+        set_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..LEVEL_MIN, ib.endpoint_id, nil)
+      else
+        device.log.warn_with({hub_logs = true}, string.format("Device reported a min level value %d that is not lower than the reported max level value %d", min, max))
+      end
+    end
+  end
+end
+
 local color_utils = require "color_utils"
 
 local function x_attr_handler(driver, device, ib, response)
@@ -411,7 +523,9 @@ local matter_driver_template = {
         [clusters.OnOff.attributes.OnOff.ID] = on_off_attr_handler,
       },
       [clusters.LevelControl.ID] = {
-        [clusters.LevelControl.attributes.CurrentLevel.ID] = level_attr_handler
+        [clusters.LevelControl.attributes.CurrentLevel.ID] = level_attr_handler,
+        [clusters.LevelControl.attributes.MaxLevel.ID] = level_bounds_handler_factory(LEVEL_MAX),
+        [clusters.LevelControl.attributes.MinLevel.ID] = level_bounds_handler_factory(LEVEL_MIN),
       },
       [clusters.ColorControl.ID] = {
         [clusters.ColorControl.attributes.CurrentHue.ID] = hue_attr_handler,
@@ -420,6 +534,8 @@ local matter_driver_template = {
         [clusters.ColorControl.attributes.CurrentX.ID] = x_attr_handler,
         [clusters.ColorControl.attributes.CurrentY.ID] = y_attr_handler,
         [clusters.ColorControl.attributes.ColorCapabilities.ID] = color_cap_attr_handler,
+        [clusters.ColorControl.attributes.ColorTempPhysicalMaxMireds.ID] = mired_bounds_handler_factory(COLOR_TEMP_MIN), -- max mireds = min kelvin
+        [clusters.ColorControl.attributes.ColorTempPhysicalMinMireds.ID] = mired_bounds_handler_factory(COLOR_TEMP_MAX), -- min mireds = max kelvin
       },
       [clusters.IlluminanceMeasurement.ID] = {
         [clusters.IlluminanceMeasurement.attributes.MeasuredValue.ID] = illuminance_attr_handler
--- a/drivers/SmartThings/matter-switch/src/test/test_matter_switch.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_matter_switch.lua
@@ -47,7 +47,7 @@ local mock_device = test.mock_device.build_test_matter_device({
           feature_map = 0, --u32 bitmap
         },
         {cluster_id = clusters.ColorControl.ID, cluster_type = "BOTH", feature_map = 31},
-        {cluster_id = clusters.LevelControl.ID, cluster_type = "SERVER"}
+        {cluster_id = clusters.LevelControl.ID, cluster_type = "SERVER", feature_map = 2}
       },
       device_types = {
         {device_type_id = 0x0100, device_type_revision = 1} -- On/Off Light
@@ -77,17 +77,17 @@ local mock_device_no_hue_sat = test.mock_device.build_test_matter_device({
   }
 })
 
+local cluster_subscribe_list = {
+  clusters.OnOff.attributes.OnOff,
+  clusters.LevelControl.attributes.CurrentLevel,
+  clusters.ColorControl.attributes.CurrentHue,
+  clusters.ColorControl.attributes.CurrentSaturation,
+  clusters.ColorControl.attributes.CurrentX,
+  clusters.ColorControl.attributes.CurrentY,
+  clusters.ColorControl.attributes.ColorTemperatureMireds,
+}
 
 local function test_init()
-  local cluster_subscribe_list = {
-    clusters.OnOff.attributes.OnOff,
-    clusters.LevelControl.attributes.CurrentLevel,
-    clusters.ColorControl.attributes.CurrentHue,
-    clusters.ColorControl.attributes.CurrentSaturation,
-    clusters.ColorControl.attributes.CurrentX,
-    clusters.ColorControl.attributes.CurrentY,
-    clusters.ColorControl.attributes.ColorTemperatureMireds,
-  }
   test.socket.matter:__set_channel_ordering("relaxed")
   local subscribe_request = cluster_subscribe_list[1]:subscribe(mock_device)
   for i, cluster in ipairs(cluster_subscribe_list) do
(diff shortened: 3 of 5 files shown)


nickolas-deboom:
This is a work in progress that shouldn't be merged until we have the chance to test with actual devices, hopefully later this week.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -262,6 +267,28 @@ local function level_attr_handler(driver, device, ib, response)
   end
 end
 
+local function level_min_attr_handler(driver, device, ib, response)
+  if ib.data.value ~= nil then
+    local level = math.floor((ib.data.value / 254.0 * 100) + 0.5)
+    device:set_field(LEVEL_MIN, level)
+    local level_max = device:get_field(LEVEL_MAX)
+    if level_max then
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.switchLevel.levelRange({minimum = level, maximum = level_max}))
```
tpmanley, line 276: Once we emit the event I think we could probably remove the persisted fields since we only need to temporarily remember one while we wait for the other.

nickolas-deboom: Made this change in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1220/commits/c7bfc422f8876abcd8f5cbe746aeef3c5aabfbbc

drivers/SmartThings/matter-switch/src/test/test_matter_switch.lua
```diff
@@ -548,4 +548,70 @@ test.register_message_test(
...
 
+test.register_message_test(
+  "Min and max color temperature attributes set capability constraint",
+  {
+    {
+      channel = "matter",
+      direction = "receive",
+      message = {
+        mock_device.id,
+        clusters.ColorControl.attributes.ColorTempPhysicalMinMireds:build_test_report_data(mock_device, 1, 500)
```
tpmanley, line 559: The min value should be less than the max value. I suggest using 153 and 555 which should translate to 6500 K and 1800 K respectively.

tpmanley:
I'd suggest trying to match https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1248 more closely. It has one handler for both min/max mireds and I think handles the conversion to K a little bit better. Plus it's nice to have consistency between the drivers.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -262,6 +267,28 @@ local function level_attr_handler(driver, device, ib, response)
   end
 end
 
+local function level_min_attr_handler(driver, device, ib, response)
+  if ib.data.value ~= nil then
+    local level = math.floor((ib.data.value / 254.0 * 100) + 0.5)
```
wkhenon, line 272: Could be helpful to add a description for the operations done to obtain `level`

greens (approved):

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -321,56 +309,42 @@ local function temp_attr_handler(driver, device, ib, response)
...
-      device:set_field(COLOR_TEMP_MIN, nil)
-      device:set_field(COLOR_TEMP_MAX, nil)
+local level_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value ~= nil then
+      local level = math.floor((ib.data.value / 254.0 * 100) + 0.5)
+      device:set_field(LEVEL_BOUND_RECEIVED..minOrMax, level)
+      local min = device:get_field(LEVEL_BOUND_RECEIVED..LEVEL_MIN)
+      local max = device:get_field(LEVEL_BOUND_RECEIVED..LEVEL_MAX)
+      if min ~= nil and max ~= nil and min < max then
```
gharveymn, line 339: If we get something outside of our expected bounds we should probably log it. Also, is `min == max` a valid scenario?

drivers/SmartThings/matter-switch/profiles/light-color-level.yml
```diff
@@ -15,7 +15,4 @@ components:
   - id: refresh
     version: 1
   categories:
-  - name: Light
-metadata:
-  mnmn: SmartThingsEdge
-  vid: generic-rgbw-color-bulb-2580K-7050K
+  - name: Light
```
gharveymn, line 18: Missing newline to end the file

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -294,6 +309,42 @@ local function temp_attr_handler(driver, device, ib, response)
...
+        device:set_field(COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MIN, nil)
+      end
+    end
+  end
+end
+
+local level_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value ~= nil then
+      local level = math.floor((ib.data.value / 254.0 * 100) + 0.5)
```
gharveymn, line 335: Nit: Readability

```suggestion
      local level = utils.round(ib.data.value / 254.0 * 100)
```

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -294,6 +309,42 @@ local function temp_attr_handler(driver, device, ib, response)
...
+        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = min, maximum = max} }))
+        device:set_field(COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MAX, nil)
+        device:set_field(COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MIN, nil)
+      end
+    end
+  end
+end
+
+local level_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
```
gharveymn, line 333: This is fine, but I would argue that you can still be DRY by putting everything below the the call to `set_field` into a common function, and keeping the handlers for min and max split, which would be more readable (IMO).

drivers/SmartThings/matter-switch/src/init.lua
greens: `min == max` shouldn't be valid, no

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -339,9 +390,23 @@ local function info_changed(driver, device, event, args)
   end
 end
 
+local function device_added(driver, device, event, args)
+  local limit_read = im.InteractionRequest(im.InteractionRequest.RequestType.READ, {})
```
tpmanley, line 394: We really only need to do once per device. What about checking the currently capability status and only doing the reads if the attribute hasn't been set yet?

drivers/SmartThings/matter-switch/src/init.lua
tpmanley: I don't think it's very likely, but my read of the spec is that they are allowed to be equal

greens: I consider "allowed by the spec" and "allowed by us" to be two different things, frankly. I think we'd do better to stick to a default presentation than to show a slider with only one value.

drivers/SmartThings/matter-switch/src/init.lua
tpmanley: Are we purposefully not doing this for existing devices? I would think we'd want to do it for all devices once

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I addressed this in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1220/commits/f04cf2ec2c69373a72d337e4376c0565baa95f56

drivers/SmartThings/matter-switch/profiles/light-color-level.yml
nickolas-deboom: Fixed in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1220/commits/f04cf2ec2c69373a72d337e4376c0565baa95f56

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1220/commits/f04cf2ec2c69373a72d337e4376c0565baa95f56

drivers/SmartThings/matter-switch/src/init.lua
greens: My intention with my code changes was that the code for the behavior we want to see going forward would be put in first, and later there would be some additional migration code added that's only live for a week or two before being removed.

Migration code like moving off of old profiles, querying for the bounds, etc.

greens (approved):

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: This was moved to `init` in b76b11afefc0552a0e3cf4329c6381d570277552 so that it will work for existing devices as well as new devices.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: In order to stay aligned with the Zigbee implementation and the other matter drivers, I would prefer to leave this as-is for now.

drivers/SmartThings/matter-switch/src/test/test_matter_switch.lua
nickolas-deboom: I updated the test to use these values in 7be32c0a41199c9d51ee83463fc6dd9800c38304 as suggested.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -146,6 +162,23 @@ local function device_init(driver, device)
     device:set_endpoint_to_component_fn(endpoint_to_component)
     device:set_find_child(find_child)
     device:subscribe()
+
+    if not device:get_field(BOUNDS_CHECKED) then
+      local limit_read
```
ctowns, line 167: You could just initialize the `limit_read` variable up here
```
limit_read = im.InteractionRequest(im.InteractionRequest.RequestType.READ, {})
```

And then merge the Color Temperature and Level limit reads into one read request you send at the end of the function instead of two separate reads.

The `BOUNDS_CHECKED` flag you have added here will properly address [Tom's comment ](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1220#discussion_r1513389233) about doing this once for each device.

nickolas-deboom: I initially had it that way and it caused failures in other tests within the matter-switch driver, because some of the mock devices in those tests support neither the colorTemperature nor the switchLevel capability. This caused it to send a blank InteractionRequest in those tests. I decided to make the changes in c1d9f8213eb383876938c93f5b7e0b657a96ea42 to ensure that it would only send read requests if the capabilities are actually supported.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Also, the BOUNDS_CHECKED flag was introduced to ensure that it will only occur once per device.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -294,6 +335,51 @@ local function temp_attr_handler(driver, device, ib, response)
...
+  end
+end
+
+local level_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value ~= nil then
+      -- Convert level from given range of 0-254 to range of 0-100.
+      local level = utils.round(ib.data.value / 254.0 * 100)
+      set_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..minOrMax, ib.endpoint_id, level, false)
+      local min = get_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..LEVEL_MIN, ib.endpoint_id)
```
tpmanley, line 368: If the devices are following the spec then the minimum level should not be less than 1 but even so I think we should make sure it is 1 or higher to be on the safe side.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -294,6 +335,51 @@ local function temp_attr_handler(driver, device, ib, response)
...
+      local max = get_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..LEVEL_MAX, ib.endpoint_id)
```
tpmanley, line 369: Same with max - it shouldn't be higher than 100 but to be safe we should cap it

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Added a comment in 2cd1a7ca5dbafa8046c75650aa9a59365a24c2ce.

drivers/SmartThings/matter-switch/src/init.lua
gharveymn: Can we do a check for emptiness of the request and them omit sending it if empty?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -294,6 +335,51 @@ local function temp_attr_handler(driver, device, ib, response)
...
+      end
+      local temp_in_kelvin = mired_to_kelvin(ib.data.value)
+      set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp_in_kelvin, false)
+      local min = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MIN, ib.endpoint_id)
+      local max = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MAX, ib.endpoint_id)
+      if min ~= nil and max ~= nil then
+        if min < max then
+          device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = min, maximum = max} }))
+          set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MAX, ib.endpoint_id, nil, false)
+          set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MIN, ib.endpoint_id, nil, false)
```
gharveymn, line 353: I'm assuming this is to accommodate for the case where the bounds are changing. However, an event won't be emitted unless both of the bounds change. So, I think rather than unsetting the fields, you can perform a check to see if the field changed in addition to the nil checks.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -294,6 +335,51 @@ local function temp_attr_handler(driver, device, ib, response)
   end
 end
 
+local mired_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value ~= nil then
+      if (ib.data.value < COLOR_TEMPERATURE_MIRED_MIN or ib.data.value > COLOR_TEMPERATURE_MIRED_MAX) then
+        device.log.warn_with({hub_logs = true}, string.format("Device reported a min or max color temperature %d mired outside of supported capability range", ib.data.value))
+        return
```
gharveymn, line 343: Are we sure we want to ignore values outside of the range rather than clamp them? I feel like clamping might be better UX.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -294,6 +335,51 @@ local function temp_attr_handler(driver, device, ib, response)
   end
 end
 
+local mired_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value ~= nil then
```
gharveymn, line 340: Nit: Consider using an early return for readability.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -32,13 +33,36 @@ local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN
...
+  device:set_field(string.format("%s_%d", field, endpoint), value, {persist = persist})
+end
+
 local function convert_huesat_st_to_matter(val)
   return math.floor((val * 0xFE) / 100.0 + 0.5)
 end
 
+local function mired_to_kelvin(value)
+  local CONVERSION_CONSTANT = 1000000
+  if value == 0 then value = 1 end -- shouldn't happen, but has
```
gharveymn, line 60: Log needed here.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -32,13 +33,36 @@ local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN
 
 local SWITCH_INITIALIZED = "__switch_intialized"
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
+local BOUNDS_CHECKED = "__bounds_checked"
+local COLOR_TEMP_BOUND_RECEIVED = "colorTemp_bound_received"
```
gharveymn, line 37: Double underscore needed (idk the policy)?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -32,13 +33,36 @@ local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN
...
+local COLOR_TEMP_MIN = "__color_temp_min"
+local COLOR_TEMP_MAX = "__color_temp_max"
+local LEVEL_BOUND_RECEIVED = "level_bound_received"
```
gharveymn, line 40: Ditto.

gharveymn (commented):
Sorry for the delay. Almost there!

drivers/SmartThings/matter-switch/src/init.lua
ctowns: These min and max values are defined based on the [colorTemperature's default range](https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/8f6538bf03eacc140c38954299184bde9599886c/capabilities/color-temperature.src/color-temperature.yml#L11) of `1-30,000`. This is not a range that is used in practice, as most bulbs fall in the 2000K-8000K range. So, a value outside of this range is almost surely incorrect, and I think ignoring it is probably best for that reason. Additionally, a value of 0 mireds is impossible because it would cause a divide by zero error when converting to kelvin.

drivers/SmartThings/matter-switch/src/init.lua
ctowns: The reasoning for setting these to nil was just to save some memory so these values are no longer stored on the device object once they are not longer needed. As it stands right now, the device will only have its bound checked once and is not set up to handle dynamic bounds changing. I wouldn't expect that to be a common use case though.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -32,13 +33,36 @@ local COLOR_TEMPERATURE_MIRED_MIN = CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN
...
+local LEVEL_MAX = "__level_max"
 local AGGREGATOR_DEVICE_TYPE_ID = 0x000E
 local detect_matter_thing
 
+local function get_field_for_endpoint(device, field, endpoint)
+  return device:get_field(string.format("%s_%d", field, endpoint))
+end
+
+local function set_field_for_endpoint(device, field, endpoint, value, persist)
+  device:set_field(string.format("%s_%d", field, endpoint), value, {persist = persist})
```
gharveymn, line 51: This is a convenience function, so we should probably just forward the arguments.

```suggestion
local function set_field_for_endpoint(device, field, endpoint, value, additional_params)
  device:set_field(string.format("%s_%d", field, endpoint), value, additional_params)
```

gharveymn: I say this because the extra argument `, false` on calls to this is a little noisy and not super necessary since the default is reasonable. `additional_params` forces you to be explicit by making you write out `{persist = true}`.

drivers/SmartThings/matter-switch/src/init.lua
gharveymn: Does the UI use the default range if this value gets dropped anyway? If so, then my point was moot anyway.

ctowns: Correct! The UI will use the default range of 1-30000 if we do not provide a value. At the moment, we do not set a value unless we have received a valid min and max value. I suppose we could set just one or the other if we do not receive a valid value for both by clamping the values as you mentioned above. That seems appropriate to me - it seems better to provide what we can even if we don't receive valid value for both min and max - would you agree?

gharveymn: Yeah, I agree that seems like the best course of action.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated code to check for emptiness of the request as suggested in 775564f06809491f7f3c5359f887bf7864edad48.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Made this update in 775564f06809491f7f3c5359f887bf7864edad48. Also added a check for the lighting feature, since MinLevel has a lower bound of 1 when lighting is enabled and 0 when it is not.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Made this update in 775564f06809491f7f3c5359f887bf7864edad48.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I updated both new handler functions to clamp the values as suggested. See 775564f06809491f7f3c5359f887bf7864edad48.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated as suggested. See 775564f06809491f7f3c5359f887bf7864edad48.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Added a log in 775564f06809491f7f3c5359f887bf7864edad48.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated to use double underscores in 775564f06809491f7f3c5359f887bf7864edad48.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated to use double underscores in 775564f06809491f7f3c5359f887bf7864edad48.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated as suggested. See 775564f06809491f7f3c5359f887bf7864edad48.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -294,6 +340,72 @@ local function temp_attr_handler(driver, device, ib, response)
...
+    if (color_temp_mireds < COLOR_TEMPERATURE_MIRED_MIN or color_temp_mireds > COLOR_TEMPERATURE_MIRED_MAX) then
+      device.log.warn_with({hub_logs = true}, string.format("Device reported a min or max color temperature %d mired outside of supported capability range", color_temp_mireds))
+      if (minOrMax == COLOR_TEMP_MIN) then
+        color_temp_mireds = COLOR_TEMPERATURE_MIRED_MIN
+      else
+        color_temp_mireds = COLOR_TEMPERATURE_MIRED_MAX
+      end
+    end
+    local temp_in_kelvin = mired_to_kelvin(color_temp_mireds)
+    set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp_in_kelvin, {persist = true})
```
gharveymn, line 358: Is this change to make it persistent intentional?

nickolas-deboom: It was not, and I removed it in the last commit: 8db6cbd25f144d5dac42038ee0ed766cabc54a3d.

gharveymn (approved):
Nice work Nick

ctowns (approved):
Looks great Nick, thanks for all your hard work to get this to the finish line!

Also, make sure you squash all your commits before merging :)

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -54,10 +65,29 @@ local device_type_profile_map = {
...
+local function set_field_for_endpoint(device, field, endpoint, value, additional_params)
+  device:set_field(string.format("%s_%d", field, endpoint), value, additional_params)
+end
+
 local function convert_huesat_st_to_matter(val)
   return math.floor((val * 0xFE) / 100.0 + 0.5)
 end
 
+local function mired_to_kelvin(value)
+  local CONVERSION_CONSTANT = 1000000
```
tpmanley, line 81: Sorry I was a bit late on the review and I realize this has already been merged, but I do have a few comments.

`CONVERSION_CONSTANT` is already defined globally so we don't need to redefine it here. It'd be good to also give it a more specific name to indicate what it's for.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -335,6 +381,72 @@ local function temp_attr_handler(driver, device, ib, response)
...
+local mired_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value == nil then
+      return
+    end
+    local color_temp_mireds = ib.data.value
+    if (color_temp_mireds < COLOR_TEMPERATURE_MIRED_MIN or color_temp_mireds > COLOR_TEMPERATURE_MIRED_MAX) then
+      device.log.warn_with({hub_logs = true}, string.format("Device reported a min or max color temperature %d mired outside of supported capability range", color_temp_mireds))
+      if (minOrMax == COLOR_TEMP_MIN) then
+        color_temp_mireds = COLOR_TEMPERATURE_MIRED_MIN
```
tpmanley, line 393: In the `COLOR_TEMP_MIN` case shouldn't this be set to `COLOR_TEMPERATURE_MIRED_MAX` since minimum K is maximum mireds?

tpmanley: It'd also be good to add tests for this case too

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -335,6 +381,72 @@ local function temp_attr_handler(driver, device, ib, response)
...
+      end
+    end
+    local temp_in_kelvin = mired_to_kelvin(color_temp_mireds)
+    set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp_in_kelvin)
+    local min = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MIN, ib.endpoint_id)
+    local max = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MAX, ib.endpoint_id)
+    if min ~= nil and max ~= nil then
+      if min < max then
+        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = min, maximum = max} }))
+        set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MAX, ib.endpoint_id, nil)
```
tpmanley, line 405: Is there a reason not to `nil` these out in the `else` case as well?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -335,6 +381,72 @@ local function temp_attr_handler(driver, device, ib, response)
...
+end
+
+local level_bounds_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value == nil then
+      return
+    end
+    local current_level = ib.data.value
+    local lighting_endpoints = device:get_endpoints(clusters.LevelControl.ID, {feature_bitmap = clusters.LevelControl.FeatureMap.LIGHTING})
+    if ((#lighting_endpoints ~= 0 and current_level < SWITCH_LEVEL_LIGHTING_MIN) or current_level < SWITCH_LEVEL_MIN or current_level > SWITCH_LEVEL_MAX) then
```
tpmanley, line 421: Can you add test cases for the different branches in this logic?

Also I think we need to be checking the endpoint reporting min or max attribute for the lighting endpoint. Right now I don't think it would work if a device happened to have multiple endpoints with the level control cluster but a mix of support for the lighting feature. That's probably unlikely but I think it should be pretty easy to do this check.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -335,6 +381,72 @@ local function temp_attr_handler(driver, device, ib, response)
...
+    end
+    -- Convert level from given range of 0-254 to range of 0-100.
+    local level = utils.round(current_level / 254.0 * 100)
+    set_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..minOrMax, ib.endpoint_id, level)
+    local min = get_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..LEVEL_MIN, ib.endpoint_id)
+    local max = get_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..LEVEL_MAX, ib.endpoint_id)
+    if min ~= nil and max ~= nil then
+      if min < max then
+        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.switchLevel.levelRange({ value = {minimum = min, maximum = max} }))
+        set_field_for_endpoint(device, LEVEL_BOUND_RECEIVED..LEVEL_MAX, ib.endpoint_id, nil)
```
tpmanley, line 441: Same question as before about whether these should be nil'd in the `else` case as well

PR #1223, opened 2024-02-16
Matter Thermostat: Heat/cool setpoint and temp limits dynamic constraints

nickolas-deboom:
Subscribe to the temperature limit and heat/cool setpoint limit attributes and set capability constraints based on the limits.


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -171,11 +176,16 @@ local function do_configure(driver, device)
   if #auto_eps ~= 0 and device:get_field(setpoint_limit_device_field.MIN_DEADBAND) == nil then
     setpoint_limit_read:merge(clusters.Thermostat.attributes.MinSetpointDeadBand:read())
   end
+  if #temp_eps ~= 0 then
```
ctowns, line 179: Similarly to what you've done in the switch and sensor drivers, let's move any limit reads to the `init` function so that we will update devices already in the field as well! `doConfigure` in this instance will essentially only run for new device adds, but moving this logic to `init` will allow us to set these limits for devices already in the field without requiring a fresh onboard.

You can move the TemperatureMeasurement limit reads that you added, as well as the Thermostat setpoint reads that are also in this `doConfigure` function. This way, all of these capability ranges can be set for devices already joined to the platform.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -374,11 +405,15 @@ local function set_setpoint(setpoint)
   end
 end
 
-local function setpoint_limit_handler(limit_field)
+local function setpoint_limit_handler(limit_field, limit_field_opposite, attribute)
```
ctowns, line 408: You could modify this to follow the same strategy of indicating `min/max` in the parameters as you have done above for the `temp_attr_handler_factory`

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Moved all of the limit reads form `doConfigure` into `init` in c33ee3ad996ac550630fc72a2fd778f617096866. Updated the test cases as needed.

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Updated as suggested in c33ee3ad996ac550630fc72a2fd778f617096866.

drivers/SmartThings/matter-thermostat/src/test/test_matter_thermo_setpoint_limits.lua
```diff
@@ -128,6 +113,10 @@ end
 test.register_coroutine_test(
   "Heat setpoint lower than min",
   function()
+    test.socket.matter:__queue_receive({
```
ctowns, line 116: I don't think the `queue_receive` calls you have added here and in the following 5 tests are doing anything at the moment. Since you are only receiving either the min or the max in each case, the setpoint limits are not actually set because we only emit the setpoint limit capability event when both the min and max have been received. So, you can remove these from the following test cases.

nickolas-deboom: These test cases need the heating and cooling setpoint limits to be defined in order to verify the functionality of `setHeatingSetpoint` and `setCoolingSetpoint`. Since we removed the `queue_receive` calls from `test_init`, the setpoint limits are not being set for the mock device and so the test cases fail. We could move the `queue_receive` calls back to `test_init`, but then we would also need to check that the capabilities for the heating and cooling setpoint ranges are being sent. If we go that route, then would it make sense to remove the three new test cases added by this PR?

ctowns: Thank you for clarifying, I understand why these are here now! I think what you've done here is good, so you can keep the test cases as they are, including the 3 test cases you've added.

ctowns (commented):
With these capability constraint changes, you should be able to modify the `set_setpoint` function to get the min and max setpoint values via the `device:get_latest_state` function. This will allow you to get these values from the capability rather than having to store them in the device object. This will also simplify the `set_setpoint` function such that you will only need to check if the new setpoint value is in the deadband, because it should theoretically already be guanrateed to be within the min and max value given the capability constraints have been applied correctly.

gharveymn (approved):
Looks good, pending analogous changes in consideration of feedback on #1220

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -374,11 +418,41 @@ local function set_setpoint(setpoint)
...
 
-local function setpoint_limit_handler(limit_field)
+local heating_setpoint_limit_handler_factory = function(minOrMax)
   return function(driver, device, ib, response)
-    local val = ib.data.value / 100.0
-    log.info("Setting " .. limit_field .. " to " .. string.format("%s", val))
-    device:set_field(limit_field, val, { persist = true })
+    if ib.data.value ~= nil then
+      local val = ib.data.value / 100.0
+      print("setting %s to %d", minOrMax, val)
```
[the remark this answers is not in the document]
nickolas-deboom: Fixed in 37c7fc196b6d762425671b2e372b379ae4d84f61.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -374,11 +418,41 @@ local function set_setpoint(setpoint)
...
+      end
+    end
+  end
+end
+
+local cooling_setpoint_limit_handler_factory = function(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value ~= nil then
+      local val = ib.data.value / 100.0
+      print("setting %s to %d", minOrMax, val)
```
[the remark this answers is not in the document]
nickolas-deboom: Fixed in 37c7fc196b6d762425671b2e372b379ae4d84f61.

drivers/SmartThings/matter-thermostat/src/test/test_matter_thermo_featuremap.lua
```diff
@@ -151,6 +151,9 @@ local cluster_subscribe_list_no_battery = {
 }
 
 local function test_init()
+  mock_device:set_field("BOUNDS_CHECKED", 1, {persist = true})
```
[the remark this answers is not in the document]
nickolas-deboom: Added a comment in 37c7fc196b6d762425671b2e372b379ae4d84f61.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -328,12 +373,24 @@ local function set_setpoint(setpoint)
       {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.AUTOMODE}
     ) > 0
 
+    local cooling_setpoint_range = device:get_latest_state(
```
ctowns, line 376: What is the reason for using 0 and 100 as the default min and max? Also, what are the expected units? I think there are a few issues with unit ambiguity that we will need to address

ctowns: @nickolas-deboom the `get_latest_state()` function returns two values as described [here](https://developer.smartthings.com/docs/edge-device-drivers/device.html?highlight=get_latest#st.Device.get_latest_state). The first return value is the state.value present for the attribute, the second return is the state. Does the second return value include the units? You can look at the examples a few lines above this to see how this might work by getting the unit from the second return value and then doing the conversion as needed: 
```
    local cached_heating_val, heating_setpoint = device:get_latest_state(
      cmd.component, capabilities.thermostatHeatingSetpoint.ID,
      capabilities.thermostatHeatingSetpoint.heatingSetpoint.NAME,
      0, { value = 0, unit = "C" }
    )
    if heating_setpoint and heating_setpoint.unit == "F" then
      cached_heating_val = utils.f_to_c(cached_heating_val)
    end
```

nickolas-deboom: The second return value indeed includes the units. I made this update in 7d6d886571b626d615efd8660e077053ea45e6ff.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -374,11 +436,51 @@ local function set_setpoint(setpoint)
...
-    device:set_field(limit_field, val, { persist = true })
+    if val >= 40 then -- assume this is a fahrenheit value
+      val = utils.f_to_c(val)
+    end
+    device:set_field(minOrMax, val)
+    local min = device:get_field(setpoint_limit_device_field.MIN_HEAT)
+    local max = device:get_field(setpoint_limit_device_field.MAX_HEAT)
+    if min ~= nil and max ~= nil then
+      if min < max then
+        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.thermostatHeatingSetpoint.heatingSetpointRange({ value = { minimum = min, maximum = max }, unit = "C" }))
```
ctowns, line 453: The temperature conversions for tempRange (and the setpoint values as well) are supported in FW 53+. Therefore, we won't want to send these requests for cases where the hub is running FW 53 or greater. We could gate this on lua libs api version 10 since that is the api version for lua libs packaged with FW 53.

Caveat: Lua libs version until now has essentially been tied to FW version since we use the lua libs packaged in FW. However, this parity may be broken with async lua libs releases, so this is something we will need to consider. The change to support these temp conversion is not technically in the lua libs, it is in the FW.

nickolas-deboom:
The update made in cf3d46dee6736fb1545a771e32ab5b6760b304fc uses subscriptions for the temp range and setpoint range limits rather than reading them during init. This change was made to resolve an issue with InteractionQueueFull exceptions seen while onboarding a large number of devices at the same time.

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: As discussed, added a check for RPC version in f56fc0de8347451163d76740b9781153ce65328c.

PR #1224, opened 2024-02-16
Matter Sensor:  Temp limits dynamic constraints

nickolas-deboom:
Subscribe to the temperature limit attributes and set capability constraint based on the limits.

### DIFF
--- a/drivers/SmartThings/matter-sensor/src/init.lua
+++ b/drivers/SmartThings/matter-sensor/src/init.lua
@@ -17,6 +17,7 @@ local log = require "log"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
 local utils = require "st.utils"
+local version = require "version"
 
 -- This can be removed once LuaLibs supports the PressureMeasurement cluster
 if not pcall(function(cluster) return clusters[cluster] end,
@@ -25,8 +26,19 @@ if not pcall(function(cluster) return clusters[cluster] end,
 end
 
 local BATTERY_CHECKED = "__battery_checked"
+local TEMP_BOUND_RECEIVED = "__temp_bound_received"
+local TEMP_MIN = "__temp_min"
+local TEMP_MAX = "__temp_max"
 
 local HUE_MANUFACTURER_ID = 0x100B
+
+local function get_field_for_endpoint(device, field, endpoint)
+  return device:get_field(string.format("%s_%d", field, endpoint))
+end
+
+local function set_field_for_endpoint(device, field, endpoint, value, additional_params)
+  device:set_field(string.format("%s_%d", field, endpoint), value, additional_params)
+end
 
 local function supports_battery_percentage_remaining(device)
   local battery_eps = device:get_endpoints(clusters.PowerSource.ID,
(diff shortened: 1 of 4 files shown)


drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -91,6 +107,27 @@ local function temperature_attr_handler(driver, device, ib, response)
...
+  return function(driver, device, ib, response)
+    if ib.data.value ~= nil then
+      local temp = ib.data.value / 100.0
+      local unit = "C"
+      device:set_field(TEMP_BOUND_RECEIVED..minOrMax, temp)
+      local min = device:get_field(TEMP_BOUND_RECEIVED..TEMP_MIN)
+      local max = device:get_field(TEMP_BOUND_RECEIVED..TEMP_MAX)
+      if min ~= nil and max ~= nil then
+        if min < max then
+          device:emit_event_for_endpoint(ib.endpoint_id, capabilities.temperatureMeasurement.temperatureRange({ value = { minimum = min, maximum = max }, unit = unit }))
```
gharveymn, line 120: There doesn't seem to be any mechanism preventing multiple endpoints from interfering with one another. Perhaps the fields need to be scoped to an endpoint?

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -15,13 +15,26 @@
...
 
 local HUE_MANUFACTURER_ID = 0x100B
 
+local function get_field_for_endpoint(device, field, endpoint)
+  return device:get_field(string.format("%s_%d", field, endpoint))
+end
+
+local function set_field_for_endpoint(device, field, endpoint, value, persist)
+  device:set_field(string.format("%s_%d", field, endpoint), value, {persist = persist})
+end
```
gharveymn, line 36: Can we add these to `Device` in the lua libraries as a follow-up?

ctowns: Good thinking - created a ticket for this here: https://smartthings.atlassian.net/browse/CHAD-12925

gharveymn (approved):
Same as #1223; looks good pending #1220.

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -91,6 +117,28 @@ local function temperature_attr_handler(driver, device, ib, response)
...
+      return
+    end
+    local temp = ib.data.value / 100.0
+    local unit = "C"
+    set_field_for_endpoint(device, TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp)
+    local min = get_field_for_endpoint(device, TEMP_BOUND_RECEIVED..TEMP_MIN, ib.endpoint_id)
+    local max = get_field_for_endpoint(device, TEMP_BOUND_RECEIVED..TEMP_MAX, ib.endpoint_id)
+    if min ~= nil and max ~= nil then
+      if min < max then
+        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.temperatureMeasurement.temperatureRange({ value = { minimum = min, maximum = max }, unit = unit }))
```
ctowns, line 132: The temperature conversions for tempRange (and the setpoint values as well) are supported in FW 53+. Therefore, we won't want to send these requests for cases where the hub is running FW 53 or greater. We could gate this on lua libs api version 10 since that is the api version for lua libs packaged with FW 53.

Caveat: Lua libs version until now has essentially been tied to FW version since we use the lua libs packaged in FW. However, this parity may be broken with async lua libs releases, so this is something we will need to consider. The change to support these temp conversion is not technically in the lua libs, it is in the FW.

This will be the same for the `matter-thermostat` PR as well: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1223

nickolas-deboom:
The update made in 82c59e6b24dc3e4fd77bfd1fd2e17ed6d20ffedb uses subscriptions for the temp range limits rather than reading them during init. This change was made to resolve an issue with InteractionQueueFull exceptions seen while onboarding a large number of devices at the same time.

drivers/SmartThings/matter-sensor/src/test/test_matter_sensor.lua
```diff
@@ -18,6 +18,8 @@ local capabilities = require "st.capabilities"
 local t_utils = require "integration_test.utils"
 local clusters = require "st.matter.clusters"
 
+test.set_rpc_version(4)
```
ctowns, line 21: Can you add a test case that checks that nothing happens when RPC version < 4?

drivers/SmartThings/matter-sensor/src/init.lua
nickolas-deboom: As discussed, added a check for RPC version in 2afbc40e1f113c7c03763e87d4061378c41ddc1f.

drivers/SmartThings/matter-sensor/src/test/test_matter_sensor.lua
nickolas-deboom: Added a new test case in 4c9707b7c23b9fbbfe370a4ab591bc9c68774041.

ctowns (commented):
This looks good, just add the unit test for if RPC version < 4 that you did for matter-thermostat (https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/4c9707b7c23b9fbbfe370a4ab591bc9c68774041) and I'll approve this!

nickolas-deboom:
> This looks good, just add the unit test for if RPC version < 4 that you did for matter-thermostat ([4c9707b](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/4c9707b7c23b9fbbfe370a4ab591bc9c68774041)) and I'll approve this!

This unit test was added here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1224/files#diff-915816924e061c2f6111072dae683786dc77bf79e5e5b37df1dfb2bed732afa5

ctowns (approved):

PR #1227, opened 2024-02-20
SmartSense multipurpose: negative temperature issue

Mariano-Github:
When the device reports negative temperature values in Int16 data type, the 2's complement of the value sent by the device must be performed


tpmanley:
Good catch @Mariano-Github ! In order to get this merged could you sign the CLA and point the PR against the `main` branch instead of `production`? It would also be great if you could add a unit test to https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-contact/src/test/test_smartsense_multi.lua if possible so the negative temperature case is tested.

Mariano-Github:
Hi,
I changed it to the main branch and changes appeared in a file that has nothing to do with it and I left it as it was. Sorry, I don't know how this works.

I don't know what you mean to do with the Test module either.
The changes were tested by a user with a real device, I do not use the test modules

I already signed the CLA

Mariano-Github:
Sorry, you can close the pull request and incorporate the modification as best for you.

Greetings

PR #1235, opened 2024-02-22
Eve Energy: fix can handle and get switch state on child add

ctowns:
When a child device is added, request the switch state. This is done to account for a race condition where a subscription report arrives before the child device has been created and the child device misses the initial state.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/eve-energy/init.lua
+++ b/drivers/SmartThings/matter-switch/src/eve-energy/init.lua
@@ -50,9 +50,10 @@ local REPORT_TIMEOUT = (15 * 60) -- Report the value each 15 minutes
 -------------------------------------------------------------------------------------
 
 local function is_eve_energy_products(opts, driver, device)
-  -- this sub driver does not support child devices
-  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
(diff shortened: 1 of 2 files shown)


greens (commented):
I'm going to be that guy and suggest you write a test.

ctowns:
@greens Good call out on the test - I wrote a test to check this in the sub driver and this revealed that the `can_handle` check for the sub driver was failing for child devices. This means the added event was actually defaulting to being handled by the parent driver.

Technically, this means that this PR or https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1213 could address the child initial refresh issue since this creates an added handler in the parent driver, which would be used by child devices in this driver when the `can_handle` function fails. However, it would be better fix the `can_handle` function as I have done here. My only concern is that it may not be safe to call the `get_parent_device` function in the `can_handle` function due to race conditions on device creation. Tomorrow I will confirm that it is safe to call `get_parent_device()` in the `can_handle()` function, and if it is we can merge this PR after it has been tested with a real device.

lelandblue:
Not urgent @ctowns - Can you please review if this PR can be closed ? or if we want to keep it ?

ctowns:
See PR#1213

PR #1237, opened 2024-02-23
matter-room-air-conditioner: Fix fan control percent feature

Jaehoon-You:
IOTE-3542
IOTE-3611


HunsupJung (approved):
Looks good to me :)

PR #1248, opened 2024-02-28
CHAD-12653 Zigbee ColorTemperature: add dynamic range handling

greens:
Uses the PhysicalMax and PhysicalMin mired values from the color control cluster to dynamically set the min/max range for zigbee bulbs in the base driver.


drivers/SmartThings/zigbee-switch/src/test/test_all_capability_zigbee_bulb.lua
```diff
@@ -189,6 +189,32 @@ test.register_message_test(
     }
 )
 
+test.register_coroutine_test(
+  "Max and min mired reports should generate a range report after both are received",
+  function ()
+    test.socket.zigbee:__queue_receive({
+      mock_device.id,
+      ColorControl.attributes.ColorTempPhysicalMinMireds:build_test_attr_report(mock_device, 555) -- ~1800 K
```
tpmanley, line 197: I believe these are flipped, so a good test value for `ColorTempPhysicalMinMireds` is 154 which is roughly 6500K and a good value for `ColorTempPhysicalMaxMireds` is 555 which is roughly 1800K

greens: i.e. the min mireds corresponds to the max kelvin and vice versa

you are absolutely right. what cursed units

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -38,6 +54,29 @@ local do_configure = function(self, device)
...
+    if temp_in_kelvin > SANITY_CHECK_MIN_KELVIN and temp_in_kelvin < SANITY_CHECK_MAX_KELVIN then
+      device:set_field(BOUND_RECEIVED..minOrMax, temp_in_kelvin)
+    end
+
+    local min = device:get_field(BOUND_RECEIVED..MIN)
+    local max = device:get_field(BOUND_RECEIVED..MAX)
+    if min ~= nil and max ~= nil and min < max then
+      device:emit_event_for_endpoint(endpoint_id, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = min, maximum = max}}))
+      device:set_field(BOUND_RECEIVED..MAX, nil)
+      device:set_field(BOUND_RECEIVED..MAX)
```
ctowns, line 77:
```suggestion
      device:set_field(BOUND_RECEIVED..MIN, nil)
```

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -38,6 +54,29 @@ local do_configure = function(self, device)
...
+  end
+end
+
+local mired_bounds_handler_factory = function(minOrMax)
+  return function(self, device, value, zb_rx)
+    local endpoint_id = zb_rx.address_header.src_endpoint.value
+    local temp_in_kelvin = mired_to_kelvin(value.value)
+    if temp_in_kelvin > SANITY_CHECK_MIN_KELVIN and temp_in_kelvin < SANITY_CHECK_MAX_KELVIN then
+      device:set_field(BOUND_RECEIVED..minOrMax, temp_in_kelvin)
+    end
```
tpmanley, line 70: I think a log message in the else case would be helpful here

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -38,6 +54,31 @@ local do_configure = function(self, device)
...
+    else
+      device.log.warn("Device reported a min or max color temp value outside of reasonable bounds: "..temp_in_kelvin..'K')
+    end
+
+    local min = device:get_field(BOUND_RECEIVED..MIN)
+    local max = device:get_field(BOUND_RECEIVED..MAX)
+    if min ~= nil and max ~= nil and min < max then
+      device:emit_event_for_endpoint(endpoint_id, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = min, maximum = max}}))
+      device:set_field(BOUND_RECEIVED..MAX, nil)
+      device:set_field(BOUND_RECEIVED..MAX, nil)
```
ctowns, line 79:
```suggestion
      device:set_field(BOUND_RECEIVED..MIN, nil)
```

tpmanley (approved):

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -19,7 +19,23 @@ local clusters = require "st.zigbee.zcl.clusters"
 local configurationMap = require "configurations"
 local SimpleMetering = clusters.SimpleMetering
 local ElectricalMeasurement = clusters.ElectricalMeasurement
+local ColorControl = clusters.ColorControl
 local preferences = require "preferences"
+local utils = require "st.utils"
+
+local SANITY_CHECK_MIN_KELVIN = 1000
+local SANITY_CHECK_MAX_KELVIN = 10000
```
wkhenon, line 27: Want to confirm that this `1,000K - 10,000K` range are the only values that we want for ColorTemperature. The spec states  a range from `15.32K - 1,000,000K`:

> _**ZCL Rev. 8 - 5.2.2.2.1.8 ColorTemperatureMireds Attribute**_
Color temperature in kelvins = 1,000,000 / ColorTemperatureMireds, where ColorTemperatureMireds is in
the range 1 to 65279 mireds inclusive, giving a color temperature range from 1,000,000 kelvins to 15.32
kelvins.

greens: Our capability only supports from 1 to 30,000 so I'll probably go with that.

In reality, I don't think we've seen any bulbs outside of this current range.

wkhenon (approved):

PR #1251, opened 2024-03-04
matter-sensor: Add support for Pressure Sensor

tpmanley:
This includes support for a standalone pressure sensor and also a temp, humidity and pressure sensor, both w/ or w/o battery.


drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -86,15 +96,20 @@ local function illuminance_attr_handler(driver, device, ib, response)
 end
 
 local function temperature_attr_handler(driver, device, ib, response)
-  local temp = ib.data.value / 100.0
-  local unit = "C"
-  device:emit_event_for_endpoint(ib.endpoint_id, capabilities.temperatureMeasurement.temperature({value = temp, unit = unit}))
+  local measured_value = ib.data.value
+  if measured_value ~= nil then
```
tpmanley, line 100: This isn't directly related to the pressure sensor and perhaps I should split it off into a separate PR, but the `MeasuredValue` attribute in the `TemperatureMeasurement` and `RelativeHumidityMeasurement` clusters are nullable so not having the check leads to a driver crash.

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -115,6 +130,15 @@ local function occupancy_attr_handler(driver, device, ib, response)
   device:emit_event(ib.data.value == 0x01 and capabilities.motionSensor.motion.active() or capabilities.motionSensor.motion.inactive())
 end
 
+local function pressure_attr_handler(driver, device, ib, response)
+  local measured_value = ib.data.value
+  if measured_value ~= nil then
+    local kPa = math.floor(measured_value / 10.0)
```
greens, line 136: we do have `utils.round`

greens (commented):
I understand the desire to get this out before the lua libs have it but this is still kind of gross.

tpmanley:
> I understand the desire to get this out before the lua libs have it but this is still kind of gross.

It's not ideal, but it's what we have to work with. Doing it this way meant I could isolate the changes to just adding the cluster definition and a couple of lines in the init to make switching it to the LuaLibs version easier down the road. I could do a much more scaled back version of the cluster to just the bits that the driver needs if you think that's preferable.

greens (approved):

drivers/SmartThings/matter-sensor/src/test/test_matter_pressure_sensor.lua
```diff
@@ -0,0 +1,156 @@
...
+local function subscribe_on_init(dev)
+  local subscribe_request = PressureMeasurementCluster.attributes.MeasuredValue:subscribe(mock_device)
+  subscribe_request:merge(clusters.PowerSource.attributes.BatPercentRemaining:subscribe(mock_device))
+  return subscribe_request
+end
+
+local function test_init()
+  test.socket.matter:__expect_send({mock_device.id, subscribe_on_init(mock_device)})
+  test.mock_device.add_test_device(mock_device)
+  -- don't check the battery for this device because we are using the catch-all "sensor.yml" profile just for testing
```
ctowns, line 59: nit: 
```suggestion
  -- don't check the battery for this device since we are just testing the "pressure-battery" profile specifically
```

ctowns (approved):

drivers/SmartThings/matter-sensor/src/test/test_matter_pressure_sensor.lua
tpmanley: fixed

PR #1252, opened 2024-03-05
Update JBL lunchbox, to utilize new luncheon library

cjswedes:


dljsjr (approved):

varzac (approved):
Have we done a search on our EdgeDrivers rep to identify if there are any additional cases that we need to address?

drivers/SmartThings/jbl/src/lunchbox/rest.lua
```diff
@@ -14,16 +23,14 @@ local RestCallStates = {
...
 
   if client.base_url.scheme == "https" then
-    default_port = 443
+    port = 443
     use_ssl = true
   end
 
-  local port = client.base_url.port or default_port
-
   local sock, err = client.socket_builder(client.base_url.host, port, use_ssl)
```
sy39ju, line 34: I think this line is required to use `4443` port.
```
local port = client.base_url.port or default_port
```

# peer verification fails
$ curl https://192.168.0.111:4443/authcode --cacert  jbl/src/selfSignedRoot.crt 
curl: (60) SSL: no alternative certificate subject name matches target host name '192.168.0.111'
More details here: https://curl.haxx.se/docs/sslcerts.html
...

# By insecure option (no peer verification), it works.
$ curl https://192.168.0.111:4443/authcode  -k
{"token": "ZmFrZSB0b2tlbg=="}

# In case of 443 port, client verification is required but AFAIK there is no client certificate. So we need to use `4443` port.
$ curl -k https://192.168.0.111:443/authcode   
curl: (56) OpenSSL SSL_read: error:1409445C:SSL routines:ssl3_read_bytes:tlsv13 alert certificate required, errno 0

dljsjr: @sy39ju Here are my thoughts:

1. It looks like this code was already using 443 as the default
2. 443 is the default port for HTTPS, similar to how 80 is the default port for HTTP
3. Using a port other than 443 is dependent on what port the server uses
4. Setting the port to 4443 should be done when the constructor is called, not by overriding the default port typically used by the protocol.

PR #1253, opened 2024-03-05
fix(hue): Address various nil field index issues

dljsjr:
The majority of these fixes are concurrency races:

Due to Hue's heavy usage of additional Cosock coroutines, it can be the
case that functions that have been passed references to device records
will be operating on a device record for a deleted device. This can cause
unexpected nil index errors and crashes.

We attempt to minimize that here with heuristics; in particular, a device
that has been deleted will return `nil` for *any* of its table keys. So
we check a field that we *know* a device record should have, its `.id` field.

We wrap all handlers in a closure that does a sanity check on the status of
the device record using this heuristic, and we wrap the handler itself in a
pcall.

We also address some one-off locations where nil field indexes can occur
for other reasons, and introduce some additional defensiveness around other
places where we've observed nil index crashes in crash metrics.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/hue/api.lua
+++ b/drivers/SmartThings/philips-hue/src/hue/api.lua
@@ -29,9 +29,20 @@ local ControlMessageBuilders = {
   end
 }
 
+local function try_send(instance, message)
+  if not instance._ctrl_tx then
+    log.error(st_utils.stringify_table(message, "Couldn't send the followings due to closed transmit channel", false))
+  end
+
+  local success, err = pcall(instance._ctrl_tx.send, instance._ctrl_tx, message)
+  if not success then
+    log.error(string.format("Failed to transmit Hue Control Message: %s", err))
+  end
+end
+
 local function do_shutdown(instance)
   if instance._running then
-    instance._ctrl_tx:send(ControlMessageBuilders.Shutdown())
+    try_send(instance, ControlMessageBuilders.Shutdown())
     instance._running = false
   end
 end
@@ -178,7 +189,7 @@ end
 
 function PhilipsHueApi:update_connection(hub_base_url, api_key)
   local msg = ControlMessageBuilders.Update(hub_base_url, api_key)
-  self._ctrl_tx:send(msg)
+  try_send(self, msg)
 end
 
 ---@return table|nil response REST response, nil if error
@@ -187,7 +198,7 @@ local function do_get(instance, path)
   local reply_tx, reply_rx = channel.new()
   reply_rx:settimeout(10)
   local msg = ControlMessageBuilders.Get(path, reply_tx);
-  instance._ctrl_tx:send(msg)
+  try_send(instance, msg)
   local recv, err = reply_rx:receive()
   if err ~= nil then
     instance.client:close_socket()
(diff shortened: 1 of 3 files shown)


varzac (approved):
I think we will want to re-visit the details of how a deleted device object is handled, but it is a tricky situation.

greens (approved):

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -1401,15 +1401,30 @@ cosock.spawn(function()
...
-local switch_on_handler = handlers.switch_on_handler
-local switch_off_handler = handlers.switch_off_handler
-local switch_level_handler = handlers.switch_level_handler
-local set_color_temp_handler = handlers.set_color_temp_handler
+local refresh_handler = safe_wrap_handler(handlers.refresh_handler)
+local switch_on_handler = safe_wrap_handler(handlers.switch_on_handler)
+local switch_off_handler = safe_wrap_handler(handlers.switch_off_handler)
+local switch_level_handler = safe_wrap_handler(handlers.switch_level_handler)
+local set_color_handler = safe_wrap_handler(handlers.set_color_handler)
+local set_color_temp_handler = safe_wrap_handler(handlers.set_color_temp_handler)
```
cjswedes, line 1427: Are these handlers not being run on the device thread meaning they are already being wrapped in a pcall?

cjswedes: Ahh okay, I understand now why they are need

cjswedes (approved):

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -762,24 +789,27 @@ local function do_bridge_network_init(driver, bridge_device, bridge_url, api_key
...
-                child_device:online()
-                child_device:set_field(Fields.IS_ONLINE, true)
-              elseif status.status == "connectivity_issue" then
-                child_device.log.info_with({hub_logs=true}, "Marking Offline after SSE Reconnect")
-                child_device:set_field(Fields.IS_ONLINE, false)
-                child_device:offline()
+            for _, status in ipairs(connectivity_status.data) do
+              local hue_device_id = (status.owner and status.owner.rid) or ""
+              log.trace(string.format("Checking connectivity status for device resource id %s", hue_device_id))
+              local child_device = child_device_map[hue_device_id]
```
dljsjr, line 795: I believe this cached value is where we were sending device health changes for deleted devices that @NoahCornell  was seeing, but I'm not fully confident because I would expect to get crashes on indexing in to the device-scoped log first. So I'm not 100% on that yet.

drivers/SmartThings/philips-hue/src/hue/api.lua
```diff
@@ -29,9 +29,20 @@ local ControlMessageBuilders = {
   end
 }
 
+local function try_send(instance, message)
+  if not instance._ctrl_tx then
+    log.error(st_utils.stringify_table(message, "Couldn't send the followings due to closed transmit channel", false))
+  end
+
+  local success, err = pcall(instance._ctrl_tx:send(message))
```
varzac, line 37: I don't think this is the syntax you want.  I think this will try to call the function returned by the send command.

dljsjr: Oof, yeah, bad copy/paste, good catch.

NoahCornell (approved):

PR #1261, opened 2024-03-07
Cannot onboading JBL 300

kdbang:
Cannot onboading JBL 300 with 0.50.10 hubcore

### DIFF
--- a/drivers/SmartThings/jbl/src/lunchbox/rest.lua
+++ b/drivers/SmartThings/jbl/src/lunchbox/rest.lua
@@ -144,6 +144,38 @@ local function parse_chunked_response(original_response, sock)
   return full_response
 end
 
+local function recv_additional_response(original_response, sock)
+  local full_response = Response.new(original_response.status, nil)
+  local headers = original_response:get_headers()
+  local content_length_str = headers:get_one("Content-Length")
+  local content_length = nil
+  local bytes_read = 0
+  if content_length_str then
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/jbl/src/lunchbox/rest.lua
```diff
@@ -158,7 +190,9 @@ local function handle_response(sock)
   if initial_recv ~= nil then
     local headers = initial_recv:get_headers()
 
-    if headers and headers:get_one("Transfer-Encoding") == "chunked" then
+    if headers:get_one("Content-Length") then
```
varzac, line 193: I think this will be a problem if the headers are empty, and it should be `if headers and headers:get_one` like the one below.

varzac: This also seems like it would never go to the `elseif` below.  Was it ever the case that the `Content-Length` header wasn't set, but the `Transfer-Encoding` was?

varzac:
We will move forward with these changes as they are given that we have tested on 50 and 52, but I think as it is this is bypassing the chunked encoding that was added so it didn't have to be done manually.  But we can investigate and make updates later after the firmware release.

cjswedes (approved):
I agree with Zach's comments, but am hesitant to make any changes and hotfix without a test. So Giving my approval based on the testing done.

PR #1266, opened 2024-03-08
Fix incorrect operational state issue by workaround

HunsupJung:
Operational state is mismatching with actual status of window covering

### DIFF
--- a/drivers/SmartThings/matter-window-covering/src/init.lua
+++ b/drivers/SmartThings/matter-window-covering/src/init.lua
@@ -21,14 +21,14 @@ local MatterDriver = require "st.matter.driver"
 
 local DEFAULT_LEVEL = 0
 local PROFILE_MATCHED = "__profile_matched"
+local IS_MOVING = "__is_moving"
+local EVENT_STATE = "__event_state"
 
-local OperationalStateEventEnum = {}
-OperationalStateEventEnum.NO_EVENT = 0x00
-OperationalStateEventEnum.CURRENT_POSITION_EVENT = 0x01
-OperationalStateEventEnum.OPERATIONAL_STATE_EVENT = 0x02
-
-local event_state = OperationalStateEventEnum.NO_EVENT
-local is_opening_closing = false
+local WindowCoveringEventEnum = {
+  NO_EVENT = 0x00,
+  CURRENT_POSITION_EVENT = 0x01,
+  OPERATIONAL_STATE_EVENT = 0x02
+}
 
 local function find_default_endpoint(device, cluster)
   local res = device.MATTER_DEFAULT_ENDPOINT
@@ -62,6 +62,8 @@ local function match_profile(device)
 end
 
 local function device_init(driver, device)
+  device:set_field(EVENT_STATE, WindowCoveringEventEnum.NO_EVENT)
+  device:set_field(IS_MOVING, false)
   if not device:get_field(PROFILE_MATCHED) then
     match_profile(device)
   end
@@ -142,8 +144,8 @@ local function current_pos_handler(driver, device, ib, response)
       ib.endpoint_id, capabilities.windowShadeLevel.shadeLevel(position)
     )
   end
-  if event_state == OperationalStateEventEnum.OPERATIONAL_STATE_EVENT then
-    if is_opening_closing == false then
+  if device:get_field(EVENT_STATE) == WindowCoveringEventEnum.OPERATIONAL_STATE_EVENT then
+    if not device:get_field(IS_MOVING) then
       if position == 0 then
         device:emit_event_for_endpoint(ib.endpoint_id, capabilities.windowShade.windowShade.closed())
       elseif position == 100 then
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -22,6 +22,9 @@ local MatterDriver = require "st.matter.driver"
 local DEFAULT_LEVEL = 0
 local PROFILE_MATCHED = "__profile_matched"
 
+local cap_call_dirty_bit = 0 -- BIT0: Current Position Attribute, BIT1:Operational Status Attribute
+local is_opening_closing = 0
```
yhoyoon, line 26: Could you please use `true` and `false` instead of integer values?

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -22,6 +22,9 @@ local MatterDriver = require "st.matter.driver"
 local DEFAULT_LEVEL = 0
 local PROFILE_MATCHED = "__profile_matched"
 
+local cap_call_dirty_bit = 0 -- BIT0: Current Position Attribute, BIT1:Operational Status Attribute
```
yhoyoon, line 25: Please replace this number with a constant that has a human-readable name explaining the meaning of the number.

greens (commented):
I wonder if the reversal issue could be solved better by including the `reverse` preference in all the window shade profiles and handling it correctly in the driver?

It's a canonical preference that is fully localized.

drivers/SmartThings/matter-window-covering/src/init.lua
greens: also it seems to be used as an enum, so it may be better to define the possible values.

HunsupJung:
This pr is related to [MTR-606](https://smartthings.atlassian.net/browse/MTR-606), [MTR-607](https://smartthings.atlassian.net/browse/MTR-607).

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -22,6 +22,14 @@ local MatterDriver = require "st.matter.driver"
 local DEFAULT_LEVEL = 0
 local PROFILE_MATCHED = "__profile_matched"
 
+local OperationalStateEventEnum = {}
+OperationalStateEventEnum.NO_EVENT = 0x00
+OperationalStateEventEnum.CURRENT_POSITION_EVENT = 0x01
+OperationalStateEventEnum.OPERATIONAL_STATE_EVENT = 0x02
```
greens, line 28:
```suggestion
local OperationalStateEventEnum = {
  NO_EVENT = 0x00
  CURRENT_POSITION_EVENT = 0x01
  OPERATIONAL_STATE_EVENT = 0x02
}
```

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -22,6 +22,14 @@ local MatterDriver = require "st.matter.driver"
...
+
+local event_state = OperationalStateEventEnum.NO_EVENT
+local is_opening_closing = false
```
greens, line 31: would this more accurately be named `is_moving`?

drivers/SmartThings/matter-window-covering/src/init.lua
greens: Alternatively, I believe you could be using `st.matter.clusters.OperationalState.types.OperationalStateEnum` here, which has values for `STOPPED`, `RUNNING`, `PAUSED`, and `ERROR`

greens (changes requested):
This is a lot of complexity to add without including any test cases.

drivers/SmartThings/matter-window-covering/src/init.lua
HunsupJung: Thank you for your recommend. I think OperationStateEventEnum should be changed to EventEnum.

drivers/SmartThings/matter-window-covering/src/init.lua
HunsupJung: I didn't know about it. Thank you~!! I'll modify it.

greens (commented):
Please include some tests for the new behavior.

HunsupJung:
Thank you for your reviewing :) And I'll update the tests. so I want you to review one more time 🙏 Thanks, again~!

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -250,6 +310,36 @@ test.register_coroutine_test(
   end
 )
 
+test.register_coroutine_test(
+  "WindowCovering OperationalStatus partially open", function()
```
greens, line 314: please give your tests unique names

greens (approved):
LGTM

@ctowns do you want to take a look?

PR #1268, opened 2024-03-08
audio TTS fix on samsung speakers

raj-1219:
Ticket: https://smartthings.atlassian.net/browse/BEE-4280

Currently, the notification could only be played with 'http' method. In order to make the TTS work uniformly, direct fallback method has been adopted to make it work without fail.

### DIFF
--- a/drivers/SmartThings/samsung-audio/README.md
+++ b/drivers/SmartThings/samsung-audio/README.md
@@ -1,0 +1,15 @@
+# Samsung-Audio integration documentation
+
+The purpose of this readme is to document the behavior of various samsung lan speaker devices,
+since during the development of the driver there have been some weird behavior.
+The document will also explain any mitigations the driver makes to provide a
+consistent integration despite the weird behavior.
+
+## Device testing notes
+
+Audio Speaker Device shows some wierd behavior while responding to various commands. Due to non uniform response by devices, device state mismatch can be
+observed sometimes in the ST app.
+
+## Samsung Audio TTS Notification
+
+Currently, Audio TTS notification is played with certain non expected behaviors. For example, when the speakers are in paused state, the notification plays but the queued music resumes playing. Similarly, when speakers are in mute state, it will unmute and play the notification but the speakers will remain unmuted after this. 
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/samsung-audio/README.md
```diff
@@ -0,0 +1,15 @@
+# Samsung-Audio integration documentation
+
+The purpose of this readme is to document the behavior of various samsung lan speaker devices,
+since during the development of the driver there have been some weird behavior.
+The document will also explain any mitigations the driver makes to provide a
+consistent integration despite the weird behavior.
+
+## Device testing notes
+
+Audio Speaker Device shows some wierd behavior while responding to various commands. Due to non uniform response by devices, device state mismatch can be
```
cjswedes, line 10: nit:
```suggestion
Audio Speaker Device shows some weird behavior while responding to various commands. Due to non uniform response by devices, device state mismatch can be
```

cjswedes (approved):

drivers/SmartThings/samsung-audio/src/command.lua
```diff
@@ -388,22 +367,7 @@ function Command.play_streaming_uri(ip, uri)
...
-    log.debug("resp method and error code : ", utils.stringify_table(resp))
-    if resp.method == "ErrorEvent" then
-      log.info(string.format("Recieved error code %s",resp.err_code))
-      response_map= fallback_to_http(ip,uri)
-    end
-    if resp.method == "PausePlaybackEvent" then
-      log.info(string.format("Recieved error code %s",resp.err_code))
-      response_map= fallback_to_http(ip,uri)
-    end
+    response_map = fallback_to_http(ip, uri)
```
KishorSarswatSam, line 370: Should we remove the function and just format the URI here? Since we are not directly calling ```handle_http_request()``` directly anyway.

KishorSarswatSam (approved):
Thanks for making the changes @raj-1219

drivers/SmartThings/samsung-audio/src/command.lua
raj-1219: I kept in order to maintain the understanding of how we reached to the conclusion of falling back to http. It is just for understanding.

PR #1278, opened 2024-03-19
Requset to add a fingerprints for Kichler Zigbee fan light

seungkwan-choi:
This is new project for Kichler.
It's a Zigbee fan and the model no. is KICHLER-FANLIGHT-Z-301.
I want to add Kichler to the fingerprints of a Zigbee fan.
The function of the option has been added to the fan light
Please review it.

### DIFF
--- a/drivers/SmartThings/zigbee-fan/profiles/kichler-fan-light.yml
+++ b/drivers/SmartThings/zigbee-fan/profiles/kichler-fan-light.yml
@@ -1,0 +1,67 @@
+name: kichler-fan-light
+components:
+  - id: main
+    label: Fan 
+    capabilities:
+      - id: switch
+        version: 1
+      - id: fanSpeed
+        version: 1
+        config:
+          values:
+            - key: "fanSpeed.value"
+              range: [0, 3]
+      - id: firmwareUpdate
+        version: 1
+      - id: refresh
+        version: 1
+    categories:
+      - name: Fan
+  - id: light
+    label: Light
+    capabilities:
+      - id: switch
+        version: 1
+      - id: switchLevel
+        version: 1
+        config:
+          values:
+            - key: "level.value"
+              range: [0, 100]
+      - id: refresh
+        version: 1
+    categories:
+      - name: Light
+preferences:
+  - name: "trim"
+    title: "Minimum Dimming Level"
+    description: "Set the lowest dimming level"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        10 : "10%"
+        15 : "15%"
+        20 : "20%"
+        25 : "25%"
+      default: 10
+  - name: "breezemode"
+    title: "Breeze Mode"
+    description: "Set the lowest fan speed to breeze mode"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        0 : "Disable"
+        1 : "Enable"
+      default: 0
+  - name: "fandirection"
+    title: "Fan Direction"
+    description: "Set Fan Direction : Cannot change it again for 10 seconds after the change. Please set it again after 10 seconds."
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        0 : "Reverse"
+        1 : "Forward"
+      default: 1
--- a/drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
+++ b/drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
@@ -113,14 +113,10 @@ local function info_changed(driver, device, event, args)
       local newlevel = math.floor((current_level/100.0) * 254 )
       if old_trim < new_trim then
         if new_trim >= current_level then
-          local newlevel = math.floor((new_trim/100.0) * 254 )
-          device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
-        else
-          device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
+          newlevel = math.floor((new_trim/100.0) * 254 )
         end
-      else
-        device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
       end
+      device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
     end
     if device.preferences.breezemode ~= args.old_st_store.preferences.breezemode then
       local speed = device:get_field('LAST_FAN_SPD') or 1
--- a/drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
+++ b/drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
@@ -134,7 +134,7 @@ local function info_changed(driver, device, event, args)
       local time_difference = os.difftime(current_fandirection_time, send_fandirection_time) or 0
       if time_difference >= 10 or send_fandirection_time == 0 then
         device:send(FanControl.attributes.FanMode:write(device, 6))
-        device:set_field('FANDIRECTION_SENDTIME', current_fandirection_time, {persist = true})
+        device:set_field('FANDIRECTION_SENDTIME', current_fandirection_time, {persist = false})
       end
     end
   end
--- a/drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
+++ b/drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
@@ -1,0 +1,187 @@
+-- Copyright 2024 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+local clusters = require "st.zigbee.zcl.clusters"
+local capabilities = require "st.capabilities"
+local FanControl = clusters.FanControl
+local Level = clusters.Level
+local OnOff = clusters.OnOff
+
+local FINGERPRINTS = {
+  { mfr = "KICHLER", model = "KICHLER-FANLIGHT-Z-301" },
+}
+
+local function can_handle_itm_fanlight(opts, driver, device)
+  for _, fingerprint in ipairs(FINGERPRINTS) do
+    if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
+      return true
+    end
+  end
+  return false
+end
+
+-- CAPABILITY HANDLERS
+local function on_handler(driver, device, command)
+  local last_level = device:get_field('LAST_DIM_LEVEL') or 100
+  local level = math.floor((last_level/100.0) * 254 )
+  if command.component == 'light' then
+    device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, level, command.args.rate or 0xFFFF))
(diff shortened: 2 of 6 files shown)


drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
```diff
@@ -0,0 +1,187 @@
...
+      return true
+    end
+  end
+  return false
+end
+
+-- CAPABILITY HANDLERS
+local function on_handler(driver, device, command)
+  local last_level = device:get_field('LAST_DIM_LEVEL') or 100
+  local level = math.floor((last_level/100.0) * 254 )
```
greens, line 36: nit: these should be moved inside the `if` statement

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
```diff
@@ -0,0 +1,187 @@
...
+    if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
+      return true
+    end
+  end
+  return false
+end
+
+-- CAPABILITY HANDLERS
+local function on_handler(driver, device, command)
+  local last_level = device:get_field('LAST_DIM_LEVEL') or 100
```
greens, line 35: as far as I can tell, you maintain the `level` value even when receiving an `off` report. Unless the device separately reports a `0` level when turning off, you should be able to use `device:get_latest_state` instead of setting an additional field here.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
```diff
@@ -0,0 +1,187 @@
...
+  device:send(FanControl.attributes.FanMode:read(device))
+end
+
+local function switch_level_handler(driver, device, command)
+  local trim_level = tonumber(device.preferences.trim) or 10
+  if command.args.level <= trim_level and command.args.level >= 1 then
+    local level = math.floor((trim_level/100.0) * 254 )
+    device:emit_component_event(device.profile.components.light, capabilities.switchLevel.level(command.args.level))
+    device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, level, command.args.rate or 0xFFFF))
+    device:emit_component_event(device.profile.components.light, capabilities.switchLevel.level(trim_level))
```
greens, line 65: What's the intent behind these two `level` events from the same component here?

greens:
Please work to raise the code coverage to at least 90%

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
seungkwan-choi: When i apply the code that was guided to me, if i turn on the light after turning it off, i can't retrieve the last value (off), and instead retrieve the off state. so,  I will continue to use the set/get_field as you are now.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
seungkwan-choi: I will change it.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
seungkwan-choi: This is to use the Trim function. This trim is used to limit the minimum value of dimming.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
greens: were you getting the latest value of the `switch` capability or the `switchLevel` capability?

```
device:get_latest_state("main", capabilities.switchLevel.ID, capabilities.switchLevel.level.NAME)
```

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
greens: I understand the point of the trim function, but it seems like you'd want to only emit the trimmed value, not both the untrimmed and trimmed value.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
```diff
@@ -0,0 +1,187 @@
...
+          local newlevel = math.floor((new_trim/100.0) * 254 )
+          device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
+        else
+          local newlevel = math.floor((current_level/100.0) * 254 )
+          device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
+        end
+      else
+        local newlevel = math.floor((current_level/100.0) * 254 )
+        device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
+      end
```
greens, line 124: this seems unneeded

greens, line 124:
```suggestion
      local newlevel = math.floor((current_level/100.0) * 254 )
      if old_trim < new_trim then
        if new_trim >= current_level then
          local newlevel = math.floor((new_trim/100.0) * 254 )
        end
      device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
```

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
seungkwan-choi: The use of the switchlevel is because it is synchronized with the main switch.
The proposed get_lastest_state maintains "off" if you turn it on after turning it off, as the last state value is "off".
Therefore, we intend to use set/get field instead of using get_lastest_state.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
seungkwan-choi: I would like to refresh the value below the trim setting value so that it is not reflected in the UI. If you set it to trim 25% and control it to 10%, it will be 25%.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
seungkwan-choi: I modified it with the code you suggested

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
greens: So why send both the 10% event and the 25% event? Wouldn't you want to just send the 25% event?

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
```diff
@@ -0,0 +1,183 @@
...
+      if old_trim < new_trim then
+        if new_trim >= current_level then
+          local newlevel = math.floor((new_trim/100.0) * 254 )
+          device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
+        else
+          device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
+        end
+      else
+        device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
+      end
```
greens, line 123:
```suggestion
      if old_trim < new_trim then
        if new_trim >= current_level then
          newlevel = math.floor((new_trim/100.0) * 254 )
        end
      end
      device:send_to_component('light', Level.server.commands.MoveToLevelWithOnOff(device, newlevel, 0))
```

drivers/SmartThings/zigbee-fan/profiles/kichler-fan-light.yml
```diff
@@ -0,0 +1,67 @@
...
+    description: "Set the lowest dimming level"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        10 : "10%"
+        15 : "15%"
+        20 : "20%"
+        25 : "25%"
+      default: 10
```
greens, line 47: In relation to this preference, I'm happy to direct you to this new feature we added, which is the ability to set min/max values for the switch level attribute. You may find that using this is simpler than having to do the trimming yourself.

Here's an example of how to emit this event:

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1220/files#diff-7ab55ddbc6a9760ea5b6f928abee244c3142c74389743ce05b895b938ad81f10R372

The min/max _should_ be displayed and carried across automations as well.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
seungkwan-choi: 24캡처

Set Trim to 25%.
If the user has changed to 10% in the UI, the intention is to express it as 25%.
However, if you adjust 25% and then adjust it to less than that (e.g., 20%), it is to inform you of the event that it has been changed to go back to 25%

drivers/SmartThings/zigbee-fan/profiles/kichler-fan-light.yml
seungkwan-choi: It is expected that there will be models derived from this model. We will try to use the code you suggested at that time. Thank you.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
seungkwan-choi: I changed it with the code you suggested

greens (commented):
This looks good but could you write a test that verifies the setlevel command after setting the trim above 0?

Similarly a test that checks the events that are sent after the trim is set above the current level of the device?

We prefer to have code coverage over 90% and you're currently at 87%.

seungkwan-choi:
The test code has been modified.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
```diff
@@ -0,0 +1,179 @@
...
+    end
+    if device.preferences.fandirection ~= args.old_st_store.preferences.fandirection then
+      local flag_direction = device:get_field('FANDIRECTION_FLAG') or 0
+      local send_fandirection_time = device:get_field('FANDIRECTION_SENDTIME') or 0
+      local current_fandirection_time = os.time()
+      local time_difference = os.difftime(current_fandirection_time, send_fandirection_time) or 0
+      if time_difference >= 10 or send_fandirection_time == 0 then
+        device:send(FanControl.attributes.FanMode:write(device, 6))
+        device:set_field('FANDIRECTION_SENDTIME', current_fandirection_time, {persist = true})
```
greens, line 138: this value doesn't need to persist

it takes more than 10s for the hub to reboot

seungkwan-choi: The purpose of using the following code, regardless of the hub, is to prevent sending commands if a command has been sent within 10 seconds. (This is related to protecting the fan driving circuit.)

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
```diff
@@ -0,0 +1,179 @@
...
+      else
+        device:set_field('FANDIRECTION_FLAG', flag_direction + 1 , {persist = true})
```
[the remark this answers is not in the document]
seungkwan-choi: The test code has been modified.

drivers/SmartThings/zigbee-fan/src/kichler-fan-light/init.lua
greens: I understand, but the `persist = true` makes the value durable through hub restarts.

Because hub restarts take more than 10s, you do not need to have this flag persist, because it is guaranteed to have been more than 10s since you last sent the write on hub restart. Just keeping it in memory should be enough.

wkhenon (approved):

greens (approved):

PR #1283, opened 2024-03-20
Add DeepSmart 

refuseRed:
Add DeepSmart


drivers/SmartThings/deepsmart/src/client/httpclient.lua
```diff
@@ -0,0 +1,251 @@
+local log = require('log')
+local cosock = require "cosock"
+local http = cosock.asyncify "socket.http"
+local ltn12 = require('ltn12')
+local RestClient = require "lunchbox.rest"
+local utils = require "utils.utils"
+local json = require('dkjson')
```
varzac, line 7: You should use `st.json` instead of `dkjson` for this functionality in drivers

drivers/SmartThings/deepsmart/src/client/httpclient.lua
```diff
@@ -0,0 +1,251 @@
...
+      client:shutdown()
+      self.client = RestClient.new("https://"..self.ip, HttpClient.labeled_socket_builder(self.wiser_index_code))
+  end
+  return process_rest_response(response,err,partial)
+end
+
+function HttpClient:do_post(url, content)
+  -- get url
+  local client = self.client
+  if (client == nil) then
```
varzac, line 78: It seems like there should be an early return in the case client is nil and not just a warn log as you use `client` below.

cjswedes (commented):
I haven't reviewed everything, as I expect some changes before I can go too much further. I was hoping that we could get a high level description of how these devices are supposed to work. You could add this in a README file, or block comments in the relevant code files. Here are the questions that I think are helpful for anyone to know before trying to review the code:
* How are devices discovered? Discover the bridge and then other devices via communication with the bridge or is everything over ssdp?
* What does the bridge do for the other devices, and how does the driver interact with the bridge?
* How are capability commands handled? (i.e. communication via bridge, individual devices, over https?, etc)
* How are manual updates that happen to the devices reported to the driver? (is it polling?)
* Is there a deepsmart api reference somewhere?
* What's a wiser? What's dp2knx/dpenum?

drivers/SmartThings/deepsmart/src/client/httpclient.lua
refuseRed: We are already modifying the test

drivers/SmartThings/deepsmart/profiles/Ac.yml
```diff
@@ -0,0 +1,22 @@
+name: Ac.v1
+components:
+- id: main
+  capabilities:
+  - id: switch
+    version: 1
+  - id: temperatureMeasurement
+    version: 1
+  - id: thermostatHeatingSetpoint
```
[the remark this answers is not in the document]
refuseRed: We are already modifying the test

drivers/SmartThings/deepsmart/src/ssdp.lua
```diff
@@ -0,0 +1,88 @@
+local socket = require('socket')
+local cosock = require "cosock"
+local log = require('log')
+local config = require('config')
+local utils = require('utils.utils')
+
+
+DEEPSMART_SSDP_SEARCH_TERM = "DEEPSMART-ARM"
```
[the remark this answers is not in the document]
refuseRed: We are already modifying the test

drivers/SmartThings/deepsmart/profiles/Ac.yml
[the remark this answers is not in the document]
refuseRed: Accessing bridge through HTTPS to obtain devices

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,776 @@
...
+local Devices = require('deepsmart.devices')
+local dp2knx = require('deepsmart.dp2knx')
+local dpenum = require('deepsmart.dpenum')
+local utils = require('utils.utils')
+local config = require('config')
+local ssdp = require('ssdp')
+local Wisers = {}
+
+
+Wisers.wisers = {}
```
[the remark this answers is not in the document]
refuseRed: it's bridge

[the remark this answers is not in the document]
cjswedes: Can you explain how the wisers and the devices work together on the network? Is it expected that there are more than 1 wiser/bridge? Does the driver only communicate with the wiser/bridge, or does it communicate with the devices as well?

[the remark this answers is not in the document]
lelandblue: Hello @refuseRed Can you please see the question above when your schedule allows?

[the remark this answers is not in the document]
refuseRed: Support multiple gateway access SSDP searches for all gateways within the local area network. It obtains all devices under the gateway through HTTPS and obtains the properties of all current devices through HTTPS. Then, every 2 seconds, it obtains devices with changed properties in the gateway through HTTPS. If there are any, it obtains all properties of the changed devices through HTTPS. Control is also achieved through HTTPS controlling sub devices through the gateway

drivers/SmartThings/deepsmart/profiles/Ac.yml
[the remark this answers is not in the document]
refuseRed: SSDP searches for all gateways within the local area network. It obtains all devices under the gateway through HTTPS and obtains the properties of all current devices through HTTPS. Then, every 2 seconds, it obtains devices with changed properties in the gateway through HTTPS. If there are any, it obtains all properties of the changed devices through HTTPS. 
Control is also achieved through HTTPS controlling sub devices through the gateway.
Support multiple gateway access.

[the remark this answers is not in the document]
refuseRed: 1 discover the bridge and use https api to get devices via the bridge
2 driver uses https api(get all devices/control device/query device) to interact with bridge
3 via bridge over https
4 driver polls the updated devices every 2 seconds, and then query polled devices over https
4 no, 6 apis
 1)  get all devices in the bridge 
 GET https://ip/homecontroller/api/v1/config/devices
 2) get device configs
 GET https://ip/homecontroller/api/v1/config/dp2knx
3) get device enum configGET https://ip/homecontroller/api/v1/config/dpsamenum
4) get updated devices since given time
POST https://ip/homecontroller/api/v1/config/changeddevs{"command":"getchangeddevs","params":{"time":"1712110490"}}
5) query device(addr is knx address like a/b/c)
POST https://ip/homecontroller/api/v1/query
{"command":"query","addrs":[1,2,3]}
6) control device
POST https://ip/homecontroller/api/v1/control
{"command":"control","cmslist":[{"addr":1,"type":0,"value":1,"delay":0}]}

6 bridge
   dp2knx is device config
   devices in bridge are all confided by TSL(things specification language)
   pid1 is one kind of ac
  pid2 is another kind of ac
   but in St pid1&&pid2 are same so we make the dp2knx to figure out how different ACs connected to st
  dpenum is properties config. for example, ac has several  modes(hot wind cold dry auto wet…), pid1 may have the value 0,1,2,3,4,5…
but pid2 may have the value 1,3,5,7,9,11…
we use dpenum to make different acs can be controlled by the same st mode value

drivers/SmartThings/deepsmart/src/lifecycles.lua
```diff
@@ -0,0 +1,68 @@
...
+    local supported_fan_modes = {'low','medium','high','auto'}
+    device:emit_event(caps.airConditionerFanMode.supportedAcFanModes(supported_fan_modes, { visibility = { displayed = false } }))
+  end
+  -- Once device has been created
+  -- at API level, poll its state
+  -- via refresh command and send
+  -- request to share server's ip
+  -- and port to the device os it
+  -- can communicate back.
+  commands.refresh(nil, device)
```
cjswedes, line 51:
```suggestion
  driver:inject_capability_command(device, {
    capability = capabilities.refresh.ID,
    command = capabilities.refresh.commands.refresh.NAME,
    args = {}
  })
```

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,825 @@
...
+  -- init all wisers
+  for i,v in pairs(wisers) do
+    log.info('add wiser '..v.uuid..' ip:port('..v.ip..':'..v.port..')')
+    local wiser = Wisers.get_wiser(v.uuid)
+    if (not is_add and wiser == nil) then
+      log.trace('ignore new wiser '..v.uuid)
+      goto continue
+    end
+    -- first check wiser ip, if ip is same then do nothing
+    if (wiser ~= nil and wiser.ip == v.ip) then
```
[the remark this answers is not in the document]
refuseRed: Unified response：
1 use string.match to extract uuid
2 port is not used. delete it
3  it happens when dhcp changes the bridge's ip or mannul static ip change or router changes the bridge's network segment
4 check uuid nil and ip must be valid ipv4
5 correct the code comments. in init no ssdp search is called. the driver load wisers and devices from datastore
6 wiser loop will poll updated devices by given time every 2 seconds. 
7 fix it. bridge device do not reload fevices&&config, it uses the devices&&config loaded from wiser's datastore, only discover or refresh will reload devixes&&config. edge devices will query properties from bridge
8 use the data saved when first added to check it's device or wiser

drivers/SmartThings/deepsmart/src/lifecycles.lua
refuseRed: comment modified

drivers/SmartThings/deepsmart/src/discovery.lua
```diff
@@ -0,0 +1,51 @@
+local log = require('log')
+local Wisers = require "deepsmart.wisers"
+local ssdp = require('ssdp')
+local utils = require('utils')
+local disco = {}
+-----------------------
+disco.ssdp_discovery_callback = function(uuid, ip, is_add)
+  if (not utils.is_valid_ipv4(ip) or uuid == nil) then
+    log.warn('ip or uuid is invalid')
```
cjswedes, line 9:
```suggestion
    log.warn(string.format('ip (%s) or uuid (%s) is invalid', ip, uuid))

```

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -338,30 +334,30 @@ function Wisers.add_wisers(wisers, is_add)
...
     Wisers.driver.datastore[v.uuid].wiser_index_code = v.uuid
     wiser["indexCode"] = v.uuid
     wiser["wiser_index_code"] = v.uuid
     wiser["last_active_time"] = "0"
     -- deepsmart wiser id is 8bytes integer
     wiser["wiser_bytes"] = utils.tonumber(v.uuid)
     wiser["ip"] = v.ip
-    wiser["port"] = v.port
-    wiser["api"] = api.client(v.uuid, v.ip)
+    --wiser["api"] = api.client(v.uuid, v.ip)
```
cjswedes, line 360: whats the reason for re-creating the rest client when needed rather than storing it on the object? With some of the repeated calls to get updated devices every 2 seconds, I would think this is something we want to keep around rather than pay the penalty of re-creating and then garbage collecting the tables all the time.

drivers/SmartThings/deepsmart/src/lifecycles.lua
```diff
@@ -19,19 +24,34 @@ function lifecycle_handler.init(driver, device)
...
+  })
   end
 end
 
 function lifecycle_handler.added(driver, device)
   log.info('device '..device.id..' added')
   if (device.parent_assigned_child_key == nil) then
+    log.info('bridge device '..device.id..' added')
+    -- for bridge save some infos to bridge device
+    device:set_field(config.FIELD.TYPE, "bridge", { persist = true })
```
cjswedes, line 41: Since `FIELD.TYPE` is only one of two values set based on parent_assigned_child_key, I think we should check that field rather than adding another field to every device.

cjswedes: `FIELD.UUID` also seems unnecessary since the information is already present on all device objects.

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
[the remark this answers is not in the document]
cjswedes: @refuseRed can you apply your responses in the comments where they are relevant. I think there was some correlation, but I lost it after 4/5. It is just easier to discuss individual comments that way. I have gone through and resolved comments that I can clearly see were addressed.

drivers/SmartThings/deepsmart/src/utils.lua
```diff
@@ -230,6 +230,20 @@ function utils.deep_table_eq(tbl1, tbl2)
   return false
 end
 
+-- check if valid ipv4
+function utils.is_valid_ipv4(ip)
```
cjswedes, line 234: This function already exists in our `net_utils` module: `require "net_utils"`
```lua
--- Validate dotted-decimal IPv4 address string
---@param ipv4_addr string address to validate
---@return true if valid IPv4 address, false otherwise
function net_utils.validate_ipv4_string(ipv4_addr)
```

drivers/SmartThings/deepsmart/src/discovery.lua
```diff
@@ -1,12 +1,13 @@
...
-local utils = require('utils')
+local net_utils = require('st.net_utils')
+
 local disco = {}
 -----------------------
 disco.ssdp_discovery_callback = function(uuid, ip, is_add)
-  if (not utils.is_valid_ipv4(ip) or uuid == nil) then
-    log.warn('ip or uuid is invalid')
+  if (not net_utils.validate_ipv4_string(ip) or uuid == nil) then
+    log.warn(string.format('ip (%s) or uuid (%s) is invalid', (ip or "nil"), (uuid or "nil")))
```
cjswedes, line 10: nit: nil will be formatted by `%s` as the string "nil"
```suggestion
    log.warn(string.format('ip (%s) or uuid (%s) is invalid', ip, uuid ))
```

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,825 @@
...
+  return dev_id,idx
+end
+---------------
+-- Parse protocol
+function Wisers.get_wiser(wiser_index_code)
+  return Wisers.wisers[wiser_index_code]
+end
+
+-------------
+-- it runs every 30s
```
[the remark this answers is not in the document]
cjswedes: @refuseRed I will not approve if this is left in.

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,825 @@
...
+      wiser["dp2knx"] = dp2knx.load_config(v.uuid, nil)
+      wiser["dpenum"] = dpenum.load_config(v.uuid, nil)
+      wiser.state = 1
+      wiser.ctrlmap = {}
+      Wisers.driver:call_on_schedule(
+      2,
+      function ()
+        return Wisers.wiser_loop(v.uuid)
+      end,
+      'wiser loop')
```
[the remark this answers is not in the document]
cjswedes: @refuseRed I am trying to understand whether the 2 second polling is something we are forced to do because that is how the devices are designed, or if it was a design choice.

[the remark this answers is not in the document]
refuseRed: wiser loop will poll updated devices by given time every 2 seconds. wiser loop will poll updated devices by given time every 2 seconds.

drivers/SmartThings/deepsmart/src/lifecycles.lua
```diff
@@ -0,0 +1,68 @@
...
+  log.info('device '..device.id..' init')
+  if (device.parent_assigned_child_key ~= nil) then
+    log.debug('save dev '..device.id..' networkid '..device.parent_assigned_child_key)
+    wisers.idmap[device.parent_assigned_child_key] = device.id
+    -------------------
+    -- Set up scheduled
+    -- services once the
+    -- driver gets
+    -- initialized.
+    -- Refresh schedule
```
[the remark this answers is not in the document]
refuseRed: fix it. bridge device do not reload fevices&&config, it uses the devices&&config loaded from wiser's datastore, only discover or refresh will reload devixes&&config. edge devices will query properties from bridge

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,825 @@
...
+    end
+  end
+  log.info('refresh wiser devices success')
+  return true,nil
+end
+
+------------
+-- init wisers mgr
+-- 1 init default dp2knx&&dpenum
+-- 2 discover wisers && add wisers
```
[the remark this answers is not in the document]
refuseRed: correct the code comments. in init no ssdp search is called. the driver load wisers and devices from datastore。
 wiser loop will poll updated devices by given time every 2 seconds.

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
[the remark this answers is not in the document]
refuseRed: check uuid nil and ip must be valid ipv4

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,825 @@
...
+        label = "DEEPSMART KNX bridge-"..v.uuid,
+        profile = "Deepsmart.bridge",
+        manufacturer = "DEEPSMART",
+        model = "KNX gateway",
+        vendor_provided_label = "DEEPSMART KNX bridge"
+      }
+      log.trace('create bridge device')
+      Wisers.driver:try_create_device(wiser_device_msg)
+      log.trace('after bridge device')
+      local devices = Wisers.driver:get_devices()
```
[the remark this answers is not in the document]
cjswedes: @refuseRed ^^

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,825 @@
...
+    if (wiser ~= nil and wiser.ip == v.ip) then
+      log.trace('wiser '..v.uuid..' is same')
+      goto continue
+    end
+    -- if wiser is wiser then create new wiser
+    if (wiser == nil) then
+      Wisers.wisers[v.uuid] = {}
+      wiser = Wisers.wisers[v.uuid]
+      wiser["dp2knx"] = dp2knx.load_config(v.uuid, nil)
+      wiser["dpenum"] = dpenum.load_config(v.uuid, nil)
```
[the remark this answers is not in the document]
cjswedes: @refuseRed ^^

drivers/SmartThings/deepsmart/src/init.lua
```diff
@@ -0,0 +1,95 @@
...
+  end
+  if (temp ~= nil) then
+    log.info('device '..device.parent_assigned_child_key..' report temp '..temp)
+    device:emit_event(caps.temperatureMeasurement.temperature({value=temp,unit='C'}))
+  end
+  return true,nil
+end
+
+
+wisers.init(driver, discovery)
```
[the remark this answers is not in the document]
cjswedes: @refuseRed Do you have any reason for why initialization is not being done on a device by device basis during the init/added handlers?

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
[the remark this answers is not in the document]
cjswedes: Is the approach to poll the wisers/bridges something that is required by the wisers, or is it a design choice you made? Does there exist any other mechanism for having the device tell the driver of device updates without the driver having to poll for them (i.e. websocket, sse, other listening server setup, etc)?

cjswedes:
The main two issues I see which need to be addressed are:
1. the periodic multicast discovery running every 30 seconds
2. the way devices are initialized when first added and when the driver starts up do not follow existing patterns where it is done on a device by device basis in the init/added lifecycle handlers. See the Hue (most relevant), Sonos, or Bose drivers for example on how this works.

refuseRed:
> The main two issues I see which need to be addressed are:
> 
> 1. the periodic multicast discovery running every 30 seconds
> 2. the way devices are initialized when first added and when the driver starts up do not follow existing patterns where it is done on a device by device basis in the init/added lifecycle handlers. See the Hue (most relevant), Sonos, or Bose drivers for example on how this works.

1 periodic discovery is used to check whether the bridge is alive and the ip is same with previous ip
2 delete old wisers.init() function. bridge and devices are all initialized in lifecycle's init function

drivers/SmartThings/deepsmart/src/commands.lua
```diff
@@ -93,6 +93,23 @@ function command_handler.set_thermostat_mode(driver, device, command)
...
+----------------
+function command_handler.set_thermostat_mode(driver, device, command)
+  -- send command
+  local success = wisers.control(device, command, {config.HEATER.ONOFF})
+  -- Check if success
+  if success then
+    if command.arg.mode == 'off' then
+      return device:emit_event(caps.switch.switch.off())
+    end
+    return device:emit_event(caps.switch.switch.on())
```
cjswedes, line 107: Why are switch events emitted here? The app will timeout on set thermostat mode commands if no thermostat mode event is emitted, which is a bad user experience. Perhaps either ThermostatMode should not be a part of the device profile? https://developer.smartthings.com/docs/devices/capabilities/capabilities-reference/#thermostatMode

cjswedes (commented):
Using ssdp to detect ip changes is fine, but using it to check online/offline status for the bridge is not. We do ip change detection on a 10 minute interval for sonos, and would like to see this interval used here. Online/offline state determination can be done by checking the status of the devices (which happens every 2 seconds); if those requests fail, then the device is offline.

drivers/SmartThings/deepsmart/src/commands.lua
liuyd96: > Perhaps either ThermostatMode should not be a part of the device profile? https://developer.smartthings.com/docs/devices/capabilities/capabilities-reference/#thermostatMode

Hi @cjswedes 

profiles/Heater.yml is used for Underfloor Heating Thermostats devices.
So we set category of profile to *Thermostat*, we also noticed [Required-capabilities](https://developer.smartthings.com/docs/certification/required-capabilities) url.

There are 4 required-capabilities of *Thermostat*:

1. temperatureMeasurement
2. thermostatMode
3. thermostatCoolingSetpoint or thermostatHeatingSetpoint or both
4. thermostatOperatingState 

So I think we shouldn't remove "ThermostatMode" from profile.

@refuseRed 

The overall structure is that,    

ST cloud <----Send Capability----> ST Hub <----Call REST API----> DeepSmart bridge -- Bridged devices.

In most cases we should convert the capability to DeepSmart restAPI in capability handler.

"ThermostatMode" has supported *off* and *heat* commands already, 
I think we could replace "switch" capability with above 2 commands.

Thanks.

drivers/SmartThings/deepsmart/src/lifecycles.lua
```diff
@@ -13,28 +13,26 @@ function lifecycle_handler.init(driver, device)
...
+  if (is_bridge) then
+    -- just add wiser(bridge) to wisers
+    wisers.add_wiser(device.device_network_id, device:get_field(config.FIELD.IP), device)
+  else
     log.debug('save dev '..device.id..' networkid '..device.parent_assigned_child_key)
     wisers.idmap[device.parent_assigned_child_key] = device.id
+    -- for heater set mode to heat
+    local devtype = wisers.get_dev_type(device.parent_assigned_child_key)
+    if devtype == config.ENUM.HEATER then
+      device:emit_event(caps.thermostatMode.thermostatMode.heat())
```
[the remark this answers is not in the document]
refuseRed: 1、 heater delete switch capability. use thermostat mode to control on/off
2、 discovery schedule for check ip move to lifestyle's bridge added/inited

drivers/SmartThings/deepsmart/src/lifecycles.lua
```diff
@@ -0,0 +1,68 @@
...
+    -- driver gets
+    -- initialized.
+    -- Refresh schedule
+    log.info('device '..device.parent_assigned_child_key..' in lifecycle init')
+  end
+end
+
+function lifecycle_handler.added(driver, device)
+  log.info('device '..device.id..' added')
+  if (device.parent_assigned_child_key == nil) then
```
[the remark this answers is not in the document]
refuseRed: discovery schedule for check ip move to lifestyle's bridge added/inited

drivers/SmartThings/deepsmart/src/lifecycles.lua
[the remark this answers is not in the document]
refuseRed: heater delete switch capability. use thermostat mode to control on/off

drivers/SmartThings/deepsmart/src/lifecycles.lua
```diff
@@ -16,11 +16,23 @@ function lifecycle_handler.init(driver, device)
...
+    if (wisers.ip_check_state == nil) then
+      wisers.ip_check_state = 1
+      -- start ssdp schedule to check bridge ip
+      driver:call_on_schedule(
+      600,
+      function ()
+        return wisers.check_ip(discovery)
+      end,
+      'Check ip schedule')
+    end
```
cjswedes, line 29: This can exist outside the handler; you dont need to prevent it from running when there isn't any devices, because the driver should be running only when there are device, or when it is actively trying to discover the first device (which will complete before the 10 minutes is up).

The benefit of that is you dont need `wisers.ip_check_state` field to prevent it from being setup twice for the first device, and it clearer that only one task is being run in the driver (rather than a task per device).

drivers/SmartThings/deepsmart/src/commands.lua
```diff
@@ -97,9 +97,18 @@ end
...
 ----------------
 function command_handler.set_thermostat_mode(driver, device, command)
-  -- just return success as mode is only heat
-  if command.arg.mode == 'heat' then
-    return device:emit_event(caps.thermostatMode.thermostatMode.heat())
+  -- send command
+  local success = wisers.control(device, command, {config.HEATER.ONOFF})
+  -- Check if success
+  if success then
+    if on_off == 'off' then
```
cjswedes, line 104: where is `on_off` defined?

drivers/SmartThings/deepsmart/src/deepsmart/dp2knx.lua
```diff
@@ -0,0 +1,177 @@
...
+  -- for pid n0ykidbbn3mqvxbx, it will split to 3 switchs to samsung
+  -- so idx(1,2,3) is used for sub devices
+  local pid2 = {pid='cekfhkz5',type=config.ENUM.AC, devs={[1]={[1]=0,[4]=1,[5]=2,[2]=3,[3]=4,[101]=5}}}
+  local pid4 = {pid='drrearrq',type=config.ENUM.HEATER, devs={[1]={[1]=0,[24]=2,[16]=3}}}
+  local pid53 = {pid='w35ineur',type=config.ENUM.NEWFAN, devs={[1]={[1]=0,[12]=1}}}
+  Dp2Knx.pids[pid2.pid] = pid2
+  Dp2Knx.pids[pid4.pid] = pid4
+  Dp2Knx.pids[pid53.pid] = pid53
+  return 0
+end
```
cjswedes, line 29:
```suggestion
-- for pid n0ykidbbn3mqvxbx, it will split to 3 switchs to samsung
-- so idx(1,2,3) is used for sub devices
Dp2Knx.pids = {
  ['cekfhkz5'] = {pid='cekfhkz5',type=config.ENUM.AC, devs={[1]={[1]=0,[4]=1,[5]=2,[2]=3,[3]=4,[101]=5}}},
  ['drrearrq'] = {pid='drrearrq',type=config.ENUM.HEATER, devs={[1]={[1]=0,[24]=2,[16]=3}}},
  ['w35ineur'] = {pid='w35ineur',type=config.ENUM.NEWFAN, devs={[1]={[1]=0,[12]=1}}}
}
```
The init function for dp2kn and dpenum can both be removed

drivers/SmartThings/deepsmart/src/commands.lua
refuseRed: move it to init.lua and delete ip_check_state

drivers/SmartThings/deepsmart/src/deepsmart/dp2knx.lua
refuseRed: delete init function and init in default constructor

drivers/SmartThings/deepsmart/src/lifecycles.lua
refuseRed: move it to init.lua and delete ip_check_state

drivers/SmartThings/deepsmart/src/commands.lua
refuseRed: use command.argument.mode

drivers/SmartThings/deepsmart/src/lifecycles.lua
cjswedes: This should be removed from the init lifecycle handler as well.

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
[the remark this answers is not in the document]
liuyd96: Hi @cjswedes 

> Is the approach to poll the wisers/bridges something that is required by the wisers, or is it a design choice you made?

I thought it's a design choice. I'll try to explain the background clearly in my reply below. 

> Does there exist any other mechanism for having the device tell the driver of device updates without the driver having to poll for them (i.e. websocket, sse, other listening server setup, etc)?

DeepSmart bridge supports `http/https/websocket/socket`

We (SRC-N and HQ) did some simple codes review before DeepSmart submitted this PR.
In the code version before the PR submission, DeepSmart use `http` and `socket` to communicate with their bridge.

We thought it's not safe so we required DeepSmart to use TLS to encrypt communication.
Since their bridge doesn't support WWS, so `https` is the only choice. 

But replacing `socket` with `https` introduced a problem that bridged devices can't update their status to ST Station actively. To solve this problem, DeepSmart team added an feature to log all updated devices in DeepSmart bridge.
The updated devices list is exposed through an https url.

Changed Dev <--- Report to DeepSmart Bridge ---> DeepSmart <--- HTTPS URL (Changed dev list) ---> ST Hub

So ST HUB needs to query changed dev list actively. That is why we need to run  `Wisers.wiser_loop` periodically.

drivers/SmartThings/deepsmart/src/lifecycles.lua
refuseRed: remove it

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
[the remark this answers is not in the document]
refuseRed: fix it. return after try_create_device. in lifecycle's added function we call the function add_wiser to create the new wiser

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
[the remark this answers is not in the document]
cjswedes: Thank you for the context. Its a shame to do a polling approach when something else exists, but since WWS isn't available on the device and the tls encryption is required, I understand why you are doing the polling approach.

Ill admit that its weird wws isn't supported by the devices since the tls handshake is supported; my understanding is that wws just requires the tls handshake prior to doing the websocket handshake. I know the hub supports it, and sonos is an example of our use of it. ¯\\_(ツ)_/¯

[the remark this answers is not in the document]
liuyd96: Hi, @cjswedes 

Thanks for your quick response. 

> Ill admit that its weird wws isn't supported by the devices since the tls handshake is supported; my understanding is that wws just requires the tls handshake prior to doing the websocket handshake.

Yes. I'll confirm it with DeepSmart asap. 

In the previous discussion with DeepSmart, we saw that only the above 4 communications methods `(http/https/websocket/socket)` were supported, so we asked DeepSmart to change `socket` to `https` directly.
We have no further confirmation regarding `WWS` support issues. This is indeed an omission.

However, even if DeepSmart Bridge truly supports the feature, it might be too late to make changes according to the current schedule. If possible, could we consider iterating on this in the future ?

CC @lelandblue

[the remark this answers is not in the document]
cjswedes: > However, even if DeepSmart Bridge truly supports the feature, it might be too late to make changes according to the current schedule. If possible, could we consider iterating on this in the future ?

I would be fine with iterating on it and providing secure websocket support for manual state updates in the future.

liuyd96:
Hi @refuseRed 

Please move your driver dir `deepsmart` from `SmartThingsEdgeDrivers/drivers` to `SmartThingsEdgeDrivers/`
Just like `SmartThingsEdgeDrivers/drivers/Aqara`

drivers/SmartThings/deepsmart/src/deepsmart/wisers.lua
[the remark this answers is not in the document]
liuyd96: > > However, even if DeepSmart Bridge truly supports the feature, it might be too late to make changes according to the current schedule. If possible, could we consider iterating on this in the future ?
> 
> I would be fine with iterating on it and providing secure websocket support for manual state updates in the future.

Hi @cjswedes 

I've confimed with DeepSmart that their Bridge does not support `Websocket over TLS`. 
The library they currently use for firmware development does not support it.

refuseRed:
> Hi @refuseRed
> 
> Please move your driver dir `deepsmart` from `SmartThingsEdgeDrivers/drivers` to `SmartThingsEdgeDrivers/` Just like `SmartThingsEdgeDrivers/drivers/Aqara`

Ok, adjusted

cjswedes (commented):
I am good with merging this; only 2 more things.
1. The folder structure is currently `/drivers/DeepSmart/src/` but I think we should match the Aqara structure where there is an overarching directory for deepsmart: `/driver/DeepSmart/deepsmart/src`
2. Commits should be squashed down to a single commit prior to merging

Once those things are done I will approve and @lelandblue can merge when appropriate.

liuyd96:
Hi @refuseRed 

> Commits should be squashed down to a single commit prior to merging, as well as be rebased on top of the upstream branch

1. Add upstream repo
```bash
❯ git remote add upstream https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers.git
```

2. Rebase your branch with upstream
```bash
❯ git pull --rebase upstream main
From https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers
 * branch              main       -> FETCH_HEAD
Current branch dev is up to date.
```

3. Check commit info
```bash
❯ git log --oneline
9c002b5d (HEAD -> dev) optimization
0f6fed98 deepsmart Change to DeepSmart
7def2870 update
43dc63d1 update
3e5c48a9 update
6412b586 update
c6eb992c update
37e9f705 optimization
a92744c8 optimization
57d8862b optimization
3b3f9d2e optimization
7cf8fce6 optimization
1e3a5c38 optimization
485ada2e update
1c21ce9c optimization
242af490 Add DeepSmart Hub
90206d9f (upstream/main, main) Merge pull request #1362 from SmartThingsCommunity/bugfix/multi-garage <------ This is upstream base commit
---snip---
```

4. Squash your commit
```bash
❯ git rebase -i 90206d9f

 1 pick 242af490 Add DeepSmart Hub
  2 pick 1c21ce9c optimization
  3 pick 485ada2e update
  4 pick 1e3a5c38 optimization
  5 pick 7cf8fce6 optimization
  6 pick 3b3f9d2e optimization
  7 pick 57d8862b optimization
  8 pick a92744c8 optimization
  9 pick 37e9f705 optimization
 10 pick c6eb992c update
 11 pick 6412b586 update
 12 pick 3e5c48a9 update
 13 pick 43dc63d1 update
 14 pick 7def2870 update
 15 pick 0f6fed98 deepsmart Change to DeepSmart
 16 pick 9c002b5d optimization
 17
 18 # Rebase 90206d9f..9c002b5d onto 9c002b5d (16 commands)
 19 #
 ---snip---
```

5. Edit pick to 's' from line 2 to line 16, then save
```bash
1 # This is a combination of 16 commits.
  2 # This is the 1st commit message:
  3
  4 Add DeepSmart Hub
  5
  6 Signed-off-by: “liuzhonghua” <m17316396515@163.com>
  7
  8 # This is the commit message #2:
  9
 10 optimization
 11
 12 Signed-off-by: “liuzhonghua” <m17316396515@163.com>
 13
 14 # This is the commit message #3:
 15
 16 update
 17
 18 Signed-off-by: “liuzhonghua” <m17316396515@163.com>
 19
 20 # This is the commit message #4:
 21
 22 optimization
 23
 24 Signed-off-by: “liuzhonghua” <m17316396515@163.com>
 25
 26 # This is the commit message #5:
 27
 28 optimization
 29
 30 Signed-off-by: “liuzhonghua” <m17316396515@163.com>
 31
 32 # This is the commit message #6:
 33
 34 optimization
 35
 36 Signed-off-by: “liuzhonghua” <m17316396515@163.com>
```

6. Squash commit info, then save.

Now you have squashed commit successfully.
```
❯ git log --oneline
38eaad6c (HEAD -> dev) Add DeepSmart Hub
90206d9f (upstream/main, main) Merge pull request #1362 from SmartThingsCommunity/bugfix/multi-garage

```

cjswedes (approved):

cjswedes:
Looks like luacheck isn't passing. We would like to keep this passing, so please make the changes to address it. You can view the errors in CI, but here they are as well:
```
Checking drivers/DeepSmart/deepsmart/src/deepsmart/api.lua 1 warning

    drivers/DeepSmart/deepsmart/src/deepsmart/api.lua:185:5: (W511) unreachable code

Checking drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua 4 warnings

    drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua:2:7: (W211) unused variable socket
    drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua:8:7: (W211) unused variable utils
    drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua:63:34: (W113) accessing undefined variable wiser_index_code
    drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua:65:24: (W113) accessing undefined variable wiser_index_code

Checking drivers/DeepSmart/deepsmart/src/discovery.lua 1 warning

    drivers/DeepSmart/deepsmart/src/discovery.lua:25:26: (W614) trailing whitespace in a comment

Checking drivers/DeepSmart/deepsmart/src/init.lua 2 warnings

    drivers/DeepSmart/deepsmart/src/init.lua:4:7: (W211) unused variable dp2knx
    drivers/DeepSmart/deepsmart/src/init.lua:5:7: (W211) unused variable dpenum

Checking drivers/DeepSmart/deepsmart/src/lifecycles.lua 3 warnings

    drivers/DeepSmart/deepsmart/src/lifecycles.lua:1:7: (W211) unused variable commands
    drivers/DeepSmart/deepsmart/src/lifecycles.lua:13:46: (W614) trailing whitespace in a comment
    drivers/DeepSmart/deepsmart/src/lifecycles.lua:42:59: (W612) line contains trailing whitespace
```

You can run luacheck locally by [installing with luarocks](https://luarocks.org/modules/mpeterv/luacheck), and then using this command from the `SmartThingsEdgeDrivers/` directory: `luacheck --config .github/workflows/.luacheckrc drivers/**`

liuyd96:
Hi @refuseRed 

There‘s still a warning needs to be solved.
```
❯ luacheck --config .github/workflows/.luacheckrc drivers/DeepSmart/*
Checking drivers/DeepSmart/deepsmart/src/commands.lua OK
Checking drivers/DeepSmart/deepsmart/src/config.lua OK
Checking drivers/DeepSmart/deepsmart/src/deepsmart/api.lua 1 warning

    drivers/DeepSmart/deepsmart/src/deepsmart/api.lua:173:3: (W512) loop is executed at most once

Checking drivers/DeepSmart/deepsmart/src/deepsmart/devices.lua OK
Checking drivers/DeepSmart/deepsmart/src/deepsmart/dp2knx.lua OK
Checking drivers/DeepSmart/deepsmart/src/deepsmart/dpenum.lua OK
Checking drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua OK
Checking drivers/DeepSmart/deepsmart/src/discovery.lua OK
Checking drivers/DeepSmart/deepsmart/src/init.lua OK
Checking drivers/DeepSmart/deepsmart/src/lifecycles.lua OK
Checking drivers/DeepSmart/deepsmart/src/ssdp.lua OK
Checking drivers/DeepSmart/deepsmart/src/utils.lua OK

Total: 1 warning / 0 errors in 12 files

```

drivers/DeepSmart/deepsmart/src/lifecycles.lua
```diff
@@ -0,0 +1,83 @@
...
+  -- report test data
+  --wisers.default_report(driver, device)
+  local is_bridge = wisers.is_device_bridge(device)
+  log.info('device '..device.id..' in init')
+  -- for bridge do not reload devices&&config
+  -- just use the devices&&config loaded from wiser's datastore
+  -- only discover or refresh bridge will reload devices&&config
+  if (is_bridge) then
+    -- just add wiser(bridge) to wisers
+    wisers.add_wiser(device.device_network_id, device:get_field(config.FIELD.IP), device)
```
liuyd96, line 17: You have to refresh your wiser again othersize your edge devices can't be added
```
wisers.refresh_wiser(device.device_network_id)
```

liuyd96:
@refuseRed 
```text
❯ luacheck --config .github/workflows/.luacheckrc drivers/DeepSmart/*
Checking drivers/DeepSmart/deepsmart/src/commands.lua OK
Checking drivers/DeepSmart/deepsmart/src/config.lua OK
Checking drivers/DeepSmart/deepsmart/src/deepsmart/api.lua OK
Checking drivers/DeepSmart/deepsmart/src/deepsmart/devices.lua OK
Checking drivers/DeepSmart/deepsmart/src/deepsmart/dp2knx.lua OK
Checking drivers/DeepSmart/deepsmart/src/deepsmart/dpenum.lua OK
Checking drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua 3 warnings

    drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua:40:14: (W311) value assigned to variable endpos is unused
    drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua:387:30: (W113) accessing undefined variable uuid
    drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua:389:17: (W211) unused variable dev

Checking drivers/DeepSmart/deepsmart/src/discovery.lua OK
Checking drivers/DeepSmart/deepsmart/src/init.lua OK
Checking drivers/DeepSmart/deepsmart/src/lifecycles.lua OK
Checking drivers/DeepSmart/deepsmart/src/ssdp.lua OK
Checking drivers/DeepSmart/deepsmart/src/utils.lua OK

Total: 3 warnings / 0 errors in 12 files

```

refuseRed:
Update point:
1. Change the search sub device to gateway and add it successfully before searching again
2 device types resolved from uuid no longer using storage
When controlling, change arg to args
Change the search gateway to broadcast to prevent WiFi routers from not supporting multicast
5. Do not update devices when querying gateway devices fails
6. Do not delete devices during search (can only be deleted on the app)
7. Add device queries and initial settings during bridge initialization

refuseRed:
Update point:
1. Change the search sub device to gateway and add it successfully before searching again
2 device types resolved from uuid no longer using storage
When controlling, change arg to args
Change the search gateway to broadcast to prevent WiFi routers from not supporting multicast
5. Do not update devices when querying gateway devices fails
6. Do not delete devices during search (can only be deleted on the app)
7. Add device queries and initial settings during bridge initialization

refuseRed:
Update point:
1. Change the search sub device to gateway and add it successfully before searching again
2. device types resolved from uuid no longer using storage
3. When controlling, change arg to args
4. Change the search gateway to broadcast to prevent WiFi routers from not supporting multicast
5. Do not update devices when querying gateway devices fails
6. Do not delete devices during search (can only be deleted on the app)
7. Add device queries and initial settings during bridge initialization

liuyd96:
> Update point:
> 
> 1. Change the search sub device to gateway and add it successfully before searching again
> 2. device types resolved from uuid no longer using storage
> 3. When controlling, change arg to args
> 4. Change the search gateway to broadcast to prevent WiFi routers from not supporting multicast
> 5. Do not update devices when querying gateway devices fails
> 6. Do not delete devices during search (can only be deleted on the app)
> 7. Add device queries and initial settings during bridge initialization

CC @cjswedes

refuseRed:
@cjswedes again review

Hellcatlk:
We've reviewed the self-test cases for the latest code.

drivers/DeepSmart/deepsmart/src/commands.lua
```diff
@@ -0,0 +1,137 @@
...
+end
+
+----------------
+-- heater mode command
+----------------
+function command_handler.set_thermostat_mode(driver, device, command)
+  -- send command
+  local success = wisers.control(device, command, {config.HEATER.ONOFF})
+  -- Check if success
+  if success then
```
refuseRed, line 103: arg change to args

drivers/DeepSmart/deepsmart/src/ssdp.lua
```diff
@@ -0,0 +1,89 @@
...
+  end
+  -- use broadcast to find wisers as wifi router's group cast is forbidden by default(need login to router to start group cast)
+  upnp:setoption('broadcast', true)
+  local timeout = socket.gettime() + config.MC_TIMEOUT + 1
+  local mx = 2
+  -- broadcasting request
+  log.info('===== SCANNING NETWORK...')
+  local multicast_msg = table.concat({
+    "M-SEARCH * HTTP/1.1",
+    "HOST: 255.255.255.255:1900",
```
refuseRed, line 46: use broadcast to search bridge while wifi routers will disable group cast by default

drivers/DeepSmart/deepsmart/src/lifecycles.lua
refuseRed: add refresh_wiser

drivers/DeepSmart/deepsmart/src/lifecycles.lua
```diff
@@ -0,0 +1,123 @@
...
+    -- load devices
+    wisers.refresh_wiser(device.device_network_id)
+    -- refresh all devices
```
refuseRed, line 20: add devices init in bridge lifecycle-init.

drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,834 @@
...
+  -- if wiser ip is not changed then do nothing
+  local deepsmartapi = wiser.api
+  if (deepsmartapi == nil) then
+    log.trace('wiser api is not inited')
+    return false
+  end
+  -- get devices
+  local configstr = deepsmartapi:load_config()
+  log.trace(string.format('load config (%s)', configstr))
+  if (configstr ~= nil and configstr ~= wiser.bridge:get_field(config.FIELD.DEVICES)) then
```
refuseRed, line 452: if configstr is nil means load config fail, do not update devices(nil config will clear all devices)

drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,834 @@
...
+
+
+--------------
+-- get devtype by device_network_id
+--------------
+function Wisers.get_dev_type(device_network_id)
+  local dev_id,_,_ = Wisers.parse_device_network_id(device_network_id)
+  local wiser,dev = Wisers.get_device_wiser(dev_id)
+  if (wiser == nil) then
+    -- when device is not exist in bridge delete from hub
```
refuseRed, line 605: old code will delete device when get wiser nil. New code remove the operation.

drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua
```diff
@@ -0,0 +1,834 @@
...
+  local dev_id = device_network_id
+  local idx = 1
+  -- if _ exists then idx exists
+  local beginpos,endpos = string.find(dev_id, '_')
+  if (beginpos ~= nil) then
+    dev_id = string.sub(device_network_id, 1, beginpos-1)
+    local str = string.sub(device_network_id, endpos+1)
+    idx = tonumber(str)
+  end
+  -- find pid
```
refuseRed, line 35: parse pid from the device_network_id and return it

lelandblue:
We will be re-reviewing this code as soon as we can, thank you!

drivers/DeepSmart/deepsmart/src/ssdp.lua
cjswedes: Can you explain why this is needed? Was it just one network that had multicast disabled? We have used multicast rather than broadcast for all other SSDP use cases, and have not had a problem or user complaint; also the SSDP rfc specifies the multicast address 239.255.255.250 should be used for SSDP. 

Making this change adds more concerns that the driver's discovery and periodic IP address checks will have adverse network affects in large networks.

drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua
cjswedes: Should this `return false` if the configstr is nil? I dont think this function is "successful" if that fails, but I could be misunderstanding.

drivers/DeepSmart/deepsmart/src/deepsmart/api.lua
```diff
@@ -0,0 +1,235 @@
...
+    retry = retry - 1
+  end
+  return false
+end
+------------
+-- del device from wiser
+-- todo
+------------
+function Api:del_device(device)
+  -- todo
```
cjswedes, line 230: Is this needed?

drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua
cjswedes: Without this, how will a device removed outside of the SmartThings platform be reflected in the smartthings app?

drivers/DeepSmart/deepsmart/src/lifecycles.lua
cjswedes: > bridge is the last one to enter the function(as the order device1,device2...bridge)

This ordering is not a guarantee by the platform. I dont think that assumption is built in, but your comment suggests otherwise. Will everything still work if the bridge init event is handled before the device init events?

drivers/DeepSmart/deepsmart/src/lifecycles.lua
```diff
@@ -0,0 +1,123 @@
...
+    local devices = wisers.driver:get_devices()
+    for i,v in pairs(devices) do
```
cjswedes, line 22: Is it possible for there to be multiple bridges in the driver? If so, I think this will not work since the first bridge will be initializing all the devices for the other bridges as well.

refuseRed: Check if the device belongs to the current portfolio

drivers/DeepSmart/deepsmart/src/ssdp.lua
refuseRed: The on-site router model does not support multicast

drivers/DeepSmart/deepsmart/src/lifecycles.lua
refuseRed: Delete the init device and initialize the logical device when added

drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua
refuseRed: Change to return false implementation

drivers/DeepSmart/deepsmart/src/deepsmart/wisers.lua
refuseRed: Currently, devices can only be deleted through the SmartThings app

drivers/DeepSmart/deepsmart/src/ssdp.lua
cjswedes: Did this actually fix the issue? The `search-parameters.yml` file specifies the ssdp search term, and the hub uses the multicast address, so if devices are not responding to that request, the driver will never start.

drivers/DeepSmart/deepsmart/src/lifecycles.lua
cjswedes: Ill defer to the testing here

drivers/DeepSmart/deepsmart/src/ssdp.lua
refuseRed: Only when the gateway responds to the ssdp query request, the device does not need to respond. ssdp will search for all devices on the gateway after finding it

cjswedes: If the gateway responds to the multicast request (ssdp query), why would the broadcast need to be used. Also, FWIW broadcast is not a socket option that we allow for the sockets in lua drivers, which I believe is needed to the broadcast addr.

refuseRed: Gateway can respond to both broadcast and multicast ssdp requests
Many WiFi routers have disabled multicast, so they can only send ssdp request data through broadcasting

cjswedes: > Many WiFi routers have disabled multicast, so they can only send ssdp request data through broadcasting

The reason I dont understand this change, is that if a gateway has disabled multicast, the driver will not start. The hub uses the ssdp search term in `search-paremeters.yml` with the multicast address to determine if the driver will be started. So if that is disabled on the gateway, the hub will never think there is a device on the network, and this driver will never be run.

refuseRed: 1. The router does not allow multicast, and the hub cannot find the gateway ；
2. The driver will run, but the driver cannot search for the gateway ；

refuseRed: After our testing cases and testing process, the problem you mentioned does not exist. Our testing results are all okay

liuyd96: > After our testing cases and testing process, the problem you mentioned does not exist. Our testing results are all okay

Hi @cjswedes 

DeepSmart claims that anything goes well. 
If there are no other unresolved issues, can we approve this PR first and let other departments test it?

We have to a get an approval of this PR, then ask other departments to do certification test.

liuyd96:
> I will let @lelandblue merge when appropriate.

We have asked other departments to test ASAP. 

Thank you so much.

PR #1288, opened 2024-03-26
CHAD-12899 Lazy-loading for zigbee-switch

greens:


wkhenon (approved):

drivers/SmartThings/zigbee-switch/src/ge-link-bulb/init.lua
```diff
@@ -28,7 +28,8 @@ local GE_LINK_BULB_FINGERPRINTS = {
 }
 
 local function can_handle_ge_link_bulb(opts, driver, device)
-  return (GE_LINK_BULB_FINGERPRINTS[device:get_manufacturer()] or {})[device:get_model()] or false
+  local subdriver = require("ge-link-bulb")
```
ctowns, line 31: I have not investigated this, but I wonder if there are any memory considerations with having the whole sub driver loaded through this `require` even if we will ultimately just return `false`. In other `can_handle` functions, the sub driver is only loaded if the prerequisite `can_handle` checks for mfr/model are true.

These can handle functions can be called quite often, so if this does load the subdriver each time this could cause memory thrashing

greens: You're probably right.

greens: I think I got all of them.

ctowns (approved):
Looks good to me!

PR #1296, opened 2024-04-02
[Aqara] Aqara doorlock K100

seojune79:
add aqara-lock package and K100 driver

### DIFF
--- a/drivers/Aqara/aqara-lock/capabilities/lockCredentialInfo.yaml
+++ b/drivers/Aqara/aqara-lock/capabilities/lockCredentialInfo.yaml
@@ -1,0 +1,133 @@
+id: stse.lockCredentialInfo
+version: 1
+status: proposed
+name: Lock Credential Info
+ephemeral: false
+attributes:
+  credentialInfo:
+    schema:
+      type: object
+      properties:
+        value:
+          type: array
+          items:
+            title: Credential info
+            type: object
+            additionalProperties: false
+            properties:
+              credentialId:
+                type: integer
+                minimum: 0
+              credentialType:
+                type: string
+                enum:
+                  - keypad
+                  - rfid
+                  - fingerprint
+              userId:
+                type: string
+                maxLength: 255
+              userLabel:
+                type: string
+                maxLength: 255
+              userType:
+                type: string
+                enum:
+                  - unknown
+                  - administrator
+                  - regularUser
+                  - frequentVisitors
+                  - host
+            required:
+              - credentialId
+              - credentialType
+              - userId
+              - userLabel
+              - userType
+      additionalProperties: false
+      required:
+        - value
+    enumCommands: []
+commands:
+  syncAll:
+    name: syncAll
+    arguments:
+      - name: credentialInfo
+        optional: false
+        schema:
+          title: Credential info
+          type: array
+          items:
+            properties:
+              credentialId:
+                type: integer
+                minimum: 0
+              credentialType:
+                type: string
+                enum:
+                  - keypad
+                  - rfid
+                  - fingerprint
+              userId:
+                type: string
+                maxLength: 255
+              userLabel:
+                type: string
+                maxLength: 255
+              userType:
+                type: string
+                enum:
+                  - unknown
+                  - administrator
+                  - regularUser
+                  - frequentVisitors
+                  - host
+  upsert:
+    name: upsert
+    arguments:
+      - name: credentialInfo
+        optional: false
+        schema:
+          title: Credential info
+          type: array
+          items:
+            properties:
+              credentialId:
+                type: integer
+                minimum: 0
+              credentialType:
+                type: string
+                enum:
+                  - keypad
+                  - rfid
+                  - fingerprint
+              userId:
+                type: string
+                maxLength: 255
+              userLabel:
+                type: string
+                maxLength: 255
+              userType:
+                type: string
+                enum:
+                  - unknown
+                  - administrator
+                  - regularUser
+                  - frequentVisitors
+                  - host
+  deleteUser:
+    name: deleteUser
+    arguments:
+      - name: userId
+        optional: false
+        schema:
+          type: integer
+          minimum: 0
+  deleteCredential:
+    name: deleteCredential
+    arguments:
+      - name: credentialId
+        optional: false
+        schema:
+          type: integer
+          minimum: 0
--- a/drivers/Aqara/aqara-lock/config.yml
+++ b/drivers/Aqara/aqara-lock/config.yml
@@ -1,0 +1,6 @@
+name: 'Aqara Lock'
+packageKey: 'aqara-lock'
+permissions:
+  zigbee: {}
+description: "SmartThings driver for Aqara Doorlock"
+vendorSupportInformation: "https://www.aqara.com/en/support/"
--- a/drivers/Aqara/aqara-lock/fingerprints.yml
+++ b/drivers/Aqara/aqara-lock/fingerprints.yml
@@ -1,0 +1,6 @@
+zigbeeManufacturer:
+  - id: "LUMI/aqara.lock.akr011"
+    deviceLabel: "Aqara Doorlock K100"
+    manufacturer: Lumi
+    model: aqara.lock.akr011
+    deviceProfileName: aqara-lock-battery
--- a/drivers/Aqara/aqara-lock/profiles/aqara-lock-battery.yml
+++ b/drivers/Aqara/aqara-lock/profiles/aqara-lock-battery.yml
@@ -1,0 +1,23 @@
+name: aqara-lock-battery
+components:
+- id: main
+  capabilities:
+  - id: lock
+    version: 1
+  - id: battery
+    version: 1
+  - id: tamperAlert
+    version: 1
+  - id: remoteControlStatus
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: stse.lockCredentialInfo
+    version: 1
+  - id: stse.lockSystemInfo
+    version: 1
+  categories:
+  - name: SmartLock
+metadata:
+  mnmn: "SolutionsEngineering"
+  vid: "SmartThings-smartthings-Aqara_K100"
--- a/drivers/Aqara/aqara-lock/src/credential_utils.lua
+++ b/drivers/Aqara/aqara-lock/src/credential_utils.lua
@@ -1,0 +1,151 @@
+local capabilities = require "st.capabilities"
+local utils = require "st.utils"
+
+local remoteControlStatus = capabilities.remoteControlStatus
+local lockCredentialInfo = capabilities["stse.lockCredentialInfo"]
+
+local credential_utils = {}
+
+local HOST_COUNT = "__host_count"
+
+local function save_data(driver)
+  driver.datastore:save()
+  driver.datastore:commit()
+end
+
+local function update_remote_control_status(driver, device, added)
+  local host_cnt = device:get_field(HOST_COUNT) or 0
+  if added then
+    if host_cnt == 0 then
+      device:emit_event(remoteControlStatus.remoteControlEnabled('true', { visibility = { displayed = false } }))
+    end
+    host_cnt = host_cnt + 1
+  else
+    host_cnt = host_cnt - 1
+
+    if host_cnt == 0 then
+      device:emit_event(remoteControlStatus.remoteControlEnabled('false', { visibility = { displayed = false } }))
+    end
+  end
+
+  device:set_field(HOST_COUNT, host_cnt, {persist = true})
+  save_data(driver)
+end
+
+local function sync_all_credential_info(driver, device, command)
+  for _, credentialinfo in ipairs(command.args.credentialInfo) do
+    if credentialinfo.userType == "host" then
+      update_remote_control_status(driver, device, true)
+    end
+  end
+  device:emit_event(lockCredentialInfo.credentialInfo(utils.deep_copy(command.args.credentialInfo), { visibility = { displayed = false } }))
+  save_data(driver)
+end
+
+local function upsert_credential_info(driver, device, command)
+  if #command.args.credentialInfo == 0 then
+    return
+  end
+
+  local credentialInfoTable = utils.deep_copy(device:get_latest_state("main", lockCredentialInfo.ID, lockCredentialInfo.credentialInfo.NAME, {}))
+
+  for _, credentialinfo in ipairs(command.args.credentialInfo) do
+    local exist = false
+    for i = 1, #credentialInfoTable, 1 do
+      if credentialInfoTable[i].credentialId == credentialinfo.credentialId then
+        credentialInfoTable[i] = utils.deep_copy(credentialinfo)
+        exist = true
+        break
+      end
+    end
+
+    if exist == false then
+      if credentialinfo.userType == "host" then
+        update_remote_control_status(driver, device, true)
+      end
+
+      table.insert(credentialInfoTable, credentialinfo)
+    end
+  end
+
+  device:emit_event(lockCredentialInfo.credentialInfo(utils.deep_copy(credentialInfoTable), { visibility = { displayed = false } }))
+  save_data(driver)
+end
+
+local function delete_user(driver, device, command)
+  local credentialInfoTable = utils.deep_copy(device:get_latest_state("main", lockCredentialInfo.ID, lockCredentialInfo.credentialInfo.NAME, {}))
+  if #credentialInfoTable == 0 then
+    return
+  end
+
+  local id = tostring(command.args.userId)  
+
+  for i = #credentialInfoTable, 1, -1 do
+    if id == credentialInfoTable[i].userId then
+      if credentialInfoTable[i].userType == "host" then
+        update_remote_control_status(driver, device, false)
+      end
+      table.remove(credentialInfoTable, i)
+    end
+  end
+
+  device:emit_event(lockCredentialInfo.credentialInfo(utils.deep_copy(credentialInfoTable), { visibility = { displayed = false } }))
+  save_data(driver)
+end
+
+local function delete_credential(driver, device, command)
+  local credentialInfoTable = utils.deep_copy(device:get_latest_state("main", lockCredentialInfo.ID, lockCredentialInfo.credentialInfo.NAME, {}))
+  if #credentialInfoTable == 0 then
+    return
+  end
+
+  for i = #credentialInfoTable, 1, -1 do
+    if command.args.credentialId == credentialInfoTable[i].credentialId then
+      if credentialInfoTable[i].userType == "host" then
+        update_remote_control_status(driver, device, false)
+      end
+      table.remove(credentialInfoTable, i)
+      break
+    end
+  end
+
+  device:emit_event(lockCredentialInfo.credentialInfo(utils.deep_copy(credentialInfoTable), { visibility = { displayed = false } }))
+  save_data(driver)
+end
+
+local function update_system_version(driver, device, command)
+  local version = command.args.version
+end
+
+local function find_userLabel(driver, device, value)
+  local unlockCredentialId = value&0xFFFF
+  local credentialInfoTable = utils.deep_copy(device:get_latest_state("main", lockCredentialInfo.ID, lockCredentialInfo.credentialInfo.NAME, {}))
+  for _, credentialInfo in ipairs(credentialInfoTable) do
+    if credentialInfo.credentialId == unlockCredentialId then
+      return credentialInfo.userId, credentialInfo.userLabel
+    end
+  end
+  return nil, nil
+end
+
+local function is_exist_host(device)
+  local host_cnt = device:get_field(HOST_COUNT) or 0
+  return host_cnt > 0 and true or false
+end
+
+local function set_host_count(device, value)
+  device:set_field(HOST_COUNT, 0, {persist = true})
+end
+
+credential_utils.save_data = save_data
+credential_utils.update_remote_control_status = update_remote_control_status
+credential_utils.sync_all_credential_info = sync_all_credential_info
+credential_utils.upsert_credential_info = upsert_credential_info
+credential_utils.delete_user = delete_user
+credential_utils.delete_credential = delete_credential
+credential_utils.update_system_version = update_system_version
+credential_utils.find_userLabel = find_userLabel
+credential_utils.is_exist_host = is_exist_host
+credential_utils.set_host_count = set_host_count
+
+return credential_utils
--- a/drivers/Aqara/aqara-lock/src/init.lua
+++ b/drivers/Aqara/aqara-lock/src/init.lua
@@ -1,0 +1,233 @@
+local security = require "st.security"
+local ZigbeeDriver = require "st.zigbee"
+local data_types = require "st.zigbee.data_types"
+local cluster_base = require "st.zigbee.cluster_base"
+local capabilities = require "st.capabilities"
+local base64 = require "base64"
+local credential_utils = require "credential_utils"
+
+local remoteControlStatus = capabilities.remoteControlStatus
+local lockCredentialInfo = capabilities["stse.lockCredentialInfo"]
+local Battery = capabilities.battery
+local Lock = capabilities.lock
+local TamperAlert = capabilities.tamperAlert
+
+local PRI_CLU = 0xFCC0
+local PRI_ATTR = 0xFFF3
+local MFG_CODE = 0x115F
+
+local serial_num = 0
+local seq_num = 0
+
+local SHARED_KEY = "__shared_key"
+local CLOUD_PUBLIC_KEY = "__cloud_public_key"
+
+local function my_secret_data_handler(driver, device, secret_info)
+  -- At time of writing this returns nothind beyond "secret_type = aqara"
+  local shared_key = secret_info.shared_key
+  local cloud_public_key = secret_info.cloud_public_key
+
+  device:set_field(SHARED_KEY, shared_key, { persist = true })
+  device:set_field(CLOUD_PUBLIC_KEY, cloud_public_key, { persist = true })
+  credential_utils.save_data(driver)
+
+  if cloud_public_key ~= nil then
+    local raw_data = base64.decode(cloud_public_key)
+    -- send cloud_pub_key
+    device:send(cluster_base.write_manufacturer_specific_attribute(device,
+      PRI_CLU, PRI_ATTR, MFG_CODE, data_types.OctetString, "\x3E" .. raw_data))
+  end
+end
+
+local function device_added(self, device)
+  if credential_utils.is_exist_host(device) then
+    device:emit_event(remoteControlStatus.remoteControlEnabled('true', { visibility = { displayed = false } }))
+  else
+    credential_utils.set_host_count(device, 0)
+    device:emit_event(remoteControlStatus.remoteControlEnabled('false', { visibility = { displayed = false } }))
+  end
+  device:emit_event(Battery.battery(100))
+  device:emit_event(TamperAlert.tamper("clear", { visibility = { displayed = false } }))
+  device:emit_event(Lock.lock("locked"))
+  credential_utils.save_data(self)
+end
+
+local function toValue(payload, start, length)
+  local ret = 0
+  for i = start, start + length - 1 do
+    ret = (ret << 8) + string.byte(payload, i)
+  end
+  return ret
+end
+
+local function toHex(value, length)
+  local ret = string.char(0xFF & value)
+  for i = length, 2, -1 do
+    ret = string.char(0xFF & (value >> 8 * (i - 1))) .. ret
+  end
+  return ret
+end
+
+local function event_lock_handler(driver, device, evt_name, evt_value)
+  if evt_value == 1 then
+    device:emit_event(Lock.lock(evt_name))
+    device:emit_event(TamperAlert.tamper.clear())
+    if credential_utils.is_exist_host(device) then
+      device:emit_event(remoteControlStatus.remoteControlEnabled('true', { visibility = { displayed = false } }))
+    else
+      device:emit_event(remoteControlStatus.remoteControlEnabled('false', { visibility = { displayed = false } }))
+    end
+  end
+end
+
+local function event_unlock_handler(driver, device, evt_name, evt_value)
+  local id, label
+  id, label = credential_utils.find_userLabel(driver, device, evt_value)
+  device:emit_event(Lock.lock.unlocked({ data = { method = evt_name, codeId = id, codeName = label } }))
+  device:emit_event(remoteControlStatus.remoteControlEnabled('false', { visibility = { displayed = false } }))
+  device:emit_event(TamperAlert.tamper.clear())
+end
+
+local function event_door_handler(driver, device, evt_name, evt_value)
+  if evt_value == 2 then
+    device:emit_event(Lock.lock(evt_name))
+    device:emit_event(TamperAlert.tamper.clear())
+  elseif evt_value == 4 then
+    device:emit_event(TamperAlert.tamper.detected())
+  end
+end
+
+local function event_battery_handler(driver, device, evt_name, evt_value)
+  device:emit_event(Battery.battery(evt_value))
+end
+
+local function event_temper_alert_handler(driver, device, evt_name, evt_value)
+  device:emit_event(TamperAlert.tamper.detected())
+end
+
+local resource_id = {
+  { id = "13.31.85", event_name = "locked",       event_handler = event_lock_handler },
+  { id = "13.17.85", event_name = "not fully locked", event_handler = event_door_handler },
+  { id = "13.48.85", event_name = "manual",       event_handler = event_unlock_handler },
+  { id = "13.51.85", event_name = "manual",       event_handler = event_unlock_handler },
+  { id = "13.42.85", event_name = "fingerprint",  event_handler = event_unlock_handler },
+  { id = "13.43.85", event_name = "keypad",       event_handler = event_unlock_handler },
+  { id = "13.44.85", event_name = "rfid",         event_handler = event_unlock_handler },
+  { id = "13.45.85", event_name = "bluetooth",    event_handler = event_unlock_handler },
+  { id = "13.90.85", event_name = "command",      event_handler = event_unlock_handler },
+  { id = "13.46.85", event_name = "keypad",       event_handler = event_unlock_handler },
+  { id = "13.56.85", event_name = "",             event_handler = event_battery_handler },
+  { id = "13.32.85", event_name = "",             event_handler = event_temper_alert_handler },
+}
+
+local function request_generate_shared_key(device)
+  device:send(cluster_base.write_manufacturer_specific_attribute(device,
+    PRI_CLU, PRI_ATTR, MFG_CODE, data_types.OctetString, "\x2B"))
+end
+
+local function lock_state_handler(driver, device, value, zb_rx)
+  local shared_key = device:get_field(SHARED_KEY)
+  local param = value.value
+  local command = string.sub(param, 0, 1)
+
+  if command == "\x3E" then
+    -- recv lock_pub_key
+    local locks_pub_key = string.sub(param, 2, string.len(param))
+    local mn_id = "Id3A"
+    local setup_id = "006"
+    local product_id = ""
+    local res, err = security.get_aqara_secret(device.zigbee_eui, locks_pub_key, "", mn_id, setup_id,
+      product_id)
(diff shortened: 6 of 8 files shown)


greens:
```
Checking drivers/Aqara/aqara-lock/src/credential_utils.lua 2 warnings

    drivers/Aqara/aqara-lock/src/credential_utils.lua:81:43: (W612) line contains trailing whitespace
    drivers/Aqara/aqara-lock/src/credential_utils.lua:[11](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/actions/runs/8518736412/job/23331426826?pr=1296#step:4:12)7:9: (W211) unused variable version

Checking drivers/Aqara/aqara-lock/src/init.lua    1 warning

    drivers/Aqara/aqara-lock/src/init.lua:139:16: (W211) unused variable err
```

drivers/Aqara/aqara-lock/profiles/aqara-lock-battery.yml
```diff
@@ -0,0 +1,23 @@
...
+    version: 1
+  - id: tamperAlert
+    version: 1
+  - id: remoteControlStatus
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: stse.lockCredentialInfo
+    version: 1
+  - id: stse.lockSystemInfo
```
greens, line 17: could you also include the definition for this capability?

drivers/Aqara/aqara-lock/src/credential_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+    if exist == false then
+      if credentialinfo.userType == "host" then
+        update_remote_control_status(driver, device, true)
+      end
+
+      table.insert(credentialInfoTable, credentialinfo)
+    end
+  end
+
+  device:emit_event(lockCredentialInfo.credentialInfo(utils.deep_copy(credentialInfoTable), { visibility = { displayed = false } }))
```
greens, line 71: Is there a reason `deep_copy` needs to be called here? Can't the table be passed directly?

drivers/Aqara/aqara-lock/src/credential_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+local lockCredentialInfo = capabilities["stse.lockCredentialInfo"]
+
+local credential_utils = {}
+
+local HOST_COUNT = "__host_count"
+
+local function save_data(driver)
+  driver.datastore:save()
+  driver.datastore:commit()
+end
```
greens, line 14: In general, I think this should happen for you seamlessly without needing to invoke these methods directly. @varzac may have more of an opinion, but the datastore is saved frequently.

drivers/Aqara/aqara-lock/src/credential_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+    end
+  end
+
+  device:emit_event(lockCredentialInfo.credentialInfo(utils.deep_copy(credentialInfoTable), { visibility = { displayed = false } }))
+  save_data(driver)
+end
+
+local function update_system_version(driver, device, command)
+  local version = command.args.version
+end
```
greens, line 118: this function is a no-op

drivers/Aqara/aqara-lock/src/credential_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+credential_utils.save_data = save_data
+credential_utils.update_remote_control_status = update_remote_control_status
+credential_utils.sync_all_credential_info = sync_all_credential_info
+credential_utils.upsert_credential_info = upsert_credential_info
+credential_utils.delete_user = delete_user
+credential_utils.delete_credential = delete_credential
+credential_utils.update_system_version = update_system_version
+credential_utils.find_userLabel = find_userLabel
+credential_utils.is_exist_host = is_exist_host
+credential_utils.set_host_count = set_host_count
```
greens, line 149: these lines are unnecessary if you define your functions like

`credential_utils.save_data = function(...)`

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,233 @@
...
+    local raw_data = base64.decode(cloud_public_key)
+    -- send cloud_pub_key
+    device:send(cluster_base.write_manufacturer_specific_attribute(device,
+      PRI_CLU, PRI_ATTR, MFG_CODE, data_types.OctetString, "\x3E" .. raw_data))
+  end
+end
+
+local function device_added(self, device)
+  if credential_utils.is_exist_host(device) then
+    device:emit_event(remoteControlStatus.remoteControlEnabled('true', { visibility = { displayed = false } }))
```
greens, line 44: why isn't this value's datatype a `boolean`?

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,233 @@
...
+  else
+    credential_utils.set_host_count(device, 0)
+    device:emit_event(remoteControlStatus.remoteControlEnabled('false', { visibility = { displayed = false } }))
+  end
+  device:emit_event(Battery.battery(100))
+  device:emit_event(TamperAlert.tamper("clear", { visibility = { displayed = false } }))
+  device:emit_event(Lock.lock("locked"))
```
greens, line 51:
```suggestion
  device:emit_event(TamperAlert.tamper.clear({ visibility = { displayed = false } }))
  device:emit_event(Lock.lock.locked())
```

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,233 @@
...
+  return ret
+end
+
+local function toHex(value, length)
+  local ret = string.char(0xFF & value)
+  for i = length, 2, -1 do
+    ret = string.char(0xFF & (value >> 8 * (i - 1))) .. ret
+  end
+  return ret
+end
```
greens, line 69: you might be interested in https://developer.smartthings.com/docs/edge-device-drivers/utils.html#st.utils.serialize_int and https://developer.smartthings.com/docs/edge-device-drivers/utils.html#st.utils.deserialize_int

drivers/Aqara/aqara-lock/src/credential_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+
+local function update_remote_control_status(driver, device, added)
+  local host_cnt = device:get_field(HOST_COUNT) or 0
+  if added then
+    if host_cnt == 0 then
+      device:emit_event(remoteControlStatus.remoteControlEnabled('true', { visibility = { displayed = false } }))
+    end
+    host_cnt = host_cnt + 1
+  else
+    host_cnt = host_cnt - 1
```
wkhenon, line 24: Is it possible for this `host_cnt` to go negative if the `HOST_COUNT` field is defaulted to 0?

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,233 @@
...
+    local raw_key = base64.decode(shared_key)
+    local raw_data = string.sub(param, 2, string.len(param))
+    local msg = security.decrypt_bytes(raw_data, raw_key, opts)
+    local text = string.sub(msg, 5, string.len(msg))
+    local payload = string.sub(text, 4, string.len(text))
+    local func_id = toValue(payload, 1, 1) .. "." .. toValue(payload, 2, 1) .. "." .. toValue(payload, 3, 2)
+    serial_num = toValue(msg, 3, 2)
+    seq_num = string.byte(text, 3)
+
+    for k, v in pairs(resource_id) do
```
greens, line 157: I'd simply change the `id` field to the actual key for this map here, that way you don't have to iterate through them every time.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,233 @@
...
+  { id = "13.31.85", event_name = "locked",       event_handler = event_lock_handler },
+  { id = "13.17.85", event_name = "not fully locked", event_handler = event_door_handler },
+  { id = "13.48.85", event_name = "manual",       event_handler = event_unlock_handler },
+  { id = "13.51.85", event_name = "manual",       event_handler = event_unlock_handler },
+  { id = "13.42.85", event_name = "fingerprint",  event_handler = event_unlock_handler },
+  { id = "13.43.85", event_name = "keypad",       event_handler = event_unlock_handler },
+  { id = "13.44.85", event_name = "rfid",         event_handler = event_unlock_handler },
+  { id = "13.45.85", event_name = "bluetooth",    event_handler = event_unlock_handler },
+  { id = "13.90.85", event_name = "command",      event_handler = event_unlock_handler },
+  { id = "13.46.85", event_name = "keypad",       event_handler = event_unlock_handler },
```
greens, line 118: for these method strings, we generally make a little "enum" to keep track of these strings: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1298

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,233 @@
...
+  device:emit_event(remoteControlStatus.remoteControlEnabled('false', { visibility = { displayed = false } }))
+  device:emit_event(TamperAlert.tamper.clear())
+end
+
+local function event_door_handler(driver, device, evt_name, evt_value)
+  if evt_value == 2 then
+    device:emit_event(Lock.lock(evt_name))
+    device:emit_event(TamperAlert.tamper.clear())
+  elseif evt_value == 4 then
+    device:emit_event(TamperAlert.tamper.detected())
```
greens, line 96: just checking: should a `Lock.lock("not fully locked")` event not be emitted here?

drivers/Aqara/aqara-lock/src/credential_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+    end
+
+    if exist == false then
+      if credentialinfo.userType == "host" then
+        update_remote_control_status(driver, device, true)
+      end
+
+      table.insert(credentialInfoTable, credentialinfo)
+    end
+  end
```
greens, line 69: You might find this useful: https://developer.smartthings.com/docs/edge-device-drivers/utils.html#st.utils.update

drivers/Aqara/aqara-lock/capabilities/lockCredentialInfo.yaml
```diff
@@ -0,0 +1,133 @@
...
+                maxLength: 255
+              userType:
+                type: string
+                enum:
+                  - unknown
+                  - administrator
+                  - regularUser
+                  - frequentVisitors
+                  - host
+  upsert:
```
greens, line 85: what's the difference between `upsert` and `syncAll`?

drivers/Aqara/aqara-lock/src/credential_utils.lua
tandres: It's possible for the auth data stored in the datastore to be lost if the hub loses power before the datastore is synced to nonvol. Even though this was anticipated to be exceedingly rare in the field it was determined that we would need to support a mechanism to force the immediate sync of the datastore to nonvol (the `commit`api). So these two calls are necessary to ensure that the auth data can't be lost.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,233 @@
...
+local MFG_CODE = 0x115F
+
+local serial_num = 0
+local seq_num = 0
+
+local SHARED_KEY = "__shared_key"
+local CLOUD_PUBLIC_KEY = "__cloud_public_key"
+
+local function my_secret_data_handler(driver, device, secret_info)
+  -- At time of writing this returns nothind beyond "secret_type = aqara"
```
tandres, line 26: Even though `secret_info.secret_type` is always going to be `aqara` it would be better to check in this function to make sure of that and ignore the event if it doesn't match `aqara`. Future updates to the library might include new messages where `secret_type` will be different and include different payloads. Those changes would cause this driver to crash because of library changes.

drivers/Aqara/aqara-lock/profiles/aqara-lock-battery.yml
Kwang-Hui: @seojune79 , `stse.lockSystemInfo` wouldn't used anymore, could you please remove this from profile?

drivers/Aqara/aqara-lock/src/credential_utils.lua
Kwang-Hui: @seojune79 Could you please remove this function? this wouldn't be used anymore.

drivers/Aqara/aqara-lock/src/init.lua
Kwang-Hui: it's exisitng standard capability. this code just use that.

drivers/Aqara/aqara-lock/src/credential_utils.lua
seojune79: I removed it.

drivers/Aqara/aqara-lock/profiles/aqara-lock-battery.yml
seojune79: Yes. I removed it.

drivers/Aqara/aqara-lock/src/init.lua
kdbang: https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/8f6538bf03eacc140c38954299184bde9599886c/capabilities/remote-control-status.src/remote-control-status.yml#L10

It seems that definition of the `remoteControlEnabled` is string.

drivers/Aqara/aqara-lock/src/credential_utils.lua
seojune79: I have modified the code based on your comment.

drivers/Aqara/aqara-lock/src/init.lua
seojune79: I modified it.

drivers/Aqara/aqara-lock/src/init.lua
seojune79: I modified the code to use "st.utils".

drivers/Aqara/aqara-lock/src/credential_utils.lua
seojune79: I added code to check host_cnt is 0.

drivers/Aqara/aqara-lock/src/init.lua
seojune79: That's a good idea. I revised it as your comment.

drivers/Aqara/aqara-lock/src/init.lua
seojune79: Okay. I modified it.

drivers/Aqara/aqara-lock/src/init.lua
seojune79: I have added it.

Kwang-Hui: `evt_value=4` means "Door Forced Open"

seojune79: Sorry, I misunderstood.

drivers/Aqara/aqara-lock/src/init.lua
seojune79: Current "secret_info.secret_type" is nil. I will add the code when the "secret_info.secret_type" value is updated.
```
2024-04-03T13:52:52.466126870+09:00 PRINT Aqara Lock  ----- [my_secret_data_handler] secret_info.secret_type = nil
```
```lua
local function my_secret_data_handler(driver, device, secret_info)
  print(string.format("----- [my_secret_data_handler] secret_info.secret_type = %s", secret_info.secret_type))
  local shared_key = secret_info.shared_key
  local cloud_public_key = secret_info.cloud_public_key

  device:set_field(SHARED_KEY, shared_key, { persist = true })
  device:set_field(CLOUD_PUBLIC_KEY, cloud_public_key, { persist = true })
  credential_utils.save_data(driver)

  if cloud_public_key ~= nil then
    local raw_data = base64.decode(cloud_public_key)
    -- send cloud_pub_key
    device:send(cluster_base.write_manufacturer_specific_attribute(device,
      PRI_CLU, PRI_ATTR, MFG_CODE, data_types.OctetString, "\x3E" .. raw_data))
  end
end
```

drivers/Aqara/aqara-lock/config.yml
```diff
@@ -0,0 +1,6 @@
+name: 'Aqara Lock'
+packageKey: 'aqara-lock'
+permissions:
```
varzac, line 3: This will require the aqara specific permission to give it access to the cloud APIs for the key retrieval.  @tandres can verify but I think it should be 

`aqara-provisioning: {}`

drivers/Aqara/aqara-lock/src/init.lua
tandres: Oh this is my mistake. I think the comment is from an older version of the libraries. The correct key is `secret_kind` - I had originally written it as `secret_type` at one point and it must have made it into some reference or something.

drivers/Aqara/aqara-lock/src/credential_utils.lua
```diff
@@ -0,0 +1,134 @@
...
+    if host_cnt == 0 then
+      device:emit_event(remoteControlStatus.remoteControlEnabled('true', { visibility = { displayed = false } }))
+    end
+    host_cnt = host_cnt + 1
+  else
+    if host_cnt > 0 then host_cnt = host_cnt - 1 end
+
+    if host_cnt == 0 then
+      device:emit_event(remoteControlStatus.remoteControlEnabled('false', { visibility = { displayed = false } }))
+    end
```
greens, line 27: if the intention is for this to only fire the _first_ time the `host_cnt` goes below 0, you may want to move it inside the if statement above. Otherwise you may still have an issue if there is a remove event when the value is 0.

drivers/Aqara/aqara-lock/src/credential_utils.lua
```diff
@@ -0,0 +1,134 @@
...
+      if credentialInfoTable[i].userType == "host" then
+        credential_utils.update_remote_control_status(driver, device, false)
+      end
+      table.remove(credentialInfoTable, i)
+      break
+    end
+  end
+
+  device:emit_event(lockCredentialInfo.credentialInfo(utils.deep_copy(credentialInfoTable),
+    { visibility = { displayed = false } }))
```
greens, line 109: I don't know if I saw an answer here, but is there a reason you have to use `deep_copy` here?

seojune79: Did you mean removing all the "deep_copy" in "emit_event"? I misunderstood that I should just remove the line. Let me check it.

drivers/Aqara/aqara-lock/src/credential_utils.lua
kdbang: Which issue do you mean? as far as i know, `host_cnt` and `remoteControlStatus.remoteControlEnabled` are using for displaying on/off button on device card. My intention is that if there is no "host" credentionInfo list, edge driver send `remoteControlStatus.remoteControlEnabled` with false.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,243 @@
...
+    serial_num = serial_num + 1
+    local raw_data = "\x5B" .. toHex(string.len(text), 1) .. toHex(serial_num, 2) .. text
+    for i = 1, 4 - (string.len(raw_data) % 4) do
+      raw_data = raw_data .. "\x00"
+    end
+
+    local opts = { cipher = "aes256-ecb", padding = false }
+    local raw_key = base64.decode(shared_key)
+    local result = security.encrypt_bytes(raw_data, raw_key, opts)
+    local msg = "\x93" .. result
```
FreeMasen, line 198: This is a bit complicated and might have some potential issues that would be difficult to detect at runtime. I wonder if extracting this to a helper function would be helpful with extra error logging around if any of these operations fail (for example if the inputs can't be converted to hex)

drivers/Aqara/aqara-lock/src/test/test_aqara_lock.lua
```diff
@@ -0,0 +1,153 @@
...
+local remoteControlStatus = capabilities.remoteControlStatus
+local lockCredentialInfo = capabilities["stse.lockCredentialInfo"]
+test.add_package_capability("lockCredentialInfo.yaml")
+local Battery = capabilities.battery
+local Lock = capabilities.lock
+local TamperAlert = capabilities.tamperAlert
+
+local PRI_CLU = 0xFCC0
+local PRI_ATTR = 0xFFF3
+local MFG_CODE = 0x115F
```
FreeMasen, line 17: duplicate constants, can we get these from the driver somehow?

drivers/Aqara/aqara-lock/src/test/test_aqara_lock.lua
```diff
@@ -0,0 +1,153 @@
...
+    test.socket.capability:__queue_receive({ mock_device.id,
+      { capability = lockCredentialInfo.ID, component = "main", command = "deleteCredential", args = { credentialId = "1" } } })
+  end
+)
+
+test.register_coroutine_test(
+  "lock_state_handler - shared_key is nil",
+  function()
+    local attr_report_data = {
+      { PRI_ATTR, data_types.OctetString.ID, "\x93\x15\xAA\xFE\x78\xEE\x3E\x81\x9A\x06\xE3\x9A\x62\xD3\xB1\xF1\xD4\x64\x8C\x16\x66\xAB\xE1\x69\x89\x8C\x04\x56\x8D\xAD\xEA\xDE\xF8" }
```
FreeMasen, line 142: This seems a bit magical, if this test started failing it would be hard to know why, is there any way to provide the context of what this message is supposed to represent?

drivers/Aqara/aqara-lock/fingerprints.yml
```diff
@@ -0,0 +1,6 @@
+zigbeeManufacturer:
+  - id: "LUMI/aqara.lock.akr011"
+    deviceLabel: "Aqara Doorlock K100"
+    manufacturer: Lumi
+    model: aqara.lock.akr011
+    deviceProfileName: aqara-lock-battery
```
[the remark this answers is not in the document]
seojune79: I inserted the last line.

drivers/Aqara/aqara-lock/profiles/aqara-lock-battery.yml
```diff
@@ -0,0 +1,21 @@
...
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: stse.lockCredentialInfo
+    version: 1
+  categories:
+  - name: SmartLock
+metadata:
+  mnmn: "SolutionsEngineering"
+  vid: "SmartThings-smartthings-Aqara_K100"
```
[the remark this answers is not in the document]
seojune79: I did it.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,243 @@
...
+
+local function event_lock_handler(driver, device, evt_name, evt_value)
+  if evt_value == 1 then
+    device:emit_event(Lock.lock(evt_name))
+    device:emit_event(TamperAlert.tamper.clear())
+    if credential_utils.is_exist_host(device) then
+      device:emit_event(remoteControlStatus.remoteControlEnabled('true', { visibility = { displayed = false } }))
+    else
+      device:emit_event(remoteControlStatus.remoteControlEnabled('false', { visibility = { displayed = false } }))
+    end
```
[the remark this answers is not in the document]
seojune79: Okay. I made it as a separate function.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -0,0 +1,243 @@
...
+  NO_USE = ""
+}
+
+local HANDLERS = {
+  LOCK_HANDLER = event_lock_handler,
+  DOOR_HANDLER = event_door_handler,
+  UNLOCK_HANDLER = event_unlock_handler,
+  BATTERY_HANDLER = event_battery_handler,
+  TAMPER_ALERT_HANDLER = event_tamper_alert_handler
+}
```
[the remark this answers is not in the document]
seojune79: I changed it back to the old code.

drivers/Aqara/aqara-lock/src/credential_utils.lua
```diff
@@ -0,0 +1,134 @@
...
+
+local remoteControlStatus = capabilities.remoteControlStatus
+local lockCredentialInfo = capabilities["stse.lockCredentialInfo"]
+
+local credential_utils = {}
+local HOST_COUNT = "__host_count"
+
+credential_utils.save_data = function(driver)
+  driver.datastore:save()
+  driver.datastore:commit()
```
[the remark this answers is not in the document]
kdbang: this driver saves `credentialInfoTable` like below.
[{"credentialId": 1, "credentialType": "keypad", "userId": "1", "userLabel": "Kim", "userType": "host"}]
This table is used to control user management on plugin.

drivers/Aqara/aqara-lock/src/test/test_aqara_lock.lua
seojune79: The constants is not duplicated in the "test_aqara_lock.lua" file.
Please explain the problem in detail.

drivers/Aqara/aqara-lock/src/test/test_aqara_lock.lua
seojune79: This is the sample messages sent by the door lock to generate a shared key.
This message consists of a command("\x93") and the sample public key("\x15 .. ") of a doorlock.

drivers/Aqara/aqara-lock/config.yml
kdbang: 045d9470-05ac-44ad-96c7-5232b69b336e

It seems that modifying the provisioning in already deployed driver is not allowed.

After delete the driver in private channel, I can get below error when i type `smartthings edge:drivers:package -p acceptance`
```
14:44 $ smartthings edge:drivers:package -p acceptance
    AxiosError: Request failed with status code 500: {"requestId":"1758721324210160703","error":{"code":"UnexpectedError","message":"A non-recoverable error condition occurred.","details":[]}}
    Code: ERR_BAD_RESPONSE
```
```
timestamp=05:44:59.415, level=ERROR, loggingId=1758721324210160703, logger=s.s.services.PermissionService, userUuid=, msg=, key=forbidden_internal_permission, non system user attempting to upload a driver with internal only permission: permission key=aqara-provisioning, correlationId=1758721324210160703,, 
```
```
timestamp=05:44:59.415, level=ERROR, loggingId=1758721324210160703, logger=s.m.e.h.UnhandledExceptionHandler, userUuid=, msg=, key=api_exception_unhandled,, 
smartthings.common.exceptions.ForbiddenException: Insufficient permission to upload internal only driver
	at smartthings.sprocket.services.PermissionService.validatePackagePermissions(PermissionService.kt:43)
	at smartthings.sprocket.services.DriverPackageService.newDriver(DriverPackageService.kt:179)
	at smartthings.sprocket.services.DriverPackageService.access$newDriver(DriverPackageService.kt:32)
	at smartthings.sprocket.services.DriverPackageService$newDriver$1.invokeSuspend(DriverPackageService.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at io.reactivex.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:38)
	at io.reactivex.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:26)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

@varzac Can't we make private channel with `aqara-provisioning: {}`? or is there other way to make private channel? (or need to get system authorization?

drivers/Aqara/aqara-feeder/config.yml
```diff
@@ -1,4 +1,5 @@
 name: 'Aqara Feeder'
 packageKey: 'aqara-feeder'
 permissions:
+  aqara-provisioning: {}
```
tandres, line 4: Unless a driver uses the `get_aqara_secret` API it doesn't need this permission.

drivers/Aqara/aqara-lock/config.yml
varzac: No, you cannot use that permission in a private channel.  As previously discussed to restrict access to the cloud services to only the official integration, only drivers published by SmartThings can use this permission.  Unfortunately changing permissions is also something we don't allow on a driver once it is published to protect users from a driver publisher changing permissions after they had given it access.  I think the result is that the easiest way to move forward is to address the changes in this PR.  Then open a new PR which will create a new channel, and publish the driver with the new permissions.  We will make sure to give the updated PR immediate attention and get it started through the release process ASAP to prepare for the WWST testing.

tandres: I forked this branch and opened up a PR to be sure the permission would be present. It looks like the `aqara-provisioning` permission will need to be removed from the feeder driver otherwise CI will fail: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/actions/runs/8603874779

drivers/Aqara/aqara-feeder/config.yml
tandres: The `aqara-lock` driver does need the permission though.

drivers/Aqara/aqara-lock/src/credential_utils.lua
greens: just removing the `deep_copy`. I think 

```
  device:emit_event(lockCredentialInfo.credentialInfo(credentialInfoTable,
    { visibility = { displayed = false } }))
```

should work, right?

drivers/Aqara/aqara-lock/src/credential_utils.lua
greens: I guess it doesn't matter because it will be debounced as a duplicate anyway.

tandres:
We've run into an issue on our end with the cloud infrastructure for allowing the `aqara-provisioning` permission to be attached to the driver. The deploy for this driver will fail to update the driver until we are able to address an issue with one of our CI keys being incorrect (or potentially expired). Since it's the end of the day here we might not be able to get the auth corrected until sometime tomorrow but we don't expect it to take more than a day. The `aqara-provisioning` permission can be kept off the driver until then for final testing as hubs that are pointed at lower environments do not enforce using that permission to call the `security.get_aqara_secret` api.

Kwang-Hui:
> We've run into an issue on our end with the cloud infrastructure for allowing the `aqara-provisioning` permission to be attached to the driver. The deploy for this driver will fail to update the driver until we are able to address an issue with one of our CI keys being incorrect (or potentially expired). Since it's the end of the day here we might not be able to get the auth corrected until sometime tomorrow but we don't expect it to take more than a day. The `aqara-provisioning` permission can be kept off the driver until then for final testing as hubs that are pointed at lower environments do not enforce using that permission to call the `security.get_aqara_secret` api.

Thanks for the comment. As you know certification test would be done at the production with given invitation link at this PR.
Please let me know once the infrastructure has been updated.

drivers/Aqara/aqara-lock/src/credential_utils.lua
seojune79: Yeah, you're right

tandres:
@Kwang-Hui @kdbang We are working on coming up with a way to test this with the permission included. I am curious in the mean time if the device identity endpoint used to fetch the lock keys has made it's way into prod yet or has testing only been done with hubs pointed at staging?

tandres:
We've come up with a proposal for how to move forward. For a little background the `aqara-provisioning` permission protects the `security.get_aqara_secret` api from being called by drivers that don't have the permission. The `aqara-provisioning` permission is only granted by the platform to drivers published by SmartThings. The issue that we ran into is that the CI that runs to generate an invitation link to test the driver does not run as the "smartthings user" - nor should it because anyone could open up a PR and generate a driver with the protected permission! 

But we need to verify the permission works so we're working on creating a channel to perform testing with. Even if the hub that is being used to test isn't on Prod I will verify the permission works as expected.

With that we should be good to merge once the last two items are addressed:
1. The permission needs to be moved from the feeder driver to the lock driver
2. @greens `deep_copy` feedback should be addressed to reduce memory consumption issues
Tomorrow we will create the channel and upload the driver with the permission set so we can test on the lock we have, we'll also share the invitation link so that other testing can happen too.

Unfortunately because of the permission issue if there are any changes during testing going forward we will need to do one of the following things:
1. ST US will need to manually push updates to the driver to the custom channel
2. We will need to make changes to the driver on the alpha channel
We appreciate that this will slow down testing subsequent changes but the guarded permission is new so the story around it is still being improved.

drivers/Aqara/aqara-feeder/config.yml
tandres: We found a typo in some definitions: this will need to be `aqara_provisioning` instead of `aqara-provisioning`. Sorry for the confusion.

tandres:
We've figured out the issue with the permissioning story. Once this driver has the needed change we can work to manually generate a driver ready for testing. I was able to interact with a lock on my own fork of this driver (with just the permission name changes).

Kwang-Hui:
> We've figured out the issue with the permissioning story. Once this driver has the needed changes we can work to manually generate a driver ready for testing. I was able to interact with a lock on my own fork of this driver (with just the permission name changes) here's the invite: https://bestow-regional.api.smartthings.com/invite/5z2RLEqBdDj1.

Hi @tandres , Everything seems updated now with https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1296/commits/a20e55da6f6e9a5d735ae59d92e327daea744d4d commit.
Could you help to generate a driver for testing?

varzac:
> > We've figured out the issue with the permissioning story. Once this driver has the needed changes we can work to manually generate a driver ready for testing. I was able to interact with a lock on my own fork of this driver (with just the permission name changes) here's the invite: https://bestow-regional.api.smartthings.com/invite/5z2RLEqBdDj1.
> 
> Hi @tandres , Everything seems updated now with [a20e55d](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/a20e55da6f6e9a5d735ae59d92e327daea744d4d) commit. Could you help to generate a driver for testing?

Thanks for the update, the current process requires some elevated permissions, and some other teams, and I don't know if anyone is available after hours. We will get it generated first thing tomorrow, but if you can do some validation against the channel invite TJ provided that should be the driver's it was this morning, but with the new permission included

drivers/Aqara/aqara-lock/src/credential_utils.lua
Kwang-Hui: This seems causes unexpected table corruption.

Kwang-Hui:
Hi @varzac , I could find issues when using `utils.update`, `utils.merge`.
The fix has updated by https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1296/commits/510971dfe1a201d4a5e99628f743ef0d35b9d58c . Please generate a driver for testing including https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1296/commits/510971dfe1a201d4a5e99628f743ef0d35b9d58c commit also. Thanks!

drivers/Aqara/aqara-lock/src/credential_utils.lua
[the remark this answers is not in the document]
varzac: To expand on this Robert, the issue is that there are edge cases where the details can be stored, but not persisted for the 1 hour period, and because that credential info is necessary for communicating with the lock, losing it renders the device non-functional on SmartThings.  We include rate limiting on this API to avoid too excessive use.

lelandblue:
**IMPORTANT**

In order to test this driver, you will need this updated driver sharing link : https://bestow-regional.api.smartthings.com/invite/5z2RLEqBdDj1

PR #1303, opened 2024-04-04
Matter v1.2 drivers lock poc

HunsupJung:
This pull request is just for testing Matter door lock with partners.


HunsupJung:
unnecessary

PR #1308, opened 2024-04-07
fix Aqara LED Light Bult T1 reonboarding issue

zhoujie2022:
reference:
  STSE-2616


zhoujie2022:
@greens please review this PR

drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
```diff
@@ -33,17 +33,18 @@ local function do_refresh(self, device)
...
 
   device:send(Level.attributes.OnTransitionTime:write(device, 0))
   device:send(Level.attributes.OffTransitionTime:write(device, 0))
+  device:send(ColorControl.commands.MoveToColorTemperature(device, 200, 0x0000))
+
+  do_refresh(self, device)
+end
+
+local function device_added(driver, device, event)
 end
```
greens, line 48: I think you should just be able to delete this function (and in the handlers map below). It doesn't look like the base driver defines an `added` handler so you shouldn't need this no-op to overwrite it.

greens (approved):

PR #1315, opened 2024-04-09
fix issue that Aqara Smart Plug T1 doesn't work after reonboarding

jinhk7:
This is a fix for re-onboarding scenario:
STSE-2617


greens (approved):

greens (changes requested):
your failing tests is from `test_aqara_smart_plug` which also tests the same added behavior.

greens (approved):

PR #1323, opened 2024-04-11
[CHAD-12999] fix (Zigbee Lock): Ignore unknown status for sds lock

jaemin11lee:
https://smartthings.atlassian.net/browse/CHAD-12999


greens (commented):
Please include a test case here verifying that no event is sent in the event of an unknown lock value report.

jaemin11lee:
> Please include a test case here verifying that no event is sent in the event of an unknown lock value report.

Thanks, added test case

greens (approved):

PR #1330, opened 2024-04-17
New Device (Matter Lock)Aqara U200

lelandblue:
This is a thread-based Matter lock seeking WWST certification.

### DIFF
--- a/drivers/SmartThings/matter-lock/fingerprints.yml
+++ b/drivers/SmartThings/matter-lock/fingerprints.yml
@@ -1,4 +1,10 @@
 matterManufacturer:
+#Aqara
+  - id: 4447/10242
+    deviceLabel: Aqara Smart Lock U200
+    vendorId: 0x115F
+    productId: 0x2802
+    deviceProfileName: lock-without-codes
 #Nuki
   - id: "Nuki Smart Lock Pro"
     deviceLabel: Nuki Smart Lock Pro
--- a/drivers/SmartThings/matter-lock/fingerprints.yml
+++ b/drivers/SmartThings/matter-lock/fingerprints.yml
@@ -1,6 +1,6 @@
 matterManufacturer:
 #Aqara
-  - id: 4447/10242
+  - id: "4447/10242"
     deviceLabel: Aqara Smart Lock U200
     vendorId: 0x115F
     productId: 0x2802


lelandblue:
@ctowns Will we automatically switch to a profile that supports lock codes? I believe this device supports lock codes, however that was not the profile selected during the certification process.

ctowns:
> @ctowns Will we automatically switch to a profile that supports lock codes? I believe this device supports lock codes, however that was not the profile selected during the certification process.

We will not automatically switch to a profile that supports lock codes. We currently only support switching to a profile _without lock codes_ because the generic `matter-lock` fingerprint matches to a profile with lock codes (base-lock).

Also, if we know the device does support lock codes, can we override the profile that was selected during the WWST submission? I think it would probably be best if manufacturer fingerprints matched to the correct profile if possible without relying on profile switching in the driver if we can. What do you think?

lelandblue:
> > @ctowns Will we automatically switch to a profile that supports lock codes? I believe this device supports lock codes, however that was not the profile selected during the certification process.
> 
> We will not automatically switch to a profile that supports lock codes. We currently only support switching to a profile _without lock codes_ because the generic `matter-lock` fingerprint matches to a profile with lock codes (base-lock).
> 
> Also, if we know the device does support lock codes, can we override the profile that was selected during the WWST submission? I think it would probably be best if manufacturer fingerprints matched to the correct profile if possible without relying on profile switching in the driver if we can. What do you think?

Ah - we are cautious about that for good reasons. Yes, I do think we should push back on the existing submission and recommend that they use a different profile.

drivers/SmartThings/matter-lock/fingerprints.yml
```diff
@@ -1,4 +1,10 @@
 matterManufacturer:
+#Aqara
+  - id: 4447/10242
+    deviceLabel: Aqara Smart Lock U200
+    vendorId: 0x115F
+    productId: 0x2802
+    deviceProfileName: lock-without-codes
```
lelandblue, line 7: This profile is under discussion - Do not merge

hcarter-775:
Can this ticket be closed, since we now have a fingerprint for this device?

HunsupJung:
@hcarter-775 Yes we can close this ticket~!

PR #1331, opened 2024-04-17
New Device (Matter Sensor) Aqara Motion and Light Sensor P2

lelandblue:
This is an addition of the Matter Aqara Motion and Light Sensor P2 device that is seeking WWST Certification.

### DIFF
--- a/drivers/SmartThings/matter-sensor/fingerprints.yml
+++ b/drivers/SmartThings/matter-sensor/fingerprints.yml
@@ -5,7 +5,7 @@ matterManufacturer:
     vendorId: 0x115F
     productId: 0x2002
     deviceProfileName: contact-battery
-  - id: 4447/8195
+  - id: "4447/8195"
     deviceLabel: Aqara Motion and Light Sensor P2
     vendorId: 0x115F
     productId: 0x2003
--- a/drivers/SmartThings/matter-sensor/fingerprints.yml
+++ b/drivers/SmartThings/matter-sensor/fingerprints.yml
@@ -5,6 +5,11 @@ matterManufacturer:
     vendorId: 0x115F
     productId: 0x2002
     deviceProfileName: contact-battery
+  - id: 4447/8195
+    deviceLabel: Aqara Motion and Light Sensor P2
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-sensor/fingerprints.yml
```diff
@@ -5,6 +5,11 @@ matterManufacturer:
     vendorId: 0x115F
     productId: 0x2002
     deviceProfileName: contact-battery
+  - id: 4447/8195
+    deviceLabel: Aqara Motion and Light Sensor P2
+    vendorId: 0x115F
+    productId: 0x2003
+    deviceProfileName: motion-illuminance
```
kianooshkarami, line 12: No Battery?

drivers/SmartThings/matter-sensor/fingerprints.yml
```diff
@@ -5,6 +5,11 @@ matterManufacturer:
     vendorId: 0x115F
     productId: 0x2002
     deviceProfileName: contact-battery
+  - id: 4447/8195
```
ctowns, line 8:
```suggestion
  - id: "4447/8195"
```
When the id is just a set of numbers, we usually add "quotes" around the numbers to ensure that they are interpreted as a string rather than as an equation.

drivers/SmartThings/matter-sensor/fingerprints.yml
lelandblue: Thanks for the catch. I will follow up with the Cert Manager to confirm before making another commit. According to the product page provided on the request I see that it does.

lelandblue:
Closing PR, a new PR will be created after the partner re-submits for certification with a profile that supports battery

drivers/SmartThings/matter-sensor/fingerprints.yml
lelandblue: The Cert request has been updated. So I have updated the profile for this device to a profile that supports battery.

ctowns (approved):

PR #1332, opened 2024-04-17
refactor: Prepare Hue Driver for more device types

dljsjr:
Historically, our first-party Philips Hue integration has only supported light bulbs. With the intent of adding support for a variety of other device types available in the Philips Hue ecosystem, this is a preparatory refactor.

The goal is to address some places in the driver where we assume that all non-bridge devices are light bulbs.

---

## Progress

- [x] Refactor Discovery: #1333 
- [x] Refactor Refresh Capability Handlers: #1334 
- [x] Refactor Attribute Event Emitters: #1335 
- [x] Refactor Lifecycle Handlers: #1343
- [x] Refactor SSE Callbacks
- [x] Refactor Stray Light Handler
- [x] Custodial Refactor (docstrings/variables/functions that don't need to have their logic changed, but need their names updated to not say "light")


dljsjr:
Closing in favor of including these commits as part of #1348

PR #1333, opened 2024-04-17
Refactor discovery for addition of new device types

dljsjr:
This is mostly a structural refactor as opposed to a functional one, but there are subtle changes.

First, we refactored the discovery module in to a directory, as we're going to be breaking the file up. The first extraction here is moving the handler for processing a discovered light in to its own file.

Next, we changed some names of some variables where it was assumed that we were working with lights; this is purely a comprehension/readability refactor.

The only functional change made here was a modification to the `supported_resource_types` mapping, which was originally just a map of booleans. This is now a map from a resource type to a handler function. In this case, we import the handler function from the new file mentioned above.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/disco/init.lua
+++ b/drivers/SmartThings/philips-hue/src/disco/init.lua
@@ -9,6 +9,8 @@ local Fields = require "fields"
 local HueApi = require "hue.api"
 local utils = require "utils"
 
+local LightDiscovery = require "disco.light_disco"
+
 local SERVICE_TYPE = "_hue._tcp"
 local DOMAIN = "local"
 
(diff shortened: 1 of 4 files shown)


greens (approved):

drivers/SmartThings/philips-hue/src/disco/init.lua
```diff
@@ -18,19 +20,22 @@ local DOMAIN = "local"
...
-  light_state_disco_cache = {},
+  device_state_disco_cache = {},
   ServiceType = SERVICE_TYPE,
   Domain = DOMAIN,
   discovery_active = false
 }
 
 local supported_resource_types = {
-  light = true
+  light = LightDiscovery.process_discovered_light
```
cjswedes, line 30: Is the intention that each new device type will get its own `*Discovery` module with a function referenced here to process it once the driver becomes aware of the device?

cjswedes (approved):

varzac:
What is your testing strategy on these refactors?  Are you testing each PR?  Or are you testing in aggregate?  I want to make sure we have some level of intermediate testing before testing the final refactor output.

dljsjr:
> What is your testing strategy on these refactors? Are you testing each PR? Or are you testing in aggregate? I want to make sure we have some level of intermediate testing before testing the final refactor output.

I've been running each refactor myself as I've been going on a production ST Station, testing the modified codepaths. I'd like to invite any of the other reviewers with Hue setups to do the same.

PR #1334, opened 2024-04-17
Refactor refresh capability handlers for addition of new device types

dljsjr:
Like the previous discovery refactor, this is *mostly* structural. We do some extracting of logic into new modules in preparation for the new functionality that's going to be incoming to keep the code legible and organized.

The only capability handlers that needed to be generalized were the refresh handlers, as they made light-specific assumptions in a few places.

We also do a bit of generalizing to the Hue REST API module that doesn't change functionality.

Depends on #1333


drivers/SmartThings/philips-hue/src/handlers/init.lua
greens, line 1: this probably should have been a `git mv`

drivers/SmartThings/philips-hue/src/handlers/refresh_handlers.lua
```diff
@@ -0,0 +1,256 @@
...
+  cosock.spawn(
+    function()
+      local child_devices = bridge_device:get_child_list() --[=[@as HueChildDevice[]]=]
+
+      if not bridge_device:get_field(Fields._INIT) then
+        log.warn("Bridge for lights not yet initialized, can't refresh yet.")
+        return
+      end
+
+      local hue_api = bridge_device:get_field(Fields.BRIDGE_API) --[[@as PhilipsHueApi]]
```
greens, line 25: you've got `--[=[` above and `--[[` here and these annotations seem to have broken github's syntax highlighting so just double-checking this is all correct.

drivers/SmartThings/philips-hue/src/hue/api.lua
```diff
@@ -267,28 +270,39 @@ function PhilipsHueApi.request_api_key(bridge_ip, socket_builder)
...
 
-function PhilipsHueApi:get_lights() return do_get(self, "/clip/v2/resource/light") end
+function PhilipsHueApi:get_all_reprs_for_rtype(rtype) return do_get(self, string.format("/clip/v2/resource/%s", string.lower(tostring(rtype)))) end
+
+function PhilipsHueApi:get_rtype_by_rid(rtype, rid)
+  return do_get(
+    self,
+    string.format("/clip/v2/resource/%s/%s",
+    string.lower(tostring(rtype)),
+    string.lower(tostring(rid)))
```
greens, line 280: I might add an indent here to indicate these are arguments to `string.format` and not `do_get`. I was a bit confused for a sec.

drivers/SmartThings/philips-hue/src/handlers/refresh_handlers.lua
```diff
@@ -0,0 +1,256 @@
...
+        light_device.log.debug(string.format("Zigbee Status for %s is connected", light_device.label))
+        light_device:online()
+        light_device:set_field(Fields.IS_ONLINE, true)
+      else
+        light_device.log.debug(string.format("Zigbee Status for %s is not connected", light_device.label))
+        light_device:set_field(Fields.IS_ONLINE, false)
+        light_device:offline()
+      end
+    end
+  end
```
cjswedes, line 111: Two things:
1. This seems like it could be generalized for all device types? Potentially something that could go in this refactor.
2. Should we be refreshing the state on a device that we know is unreachable by the bridge?

drivers/SmartThings/philips-hue/src/handlers/refresh_handlers.lua
```diff
@@ -56,30 +43,43 @@ function M.do_refresh_all_for_bridge(driver, bridge_device)
...
-          M.do_refresh_light(driver, device, conn_status_cache, light_status_cache)
+        -- Query for the states of all devices for a device type for a given child device,
+        -- but only the first time we encounter a device type. We cache them since we're refreshing
+        -- everything.
+        if
+            type(device_type_refresh_handlers_map[device_type]) == "function" and
+            statuses_by_device_type[device_type] == nil
+        then
+          local reprs, rest_err = hue_api:get_all_reprs_for_rtype(device_type)
+          if reprs ~= nil and rest_err == nil then
```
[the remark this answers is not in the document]
dljsjr: No, this is the success case.

In some situations, the Hue API will return JSON even when there's an HTTP error.

drivers/SmartThings/philips-hue/src/handlers/init.lua
dljsjr: It was, but it might have gotten mucked up during a rebase or something? I'm not sure.

drivers/SmartThings/philips-hue/src/handlers/refresh_handlers.lua
dljsjr: @cjswedes I started looking at this yesterday and I realized that if we build this on top of the next two PR's in the stack, we can actually implement a single refresh handler for all devices. So I'll be making this change further along.

greens (approved):

PR #1335, opened 2024-04-17
 Refactor attribute event emitters in to a module.

dljsjr:
This is a purely structural refactor.

Depends on #1334

### DIFF
--- a/drivers/SmartThings/philips-hue/src/init.lua
+++ b/drivers/SmartThings/philips-hue/src/init.lua
@@ -17,36 +17,32 @@
 --  Improvements to be made:
 --
 --  ===============================================================================================
+local Driver = require "st.driver"
+
 local cosock = require "cosock"
 local log = require "log"
-
 local capabilities = require "st.capabilities"
-local Driver = require "st.driver"
 local json = require "st.json"
 local st_utils = require "st.utils"
 -- trick to fix the VS Code Lua Language Server typechecking
 ---@type fun(val: table, name: string?, multi_line: boolean?): string
 st_utils.stringify_table = st_utils.stringify_table
 
+local Consts = require "consts"
 local Discovery = require "disco"
 local EventSource = require "lunchbox.sse.eventsource"
 local Fields = require "fields"
-local handlers = require "handlers"
 local HueApi = require "hue.api"
-local HueColorUtils = require "hue.cie_utils"
+
+local attribute_emitters = require "handlers.attribute_emitters"
+local command_handlers = require "handlers.commands"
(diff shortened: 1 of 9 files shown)


drivers/SmartThings/philips-hue/src/utils.lua
```diff
@@ -5,6 +5,11 @@ local utils = {}
 
 local MAC_ADDRESS_STR_LEN = 12
 
+
+function utils.kelvin_to_mirek(kelvin) return 1000000 / kelvin end
```
greens, line 9: isn't the unit a "mired"?

and I don't know if hue is better-behaved, but we've definitely run into situations where we get divby0 errors here before on the zigbee side

dljsjr: > isn't the unit a "mired"?

You and I actually had this conversation when I first wrote the driver a few years ago.

They're the same unit with different names. Mired is _**micro reciprocal degree**_, an acronym. Mirek is the SI Unit, but nobody uses it. [From Wikipedia](https://en.wikipedia.org/wiki/Mired):

```
The [SI](https://en.wikipedia.org/wiki/SI) term for this unit is the reciprocal megakelvin (MK−1), shortened to mirek, but this term has not gained traction.[[1]](https://en.wikipedia.org/wiki/Mired#cite_note-1)
```

The Hue API and documentation use mirek, so I’m using that for consistency here.

> and I don't know if hue is better-behaved, but we've definitely run into situations where we get divby0 errors here before on the zigbee side

We sanitize the input kelvin for capability commands and the input mireks for attribute emits before calling this function:

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1335/files#diff-ddccea1fd0c65608f45092f0f576c125a2afaebd37e7b9bb78855eece84037e5R202

greens (approved):

drivers/SmartThings/philips-hue/src/init.lua
```diff
@@ -815,7 +711,7 @@ local function do_bridge_network_init(driver, bridge_device, bridge_url, api_key
     eventsource.onerror = function()
       log.error_with({ hub_logs = true }, string.format("Hue Bridge \"%s\" Event Source Error", bridge_device.label))
 
-      for _, device_record in ipairs(bridge_device:get_child_list()) do
+      for _, device_record in ipairs(bridge_device:get_child_list() --[=[@as HueChildDevice[]]=]) do
```
cjswedes, line 714: This breaks github's syntax highlighting

cjswedes (approved):

PR #1337, opened 2024-04-18
[Aqara/Smart Pet Feeder] add new event and fix issues

seojune79:
1. add new error status event
2. fix abnormal feeding logs in history tab and life cycle issues

### DIFF
--- a/drivers/Aqara/aqara-feeder/src/init.lua
+++ b/drivers/Aqara/aqara-feeder/src/init.lua
@@ -17,7 +17,12 @@ local PRIVATE_CLUSTER_ID = 0xFCC0
 local PRIVATE_ATTR_ID = 0xFFF1
 local MFG_CODE = 0x115F
 
-local callback_timer = function(driver, device, cmd)
+local function reload_portion(device)
+  local lastPortion = device:get_latest_state("main", capabilities.feederPortion.ID,
+    capabilities.feederPortion.feedPortion.NAME) or 0
+  device:emit_event(capabilities.feederPortion.feedPortion({ value = lastPortion, unit = "servings" }))
+end
+local callback_feed = function(device)
   return function()
     device:emit_event(capabilities.feederOperatingState.feederOperatingState("idle"))
   end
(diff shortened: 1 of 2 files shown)


tpmanley (approved):

drivers/Aqara/aqara-feeder/src/init.lua
```diff
@@ -104,23 +106,25 @@ local function petFeeder_handler(driver, device, value, zb_rx)
...
   end
 end
 
 -- [[ lifecycle_handlers ]]
 local function device_added(driver, device)
-  -- private protocol enable
-  device:send(cluster_base.write_manufacturer_specific_attribute(device,
-    PRIVATE_CLUSTER_ID, 0x0009, MFG_CODE, data_types.Uint8, 1))
-  -- init
   do_payload(device, 4, 24, 85, OP_WRITE, 1, 0)
```
greens, line 128: does this private protocol need to be enabled before the `do_payload` here will actually do anything?

tpmanley (changes requested):
In my testing when the error state was hit I could no longer open the device card in the app

seojune79:
> In my testing when the error state was hit I could no longer open the device card in the app

Can you tell me how you tested it?

tpmanley:
> > In my testing when the error state was hit I could no longer open the device card in the app
> 
> Can you tell me how you tested it?

I used the button on the feeder to dispense food over and over again until it jammed up

seojune79:
> > > In my testing when the error state was hit I could no longer open the device card in the app
> > 
> > 
> > Can you tell me how you tested it?
> 
> I used the button on the feeder to dispense food over and over again until it jammed up

It's hard for me to understand that I cannot open the card because of ["device:mit_event(capabilities.feederOperatingState.feederOperatingState(evt))"](https://github.com/seojune79/SmartThingsEdgeDrivers/blob/9942b52bd4029a52078094144188d00d0baba808/drivers/Aqara/aqara-feeder/src/init.lua#L122).

I don't have the same problem.
- 40sec
- 1min 22sec
[my test](https://drive.google.com/file/d/1Xx-sHUw9sA4QvDfHbPs_prYjO5hZqzAx/view?usp=sharing)

Can you check it again?

tpmanley (approved):
I tested again and it is now working for me as well

PR #1339, opened 2024-04-19
Matter-Switch: Limit reads subscriptions

nickolas-deboom:
Subscribe to min/max color temperature and min/max switch level attributes rather than reading them during device init. This is to resolve an issue identified while investigating `InteractionQueueFull` spikes in the exception reports, where onboarding a large number of devices resulted in more attributes being read than the maximum number of attributes per read interaction (9). These attribute reads were split into multiple read requests, which exceeded the maximum number of concurrent requests (5), causing the remaining requests to be dropped.

Utilizing subscriptions for these attributes rather than manually querying the device will take advantage of the current optimizations for subscription requests and avoid this issue.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -13,7 +13,6 @@
 -- limitations under the License.
 
 local capabilities = require "st.capabilities"
-local im = require "st.matter.interaction_model"
 local log = require "log"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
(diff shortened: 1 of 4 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -39,7 +38,6 @@ local SWITCH_INITIALIZED = "__switch_intialized"
 -- in the device table for devices that joined prior to this transition, and it
 -- will not be set for new devices.
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
-local BOUNDS_CHECKED = "__bounds_checked"
```
tpmanley, line 42: It might be a good idea to include a change that deletes the field from persistent storage. That way if we later add some new code that happens that same field name we won't run into issues with stale data. It wouldn't need to live in the code forever, maybe just for a few weeks to give most hubs a chance to pick it up.

tpmanley: Gotcha, thanks!

ctowns (approved):
LGTM! Don't forget to squash your commits!

samadDotDev (approved):
For reference, this will cause wildcard subscription on the level control cluster, and in general, if a device supports all capabilities in the `subscribed_attributes` handler, then we'll be 2 clusters away from falling back to a global wildcard subscription.

Con-Tejus (approved):

gharveymn (approved):

PR #1343, opened 2024-04-24
Refactor Lifecycle Handlers

dljsjr:
Like previous refactors, this is primarily structural in nature as we extract the handlers and namespace them to make it easier to look them up generically via the service name.

One caveat to that, though, is the refactor to the `utils` module and the new techniques for determining device types.

Previously, because we only supported bridges and lights, there was a good deal of logic that checked for a device being a light simply by seeing if it was *not* a bridge, with some brittle DNI inspection heuristics.

Now that we need to support more than lights, this won't suffice.

This is further complicated because we cannot rely on the presence of any information introspected during Discovery to still be around at `added`.

We introduce a new scheme for the parent-assigned child key to solve this, as we know this will be available on the device record at the time the `added` handler is invoked.

We also maintain backward compatibility with the light devices that were either migrated or joined before this refactor took place.

Lastly, as part of the utils refactor, we harden the DNI validation logic.


drivers/SmartThings/philips-hue/src/handlers/lifecycle_handlers.lua
```diff
@@ -0,0 +1,120 @@
+local log = require "log"
```
cjswedes, line 1: I typically follow the practice of having an init.lua file in the directory rather than a lua file outside of the directory where there are submodules/other lua files. I think we have followed this for almost everything in our lua libs (except cosock) and drivers. Its not a huge deal, but just sort of weird IMO to see directories without an init.lua file in them. Was there a reason you did it the way you did?

cjswedes (approved):
I think this looks good. However, I recommend some testing for the backwards compatibility case where the IDs are mixed between what you have done here and what has been done previously. A bug there would be quite detrimental.

drivers/SmartThings/philips-hue/src/handlers/lifecycle_handlers.lua
dljsjr: Yeah, I wanted to minimize the occurrences of [this](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1334#discussion_r1578241412) after Steven's comment on the prior PR.

dljsjr:
> I think this looks good. However, I recommend some testing for the backwards compatibility case where the IDs are mixed between what you have done here and what has been done previously. A bug there would be quite detrimental.

I have in fact tested this, but it's a good call-out and I'll be testing it a few more times before I merge this.

NoahCornell (approved):

PR #1344, opened 2024-04-24
CHAD 12650 Z-Wave: Update base thermostat driver to use dynamic numeric constraints for heating/cooling setpoints

matthewdehaven:
Uses Z-Wave `ThermostatSetpoint:CapabilitiesGet` to request the thermostat setpoints for heating and cooling on initialization.


drivers/SmartThings/zwave-thermostat/src/init.lua
```diff
@@ -37,6 +48,10 @@ local function device_added(driver, device)
     device:is_cc_supported(cc.THERMOSTAT_FAN_MODE) then
     device:send(ThermostatFanMode:SupportedGet({}))
   end
+  if device_supports_thermostat_setpoint(device) then
+    device:send(ThermostatSetpointV3:CapabilitiesGet({setpoint_type = ThermostatSetpoint.setpoint_type.HEATING_1}))
+    device:send(ThermostatSetpointV3:CapabilitiesGet({setpoint_type = ThermostatSetpoint.setpoint_type.COOLING_1}))
```
cbaumler, line 53: Just curious why `device_supports_thermostat_setpoint` returns true if one or the other of the heating and cooling setpoint capabilities is supported, but here we are getting both. Would it make sense to only get what is supported?

greens (approved):

cbaumler (commented):
Nice work!

drivers/SmartThings/zwave-thermostat/src/init.lua
greens: That is a good point. We do have some heat-only thermostats.

matthewdehaven: Thanks for pointing out, I didn't think of that case. I adjusted this to be two conditions: Check for support for each heating and cooling setpoints independently before sending the request for the corresponding bounds.

cbaumler (approved):

greens (commented):
I was just thinking about this. I think we might want to have some sanity check values to gate these reports. We have seen devices report garbage on similar reports. You also may want to do some testing with a real device.

matthewdehaven:
> I was just thinking about this. I think we might want to have some sanity check values to gate these reports. We have seen devices report garbage on similar reports. You also may want to do some testing with a real device.

Okay, yes it is my intention to test with a real device. I'm working on getting access to one.

PR #1345, opened 2024-04-24
Add a Dimmer Switch in Matter fingerprints

DongHoon-Ryu:
Add to fingerprints to use Matter switch-level profile that is already registered.
  IOTE-3863, IOTE-3757

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -2191,6 +2191,11 @@ matterGeneric:
     deviceTypes:
       - id: 0x010B # Dimmable Plug-in Unit
     deviceProfileName: plug-level
+  - id: "matter/dimmer/switch"
+    deviceLabel: Matter Dimmer Switch
+    deviceTypes:
+      - id: 0x0104 # Dimmer Switch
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -2191,6 +2191,11 @@ matterGeneric:
     deviceTypes:
       - id: 0x010B # Dimmable Plug-in Unit
     deviceProfileName: plug-level
+  - id: "matter/dimmer/switch"
+    deviceLabel: Matter Dimmer Switch
+    deviceTypes:
+      - id: 0x0104 # Dimmer Switch
```
tpmanley, line 2197: Dimmer Switch only has implements the client side for the On/Off and Level Control clusters so it won't be possible for us to control those clusters since the hub is the client and it needs server clusters on the device in order to communicate with it. Therefore I do not think it makes sense to add this fingerprint.

lelandblue:
@DongHoon-Ryu Is this PR still needed? Can we close?

DongHoon-Ryu:
> @DongHoon-Ryu Is this PR still needed? Can we close?

No, It is not needed. I will close it. Thanks.

PR #1347, opened 2024-04-25
New Device (Matter Switch) Linkind Smart Light Bulb A60

lelandblue:
This PR is a for a WWST Certification request for the Linkind LS0101911261 Smart Light Bulb A60


ctowns (approved):
Approved, but let's follow up to figure out if this device provides real color temp values that can be used by the dynamic constraints functionality once we've started testing it.

PR #1350, opened 2024-04-30
Added support for the Philips Hue Smart Plug

kiera-robinson-st:
Philips Hue Smart Plug is exposed as a 'light' service so existing light code was modified to include the case for Philips Hue Smart Plug new device adoption and use


dljsjr:
Working from OneApp now:

IMG_6565

dljsjr (approved):

drivers/SmartThings/philips-hue/profiles/plug.yml
```diff
@@ -0,0 +1,12 @@
+name: plug
+components:
+- id: main
+  capabilities:
+  - id: switch
+    version: 1
+  - id: samsungim.hueSyncMode
```
tpmanley, line 7: Can the plug be used with hue sync?

dljsjr: It's unclear to me what exactly it would do, but the representation does return a `mode` value when queried.

tpmanley: OK, probably does make sense to leave it in then

drivers/SmartThings/philips-hue/profiles/plug.yml
```diff
@@ -0,0 +1,12 @@
+name: SmartPlug
```
tpmanley, line 1: It's not required, but I think we typically try to make the `name` match the filename

drivers/SmartThings/philips-hue/profiles/motion-sensor.yml
```diff
@@ -0,0 +1,16 @@
+name: motion-sensor
+components:
+- id: main
+  capabilities:
+  - id: motionSensor
+    version: 1
+  - id: battery
```
greens, line 7: tom pointed this out on one of doug's pr, but we typically put `battery` right above `refresh`

greens (commented):
hard to tell what's unique to this PR.

Maybe change the target branch to `feat/device/motion-and-contact-sensors`? I know that complicates timing, though.

dljsjr:
> hard to tell what's unique to this PR.
> 
> Maybe change the target branch to `feat/device/motion-and-contact-sensors`? I know that complicates timing, though.

Did this last week and forgot to ping you.

tpmanley (approved):

drivers/SmartThings/philips-hue/profiles/plug.yml
dljsjr: Fixed on https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1359

cjswedes (approved):

greens (approved):

PR #1353, opened 2024-05-01
Adds support for 4-button Hue remote accessories

dljsjr:
This is the Hue Dimmer Remote and the Hue Tap Dial.
    
Note that we do *not* yet have support for the rotary portion of the tap dial.


drivers/SmartThings/philips-hue/profiles/4-button-remote.yml
```diff
@@ -0,0 +1,30 @@
+name: 4-button-remote
+components:
```
varzac, line 2: @greens @tpmanley is this how we normally model buttons.  Looking briefly at Zigbee just judging on profile names it seems like we might typically use one component per button in addition to a "main" component.  In which case do we want an additional button only component here?

drivers/SmartThings/philips-hue/src/disco/button.lua
```diff
@@ -0,0 +1,148 @@
...
+      button_remote_description[button_id_key] = button_repr.data[1].id
+
+      if control_id == 1 and button_remote_description.id == nil then
+        button_remote_description.id = button_repr.data[1].id
+      end
+    end
+  end
+
+  local battery, battery_err = api_instance:get_device_power_by_id(rid_by_rtype[HueDeviceTypes.DEVICE_POWER])
+  if battery_err then return nil, battery_err end
```
varzac, line 58: Does the caching below need to happen regardless of this?  Or is it always expected that there be device power info and this is a fatal error?

drivers/SmartThings/philips-hue/src/disco/contact.lua
```diff
@@ -0,0 +1,127 @@
...
+---@param resource_id string
+---@param device_service_info HueDeviceInfo
+---@param device_state_disco_cache table<string, table>
+---@param st_metadata_callback fun(driver: HueDriver, metadata: table)?
+function M.handle_discovered_device(
+    driver, bridge_id, api_instance,
+    resource_id, device_service_info,
+    device_state_disco_cache, st_metadata_callback
+)
+  local err = select(2,
```
varzac, line 96: What is this select doing?

drivers/SmartThings/philips-hue/src/disco/light.lua
```diff
@@ -95,6 +91,7 @@ function M.handle_discovered_device(
         parent_assigned_child_key = parent_assigned_child_key
       }
 
+      log.debug(true, st_utils.stringify_table(st_metadata, "light create", true))
```
varzac, line 94: Do we want this debug log?

drivers/SmartThings/philips-hue/profiles/contact-sensor.yml
```diff
@@ -0,0 +1,14 @@
...
+  - id: contactSensor
+    version: 1
+  - id: tamperAlert
+    version: 1
+  - id: battery
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: MotionSensor
```
cjswedes, line 14:
```suggestion
  - name: ContactSensor
```

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
```diff
@@ -125,30 +112,175 @@ local function _emit_light_events_inner(light_device, light_repr)
...
+    log.debug(true, "emit power")
+    button_device:emit_event(
+      capabilities.battery.battery(
+        st_utils.clamp_value(button_info.power_state.battery_level, 0, 100)
+      )
+    )
+  end
+
+  local button_idx_map = button_device:get_field(Fields.BUTTON_INDEX_MAP)
+  if not button_idx_map then
```
varzac, line 143: should there be a return in this if statement?

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
```diff
@@ -125,30 +112,175 @@ local function _emit_light_events_inner(light_device, light_repr)
...
+    button_device.profile.components[component_idx]:emit_event(
+      capabilities.button.button.held({state_change = true})
+    )
+  end
+
+  if button_report.event == "long_release" and button_device:get_field("button_held") then
+    button_device:set_field("button_held", false)
+  end
+
+  if button_report.event == "short_release" and not button_device:get_field("button_held") then
```
varzac, line 174: So just to check, in a momentary press, the `long_press` would never come in.  Is there a `short_press` equivalent?  I still think this is the right thing here to just generate on the release, but was curious,

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
```diff
@@ -125,30 +112,175 @@ local function _emit_light_events_inner(light_device, light_repr)
...
+  end
+end
+
+function AttributeEmitters.emit_motion_sensor_attribute_events(sensor_device, sensor_info)
+  if sensor_device == nil or (sensor_device and sensor_device.id == nil) then
+    log.warn("Tried to emit attribute events for a device that has been deleted")
+    return
+  end
+
+  if sensor_info.power_state then
```
varzac, line 225: Would it make sense to pull this whole check into a 

`handle_power_state_if_needed` 

function that would do this check and emit if needed?  Or is the structure of the info not consistent enough across the devices?

drivers/SmartThings/philips-hue/src/disco/button.lua
```diff
@@ -0,0 +1,148 @@
...
+      log.error(
+        string.format(
+          "Driver does not currently support remotes with %s buttons, cannot create device", button_description.num_buttons
+        )
+      )
+      return
+    end
+    local button_profile_ref = ""
+    if button_description.num_buttons == 4 then
+      button_profile_ref = "4-button-remote"
```
tpmanley, line 126: I think we probably want to log the error case here and skip the device creation since without a profile the device isn't going to be useable. Thoughts?

drivers/SmartThings/philips-hue/profiles/4-button-remote.yml
tpmanley: It looks like we've been a bit inconsistent. For Matter we did it this way, and for Zigbee/Z-Wave we did it like you describe. Ideally we'd be consistent across all device types, but in this case we're aiming for consistency with Matter in particular since the Hue bridge can also be joined using Matter, so I'm inclined to leave it as is. Thoughts @greens @ctowns ?

greens: Yeah it's not consistent and we often get feedback about the "main" catchall button. One of the main (no pun intended) reasons we _did_ include it in some cases was because there used to be a bug where a component's capabilities would not function if the main component didn't also include that capability. It has been fixed, but left us with this confusing legacy.

I'd err on the side of not having a button component in "main" unless it's used as just one of the buttons, rather than the typical implementation, which is that it fires from any component button event.

drivers/SmartThings/philips-hue/src/utils.lua
```diff
@@ -54,6 +54,23 @@ function utils.kelvin_to_mirek(kelvin) return 1000000 / kelvin end
 
 function utils.mirek_to_kelvin(mirek) return 1000000 / mirek end
 
+-- TODO: The Hue API itself doesn't have events for multipresses, however, it will
+-- emit batched "short release" eventsource on the SSE stream if they're close together.
+-- Right now the SSE stream handling is a relatively dumb pass-through that doesn't inspect
+-- the data as it unpacks it; it just dispatches each event in the batch as it encounters it.
+-- We could implement 2x-6x press if we add some smarts to the SSE stream handling.
```
greens, line 61: Possibly, I wouldn't try to force it. There's ways to do this with just the pass-through (involving lots of timers), but I'd say if it's not distinguished in the API then you've made the right call.

drivers/SmartThings/philips-hue/profiles/4-button-remote.yml
dljsjr: The button component in `main` maps to to Button 1 in a 1, 2, or 4 button remote config; it is not a catch-all button capability, it only maps to Button 1. And as Tom indicated, I did it this way to be consistent in presentation with the Matter integration, which was a goal outlined in the PRD.

drivers/SmartThings/philips-hue/src/disco/contact.lua
dljsjr: That's a native Lua select, not an async select.

It throws away all arguments that come before the given index in a variadic return situation. The benefit here compared to binding to `_` is that you don't take up memory with a variable binding. It's basically saying we don't care about the first return value, we only care if there's an error.

drivers/SmartThings/philips-hue/src/disco/light.lua
dljsjr: This specific one may have been removed but I think I'm going to leave most in during SVT testing.

drivers/SmartThings/philips-hue/src/disco/button.lua
dljsjr: An error here would indicate a connectivity error, not a REST/HTTP error. So there would be no point in continuing becuase we would expect additional REST calls to fail, since the REST client itself has retry heuristics built in.

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
dljsjr: I believe so, yes. Good catch.

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
dljsjr: Yeah, there's no short_press equivalent.

The five supported values are:

- `initial_press`
- `short_release`
- `long_press`
- `repeat`
- `long_release`

The potential event sequences are either:

1. `initial_press` -> `short_release`, or
2. `initial_press` -> `long_press` -> `repeat` -> … -> `long_release`

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
dljsjr: I needed to finish all the devices to answer this question, and the answer is that yes it can be generalized.

The entire `sensor` class of devices (of which Signify includes Buttons) can be generalized quite a bit. But I didn't wanna back myself in to a crappy design by doing a DRY refactor too early.

drivers/SmartThings/philips-hue/src/disco/button.lua
dljsjr: Agreed, good call out.

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
tpmanley: Is `repeat` just sent every second or something if the button is held for a long time?

tpmanley (approved):

drivers/SmartThings/philips-hue/src/disco/button.lua
dljsjr: I confirmed that as long as a `GET` REST call returns JSON, you shouldn't get an error here; you'll get a valid response with errors in the JSON payload and no `data` array.

- All `GET` REST calls converge on `do_get`, which sends the HTTP request to a separate task over a channel: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/038217a0da0b337e3b3d7b569fea847d73a7d142/drivers/SmartThings/philips-hue/src/hue/api.lua#L207-L222
- The REST call has retry logic, and then tries to forward *whatever* it gets back at the end of that to `process_rest_response`: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/038217a0da0b337e3b3d7b569fea847d73a7d142/drivers/SmartThings/philips-hue/src/hue/api.lua#L167-L171
- As long as the response has a body and that body is valid JSON, it will not return an error: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/038217a0da0b337e3b3d7b569fea847d73a7d142/drivers/SmartThings/philips-hue/src/hue/api.lua#L102-L116
- Querying the Hue Bridge for a non-existent resource does give an HTTP 404, but we're not currently looking at that for error conditions; and the 404 includes a valid JSON body: 
image

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
dljsjr: The button service has a `repeat_interval` property that lets you know how often it'll be sent, but yes:

```json
{
  "errors": [],
  "data": [
    {
      "id": "8b5e4b4e-ee4b-4102-9788-1d3b29a859cc",
      "id_v1": "/sensors/9",
      "owner": {
        "rid": "621164bd-30e5-4e4c-9207-024e8f04b643",
        "rtype": "device"
      },
      "metadata": {
        "control_id": 1
      },
      "button": {
        "last_event": "short_release",
        "button_report": {
          "updated": "2024-05-02T20:22:46.673Z",
          "event": "short_release"
        },
        "repeat_interval": 800,
        "event_values": [
          "initial_press",
          "repeat",
          "short_release",
          "long_release",
          "long_press"
        ]
      },
      "type": "button"
    }
  ]
}
```

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
dljsjr: Fixed on https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1359

drivers/SmartThings/philips-hue/src/disco/button.lua
dljsjr: Fixed on https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1359

drivers/SmartThings/philips-hue/profiles/4-button-remote.yml
cjswedes: IMO the layout we did with matter is the best for the user, and I like that we follow it here. I am fine with the main component representing the first button.

drivers/SmartThings/philips-hue/src/hue_device_types.lua
```diff
@@ -11,7 +12,12 @@ local HueDeviceTypes = {
   ZIGBEE_CONNECTIVITY = "zigbee_connectivity"
 }
 
+local SupportedNumberOfButtons = {
+  [4] = true
```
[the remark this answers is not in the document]
dljsjr: In the second to last PR.

drivers/SmartThings/philips-hue/src/disco/button.lua
```diff
@@ -0,0 +1,148 @@
...
+---@return string? err nil on success
+function M.update_state_for_all_device_services(api_instance, device_service_id, bridge_id, primary_button_resource_id, cache)
+  log.debug("----------- Calling REST API")
+  local device_service_info, err = api_instance:get_device_by_id(device_service_id)
+  if err or not (device_service_info and device_service_info.data) then
+    log.error("Couldn't get device info for button, error: " .. st_utils.stringify_table(err))
+    return
+  end
+
+  log.debug("------------ _do_update")
```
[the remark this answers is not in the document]
dljsjr: Fixed on https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1359

NoahCornell (approved):

PR #1354, opened 2024-05-01
Chad 12671

matthewdehaven:
Dynamically set heat and cool min/max setpoints for ZigBee thermostats by requesting the values from the device.

### DIFF
--- a/drivers/SmartThings/zigbee-thermostat/src/init.lua
+++ b/drivers/SmartThings/zigbee-thermostat/src/init.lua
@@ -196,31 +196,29 @@ local emit_heat_setpoint_bounds_event = function (device)
   device:set_field(SETPOINT_MAX_HEAT, nil)
 end
 
-local thermostat_setpoint_min_cool_handler = function(driver, device, setpoint)
-  device:set_field(SETPOINT_MIN_COOL, setpoint.value)
-  if device:get_field(SETPOINT_MIN_COOL) and device:get_field(SETPOINT_MAX_COOL) then
-      emit_cool_setpoint_bounds_event(device)
-  end
-end
+local setpoint_limit_handler_factory = function(handled_attribute)
 
-local thermostat_setpoint_max_cool_handler = function(driver, device, setpoint)
-  device:set_field(SETPOINT_MAX_COOL, setpoint.value)
-  if device:get_field(SETPOINT_MIN_COOL) and device:get_field(SETPOINT_MAX_COOL) then
-      emit_cool_setpoint_bounds_event(device)
-  end
-end
+  local paired_attributes = {
+    [SETPOINT_MIN_HEAT] = SETPOINT_MAX_HEAT,
+    [SETPOINT_MAX_HEAT] = SETPOINT_MIN_HEAT,
+    [SETPOINT_MIN_COOL] = SETPOINT_MAX_COOL,
+    [SETPOINT_MAX_COOL] = SETPOINT_MIN_COOL
+  }
 
-local thermostat_setpoint_min_heat_handler = function(driver, device, setpoint)
-  device:set_field(SETPOINT_MIN_HEAT, setpoint.value)
-  if device:get_field(SETPOINT_MIN_HEAT) and device:get_field(SETPOINT_MAX_HEAT) then
-      emit_heat_setpoint_bounds_event(device)
-  end
-end
+  local paired_attribute = paired_attributes[handled_attribute]
+  local event_emit_function = nil
 
-local thermostat_setpoint_max_heat_handler = function(driver, device, setpoint)
-  device:set_field(SETPOINT_MAX_HEAT, setpoint.value)
-  if device:get_field(SETPOINT_MIN_HEAT) and device:get_field(SETPOINT_MAX_HEAT) then
-      emit_heat_setpoint_bounds_event(device)
+  if handled_attribute == SETPOINT_MIN_HEAT or handled_attribute == SETPOINT_MAX_HEAT then
+    event_emit_function = emit_heat_setpoint_bounds_event
+  else
+    event_emit_function = emit_cool_setpoint_bounds_event
+  end
+
+  return function(driver, device, setpoint)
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zigbee-thermostat/src/init.lua
```diff
@@ -164,6 +164,63 @@ local thermostat_mode_setter = function(mode_name)
   end
 end
 
+local thermostat_setpoint_min_cool = nil
+local thermostat_setpoint_max_cool = nil
+local thermostat_setpoint_min_heat = nil
+local thermostat_setpoint_max_heat = nil
```
greens, line 170: Unfortunately, local variables in here will be in scope for _all_ devices using this driver. In order to use device-specific values you'll need to use the data store with `device:set_field()`

greens (commented):
You'll probably want to add some querying of these attributes to the `added` handler in the base driver as well.

matthewdehaven:
> You'll probably want to add some querying of these attributes to the `added` handler in the base driver as well.

My understanding of this comment is basically that we need something to initially query these attributes when a devices is added. Correct me if I misinterpreted.

I added these attributes to the list of attributes requested in `do_refresh`, which is called by the `added` handler. My belief was that would effectively cause a query of those attributes on addition of a device. Is that incorrect or insufficient?

```
Thermostat.attributes.MinHeatSetpointLimit
Thermostat.attributes.MaxHeatSetpointLimit
Thermostat.attributes.MinCoolSetpointLimit
Thermostat.attributes.MaxCoolSetpointLimit
```

drivers/SmartThings/zigbee-thermostat/src/init.lua
matthewdehaven: I pushed an update to use `device:set_field` and `device:get_field` instead of the local variables.

drivers/SmartThings/zigbee-thermostat/src/init.lua
```diff
@@ -164,6 +169,61 @@ local thermostat_mode_setter = function(mode_name)
...
+      emit_heat_setpoint_bounds_event(device)
+  end
+end
+
+local thermostat_setpoint_max_heat_handler = function(driver, device, setpoint)
+  device:set_field(SETPOINT_MAX_HEAT, setpoint.value)
+  if device:get_field(SETPOINT_MIN_HEAT) and device:get_field(SETPOINT_MAX_HEAT) then
+      emit_heat_setpoint_bounds_event(device)
+  end
+end
```
greens, line 225: You could save yourself some repetition here if you wanted by using a factory for creating functions here. Something like:

```
local setpoint_limit_handler_factory = function(handled_attribute, paired_attribute)
  return function(driver, device, setpoint)
    device:set_field(handled_attribute, setpoint.value)
    if device:get_field(paired attribute) and device:get_field(handled_attribute) then
      emit_heat_setpoint_bounds_event(device)
    end
  end
end
```

Then in the handlers below you could do

```
[Thermostat.attributes.MinHeatSetpointLimit.ID] = setpoint_limit_handler_factory(SETPOINT_MIN_HEAT, SETPOINT_MAX HEAT),
```

matthewdehaven: Done. I created a handler factory.

greens (commented):
Alright that didn't save as much as I was hoping. What if you passed in the parts of the string instead, i.e. `setpoint_handler_factory(MIN, HEAT)`? Gives you enough information to simplify a lot I think, and the `handled_attribute` can be constructed using `"setpoint"..min_or_max..heat_or_cool`

matthewdehaven:
I tested these driver changes on a zigbee thermostat with the following the steps below.

From this testing, I learned zigbee temperatures are 100 * Celsius temperature, so I had to add a division of the received temperatures before emitting events in Celsius.

- Loaded modified driver onto the hub
- Added thermostat to the hub
- Verified in log output the min/max setpoints were received and events
  were emitted for each heating and cooling
  - `2024-05-14T18:35:19.372619969Z INFO Zigbee Thermostat -- MD  <ZigbeeDevice: e6795867-2184-4f4e-a323-243e78af10fd [0xF5EB] (Fidure Thermostat)> emitting event: {"attribute_id":"heatingSetpointRange","capability_id":"thermostatHeatingSetpoint","component_id":"main","state":{"unit":"C","value":{"maximum":30.0,"minimum":7.0}}}`
  - `2024-05-14T18:35:20.227855802Z INFO Zigbee Thermostat -- MD  <ZigbeeDevice: e6795867-2184-4f4e-a323-243e78af10fd [0xF5EB] (Fidure Thermostat)> emitting event: {"attribute_id":"coolingSetpointRange","capability_id":"thermostatCoolingSetpoint","component_id":"main","state":{"unit":"C","value":{"maximum":30.0,"minimum":7.0}}}`
- Raised and lowered the setpoints for the thermostat in the app and
  verified they could not be set higher or lower than the dynamic bounds
- Modified the driver to set the minimum heat setpoint to a 0C
- Tried again lowering the heating setpoint
- Could go below the previous 7C limit, but hit a different limit around 4C

matthewdehaven:
> Alright that didn't save as much as I was hoping. What if you passed in the parts of the string instead, i.e. `setpoint_handler_factory(MIN, HEAT)`? Gives you enough information to simplify a lot I think, and the `handled_attribute` can be constructed using `"setpoint"..min_or_max..heat_or_cool`

I implemented this approach and it cut the change down by about 20 lines

greens (approved):

lelandblue:
Hey @matthewdehaven Not urgent, I saw tthat this PR is still open. Do we want to take further action with this PR? Please share, when the time is right.

matthewdehaven:
> Hey @matthewdehaven Not urgent, I saw tthat this PR is still open. Do we want to take further action with this PR? Please share, when the time is right.

I think we should merge it in the next release. The change has been tested on an zigbee thermostat and should be ready to go. I'll need to review the edge driver release process this week to be sure I follow the process correctly to get this change in.

PR #1359, opened 2024-05-02
Addresses a handful of bugs with the new device types

dljsjr:
- Fixes an issue where compound devices wouldn't be labeled properly
- Fixes an issue where the stray device task was scanning a bridge every time it onboarded, even if there were no strays
- Fixes an issue where refresh on devices could fail early in driver start by accidentally passing in an API Key string instead of an API Rest Client instance to a function
- Fixes an issue where multi-service devices (Buttons, Motion Sensors, Contact Sensors) could sometimes fail to emit attributes if their init handler ran before their parent bridge's network connection was open.


dljsjr:
@cjswedes @greens I'm exploring a pattern in this PR that could help clean up the places in the Contact/Motion/Button files where we have stuff that's 80% copy/paste with 20% variance. I think that it could be used to generalize the attribute emitters, refresh handlers, and discovery logic. Let me know your thoughts.

tpmanley (approved):

drivers/SmartThings/philips-hue/src/disco/light.lua
```diff
@@ -29,7 +29,13 @@ local function join_light(driver, light, device_service_info, parent_device_id,
     return
   end
 
-  local device_name = light.metadata.name
+  local device_name
+  if light.metadata.name == device_service_info.metadata.name then
```
cjswedes, line 33: Im not sure if nil checks are warranted here. Indexing like this makes me concerned when they aren't present, but maybe things are guaranteed non-nil?

drivers/SmartThings/philips-hue/src/handlers/lifecycle_handlers.lua
```diff
@@ -9,6 +9,31 @@ local StrayDeviceHelper = require "stray_device_helper"
...
+          device.parent_assigned_child_key,
+          expected_parent_assigned_child_key
+        )
+      )
+      -- updating parent_assigned_child_key in metadata isn't supported
+      -- on Lua Libs API versions before 11.
+      if require("version").api <= 10 then
+        return
+      end
+      device:try_update_metadata({ parent_assigned_child_key = expected_parent_assigned_child_key })
```
cjswedes, line 32: So this function will essentially be a no-op until 0.54 FW? And we intend to add support for this in that timeframe?

drivers/SmartThings/philips-hue/src/handlers/lifecycle_handlers/button.lua
```diff
@@ -150,31 +147,14 @@ function ButtonLifecycleHandlers.init(driver, device)
...
-  end
-
-  driver.services_for_device_rid[hue_device_id] = svc_rids_for_device
-  log.debug(st_utils.stringify_table(driver.services_for_device_rid[hue_device_id], "svcs for device rid", true))
-  for rid, _ in pairs(driver.services_for_device_rid[hue_device_id]) do
-    log.debug(string.format("Button %s mapping to [%s]", device.label, rid))
-    driver.hue_identifier_to_device_record[rid] = device
+  if not button_info then
+    log.warn(string.format("Button %s parent bridge not ready, queuing refresh", device and device.label))
+    driver._devices_pending_refresh[device.id] = device
```
cjswedes, line 152: Why is this situation only possible for Contact and motion sensors? Is it related to the multiservice situation with these devices?

drivers/SmartThings/philips-hue/src/stray_device_helper.lua
```diff
@@ -103,114 +103,112 @@ function StrayDeviceHelper.spawn()
...
-          -- TODO: We can optimize around this by keeping track of whether or not this bridge
-          -- needs to be scanned (maybe skip scanning if there are no stray devices?)
-          --
-          -- @doug.stephen@smartthings.com
-          log.info(
-            string.format(
-              "Stray devices handler notified of new bridge %s, scanning bridge",
-              (msg.device.label or msg.device.device_network_id or msg.device.id or "unknown bridge")
+          if next(stray_devices) ~= nil then
+            log.info(
```
cjswedes, line 107: Do you want a log for the else case to know if this check happened, but there were no stray devices?

cjswedes (approved):

drivers/SmartThings/philips-hue/src/disco/light.lua
dljsjr: `metadata` is guaranteed on all services.
`metadata.name` is guaranteed on all `device` services.
`metadata.name` is not guaranteed on all other services, but it *is* guaranteed on `light` services. We confirmed this with Signify.

drivers/SmartThings/philips-hue/src/handlers/lifecycle_handlers.lua
dljsjr: I can move the version number out of it doesn't land, but it's implemented on that PR I have open internally for Scripting Engine that you commented on where I need to update the firmware side structs.

drivers/SmartThings/philips-hue/src/stray_device_helper.lua
dljsjr: This whole thing needs to be refactored still, but it's low priority because it can typically only get hit during migration. There are very rare edge cases where it can still happen outside of migration, but it would involve inventory corruption.

TL;DR probably, yes, but the stray resolver needs a deeper refactor in general that shouldn't block SVT.

drivers/SmartThings/philips-hue/src/handlers/lifecycle_handlers/button.lua
dljsjr: [This code](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/1dbda0f268167b54c1b1e625ae31315ffbca62fd/drivers/SmartThings/philips-hue/src/handlers/lifecycle_handlers/light.lua#L126-L136) could probably be refactored to be the same, honestly. I wanted to leave existing code as untouched as possible.

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
```diff
@@ -282,5 +283,6 @@ device_type_emitter_map[HueDeviceTypes.BUTTON] = AttributeEmitters.emit_button_a
 device_type_emitter_map[HueDeviceTypes.CONTACT] = AttributeEmitters.emit_contact_sensor_attribute_events
 device_type_emitter_map[HueDeviceTypes.LIGHT] = AttributeEmitters.emit_light_attribute_events
 device_type_emitter_map[HueDeviceTypes.MOTION] = AttributeEmitters.emit_motion_sensor_attribute_events
+device_type_emitter_map[HueDeviceTypes.CONTACT] = AttributeEmitters.emit_contact_sensor_attribute_events
```
NoahCornell, line 286: This is also done on 283

NoahCornell (approved):

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
dljsjr: Bad `rerere` playback, fixed!

PR #1360, opened 2024-05-03
Harman-Luxury Improve device update communications

alon-tchelet:
Reduced the thread responsible to poll updates from the device from one per device to a single one for the whole driver. This will help with stability when a user has a lot of devices (Harman Luxury devices in particular) connected to their hub.
Moreover, the HTTP timeout time was reduced and a timeout error will now has its own error message.

### DIFF
--- a/drivers/SmartThings/harman-luxury/src/init.lua
+++ b/drivers/SmartThings/harman-luxury/src/init.lua
@@ -17,22 +17,12 @@ local const = require "constants"
 -- Device Functions
 ----------------------------------------------------------
 
-local function stop_check_for_updates_thread(device)
-  local current_timer = device:get_field(const.UPDATE_TIMER)
-  if current_timer ~= nil then
-    log.info(string.format("create_check_for_updates_thread: dni=%s, remove old timer", device.device_network_id))
-    device.thread:cancel_timer(current_timer)
-  end
-end
-
 local function device_removed(_, device)
-  log.info("Device removed")
-  -- cancel timers
-  stop_check_for_updates_thread(device)
+  local device_dni = device.device_network_id
+  log.info(string.format("Device removed - dni=\"%s\"", device_dni))
 end
 
 local function goOffline(device)
-  stop_check_for_updates_thread(device)
   if device:get_field(const.STATUS) then
     device:emit_event(capabilities.switch.switch.off())
     device:emit_event(capabilities.mediaPlayback.playbackStatus.stopped())
@@ -210,15 +200,25 @@ local function check_for_updates(device)
   end
 end
 
-local function create_check_for_updates_thread(device)
-  -- stop old timer if one exists
-  stop_check_for_updates_thread(device)
+local function check_for_updates_manager(driver)
+  -- get all available devices
+  local devices = driver:get_devices()
 
-  log.info(string.format("create_check_for_updates_thread: dni=%s", device.device_network_id))
-  local new_timer = device.thread:call_on_schedule(const.UPDATE_INTERVAL, function()
-    check_for_updates(device)
-  end, "value_updates_timer")
-  device:set_field(const.UPDATE_TIMER, new_timer)
+  -- loop all devices and update them using polling
+  if next(devices) ~= nil then
+    for _, device in pairs(devices) do
+      local state = device:get_field(const.STATUS)
+      local device_dni = device.device_network_id
+      if state then
+        log.debug(string.format("polling values update for \"%s\"", device_dni))
+        check_for_updates(device)
+      else
+        log.debug(string.format("skipping polling values update for \"%s\" as it's offline", device_dni))
+      end
+    end
+  else
+    log.trace("No Harman Luxury devices to poll")
+  end
 end
 
 local function device_init(driver, device)
@@ -250,8 +250,6 @@ local function device_init(driver, device)
                        "NUMBER9"}))
 
   log.trace(string.format("device IP: %s", device_ip))
-
-  create_check_for_updates_thread(device)
 
   refresh(driver, device)
 end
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/harman-luxury/src/init.lua
```diff
@@ -210,15 +200,25 @@ local function check_for_updates(device)
   end
 end
 
-local function create_check_for_updates_thread(device)
-  -- stop old timer if one exists
-  stop_check_for_updates_thread(device)
+local function check_for_updates_manager(driver)
+  -- get all available drivers
```
cjswedes, line 204: nit:
```suggestion
  -- get all available devices
```

drivers/SmartThings/harman-luxury/src/init.lua
```diff
@@ -210,15 +200,25 @@ local function check_for_updates(device)
...
-  end, "value_updates_timer")
-  device:set_field(const.UPDATE_TIMER, new_timer)
+  -- loop all devices and update them using polling
+  if next(devices) ~= nil then
+    for _, device in pairs(devices) do
+      local state = device:get_field(const.STATUS)
+      local device_dni = device.device_network_id
+      if state then
+        log.debug(string.format("polling values update for \"%s\"", device_dni))
+        check_for_updates(device)
```
cjswedes, line 214: If one device is non-responsive, it will prevent updates of other devices from happening until the timeout has occurred. This is a problem with having a single "task" to handle all device updates, and IMO is a really good reason to have a single task per device to handle updates.

drivers/SmartThings/harman-luxury/src/init.lua
```diff
@@ -407,6 +405,13 @@ driver:call_on_schedule(const.HEALTH_CHEACK_INTERVAL, function()
   update_connection(driver)
 end, const.HEALTH_TIMER)
 
+-- create driver values update routine
+
+log.info("create_check_for_updates_manager for Harman Luxury devices")
+driver:call_on_schedule(const.UPDATE_INTERVAL, function()
+  check_for_updates_manager(driver)
+end, const.UPDATE_TIMER)
```
cjswedes, line 413: It seems like there is the potential for an explosion of memory/network activity if devices are not being responsive and there are multiple devices in the driver. 

Take the case where the first device in the drivers list of devices is non-responsive. Each call to `check_for_updates_manager` will take more than 5 seconds to complete, but in that time it will be called 5 times. This is not good. 

One way around this, is to spawn a cosock task that has an infinite loop to do the updates and make sure that a full iteration of devices is completed before starting another. However, I think it would be preferable to have a single "task" per device to ensure devices are updated at the update rate if possible, and when being non-responsive do not block other devices from updating.

cjswedes (commented):
The approach to have a single function called every second seems very problematic. 

I would prefer to see an approach where each device has its updates handled individually in its own cosock task. Those tasks are a loop that polls the device, and then waits 1 second before doing the next poll. The device can use a cosock channel to cancel its polling task when it is removed.

alon-tchelet:
I agree with your comments. What we previously used was closer to what you described, but it seems the hubs or a different component in the SmartThings arrangement was unable to enable multiple `driver:call_on_schedule` objects (one per device) and it would create a lot of issues.

We currently don't have the infrastructure to set up cosock communication, but ideally we would like to just update SmartThings regularly as values change triggered by the device. Samsung have recommended previously to use SSE (which we also don't currently support).
What is, in your opinion, the best way alternative for polling?

cjswedes:
> I agree with your comments. What we previously used was closer to what you described, but it seems the hubs or a different component in the SmartThings arrangement was unable to enable multiple driver:call_on_schedule objects (one per device) and it would create a lot of issues.

I'm a little curious about this. I know multiple call_on_schedule functions can be setup, this has been done in other drivers. Maybe the same issues I am reference wrt to many polls happening to the same device due to the function being triggered faster than it can finish its network request. Either way, using call_on_schedule per device isn't ideal anyway.

> We currently don't have the infrastructure to set up cosock communication, but ideally we would like to just update SmartThings regularly as values change triggered by the device

Cosock is just the library we use for executing our socket operations on coroutines in lua. I will provide an example of how this can be done to setup your polling loops.

cjswedes:
At the end of the device_init handler you can setup a cosock task to do the devices polling operation at the correct interval like this:
```lua
  local cancel_tx, rx = cosock.channel.new()
  device:set_field("polling_cancel", cancel_tx)
  cosock.spawn(function()
    rx:set_timeout(const.UPDATE_INTERVAL)
    while true do
      local cancelled, err = rx:receive() --Timeout is the 1 sec delay
      if cancelled or err ~= "timeout" or device == nil then -- defensive check for device being nil due to removal
        return
      end
      check_for_updates(device)
    end
  end, string.format("%s polling task", device.id))
```
And then when the device is removed you need to cancel that task by sending on the cancel channel, so in the device_removed handler you would add this:
```lua
  local cancel_tx = device:get_field("polling_cancel")
  if cancel_tx then
    cancel_tx:send("cancel") --anything non-nil should work
  end
```

alon-tchelet:
> I'm a little curious about this. I know multiple call_on_schedule functions can be setup, this has been done in other drivers. Maybe the same issues I am reference wrt to many polls happening to the same device due to the function being triggered faster than it can finish its network request. Either way, using call_on_schedule per device isn't ideal anyway.

Ah yes. That's possible. Perhaps I should reduce the timeout down to 1s. That's generally sufficient, but will certainly have dropped values in weaker connections.

Regarding your example, I will try to implement it and update the commits according.

alon-tchelet:
This PR is being closed as a proper fixup is done in PR #1433

PR #1362, opened 2024-05-05
zigbee-contact: Fix garage door handling for SmartSense Multi 

tpmanley:
The original SmartSense Multi has a different z-axis than the newer sensors.
When installed on the garage door it is parallel to the ground when the door
is closed and perpendicular to the ground when the door is open. For this
reason it can't share the garage door handling code that the newer sensors
use. This change implements the necessary custom handling and adds a unit
test.

https://smartthings.atlassian.net/browse/CHAD-13098

### DIFF
--- a/drivers/SmartThings/zigbee-contact/src/smartsense-multi/init.lua
+++ b/drivers/SmartThings/zigbee-contact/src/smartsense-multi/init.lua
@@ -146,7 +146,19 @@ local function xyz_handler(driver, device, zb_rx)
   local x = multi_utils.convert_to_signedInt16(zb_rx.body.zcl_body.body_bytes:byte(1), zb_rx.body.zcl_body.body_bytes:byte(2))
   local y = multi_utils.convert_to_signedInt16(zb_rx.body.zcl_body.body_bytes:byte(3), zb_rx.body.zcl_body.body_bytes:byte(4))
   local z = multi_utils.convert_to_signedInt16(zb_rx.body.zcl_body.body_bytes:byte(5), zb_rx.body.zcl_body.body_bytes:byte(6))
-  multi_utils.handle_three_axis_report(device, x, y, z)
+  device:emit_event(capabilities.threeAxis.threeAxis({value = {x, y, z}}))
+  if device.preferences["certifiedpreferences.garageSensor"] then
+    -- The sensor is mounted on the garage door vertically. Unlike the newer sensors, the z-axis is parallel to the ground
+    -- when the door is closed and perpendicular to the ground when the door is open. This is why we are using custom handling
+    -- instead of multi_utils.handle_three_axis_report. The values here were the same used in the original Groovy DTH
+    -- and in the protocol handler.
+    local abs_z = math.abs(z)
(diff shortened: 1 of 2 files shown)


tpmanley:
Once https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1361 is approved and merged I'll rebase this branch to pick that up.

drivers/SmartThings/zigbee-contact/src/multi-sensor/multi_utils.lua
```diff
@@ -138,7 +138,7 @@ multi_utils.convert_to_signedInt16 = function(byte1, byte2)
   local finalValue
   local swapped = (byte2 << 8) | byte1
   local sign_mask = 0x8000
-  local int16mask = 0xFF
+  local int16mask = 0xFFFF
```
greens, line 141: wouldn't this now be an int32 mask?

drivers/SmartThings/zigbee-contact/src/smartsense-multi/init.lua
```diff
@@ -146,7 +146,19 @@ local function xyz_handler(driver, device, zb_rx)
...
   local z = multi_utils.convert_to_signedInt16(zb_rx.body.zcl_body.body_bytes:byte(5), zb_rx.body.zcl_body.body_bytes:byte(6))
-  multi_utils.handle_three_axis_report(device, x, y, z)
+  device:emit_event(capabilities.threeAxis.threeAxis({value = {x, y, z}}))
+  if device.preferences["certifiedpreferences.garageSensor"] then
+    -- The sensor is mounted on the garage door vertically. Unlike the newer sensors, the z-axis is parallel to the ground
+    -- when the door is closed and perpendicular to the ground when the door is open. This is why we are using custom handling
+    -- instead of multi_utils.handle_three_axis_report. The values here were the same used in the original Groovy DTH
+    -- and in the protocol handler.
+    local abs_z = math.abs(z)
+    if abs_z > 825 then
```
greens, line 156: What if you just added an argument to the `multi_utils` function to use as the value here. It's a lot of code repetition to handle changing `900` to `825`.

tpmanley: In my opinion that would make the function harder to read than just copying the few lines we need right here. The change that is more important than the 900 -> 825 is treating the larger values as open instead of closed.

drivers/SmartThings/zigbee-contact/src/multi-sensor/multi_utils.lua
tpmanley: 0xFFFF is 16-bits. See https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1361 for more details about this change

drivers/SmartThings/zigbee-contact/src/smartsense-multi/init.lua
greens: ah i didn't notice the flip there

PR #1365, opened 2024-05-06
New Device (Matter Switch) Legrand Smart Lights Plug

lelandblue:
This is a WWST certification request for a New Device from Legrand, for a Smart Lights Smart Plug that uses the plug-binary profile, and according to the DCL has a Device Type ID of x10A

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -139,7 +139,12 @@ matterManufacturer:
     vendorId: 0x1339
     productId: 0x006F
     deviceProfileName: plug-binary
-
+#Legrand
+  - id: "4129/3"
+    deviceLabel: Smart Lights Smart Plug
+    vendorId: 0x1021
+    productId: 0x0003
+    deviceProfileName: plug-binary
 #Lifx
   - id: "5155/169"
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -139,7 +139,12 @@ matterManufacturer:
     vendorId: 0x1339
     productId: 0x006F
     deviceProfileName: plug-binary
-
+#Legrand
+  - id: "4129/3"
+    deviceLabel: Smart Lights Smart Plug
```
ctowns, line 144: Just confirming this isn't meant to be "Smart Light Plug" or something like that, right?

ctowns (approved):

drivers/SmartThings/matter-switch/fingerprints.yml
lelandblue: I verified   this is what the partner submitted, so I think this is intentional.

ctowns (approved):

lelandblue:
Hey @ctowns or @nickolas-deboom - Can you re-review this PR for me please, an additional commit was just added for another device from the same partner. Thank you!

ctowns (approved):

PR #1367, opened 2024-05-07
changed MA AVRs profile name to a more user friendly name

alon-tchelet:
Trying to change the Device Name as this is not user friendly, as can be seen below:
image

### DIFF
--- a/drivers/SmartThings/harman-luxury/profiles/jbl-ma-avr-series.yaml
+++ b/drivers/SmartThings/harman-luxury/profiles/jbl-ma-avr-series.yaml
@@ -1,4 +1,4 @@
-name: maX10
+name: jbl-ma-avr-series
 components:
 - id: main
   capabilities:
--- a/drivers/SmartThings/harman-luxury/src/devices.lua
+++ b/drivers/SmartThings/harman-luxury/src/devices.lua
@@ -29,22 +29,22 @@ local devices_model_info = {
     model = "L42ms",
   },
   [SetupID.MA510] = {
-    profile = "maX10",
+    profile = "jbl-ma-avr-series",
     manufacturer = "JBL",
     model = "MA510",
   },
   [SetupID.MA710] = {
-    profile = "maX10",
+    profile = "jbl-ma-avr-series",
     manufacturer = "JBL",
     model = "MA710",
   },
   [SetupID.MA7100HP] = {
-    profile = "maX10",
+    profile = "jbl-ma-avr-series",
     manufacturer = "JBL",
     model = "MA7100HP",
   },
   [SetupID.MA9100HP] = {
-    profile = "maX10",
+    profile = "jbl-ma-avr-series",
     manufacturer = "JBL",
     model = "MA9100HP",
   },


drivers/SmartThings/harman-luxury/profiles/jbl-ma-avr-series.yaml
```diff
@@ -1,4 +1,4 @@
-name: maX10
+name: jbl-ma-avr-series
```
varzac, line 1: I don't know that we allow for renaming profiles in production as this will be seen as deleting a profile and creating a new one and will orphan all existing devices.  @greens or @calebvarberg do you know if this is allowable?

alon-tchelet:
Well, the device is not in the market yet. So it won't affect any end-user.

drivers/SmartThings/harman-luxury/profiles/jbl-ma-avr-series.yaml
alon-tchelet: Well, the device is not in the market yet. So it won't affect any end-user.

varzac: But it has been published to production, which is where the restriction lies.  If this is not allowed by the platform it will require special privileges to change if that's even possible.  We'll see if this is possible given that the driver has been released.

alon-tchelet: Okay I understand. If the displayed name in the menu I shown in the picture above could be changed, this PR can be scraped.
Harman Luxury want to show under "Device Name" the device model (e.g. `JBL MA510`), but if it's impossible, as there are 4 devices using this profile, the series name will be sufficient (i.e. `JBL MA AVR Series`)

greens (commented):
what does the profile name have to do with the device name? if you want to change the device's join name you'd need to be working here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/harman-luxury/src/api/apis.lua#L63

alon-tchelet:
@greens 
I don't know what it has to do with. But the only place in our code where the term "maX10" appears in is where I changed it. The function you pointed returns the user friendly name set by the user on the JBL app or web interface. By default it will be in the format seen under the column "DEVICE LABEL" (<device-module>-<6-chars-of-mac>; e.g. "JBL MA9100HP-95...."). Hence, I believe "DEVICE NAME" is basing its value through the profile name. I don't know if it's done in the app code during/after the WWST certification process, but this is the only place this string is used.

greens:
@alon-tchelet Ah. I see. Yes, that column comes from the profile name. But that "name" is not visible to the user in many places. Mainly in the spot you've pointed out here.

The issue, as Zach pointed out, is that we cannot delete profiles once they've been created. Renaming it as you've done here will count as removing an existing profile and creating a new one, which will cause the deploy of the driver to fail.

lelandblue:
Hello @alon-tchelet The "Device Name" field that is appearing, in the screen shot you provided, is not the same name that the device will show as its default name when the device onboards to SmartThings. 

I am going to encourage someone from SmartThings to reach out regarding this PR, as there might be an alternative solution to resolving the issue you are concerned with.

alon-tchelet:
Closing this PR for now as Harman Luxury have decided not to make this change for now as it would require to apply the series to certification again

PR #1376, opened 2024-05-14
Matter Switch: add support for additional switch device types

hcarter-775:
The Dimmer Switch and Color Dimmer Switch both support the OnOff cluster as client, but some devices in the field support them as server. These changes accomodate that discrepancy. Changes for OnOff Switch device type were originally commited here:
cda431bf5496f002fab8d91d31a4df95c790abbc

https://smartthings.atlassian.net/browse/CHAD-13121

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -52,13 +52,19 @@ local COLOR_TEMP_LIGHT_DEVICE_TYPE_ID = 0x010C
 local EXTENDED_COLOR_LIGHT_DEVICE_TYPE_ID = 0x010D
 local ON_OFF_PLUG_DEVICE_TYPE_ID = 0x010A
 local DIMMABLE_PLUG_DEVICE_TYPE_ID = 0x010B
+local ON_OFF_SWITCH_ID = 0x0103
+local ON_OFF_DIMMER_SWITCH_ID = 0x0104
+local ON_OFF_COLOR_DIMMER_SWITCH_ID = 0x0105
 local device_type_profile_map = {
   [ON_OFF_LIGHT_DEVICE_TYPE_ID] = "light-binary",
   [DIMMABLE_LIGHT_DEVICE_TYPE_ID] = "light-level",
   [COLOR_TEMP_LIGHT_DEVICE_TYPE_ID] = "light-level-colorTemperature",
   [EXTENDED_COLOR_LIGHT_DEVICE_TYPE_ID] = "light-color-level",
   [ON_OFF_PLUG_DEVICE_TYPE_ID] = "plug-binary",
-  [DIMMABLE_PLUG_DEVICE_TYPE_ID] = "plug-level"
+  [DIMMABLE_PLUG_DEVICE_TYPE_ID] = "plug-level",
+  [ON_OFF_SWITCH_ID] = "switch-binary",
+  [ON_OFF_DIMMER_SWITCH_ID] = "switch-level",
+  [ON_OFF_COLOR_DIMMER_SWITCH_ID] = "switch-color-level",
 }
 local detect_matter_thing
 
@@ -124,7 +130,6 @@ end
 local function initialize_switch(driver, device)
   local switch_eps = device:get_endpoints(clusters.OnOff.ID)
   table.sort(switch_eps)
-
   -- Since we do not support bindings at the moment, we only want to count On/Off
   -- clusters that have been implemented as server. This can be removed when we have
   -- support for bindings.
@@ -159,14 +164,27 @@ local function initialize_switch(driver, device)
 
   device:set_field(SWITCH_INITIALIZED, true)
   -- The case where num_server_eps > 0 is a workaround for devices that have the On/Off
-  -- Light Switch device type but implement the On Off cluster as server (which is against the spec
+  -- Light Switch device type but implement the On Off cluster as sServer (which is against the spec
   -- for this device type). By default, we do not support On/Off Light Switch because by spec these
   -- devices need bindings to work correctly (On/Off cluster is client in this case), so this device type
   -- does not have a generic fingerprint and will join as a matter-thing. However, we have
   -- seen some devices claim to be On/Off Light Switch device type and still implement On/Off server, so this
   -- is a workaround for those devices.
   if num_server_eps > 0 and detect_matter_thing(device) == true then
-    device:try_update_metadata({profile = "switch-binary"})
+    local id = 0
+    for _,ep in ipairs(device.endpoints) do
+      -- main_endpoint only supports server cluster by definition of get_endpoints()
+      if main_endpoint == ep.endpoint_id then
+        for _, dt in ipairs(ep.device_types) do
+          id = math.max(id, dt.device_type_id)
+        end
+        break
+      end
+    end
+
+    if device_type_profile_map[id] ~= nil then
+      device:try_update_metadata({profile = device_type_profile_map[id]})
+    end
   end
 end
 
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -172,7 +172,7 @@ local function initialize_switch(driver, device)
   -- workaround for those devices.
   if num_server_eps > 0 and detect_matter_thing(device) == true then
     local id = 0
-    for _,ep in ipairs(device.endpoints) do
+    for _, ep in ipairs(device.endpoints) do
       -- main_endpoint only supports server cluster by definition of get_endpoints()
       if main_endpoint == ep.endpoint_id then
         for _, dt in ipairs(ep.device_types) do
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -177,7 +177,7 @@ local function initialize_switch(driver, device)
       if main_endpoint == ep.endpoint_id then
         for _, dt in ipairs(ep.device_types) do
           -- no device type that is not in the switch subset should be considered.
-          if (dt.device_type_id <= ON_OFF_COLOR_DIMMER_SWITCH_ID) then
+          if (ON_OFF_SWITCH_ID <= dt.device_type_id and dt.device_type_id <= ON_OFF_COLOR_DIMMER_SWITCH_ID) then
             id = math.max(id, dt.device_type_id)
           end
         end
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -163,20 +163,23 @@ local function initialize_switch(driver, device)
   end
 
   device:set_field(SWITCH_INITIALIZED, true)
-  -- The case where num_server_eps > 0 is a workaround for devices that have the On/Off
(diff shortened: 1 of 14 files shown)


ctowns (commented):
Looks great! Let's test this with the VDA and then we should be good to go

hcarter-775:
Completed testing using the following VDA devices on my personal v3:
1. Virtual OnOff Switch 
2. Virtual Dimmer Switch
3. Virtual Color Dimmer Switch
4. Virtual Bridge with multiple instances of the devices listed above, as well as instances of various Light devices. 
All functionality was preserved for previously working devices, as well as for new functionality for Dimmer and Color Dimmer Switch in the cases of a Bridge and on their own.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -158,15 +163,28 @@ local function initialize_switch(driver, device)
...
+  -- claim to be Light Switch device types and still implement their clusters as server, so this is a
+  -- workaround for those devices.
   if num_server_eps > 0 and detect_matter_thing(device) == true then
-    device:try_update_metadata({profile = "switch-binary"})
+    local id = 0
+    for _,ep in ipairs(device.endpoints) do
+      -- main_endpoint only supports server cluster by definition of get_endpoints()
+      if main_endpoint == ep.endpoint_id then
+        for _, dt in ipairs(ep.device_types) do
+          id = math.max(id, dt.device_type_id)
```
gharveymn, line 179: Will this break if the device also has some other device type with an ID higher than On/Off Light Switch and friends?

hcarter-775: No, because below that, we have this check: 
` if device_type_profile_map[id] ~= nil then
     device:try_update_metadata({profile = device_type_profile_map[id]})
 end` 
If there were a higher ID, it will not have a mapping in the device_type_profile_map array, and so nothing will be be done in that case.

Maybe a silly question, but how did you clip that code snippet in your comment?

ctowns: So this code should only run if `detect_matter_thing` is true, which means a device joined that did not match to any generic fingerprint for a device type that we support. We do not have generic fingerprints for the On/Off Light Switch and friends because we technically do not support them because these are meant to work with bindings. However, we can support these device types if they do not follow the spec and implement their application clusters as server.

If some other device type we are not accounting for comes along, then the check on line 185 will ensure we do not switch to it because it does not have a matching profile in our profile map.

For context, the reason why we check for the largest ID is because endpoints can report that they support superset/subset device types on the same endpoint: https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_SupersetDeviceTypes

Therefore, an endpoint could support On/Off Light Switch, Dimmer Switch, and Color Dimmer Switch and report all of these as device types since Color Dimmer is a super set of the others. However, it could not also have a OnOff Plug device type on the same endpoint because that is not a subset/superset of the other device types on the endpoint.

ctowns: > Maybe a silly question, but how did you clip that code snippet in your comment?

When you leave a comment on a specific line on a review, it will automatically clip a few lines around where you left your comment so you can see the comment in context in this view.

To leave a comment on specific line, you just need to click the blue + sign next to line number you want to leave a comment on in the "Files changed" view. Here's a little bit more context from the GitHub docs: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#starting-a-review

gharveymn: I get what you're trying to do with the max wrt the superset device type, but 

> However, it could not also have a OnOff Plug device type on the same endpoint because that is not a subset/superset of the other device types on the endpoint.

I think I missing where the spec is saying this. And, to elucidate, this is the device type list I'm wondering about:

```
0x0103 On/Off Light Switch
0x0104 Dimmer Switch
0x0105 Color Dimmer Switch
0x010A On/Off Plug-in Unit
```

Then, this code will detect `0x010A` and stuff will be broken, or am I missing something? Personally, if the spec isn't absolutely clear that this isn't allowed, I think we will have fewer potential problems in the future if we only do the max of the ones we are targeting.

ctowns: Here is the spot in the spec that defines what I am talking about for endpoints having one application device type, or a superset of device types: https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_EndpointComposition

from the spec:
Each [simple endpoint](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_SimpleEndpoint) SHALL support only one [Application device type](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_ApplicationEndpoint) with these exceptions:

The endpoint MAY support additional device types which are subsets of the [Application device](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_ApplicationEndpoint) type (the superset). See [Superset Device Types](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_SupersetDeviceTypes) for cluster requirements of superset devices.

The endpoint MAY support additional device types ([application](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_ApplicationEndpoint), [utility](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_UtilityEndpoint) or [node ](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_NodeEndpoint)device types) as defined by each additional device type.

So I don't think the collection of device types you have mentioned above would be possible because the OnOff plug in unit is not a subset of the Color Dimmer Switch (which is the superset of on/off light switch and dimmer switch).

However, I agree that this is not immediately obvious, and for the sake of readability, we could limit this section to just check for the switch device types since that is specifically what we are looking for. Does this explanation make sense though?

gharveymn: Thank you! Wtf does this even mean though:

> The endpoint MAY support additional device types ([application](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_ApplicationEndpoint), [utility](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_UtilityEndpoint) or [node ](https://librarian-regionals.smartthingsgdev.com/matter/index.html#ref_NodeEndpoint)device types) as defined by each additional device type.

Yeah, the explanation makes sense, but yeah let's just make it explicit so we have less to worry about.

ctowns: Yeah, this is not terribly clear... I believe this is referring to cases where you might have a bridged device that has a bridged node device type and an application cluster, or a case where you have an application cluster and a utility cluster (power source for example). As of right now, I do not see any device types that include as a part of their device types requirements defining another _application_ cluster on the _same endpoint_. I believe this line is just here in case that situation should ever arise, but right now in cases where any additional device types dependencies are included in the spec for a device type, those additional device types are included on separate endpoints.

Here is a somewhat relevant thread from the CSA channel that helps provide a little more context. https://csamembers.slack.com/archives/C01B01GH64E/p1684356219272049?thread_ts=1684272999.049459&cid=C01B01GH64E

However, I still agree that making this explicit seems less error prone, so I think we should make the change that Gene proposed @hcarter-775

Thanks for the great question @gharveymn !

hcarter-775: @ctowns @gharveymn I have committed a change in line 180 with a corresponding comment as to this discussion. Tests continue to pass. I have not retried VDA devices with this driver change; do you think that is necessary in this case?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -158,15 +163,31 @@ local function initialize_switch(driver, device)
...
+  -- workaround for those devices.
   if num_server_eps > 0 and detect_matter_thing(device) == true then
-    device:try_update_metadata({profile = "switch-binary"})
+    local id = 0
+    for _,ep in ipairs(device.endpoints) do
+      -- main_endpoint only supports server cluster by definition of get_endpoints()
+      if main_endpoint == ep.endpoint_id then
+        for _, dt in ipairs(ep.device_types) do
+          -- no device type that is not in the switch subset should be considered.
+          if (dt.device_type_id <= ON_OFF_COLOR_DIMMER_SWITCH_ID) then
```
ctowns, line 180: This check will still include these device types because they have a lower value than `ON_OFF_COLOR_DIMMER_SWITCH_ID = 0x0105`

```
local ON_OFF_LIGHT_DEVICE_TYPE_ID = 0x0100
local DIMMABLE_LIGHT_DEVICE_TYPE_ID = 0x0101
```

So you'll probably want to add a lower bounds on this as well.

drivers/SmartThings/matter-switch/src/init.lua
ctowns: In general, I lean on the side of caution and try to test as much as I can. You are right that this is a small change that shouldn't affect functionality given our test cases, but I'd still recommend using the VDA to join a few devices (you can join all these devices in one test by adding them as bridged devices on the virtual bridge in the VDA) and just doing a smoke test to make sure we didn't overlook anything.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: This has been added in the most recent push

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: I added a VDA bridge device (on/off switch, on/off dimmer switch, on/off color dimmer switch) last night using the edited drivers- everything worked as expected.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -158,15 +163,31 @@ local function initialize_switch(driver, device)
...
-  -- is a workaround for those devices.
+  -- for this device type). By default, we do not support Light Switch device types because by spec these
+  -- devices need bindings to work correctly (On/Off cluster is client in this case), so these device types
+  -- do not have a generic fingerprint and will join as a matter-thing. However, we have seen some devices
+  -- claim to be Light Switch device types and still implement their clusters as server, so this is a
+  -- workaround for those devices.
   if num_server_eps > 0 and detect_matter_thing(device) == true then
-    device:try_update_metadata({profile = "switch-binary"})
+    local id = 0
+    for _,ep in ipairs(device.endpoints) do
```
gharveymn, line 175: super nit:

```suggestion
    for _, ep in ipairs(device.endpoints) do
```

gharveymn (approved):

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: made that change :+1:

ctowns (approved):
Look good to me, great work @hcarter-775 ! 🎉 

One thing we may want to do is squash some of these commits together before merging. We can essentially squash all these changes into one single commit that covers all the changes. Using separate commits as you have done here is helpful for review, but now that we are ready to merge, it can be useful from a readability and maintenence perspective to squash these commits into one commit since most of the extra commits are small fixes rather than new additions to the original commit.

harireddy2002:
Hi @hcarter-775 .
Actually I need help regarding PR against smartthings drivers.
It will be helpful if we connect together. Ping me through main id: hariprasad0322@gmail.com

PR #1382, opened 2024-05-17
Add clusters for Matter 1.2 device types

ctowns:
Adds new Matter 1.2 clusters for these types directly to the driver. Cluster definitions have been stripped down to just the essentials that are used in the drivers.


drivers/SmartThings/matter-sensor/src/AirQuality/server/attributes/AcceptedCommandList.lua
```diff
@@ -0,0 +1,75 @@
+local cluster_base = require "st.matter.cluster_base"
```
tpmanley, line 1: Do we need this file? Air Quality doesn't have any commands?

drivers/SmartThings/matter-sensor/src/AirQuality/server/attributes/EventList.lua
```diff
@@ -0,0 +1,75 @@
+local cluster_base = require "st.matter.cluster_base"
```
tpmanley, line 1: Do we need this file? Air Quality doesn't have any events

ctowns: Removed in 2df4f7a8fa7e0847b68a82f2e06417ac63f957cf!

drivers/SmartThings/matter-sensor/src/AirQuality/server/attributes/AcceptedCommandList.lua
ctowns: Removed in 2df4f7a8fa7e0847b68a82f2e06417ac63f957cf!

ctowns:
Added remaining clusters for Matter 1.2 device types. So far I have tested the matter-appliance and matter-sensor drivers and they both seem to be working on a hub without Matter 1.2 clusters (i.e with old lua libs). I'll test the other drivers and report the results here as well.

ctowns:
Updated PR that was merged here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1390

PR #1390, opened 2024-05-23
Add Matter 1.2 clusters directly to drivers

ctowns:
Adds stripped down versions of clusters directly to drivers so that we do not need to depend on lua libs releases.
    
Also adds some test files for drivers where they were missing.

Each device type was tested with a hub running 52 lua libs (without Matter 1.2 clusters) while running these drivers. Each attribute was then tested by checking each value for the attribute and confirming that the hub can properly receive updates to that value and also set that value correctly where applicable. This testing is summarized here: https://smartthings.atlassian.net/wiki/spaces/~62de3b63b6b0b70770d7905d/pages/3238854850/Matter+1.2+VDA+Testing+with+Driver+Clusters


ctowns:
This still needs some clean up, but with these changes in place I am able to successfully test with the Matter 1.2 drivers on a hub with 52 lua libs (i.e. without the Matter 1.2 cluster definitions).

I will plan to do some more clean up, but this should be pretty close. With these changes in place, we shouldn't need to depend on a lua libs update to support Matter 1.2 device types on all hubs.

Overall, the driver memory footprint has some modest savings in most cases when using the built-in cluster definitions in the driver. However, there are large savings for Air Quality Sensor due to compressing the 12 optional concentration measurement clusters down and having them utilize one base ConcentrationMeasurement cluster definition. All other drivers had smaller memory decreases, but still noticeable. In all cases, the drivers are not near the driver disk space limit (different from memory limit) thanks to comment stripping and cutting out unnecessary bloat from the cluster definitions. There is probably opportunity for more trimming as well.

Here are some memory stats for each of the 4 drivers with a specific device type:
**Matter Sensor - Air Quality Sensor**

```
+-------------+----------------------------------+-------------------------------+
|    Value    | Using 53 with lua libs clusters  | Using 52 with driver clusters |
+-------------+----------------------------------+-------------------------------+
| Proto count |                             1724 |                          1286 |
| Proto size  |                          668004  |                        512640 |
| Heap        |                          3194880 |                       2744320 |
+-------------+----------------------------------+-------------------------------+

```

**Matter Thermostat - Room AC**
```
+-------------+---------------------------------+-------------------------------+
|    Value    | Using 53 with lua libs clusters | Using 52 with driver clusters |
+-------------+---------------------------------+-------------------------------+
| Proto count |                            1209 |                          1123 |
| Proto size  |                          519824 |                        489916 |
| Heap        |                         2187264 |                       2953216 |
+-------------+---------------------------------+-------------------------------+

```

**Matter RVC - Robot Vacuum Cleaner**
```

+-------------+---------------------------------+-------------------------------+
|    Value    | Using 53 with lua libs clusters | Using 52 with driver clusters |
+-------------+---------------------------------+-------------------------------+
| Proto count |                             882 |                           857 |
| Proto size  |                          417440 |                        409276 |
| Heap        |                         2523136 |                       1499136 |
+-------------+---------------------------------+-------------------------------+

```

**Matter Appliance - Dishwasher**
```

+-------------+---------------------------------+-------------------------------+
|    Value    | Using 53 with lua libs clusters | Using 52 with driver clusters |
+-------------+---------------------------------+-------------------------------+
| Proto count |                            1253 |                          1232 |
| Proto size  |                          557368 |                        554812 |
| Heap        |                         5353472 |                       2400256 |
+-------------+---------------------------------+-------------------------------+

```

drivers/SmartThings/matter-appliance/profiles/refrigerator-freezer-tn-tl.yml
```diff
@@ -2,37 +2,37 @@ name: refrigerator-freezer-tn-tl
 components:
 - id: main
   capabilities:
-  - id: mode # Refrigerator And Temperature Controlled Cabinet Mode
```
ctowns, line 5: Having comments in the profile definition causes issues when parsing this file for capability names. This results in subscription issues because we cannot properly parse the capability name when determining which subscripted attributes we will subscribe to.

cc: @yhoyoon @HunsupJung

HunsupJung: Yes, It was for implementation. It must be removed.

lelandblue:
Ticket Reference for this is :  CHAD-13240

ctowns:
Added a link to a confluence page summarizing the testing for this change here: https://smartthings.atlassian.net/wiki/spaces/~62de3b63b6b0b70770d7905d/pages/3238854850/Matter+1.2+VDA+Testing+with+Driver+Clusters

This is also linked in the top comment on this PR.

Ticket: https://smartthings.atlassian.net/browse/CHAD-13240

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
```diff
@@ -19,6 +19,11 @@ local clusters = require "st.matter.clusters"
 local log = require "log"
 local utils = require "st.utils"
 
+local version = require "version"
+clusters.RefrigeratorAlarm = require "RefrigeratorAlarm"
```
tpmanley, line 23: Needs `version.api` check

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
```diff
@@ -161,6 +166,7 @@ end
 local function refrigerator_tcc_supported_modes_attr_handler(driver, device, ib, response)
   local refrigeratorTccModeSupportedModes = {}
   for _, mode in ipairs(ib.data.elements) do
+    clusters.RefrigeratorAndTemperatureControlledCabinetMode.types.ModeOptionStruct:augment_type(mode)
```
tpmanley, line 169: I don't yet understand why, but the other files have a version check around similar code

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -15,9 +15,25 @@
 local MatterDriver = require "st.matter.driver"
 local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
+local utils = require "st.utils"
 
 local log = require "log"
 
+local error_elements = {
```
tpmanley, line 22: appears to be unused

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -174,12 +181,12 @@ local function get_device_type(driver, device)
 end
 
 local function do_configure(driver, device)
-  local heat_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.Feature.HEATING})
-  local cool_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.Feature.COOLING})
-  local auto_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.Feature.AUTOMODE})
+  local heat_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.HEATING})
```
tpmanley, line 184: This will work with both old and new (10) versions of the Lua Libs?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -20,16 +20,23 @@ local im = require "st.matter.interaction_model"
...
   [clusters.Thermostat.types.ThermostatSystemMode.AUTO]           = capabilities.thermostatMode.thermostatMode.auto,
   [clusters.Thermostat.types.ThermostatSystemMode.COOL]           = capabilities.thermostatMode.thermostatMode.cool,
   [clusters.Thermostat.types.ThermostatSystemMode.HEAT]           = capabilities.thermostatMode.thermostatMode.heat,
-  [clusters.Thermostat.types.ThermostatSystemMode.EMERGENCY_HEAT] = capabilities.thermostatMode.thermostatMode.emergencyheat,
+  [clusters.Thermostat.types.ThermostatSystemMode.EMERGENCY_HEATING] = capabilities.thermostatMode.thermostatMode.emergencyheat,
   [clusters.Thermostat.types.ThermostatSystemMode.PRECOOLING]     = capabilities.thermostatMode.thermostatMode.precooling,
   [clusters.Thermostat.types.ThermostatSystemMode.FAN_ONLY]       = capabilities.thermostatMode.thermostatMode.fanonly,
-  [clusters.Thermostat.types.ThermostatSystemMode.DRY]            = capabilities.thermostatMode.thermostatMode.dryair,
-  [clusters.Thermostat.types.ThermostatSystemMode.SLEEP]          = capabilities.thermostatMode.thermostatMode.asleep
+  [8]            = capabilities.thermostatMode.thermostatMode.dryair,
```
tpmanley, line 38: What's the story behind this change? A comment explaining why hardcoded IDs would be helpful

drivers/SmartThings/matter-appliance/src/matter-dishwasher/init.lua
```diff
@@ -95,6 +102,9 @@ end
 local function dishwasher_supported_modes_attr_handler(driver, device, ib, response)
   dishwasherModeSupportedModes = {}
   for _, mode in ipairs(ib.data.elements) do
+    if version.api < 10 then
```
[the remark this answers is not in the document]
ctowns: This is only done for older API versions because on hubs where they already have the 53 lua libs with Matter 1.2 support, then when this struct is deserialized, it will be augmented to use the appropriate type with the correct field defs when it is deserialized. However, without the 53 lua libs, it is just deserialized as a base struct data type and is not augemented to the SupportedModes type. It has all the same data in the same structure, but the struct is just missing the field defs, so augmenting here allows us to access the values of the struct using the field names rather than index directly to the value with a hardcoded number.

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: Yes, the new lua libs feature aliases that will check for the this old name for the feature ('ThermostatFeature') and map it to the new name as of the 1.2 spec (just `Feature`)

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: Added a comment and made it so these were not hardcoded

drivers/SmartThings/matter-rvc/src/init.lua
ctowns: Removed

ctowns:
I added the embedded clusters for matter-pump since they were missing in my first pass of this.

Also, I added additional handing for the get_endpoints function for cases when we are running older lua libs and must rely on the embedded clusters for things like feature maps. I added the `embedded-clusters-utils` file to have this workaround for old hubs to allow the `get_endpoints` functionality to still work

HunsupJung (approved):

drivers/SmartThings/matter-pump/src/embedded-cluster-utils.lua
```diff
@@ -0,0 +1,52 @@
...
+      if not feature_bitmap or not clus then return false end
+      return embedded_cluster.are_features_supported(feature_bitmap, clus.feature_map)
+    end
+    local eps = {}
+    for _, ep in ipairs(device.endpoints) do
+      for _, clus in ipairs(ep.clusters) do
+        if ((clus.cluster_id == cluster_id)
+              and (opts.feature_bitmap == nil or clus_has_features(clus, opts.feature_bitmap))
+              and ((opts.cluster_type == nil and clus.cluster_type == "SERVER" or clus.cluster_type == "BOTH")
+                or (opts.cluster_type == clus.cluster_type))
```
tpmanley, line 39: Could be simplified to:

`and (clus.cluster_type == opts.cluster_type or clus.cluster_type == "SERVER" or clus.cluster_type == "BOTH")`

drivers/SmartThings/matter-switch/profiles/small-profile.yml
```diff
@@ -0,0 +1,10 @@
+name: small-profile
```
tpmanley, line 1: Was this added on purpose?

tpmanley (approved):

drivers/SmartThings/matter-pump/src/embedded-cluster-utils.lua
ctowns: I think this is set up this way in case the goal is to have `opts.cluster_type == "CLIENT"` as an option, but to default to just returning server cluster endpoint if no option is provided. I don't expect we would use the client option, but I will keep it like this to maintain parity with the scripting-engine definition of this function.

drivers/SmartThings/matter-switch/profiles/small-profile.yml
ctowns: This was an accident, removed!

PR #1411, opened 2024-05-31
New Devices (AiDot Matter Color Temperature Light) for Certification

chenpeixiang:
This PR is to add new devices with new profile to WWST Certification.


drivers/SmartThings/matter-switch/profiles/light-level-colorTemperature-2700K-6500K.yml
```diff
@@ -0,0 +1,24 @@
...
+      values:
+        - key: "level.value"
+          range: [1, 100]
+  - id: colorTemperature
+    version: 1
+    config:
+      values:
+        - key: "colorTemperature.value"
+          range: [ 2700, 6500 ]
+  - id: firmwareUpdate
```
lelandblue, line 19: Should there also be a capability for Color Control?

ctowns: If the device only supports color temperature, it does not require the color control capability (which is for cases where you can change the RGB color values rather than just color temp).

drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -218,7 +218,7 @@ matterManufacturer:
     vendorId: 0x115A
     productId: 0x711
     deviceProfileName: light-color-level-2700K-6500K
- - id: "4442/72"
+  - id: "4442/72"
```
ctowns, line 221: nit: remove extra space

ctowns, line 221:
```suggestion
 - id: "4442/72"
```
nit: remove the extra space that was added here

ctowns (approved):

PR #1413, opened 2024-06-03
Add a matter-lock driver for Aqara

HunsupJung:
As a result of testing the U200 actual device, power source cluster is not supported. this patch removes battery capability.
And in the case of Matter door lock, lockAlarm capability will be used instead of tamperAlert capability, so that part has been applied.


drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -21,8 +21,11 @@ local PowerSource = clusters.PowerSource
 local capabilities = require "st.capabilities"
 local im = require "st.matter.interaction_model"
 local lock_utils = require "lock_utils"
+local device_lib = require "st.device"
 
 local INITIAL_COTA_INDEX = 1
+local AQARA_MANUFACTURER_ID = 0x115f
```
ctowns, line 27: @HunsupJung can we just add a fingerprint for this device instead? We could create a fingerprint that matches these devices to the `lock-without-codes` profile and potentially avoid any driver changes.

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -505,7 +508,27 @@ local function device_added(driver, device)
   --Note: May want to write OperatingMode to NORMAL, to attempt to ensure remote operation works
   --Note: May want to write RequirePINForRemoteOperation, to avoid cota cases if possible.
   device:emit_event(capabilities.tamperAlert.tamper.clear())
+end
+
+local function is_aqara_products(device)
```
ctowns, line 513: If we must make driver changes, then I would suggest that we move this functionality to a sub driver. In the past, we have kept manufacturer-specific functionality in sub drivers. For an example of this in matter drivers, you can see the eve-energy sub driver: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/matter-switch/src/eve-energy/init.lua

The sub driver should be relatively simple: you can use this same `is_aqara_products` function as the `can_handle` function in the sub driver, and then you can just copy the `do_configure` function you have here into the sub driver with your changes. This way, the sub driver can override the main drivers `do _Configure` function in the case that the device is an aqara lock. The rest of the functions will work normally since the sub driver will only override the `do_configure` function.

In this case, the sub driver will be very simple and will only contain the `can_handle` function and the updated `do_configure` function

ctowns:
Also it looks like many unit tests are failing because of the new `do_configure` function. You will need to update these unit tests to account for the change to separate the `device_added` and `do_configure` function.

drivers/SmartThings/matter-lock/src/init.lua
ctowns: Could you please add the corresponding fingerprint to the `fingerprints` file since it will be required for WWST anyway?

HunsupJung: I'm sorry for the late review. I will add fingerprint.

drivers/SmartThings/matter-lock/src/init.lua
HunsupJung: I was thinking that you might suggest a sub driver. I'll modify it right away.

HunsupJung:
@ctowns 
Thank you for your review. 
If DoorLock supports COTA, base-lock with Lock Codes is used, and U200 supports COTA, so I was worried.
As you said, I created a sub driver for Aqara product and updated it to include appropriate capabilities.

ctowns:
@HunsupJung I think we can move all of this functionality to the base driver without needing a sub driver, so I have made that change in the latest commit.

However, on testing this, I see that there is no capability presentation for the lockAlarm capability even though the device is using a profile with the lockAlarm capability. I am using the latest R2 app and I am in developer mode. Is this expected?

Previously, the tamperAlert capability would be visible on the device card and would indicate that there was an alarm on the lock. However, the lockAlarm does not seem to have a presentation. Is plugin work for this capability complete?

HunsupJung:
@ctowns 
Thank you so much for your review. 👍
You're right. lockAlarm has no presentation. Instead, we will provide notification to users through it.

drivers/SmartThings/matter-lock/fingerprints.yml
```diff
@@ -10,9 +10,15 @@ matterManufacturer:
     vendorId: 0x135d
     productId: 0xb0
     deviceProfileName: lock-nocodes-notamper
+#Aqara
+  - id: "4447/10242"
+    deviceLabel: Aqara Smart Lock U200
+    vendorId: 0x115F
+    productId: 0x2802
+    deviceProfileName: lock-nocodes-nobattery-lockAlarm
```
tpmanley, line 18: @HunsupJung Is the intention to later add the new lock code capability to this profile so that existing devices automatically get lock code support? If so I'd suggest not naming the profile "nocodes"

drivers/SmartThings/matter-lock/fingerprints.yml
```diff
@@ -10,9 +10,15 @@ matterManufacturer:
...
 matterGeneric:
   - id: "matter/door-lock"
     deviceLabel: Matter Door Lock
     deviceTypes:
       - id: 0x000A # Door Lock
-    deviceProfileName: base-lock
+    deviceProfileName: base-lock #TODO update if lockAlarm is visible
```
tpmanley, line 24: I think we want to match the new Aqara profile whatever we do end up using there, except probably with battery since most locks are going to be battery powered

drivers/SmartThings/matter-lock/fingerprints.yml
HunsupJung: @tpmanley , @ctowns 
If the same profile is used afterwards, it seems correct to remove nocodes from the name. How about lock-lockalarm-nobattery that simply removes nocodes from the name?

drivers/SmartThings/matter-lock/fingerprints.yml
HunsupJung: Then can we apply the lock-lockalarm profile here?

drivers/SmartThings/matter-lock/fingerprints.yml
ctowns: [d0a77c9](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1413/commits/d0a77c936e3ea9e7dfbf99220f3f0fa63051754b) Updated this to be `lock-lockalarm-nobattery` but NOT include lock codes in the profile capabilities so that we can add the new lock codes capability at a later date but still use the same profile name.

drivers/SmartThings/matter-lock/fingerprints.yml
ctowns: [d0a77c9](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1413/commits/d0a77c936e3ea9e7dfbf99220f3f0fa63051754b) updated this to use the `lock-lockalarm` capability, which is the same as the `lock-lockalarm-nobattery` used the the aqara lock but it has battery support

tpmanley (approved):

HunsupJung:
@tpmanley , @ctowns 
Thank you so much for your review~! 👍

PR #1418, opened 2024-06-04
[Aqara/K100] add OTP and notification functions

seojune79:


Kwang-Hui:
Hi team. 
Please refer `STSE-2969` for the details.

drivers/Aqara/aqara-lock/profiles/aqara-lock-battery.yml
```diff
@@ -6,6 +6,8 @@ components:
     version: 1
   - id: battery
     version: 1
+  - id: lockAlarm
+    version: 1
   - id: tamperAlert
     version: 1
```
greens, line 12: It seems like this capability is no longer being used.

greens (commented):
looks like there might be an issue with the lock alarm capability. I'm investigating

drivers/Aqara/aqara-lock/profiles/aqara-lock-battery.yml
Kwang-Hui: Actually `tamperAlert` wasn't visible plugin and automation - this is hidden by device-config due to dev request.

Isn't there any issue when removing `tamperAlert` from profile for the existing users?

greens: If it's already not visible then there is even less of a concern.

Kwang-Hui: @seojune79 Could you remove `tamperAlert` from file with relavant code?

seojune79: I updated it.

greens:
Because the LockAlarm capability is so new, it's not included in the generated code in the lua libraries release artifact. It will work on the hub itself, because that talks to the cloud.

However, for the sake of getting the tests to pass, I'd recommend including a copy of the capability with this PR.

greens:
Based on local testing, some of the existing tests for the lock will also need to be updated to reflect the new functionality.

Kwang-Hui:
@greens , Do we need capability schema yaml file just like `antiLockStatus.yaml` on the same location? and do we need additional code work based on it?

drivers/Aqara/aqara-lock/config.yml
```diff
@@ -1,7 +1,7 @@
 name: 'Aqara Lock'
 packageKey: 'aqara-lock'
 permissions:
-  aqara_provisioning: {}
+  # aqara_provisioning: {}
```
Kwang-Hui, line 4: This need to be uncommented.

seojune79: I did it.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -67,41 +69,60 @@ local function toHex(value, length)
 end
 
 local function event_lock_handler(driver, device, evt_name, evt_value)
-  if evt_value == 1 then
+  if evt_value == 0x1 then
     device:emit_event(Lock.lock(evt_name))
-    device:emit_event(TamperAlert.tamper.clear())
+    device:emit_event(LockAlarm.alarm.clear({ visibility = { displayed = false }}))
```
greens, line 74: you probably want to leave the visibility here, since this clear is coming from an actual report from the lock, we'll want it to be visible in history.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -67,41 +69,60 @@ local function toHex(value, length)
...
+  if evt_value == 0x80020000 then -- one-time password
+    id = "OTP_STANDALONE"
+    label = nil
+  else
   id, label = credential_utils.find_userLabel(driver, device, evt_value)
+  end
   device:emit_event(Lock.lock.unlocked({ data = { method = evt_name, codeId = id, codeName = label } }))
   device:emit_event(remoteControlStatus.remoteControlEnabled('false', { visibility = { displayed = false } }))
-  device:emit_event(TamperAlert.tamper.clear())
+  device:emit_event(LockAlarm.alarm.clear({ visibility = { displayed = false }}))
```
greens, line 89: same here

drivers/Aqara/aqara-lock/src/init.lua
Kwang-Hui: the "clear" comes from door lock has locked normally.
but to make `lockAlarm` as an event - not as a state -, I asked @seojune79 to hide clear event.
lockAlarm doesn't shown up at detail view and automation. but will send push notification when issued (except `clear` event).
If you strongly suggest `clear` should be shown at history. please let me know

drivers/Aqara/aqara-lock/src/init.lua
lelandblue: @seojune79 Please review when your schedule allows.

drivers/Aqara/aqara-lock/src/init.lua
greens: So how is `lockAlarm` displayed to the user?

Ultimately, it's your design decision, but I'd just expect users to find it useful to see when an alarm was cleared in history.

Kwang-Hui: I'll leave the details on `STSE-2969` ticket

drivers/Aqara/aqara-lock/src/init.lua
Kwang-Hui: @greens @lelandblue , please visit `STSE-2969` for details.

greens (approved):

PR #1423, opened 2024-06-07
Matter-Switch: Integrate matter-button driver

nickolas-deboom:
This PR was split into three separate PRs in order to simplify reviewing and also to allow more discussion and testing before supporting combination button/switch device types.

The newly created PRs are:

- [PR 1547](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1547): Merge the switch and button drivers
- [PR 1548](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1548): Support subscribing to child device attributes
- [PR 1549](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1549): Support combination button/switch device types


nickolas-deboom:
These changes were prompted by new Inovelli devices that feature button and switch functionality. Once we are able to test with these devices, there will likely be additional changes needed in this PR, however I elected to push up the changes now in order to begin getting some feedback as well as to begin the discussion on the best way to handle deprecating the matter-button driver.

lelandblue:
@nickolas-deboom -  I would encourage you to add a few additional folks like Steven or Z Varberg as reviewers as they might have additional context when trying to "depreciate a driver".

drivers/SmartThings/matter-button/src/init.lua
```diff
@@ -1,3 +1,8 @@
+-- As of July 2024, the matter-button driver has been deprecated in favor of being integrated into the matter-switch
+-- driver. All functionality and test cases present in this driver were carried over into matter-switch. Note that this
+-- won't affect devices using the button driver unless they are re-onboarded, in which case they would onboard to the
+-- switch driver, as all button fingerprints were removed from this driver and moved to the switch driver.
+
```
lelandblue, line 5: I would add additionally here that we are no longer accepting bug fixes, feature enhancements such as fingerprint adds to this driver. 

Something along those lines to make it really clear - there are no planned changes to folks who might be reading this.

nickolas-deboom: Updated the comment as suggested!

greens (approved):

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -121,98 +270,233 @@ local function assign_child_profile(device, child_ep)
...
+              component_map[string.format("switch%d", current_component_number)] = ep
+            else
+              component_map[string.format("button%d", current_component_number)] = ep
+            end
+            current_component_number = current_component_number + 1
+          else
+            component_map["main"] = ep
+          end
+          component_map_used = true
+        else -- Create child devices for non-main endpoints
```
ctowns, line 372: Update comment

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -121,98 +270,233 @@ local function assign_child_profile(device, child_ep)
...
-  if num_server_eps > 0 and detect_matter_thing(device) == true then
-    local id = 0
-    for _, ep in ipairs(device.endpoints) do
-      -- main_endpoint only supports server cluster by definition of get_endpoints()
-      if main_endpoint == ep.endpoint_id then
-        for _, dt in ipairs(ep.device_types) do
-          -- no device type that is not in the switch subset should be considered.
-          if (ON_OFF_SWITCH_ID <= dt.device_type_id and dt.device_type_id <= ON_OFF_COLOR_DIMMER_SWITCH_ID) then
-            id = math.max(id, dt.device_type_id)
+    if num_server_eps > 1  then
```
ctowns, line 392: The assumptions that `num_server_eps > 1` means that the device is parent-child is no longer true since a n-button device would have 2 server eps but be represented with an MCD profile. You can just exchange this for a boolean that is set to true when `try_create_device` is called in the conditional above

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -121,98 +270,233 @@ local function assign_child_profile(device, child_ep)
...
+      -- If the device is a parent child device, then set the find_child function on init.
+      -- This is persisted because initialize switch is only run once, but find_child function should be set
+      -- on each driver init.
+      device:set_field(IS_PARENT_CHILD_DEVICE, true, {persist = true})
+    end
+
+    device:set_field(SWITCH_INITIALIZED, true)
+
+    if #button_eps > 0 then
```
ctowns, line 401: What if the device has a mixture of buttons and switch device types? For example, a button with an LED light on it? I think we need to determine how that kind of device type should be represented

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -121,98 +270,233 @@ local function assign_child_profile(device, child_ep)
...
+        else
+          new_profile = string.format("%d-button", #button_eps)
+        end
+      elseif not battery_support then
+        -- a battery-less button/remote (either single or will use parent/child)
+        new_profile = "button"
+      end
+
+      if new_profile then
+        if #switch_eps > 0 then
```
ctowns, line 419: This would only work in the case that there is just one switch EP and it is an On/Off plug/switch, it doesn't account for more complex device types like color bulbs. This might be out of the scope of this PR, but we may just want to remove this part until we have a more comprehensive solution for these complex device types and just limit this PR to supporting the devices that are already supported by the `matter-button` driver.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -121,98 +270,233 @@ local function assign_child_profile(device, child_ep)
...
+      end
+    else
+      -- The case where num_server_eps > 0 is a workaround for devices that have a
+      -- Light Switch device type but implement the On Off cluster as server (which is against the spec
+      -- for this device type). By default, we do not support Light Switch device types because by spec these
+      -- devices need bindings to work correctly (On/Off cluster is client in this case), so these device types
+      -- do not have a generic fingerprint and will join as a matter-thing. However, we have seen some devices
+      -- claim to be Light Switch device types and still implement their clusters as server, so this is a
+      -- workaround for those devices.
+      if num_server_eps > 0 and detect_matter_thing(device) == true then
```
ctowns, line 432: You'll what to make this `num_switch_server_eps` so that we are not counting button server eps also. This logic should only affect Light Switch device types as described in the comment above, so we only want to count server eps for the light switch types when determining if we need to go down this codepath.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -121,98 +270,233 @@ local function assign_child_profile(device, child_ep)
...
+    -- that have been implemented as server. This can be removed when we have
+    -- support for bindings.
+    local num_server_eps = 0
+    local main_endpoint = find_default_endpoint(device)
+
+    for _, ep in ipairs(all_eps) do
+      if device:supports_server_cluster(clusters.OnOff.ID, ep) or device:supports_server_cluster(clusters.Switch.ID, ep) then
+        num_server_eps = num_server_eps + 1
+        -- Configure MCD for button endpoints
+        if tbl_contains(STATIC_PROFILE_SUPPORTED, #button_eps) then
```
ctowns, line 360: I think this would break in the event that we have a more complex device type such that
```
ep1 = switch
ep2 = button
ep3 = button
```
Because then the first switch EP would be assigned to main button component, and the ep3 button would be created as a child of the first two endpoints.

Overall, I don't think we have a robust plan for handling combined switch + button device types yet. If the scope of these PR is just meant to "add support for button devices to the matter switch driver, but NOT add support for combination switch + button devices" then maybe we could simplify this by handling the button and switch eps separately and not assuming that there would be a device that supports both. 

However, if the goal is to add support for switch + button combo devices, then I think we need to reconsider how we would handle that. Right now, buttons are mostly be handled as hard-coded MCD profiles. Switch devices types (plugs, bulbs, etc.) are being handled as parent-child devices. Would a combo device have to be MCD by default? Could we mix parent child with MCD (I'm not sure if this is even possible) and have an MCD 3-button parent with a child light device (or vice versa)? Right now, I don't think this could handle devices like that properly and I think this is worth starting a discussion about - unless I have already missed this somewhere!

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -226,6 +510,8 @@ end
 
 local function device_init(driver, device)
   if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    device:set_component_to_endpoint_fn(component_to_endpoint)
```
ctowns, line 513: Do these functions still need to be moved here before the `initialize_switch` call, or do they work where they were before now that we have the deferred configure for the buttons in place?

I don't think it matters where these are called in the `init` function, I just want to make sure I understand why they needed to be moved and if they still need to be for some reason.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: My thinking for this would be that if there are any button endpoints, the device will use an MCD configuration, and you can see [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/217b899ba18265f1e6123416e352a63343dc61b5/drivers/SmartThings/matter-switch/src/init.lua#L363) that a switch endpoint would use switch%d for its component name. I created an example profile of a 2 button, 1 switch device that would use this configuration. [This line](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/217b899ba18265f1e6123416e352a63343dc61b5/drivers/SmartThings/matter-switch/src/init.lua#L420) would append "-switch" to the profile name if there are switch endpoints in addition to button endpoints. If devices come out that need multiple switches or other configurations, we could update this profile selection logic and create new profiles for those devices, like we decided to do in the thermostat and sensor drivers. Do you think this approach would make sense?

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I saw this comment after the previous one, and I do think that would make sense. My only question is, I created test cases for this example 2-button 1-switch device and so if we remove this then those test cases obviously wouldn't work. Should we just go ahead and remove this for now, and also remove the test cases until we actually have a more complex device like this?

PR #1425, opened 2024-06-10
New Devices (Z-Wave Switch) Leviton ZW15S ZW6HD

lelandblue:
This PR is created as an intentional duplicate of https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/876. PR 876 can no longer accept changes, and minor change needed to be completed before it was merged so I created this PR. 

When the partner originally submitted PR 876 the new profile for leviton zw6hd, listed the capability of firmwareUpdate however that device does not have support for that capability. 

PR 876 has completed WWST Testing. So I think we will be good to merge this PR after reviews.


drivers/SmartThings/zwave-switch/profiles/leviton-zw6hd
greens, line 1: this file is missing its extension (.yml)

greens (approved):

PR #1422, opened 2024-06-06
Add drivers for Matter 1.2 Device Types

ctowns:
This is a more easily reviewable version of https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1272 that condenses everything in two commits: one which has all the driver changes, and one which includes the driver-embedded cluster definitions. This will hopefully make it easier to review the files, and also clean up the commit history.

This is the candidate for merging.

The new Matter 1.2 Device Types are sorted into different drivers as specified below. The driver channel link attached to the PR can be used to share the latest 1.2 device type drivers for testing.

### Merge Checklist

- [x] Remove duplicate profiles
- [x] Resolve luacheck warnings

### Device Types per Driver

**Matter Sensor**
- Smoke CO Alarm
- Air Quality Sensor

**Matter Thermostat**
- Air Purifier
- Room Air Conditioner

**Matter Appliance**
- Dishwasher
- Laundry Washer
- Refrigerator
- Temperature Controlled Cabinet

**Matter Robot Vacuum Cleaner**
- Robot Vacuum Cleaner

PR #1430, opened 2024-06-11
Matter Pump: add support for pump device type

ctowns:
These matter pump changes were taken from this Matter 1.2 PR so we could address some remaining open items: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1422

### DIFF
--- a/drivers/SmartThings/matter-pump/profiles/pump-level.yml
+++ b/drivers/SmartThings/matter-pump/profiles/pump-level.yml
@@ -1,0 +1,21 @@
+name: pump-level
+components:
+- id: main
+  capabilities:
+  - id: switch
+    version: 1
+  - id: switchLevel
+    version: 1
+  - id: pumpOperationMode
+    version: 1
+  - id: pumpControlMode
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Irrigation
+metadata:
+  mnmn: SmartThingsEdge
+  vid: generic-pump-level
--- a/drivers/SmartThings/matter-pump/profiles/pump-level.yml
+++ b/drivers/SmartThings/matter-pump/profiles/pump-level.yml
@@ -15,7 +15,7 @@ components:
   - id: refresh
     version: 1
   categories:
-  - name: Irrigation
+  - name: Pump
 metadata:
   mnmn: SmartThingsEdge
   vid: generic-pump-level
--- a/drivers/SmartThings/matter-pump/profiles/pump-only.yml
+++ b/drivers/SmartThings/matter-pump/profiles/pump-only.yml
@@ -13,7 +13,7 @@ components:
   - id: refresh
     version: 1
   categories:
-  - name: Irrigation
+  - name: Pump
 metadata:
   mnmn: SmartThingsEdge
   vid: generic-pump-level
--- a/drivers/SmartThings/matter-pump/profiles/pump-only.yml
+++ b/drivers/SmartThings/matter-pump/profiles/pump-only.yml
@@ -1,0 +1,19 @@
+name: pump-only
+components:
+- id: main
+  capabilities:
+  - id: switch
+    version: 1
+  - id: pumpOperationMode
+    version: 1
+  - id: pumpControlMode
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Irrigation
+metadata:
+  mnmn: SmartThingsEdge
+  vid: generic-pump-level
--- a/drivers/SmartThings/matter-pump/config.yml
+++ b/drivers/SmartThings/matter-pump/config.yml
@@ -1,0 +1,4 @@
+name: 'Matter Pump'
+packageKey: 'matter-pump'
+permissions:
+  matter: {}
--- a/drivers/SmartThings/matter-pump/fingerprints.yml
+++ b/drivers/SmartThings/matter-pump/fingerprints.yml
@@ -1,0 +1,6 @@
+matterGeneric:
+  - id: "matter/pump"
+    deviceLabel: Matter Pump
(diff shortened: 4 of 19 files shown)


drivers/SmartThings/matter-pump/profiles/pump-level.yml
```diff
@@ -0,0 +1,21 @@
...
+  - id: pumpOperationMode
+    version: 1
+  - id: pumpControlMode
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Irrigation
```
ctowns, line 18: Shouldn't we use the Pump category @HunsupJung ?

Comment copied from here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1422#discussion_r1632061585

drivers/SmartThings/matter-pump/profiles/pump-level.yml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
```
ctowns, line 19: @HunsupJung why is pump using a custom presentation?

Comment copied from here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1422#discussion_r1632061902

drivers/SmartThings/matter-pump/profiles/pump-only.yml
```diff
@@ -0,0 +1,19 @@
...
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Irrigation
+metadata:
+  mnmn: SmartThingsEdge
+  vid: generic-pump-level
```
tpmanley, line 19: If a custom vid is needed, is this the core one? This profile doesn't have level

drivers/SmartThings/matter-pump/profiles/pump-level.yml
HunsupJung: HJ commented what the process was on following ticket.
https://smartthings.atlassian.net/browse/CHAD-13354?focusedCommentId=1456242

drivers/SmartThings/matter-pump/profiles/pump-level.yml
HunsupJung: It's related naming of switchLevel capability. switchLevel Capability is displayed as `dimmer`. so we needed to modify it to `level` through device configuration.

drivers/SmartThings/matter-pump/profiles/pump-only.yml
HunsupJung: The reason why the same device configuration is applied to all profiles is that the presentation id of the card is changed when a profile which device configuration is applied is changed from a profile which device configuration is not applied in the driver. and There is no way to check if the presentation id has been changed in the app, so we applied device configuration to all profiles. this is the way proposed by the presentation team, so please let me know if there is a better way.

HunsupJung (approved):

drivers/SmartThings/matter-pump/profiles/pump-level.yml
tpmanley: I see. Seems like that's the best option for now but let's talk internally about longer term options so this isn't necessary

drivers/SmartThings/matter-pump/profiles/pump-only.yml
tpmanley: I don't think I was aware of this issue. Let's talk about this internally to see if there are other options

drivers/SmartThings/matter-pump/profiles/pump-level.yml
tpmanley: I responded on the ticket and we can discuss further there

drivers/SmartThings/matter-pump/profiles/pump-level.yml
HunsupJung: @tpmanley 
Yes, please tell me if there is any part that needs to be proceeded.

drivers/SmartThings/matter-pump/profiles/pump-level.yml
ctowns: I have updated the profiles to use the `Pump` category based on internal discussion here: https://smartthings.atlassian.net/browse/CHAD-13354?focusedCommentId=1456242

drivers/SmartThings/matter-pump/profiles/pump-level.yml
ctowns: Let's plan to wrap up discussion on this topic this week so we can get this driver to beta next week.

ctowns: I think we will need to move forward with the metadata for now to modify the presentation to say `level`

drivers/SmartThings/matter-pump/profiles/pump-only.yml
ctowns: @HunsupJung does this mean that both profiles will use the `generic-pump-level` metadata, but the level will still not be displayed for the `pump-only` profile?

HunsupJung: Yes, that is correct.

ctowns: If we move forward with the custom metadata, will be we able to set things like `capabilities.switchLevel.levelRange()` in the driver, or will the metadata override that? This is similar to the question I asked here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1458#discussion_r1653382008

ctowns: I have tested this and the answer is yes, we will still be able to use things like the `capabilities.switchLevel.levelRange()` as needed

PR #1433, opened 2024-06-11
Harman Luxury Change from REST HTTP API to WebSckets

alon-tchelet:
This PR is fixing pretty much all the issues previously experienced with the driver as the vast majority of issues came from random timeouts and the lack of lua sockets usage.

By moving all the communication between the device and the hub to websockets we now have a stable bidirectional connection based on lua sockets. The old inefficient and blocking REST API polling was removed and the device is now able to give live updates without being polled by the hub. This results in a much more relaxed and quieter network activity by the device and hub and a smoother user experience.

Note: This PR is replacing previous improvement attempt in now closed PR #1360


cjswedes (commented):
The usage of websockets is really good for this driver, and I think it is worthwhile to pursue it. Unfortunately the lustre library used here is out of date; I assume you copied from bose which is also out of date and we intend to update some day. You should instead use the [version](https://github.com/cosock/lustre/releases/tag/v0.1.0-4) that is made available in the lua libraries so that you are using the most recent version which has several bug fixes, and so you dont have to include any of the lustre files in the driver package. If you want an example of using the newer api, which isn't callback based, sonos is really the only example (sonos packages its own lustre but it is almost identical to what is in the lua libs with only a small change to the constructor api iirc).

Sonos has a `Router`  [which is a single cosock task](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/f5d1eb4757cfdd4e6bcb89f753db35a2af71f607/drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua#L37-L132) that manages all sending and receiving for all the websockets used in the driver. You could follow that approach or another option is to have a single task for managing each websocket individually.

drivers/SmartThings/harman-luxury/src/api/apis.lua
```diff
@@ -29,27 +20,6 @@ local APIs = {}
...
---- get Member ID from Harman Luxury on ip
----@param ip string
----@return string|nil, nil|string
-function APIs.GetMemberId(ip)
-  return get.String(ip, MEMBER_ID_PATH)
-end
-
 --- get device manufacturer name from Harman Luxury on ip
 ---@param ip string
 ---@return string|nil, nil|string
```
dljsjr, line 25: Are these REST calls still here because they're needed before the REST connection is open, or because there's no equivalent WebSocket message?

drivers/SmartThings/harman-luxury/src/api/nsdk.lua
```diff
@@ -88,7 +88,12 @@ local function handleReply(func_name, u, sink, code, valLocationFunc)
     end
     return nil, err
   else -- UNKNOWN VALUE
-    local err = string.format("Error in %s: Unknown return value: code: %s, sink: %s", func_name, code, sink)
+    local err
+    if string.find(code, "timeout") then
```
dljsjr, line 92: Very minor nit: Since `code` can be both an integer and a string, the type signature on the function doc comments should be `---@param code integer|string`

dljsjr:
I would also advocate for using the `cosock`/`select` based `lustre` instead of the callback based one. As Carter mentioned, it's included in the Lua Libs, so you'd no longer need to vendor it, but you will need to adjust the API usage.

> (sonos packages its own lustre but it is almost identical to what is in the lua libs with only a small change to the constructor api iirc)

As Carter mentioned, the patch to `lustre` in the Sonos driver is a minor tweak to the constructors. We need to be able to pass in our own TCP socket instead of having `lustre` create one implicitly. This is because the default `lustre` package doesn't work with `wss://` out of the box, but the only thing needed for `wss://` is passing in a TLS-wrapped lua socket. So the Sonos version of `lustre` modifies the constructor to allow for passing in our own socket.

If that's too large of a refactor for your timelines on this PR, it's not a blocker. But it is something I would recommend you keep on your radar, in the interest of keeping the code size and memory usage for the driver at a minimum by using as much shared code as possible.

drivers/SmartThings/harman-luxury/src/disco.lua
```diff
@@ -41,20 +32,12 @@ function Discovery.set_device_field(driver, device)
   local device_cache_value = driver.datastore.discovery_cache[device.device_network_id]
 
   -- persistent fields
-  device:set_field(const.STATUS, true, {
-    persist = true,
-  })
   device:set_field(const.IP, device_cache_value.ip, {
```
dljsjr, line 35: Does this ever get updated if the IP address of the speaker changes due to DHCP?

drivers/SmartThings/harman-luxury/src/hl_websocket.lua
```diff
@@ -0,0 +1,268 @@
...
+
+--- a websocket to get updates from Harman Luxury devices
+--- @class harman-luxury.HLWebsocket
+--- @field driver table the driver the device is a memeber of
+--- @field device table the device the websocket is connected to
+--- @field websocket table|nil the websocket connection to the device
+local HLWebsocket = {}
+HLWebsocket.__index = HLWebsocket
+
+--- hendles capabilities and sends the commands to the device
```
dljsjr, line 21: typo: `hendles` -> `handles`

drivers/SmartThings/harman-luxury/src/disco.lua
alon-tchelet: Good point. I forgot taking this into account when changing to WebSockets. I'll add some logic to handle this scenario as I switch to the lua library's lustre library

alon-tchelet:
Hi @dljsjr and @cjswedes ,
I pushed the reworked version. Could you have a look?
I do have an issue with the `keep_alive` config of the websocket. I obviously need it to inspect if the device is still online, but it seems that when I provide a number value I get this error:
```
FATAL Harman Luxury  runtime error: [string "cosock.lua"]:313: [string "lustre/ws.lua"]:299: attempt to index a nil value (local 'recv')
stack traceback:
	[string "lustre/ws.lua"]:299: in method '_handle_recvs'
	[string "lustre/ws.lua"]:286: in method '_receive_loop'
	[string "lustre/ws.lua"]:178: in function <[string "lustre/ws.lua"]:177>
stack traceback:
	[C]: in function 'error'
	[string "cosock.lua"]:313: in field 'run'
	[string "st/driver.lua"]:1016: in method 'run'
	[string "init.lua"]:207: in main chunk
```
I don't understand what I'm missing that causes it to try to access a `nil` value.

PS, I will clean up the commit messages when the changes are accepted

drivers/SmartThings/harman-luxury/src/lustre/ws.lua
```diff
@@ -0,0 +1,655 @@
...
+    order = not order
+    local recv, _, err = socket.select(rs, nil,
+      self.config._keep_alive)
+    log.debug((recv and "recv") or "~recv",
+      err or "")
+    if not recv then
+      if self:_handle_select_err(loop_state, err) then
+        return
+      end
+    elseif self:_handle_recvs(loop_state, recv, 1) then
```
dljsjr, line 285: I'm the person that was helping you out on the forums :)

You can fix this issue by only including this file while preserving the directory structure. You don't need to include the entire library.

alon-tchelet: Okay. I'll modify  it

dljsjr (approved):
Approved pending the change that @alon-tchelet is already making 🎉

alon-tchelet:
Cleaned the commits. I will update when the self verification is complete

cjswedes (approved):

cjswedes:
@alon-tchelet we'd like to get this merged by 11am CST 7/2. Can you squash your commits down to remove the fixup commits?

alon-tchelet:
Sorry, I did it locally and forgot pushing to remote. All good now.
Do you know how long will it be until this driver will get to end users?

cjswedes:
If all goes well with the release. Itll go to the beta channel today, and the production channel next week Monday.

PR #1437, opened 2024-06-12
update popp thermostat

mh-zwave:


Brianj94:
Hi Team, this one is coming from Aeotec. They are looking to make a minor fix for the /popp subdriver. This device is already WWST certified. Please let us know if there are any issues with the change they would like to make.

greens:
Looks like you'll also need to update the test.

I'd also suggest adding a few more. I'd suggest adding one that sets the mode. That way this error could have been detected earlier.

greens (approved):

PR #1439, opened 2024-06-12
Matter Pump: improve support for OperatingMode handling

hcarter-775:


drivers/SmartThings/matter-pump/profiles/pump-level.yml
```diff
@@ -6,6 +6,10 @@ components:
     version: 1
   - id: switchLevel
     version: 1
+    config:
+      values:
+        - key: "level.value"
+          range: [1, 100]
```
HunsupJung, line 12: From Presentation team, I heard and checked when embedded config is applied, metadata cannot be applied.

drivers/SmartThings/matter-pump/src/init.lua
```diff
@@ -175,9 +193,11 @@ local function on_off_attr_handler(driver, device, ib, response)
 end
 
 local function level_attr_handler(driver, device, ib, response)
-  if ib.data.value ~= nil then
-    local level = math.floor((ib.data.value / 254.0 * 100) + 0.5)
-    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.switchLevel.level(level))
+  if ib.data.value then
+      local max_level = get_field_for_endpoint(device, LEVEL_MAX, ib.endpoint_id) or DEFAULT_MAX_LEVEL
```
ctowns, line 197: nit: unindent this by one tab (2 spaces)

drivers/SmartThings/matter-pump/src/init.lua
```diff
@@ -175,9 +193,11 @@ local function on_off_attr_handler(driver, device, ib, response)
 end
 
 local function level_attr_handler(driver, device, ib, response)
-  if ib.data.value ~= nil then
-    local level = math.floor((ib.data.value / 254.0 * 100) + 0.5)
-    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.switchLevel.level(level))
+  if ib.data.value then
+      local max_level = get_field_for_endpoint(device, LEVEL_MAX, ib.endpoint_id) or DEFAULT_MAX_LEVEL
+      local level = math.floor((ib.data.value / max_level * MAX_CAP_SWITCH_LEVEL) + 0.5)
+      level = math.max(level, MIN_CAP_SWITCH_LEVEL)
```
ctowns, line 199: This is the value we actually receive from the device, so I don't think we should change this value. If the device gives us 0, then that should be considered valid if that is how the device works

hcarter-775:
Similar merge was made earlier today (6/26). Closing this PR.

PR #1440, opened 2024-06-13
Use component_to_endpoint to get default endpoint

ctowns:
Currently, the endpoint is hardcoded to the matter default endpoint (1) for some of the appliance device types. This is the default assumption for simple single component devices. However, we should use the `component_to_endpoint` function instead for more robust endpoint handling. In these cases, `component_to_endpoint` will essentially call `find_default_endpoint` and return the first nonzero endpoint on the matter device. In many cases it could be 1, but this change will make the endpoint handling more generic.

### DIFF
--- a/drivers/SmartThings/matter-appliance/src/init.lua
+++ b/drivers/SmartThings/matter-appliance/src/init.lua
@@ -247,8 +247,8 @@ local function handle_temperature_setpoint(driver, device, cmd)
     return
   end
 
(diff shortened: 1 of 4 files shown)


nickolas-deboom (approved):

HunsupJung:
@ctowns 
I tested your PR and there was no problem. 👍

HunsupJung (approved):

PR #1442, opened 2024-06-13
Matter-Window-Covering: Add test cases for shade status issue

nickolas-deboom:
This change adds test cases to verify the fix for an issue where the window covering operational status is not updated from "Opening" or "Closing" to "Partially Open" after the shade level is adjusted in the ST app.

### DIFF
--- a/drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
+++ b/drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
@@ -598,4 +598,188 @@ test.register_coroutine_test(
   end
 )
 
+test.register_coroutine_test(
+    "WindowCovering shade level adjusted by greater than 2%; status reflects Closing followed by Partially Open", function()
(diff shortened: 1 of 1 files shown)


tpmanley:
Can you add a unit test that would catch this issue?

nickolas-deboom:
> Can you add a unit test that would catch this issue?

I added 2 test cases in the latest commit: the first adjusts the shade level by > 2%, which passes on both the main branch and with these changes. The second adjusts the shade level by 2%, which on the main branch does not update the capability to Partially Open.

nickolas-deboom:
A similar fix to this issue was made by [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1447). I think that that solution is more complete, as it includes an update to `current_status_handler` that I believe is needed. I think that I will revert my driver changes, but I think it would still a good idea to keep the test cases updates made in this PR, or possibly include them in PR 1447.

ctowns (approved):
Thanks for adding these unit tests Nick!

PR #1445, opened 2024-06-14
Matter Switch: ensure device init is called on added

ctowns:


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -379,7 +379,10 @@ local function temp_attr_handler(driver, device, ib, response)
       return
     end
     local temp = utils.round(MIRED_KELVIN_CONVERSION_CONSTANT/ib.data.value)
-    local temp_device = find_child(device, ib.endpoint_id) or device
+    local temp_device = device
+    if device:get_field(IS_PARENT_CHILD_DEVICE) == true then
+      temp_device = find_child(device, ib.endpoint_id) or device
+    end
```
samadDotDev, line 385: Potentially a separate commit so this is more visible outside of the blanket init fix.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -510,6 +513,9 @@ local function device_added(driver, device)
   if device.network_type == device_lib.NETWORK_TYPE_CHILD then
     handle_refresh(driver, device)
   end
+
+  -- call device init in case init is not called after added due to device caching
+  device_init(driver, device)
```
samadDotDev, line 518: Just calling out here as well that this has been audited to be idempotent

samadDotDev (approved):

Con-Tejus (approved):

drivers/SmartThings/matter-switch/src/init.lua
ctowns: Done

PR #1446, opened 2024-06-15
Map doorlock that doesn't support Power source cluster to lock-lockalarm-nobattery

HunsupJung:
SecuYou releases their Matter supported door lock. securYou Smart Lock. ([Link](https://csa-iot.org/csa_product/secuyou-smart-lock/))
This door lock doesn't support Power source cluster and we need to map door lock that doesn't support Power source cluster to lock-lockalarm-nobattery.

### DIFF
--- a/drivers/SmartThings/matter-lock/src/init.lua
+++ b/drivers/SmartThings/matter-lock/src/init.lua
@@ -526,8 +526,13 @@ local function device_added(driver, device)
   --Note: May want to write RequirePINForRemoteOperation, to avoid cota cases if possible.
   device:emit_event(capabilities.tamperAlert.tamper.clear())
   device:emit_event(capabilities.lockAlarm.alarm.clear())
-  local eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
-  if #eps == 0 then
+  local pin_eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -526,8 +526,13 @@ local function device_added(driver, device)
   --Note: May want to write RequirePINForRemoteOperation, to avoid cota cases if possible.
   device:emit_event(capabilities.tamperAlert.tamper.clear())
   device:emit_event(capabilities.lockAlarm.alarm.clear())
-  local eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
-  if #eps == 0 then
+  local pin_eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
+  local battery_eps = device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
+  if #battery_eps == 0 then
+    device.log.debug("Device does not support battery. Switching profile.")
+    device:try_update_metadata({profile = "lock-lockalarm-nobattery"})
```
tpmanley, line 533: The profile will get changed again if `#pins_eps == 0` due to the code just below

HunsupJung: Yes, that is correct. In fact, because of the PR I pushed, there is no possibility of executing the code below at this time.  so.. I think below code can be removed from the device_added function.

Issue #1436, opened 2024-06-12
[BUG]Matter-window-covering update operational state incorrectly


404-code:
I found that the operational state was displayed incorrectly when the window covering stopped. It always showed "opening" or "closing" unless I refreshed it manually. Maybe there is something that needs to be fixed in the functions current_pos_handler and current_status_handler.

In this case, it would cause the bug:

When the blind starts to move, the blind only updates the operational status, setting ``IS_MOVING`` to true and ``EVENT_STATE`` to OPERATIONAL_STATE_EVENT. Until the blind stops, the operational status and current position update simultaneously (the message packet is added at the end). The message of the current position will be processed first. Because ``EVENT_STATE`` is OPERATIONAL_STATE_EVENT but ``IS_MOVING`` is true, it will set ``EVENT_STATE`` to NO_EVENT. Thus, when it processes the message of operational status, the code will not execute ``device::emit_event_for_endpoint to update the status``, because ``EVENT_STATE`` is NO_EVENT. Finally, it just sets ``IS_MOVING`` to false and shows "opening" or "closing".

If  just replace line 159 with ``device:set_field(EVENT_STATE, WindowCoveringEventEnum.CURRENT_POSITION_EVENT)``, could it cause other problems?

Looking forward to your reply. Thanks.

The message packet of the matter subscription response showed the following. It always reports the current position first:

> [1718162472.140788][34296:34298]` CHIP:EM: >>> [E:4155r S:20551 M:67760107] (S) Msg RX from 1:0000000000000001 [5F03] --- Type 0001:05 (IM:ReportData)
[1718162472.140841][34296:34298] CHIP:EM: Handling via exchange: 4155r, Delegate: 0xaaaabdd5ba30
[1718162472.140986][34296:34298] CHIP:DMG: ReportDataMessage =
[1718162472.141022][34296:34298] CHIP:DMG: {
[1718162472.141053][34296:34298] CHIP:DMG: 	SubscriptionId = 0xc267cab4,
[1718162472.141085][34296:34298] CHIP:DMG: 	AttributeReportIBs =
[1718162472.141127][34296:34298] CHIP:DMG: 	[
[1718162472.141159][34296:34298] CHIP:DMG: 		AttributeReportIB =
[1718162472.141208][34296:34298] CHIP:DMG: 		{
[1718162472.141242][34296:34298] CHIP:DMG: 			AttributeDataIB =
[...]

PR #1447, opened 2024-06-15
Fix a part of mis-implementation for window covering device type

HunsupJung:
Previously, the window covering driver was modified because the windowshade value was sometimes wrong.
And there was a part of mis-implementation.
So the issue was reported as below.
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/issues/1436

This is a patch that fix the issue.

### DIFF
--- a/drivers/SmartThings/matter-window-covering/src/init.lua
+++ b/drivers/SmartThings/matter-window-covering/src/init.lua
@@ -155,8 +155,8 @@ local function current_pos_handler(driver, device, ib, response)
       else
         device:emit_event_for_endpoint(ib.endpoint_id, capabilities.windowShade.windowShade.unknown())
       end
+      device:set_field(EVENT_STATE, WindowCoveringEventEnum.NO_EVENT)
     end
-    device:set_field(EVENT_STATE, WindowCoveringEventEnum.NO_EVENT)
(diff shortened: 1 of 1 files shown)


HunsupJung:
Please refer to following link for detail situation.
https://smartthings.atlassian.net/wiki/spaces/STHK/pages/3268051961/Matter+window+covering+device+type+issue

nickolas-deboom:
I looked into this reported bug as well and came to a similar conclusion as you did and I made a slightly different update [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1442). I think I will close that PR, because your changes also include an update to `current_status_handler` which I think are needed as well.

However, I might leave my PR up to at least include the test case updates. I added two test cases which exercise two scenarios: (1) adjusting the shadeLevel by > 2%, which works correctly on the main branch and with these changes, and (2) adjusting the shadeLevel by <= 2%, which on the main branch does not update the status to "Partially Open" but instead gets stuck on "Closing"/"Opening".

HunsupJung:
@nickolas-deboom 
Thank you for additional patch.

nickolas-deboom (approved):

PR #1448, opened 2024-06-16
Add unit tests for appliances, rvc, and room ac

ctowns:
Some of the Matter 1.2 device type unit tests were left out of the repo so that CI unit tests would not continuously fail due to missing capability definitions in the integration test framework. Once the CI is updated to use the latest lua libs, we can add these tests to the repo because the capability definitions will then be available in the CI lua libs.

Note:  This issue is strictly related to unit tests - new capabilities do not have a lua libs dependency on hubs, so even old hubs will get the new capability definitions even if they are running older lua libs.

These unit test all pass locally when running the latest lua libs.


ctowns:
We can merge these in once CI lua libs have been updated. I will follow up on what the timing will be to update the lua libs.

@HunsupJung @yhoyoon I have tried to go through and write unit tests for the major functionality of the appliance, rvc, and room ac device types, but please take a look and add additional unit tests if there are any additional cases to test. Thank you!

Note, some of these unit tests are from the original Matter 1.2 device types PR, but they were intentionally removed to prevent CI unit tests from continuously failing. However, these tests all pass when running locally with the latest lua libs.

HunsupJung (approved):

cjswedes:
I just published the 53 release of the lua libs. Are these unit test changes still needed? They are running against those libs now.

lelandblue:
@ctowns Should we keep or close this PR?

hcarter-775:
@ctowns can this ticket be closed?

PR #1449, opened 2024-06-17
Matter Lock: remove Aqara unit tests for now

ctowns:
Remove unit tests for Aqara door lock since they are using a new capability. We can add these back in once we have updated the CI.


hcarter-775 (approved):
Ok, glad this is removed for now :+1:

ctowns:
Will add these back with the 53 release

PR #1454, opened 2024-06-18
Aqara Smart Radiator Thermostat E1

Hongming6:


drivers/SmartThings/zigbee-thermostat/src/aqara/init.lua
```diff
@@ -0,0 +1,290 @@
...
+      ["13"] = 1100,
+      ["14"] = 1150,
+      ["15"] = 1200,
+      ["16"] = 1250,
+      ["17"] = 1300,
+      ["18"] = 1350,
+      ["19"] = 1400,
+      ["20"] = 1450,
+      ["21"] = 1500
+    },
```
greens, line 76: you could make `value_map` a function instead that returns `tonumber(value)*50 + 450` and then in the info changed handler, check the type of the `value_map` right after you check for `nil`.

Hongming6: Thx，pls check again~

Hongming6: Thx，pls check again~

greens (commented):
The reason your tests are failing is because the `antifreezing` thermostat mode was added _after_ the generation of the lua_lbs v53 artifact: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/releases/tag/apiv10_53

This is what the CI uses to run the tests, so it won't find the antifreezing mode locally. This isn't a problem on the hub, which pulls its capability definitions from the cloud, though.

In order to get around this, I would alter your tests so that they do not include code paths with `antifreezing`.

lelandblue:
Hello @Hongming6. Please review the comment above when you can, then we can proceed. Thank you.

wkhenon (approved):

Hongming6:
> The reason your tests are failing is because the `antifreezing` thermostat mode was added _after_ the generation of the lua_lbs v53 artifact: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/releases/tag/apiv10_53
> 
> This is what the CI uses to run the tests, so it won't find the antifreezing mode locally. This isn't a problem on the hub, which pulls its capability definitions from the cloud, though.
> 
> In order to get around this, I would alter your tests so that they do not include code paths with `antifreezing`.

Got it, you can modify the test, thanks.

Hongming6:
> Hey @Hongming6 - Sorry for the confusion. Can you please modify the tests?
@greens @lelandblue 
I blocked the associated test function so that the test would pass. Should I modify it this way?

greens:
@Hongming6 The two tests you have that reference `antifreezing`: the `added` test and the `controlSequenceOfOperation` test will probably need to be commented out, since there's no way to run them without hitting the attribute.

Hongming6:
> @Hongming6 The two tests you have that reference `antifreezing`: the `added` test and the `controlSequenceOfOperation` test will probably need to be commented out, since there's no way to run them without hitting the attribute.

Okay

greens:
Two things left, sorry:

```
Checking drivers/SmartThings/zigbee-thermostat/src/aqara/init.lua 1 warning

    drivers/SmartThings/zigbee-thermostat/src/aqara/init.lua:272:6: (W612) line contains trailing whitespace
```

And the `hardwareFault` capability is also too new. You'll need to include a copy of the capability's yaml alongside the driver here so that the tests pass.

Hongming6:
> And the `hardwareFault` capability is also too new. You'll need to include a copy of the capability's yaml alongside the driver here so that the tests pass.

I fill it in and the test still encounters an error, I'm not sure what causes it? I've pushed the relevant changes up, can you help me find out why? Or how about just commenting out this test?

drivers/SmartThings/zigbee-thermostat/src/aqara/init.lua
```diff
@@ -0,0 +1,277 @@
...
+  device:emit_component_event(device.profile.components.ChildLock, capabilities.lock.lock.unlocked())
+  device:emit_event(capabilities.hardwareFault.hardwareFault.clear())
+  device:emit_event(valveCalibration.calibrationState.calibrationPending())
+  device:emit_event(invisibleCapabilities.invisibleCapabilities({""}))
+  device:emit_event(capabilities.battery.battery(100))
+end
+
+local function thermostat_alarm_status_handler(driver, device, value, zb_rx)
+  if value.value == 1 then
+    capabilities.hardwareFault.hardwareFault.detected()
```
greens, line 128:
```suggestion
    device:emit_event(capabilities.hardwareFault.hardwareFault.detected())
```

This was the actual issue. You don't need to include the `hardwareFault` yaml, it's actually already in the release artifact.

Hongming6: Thanks,modified

greens (approved):

PR #1456, opened 2024-06-18
Matter Window Covering: Remove CurrentLevel attribute handling

nickolas-deboom:
While testing with the VDA Window Covering device it was noticed that when the device card for a window covering device is refreshed, the shade level displays as 0%. This is due to the CurrentLevel attribute, which is not a required or optional attribute for the window covering device type, overwriting the windowShadeLevel capability after it is set by CurrentPositionLiftPercent100ths. This change removes the CurrentLevel attribute from the matter-window-covering driver.

These changes were tested using the VDA Window Covering virtual device by verifying that all capabilities display correctly, all commands function correctly, and that upon a refresh the shadeLevel is not reset back to 0%.

[MTR-715](https://smartthings.atlassian.net/browse/MTR-715)

[MTR-715]: https://smartthings.atlassian.net/browse/MTR-715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hcarter-775:
What testing have you done to make sure that things are still working correctly?

nickolas-deboom:
> What testing have you done to make sure that things are still working correctly?

Hey Harrison, I realized that I never responded to your comment, but I updated the description of this PR with some info on how this was tested!

nickolas-deboom:
> This changes makes sense to me, as the [WindowCovering device type](https://librarian-regionals.smartthingsgdev.com/matter/Matter_1-3/device_library.html#ref_WindowCovering) does not specify a requirement for the LevelControl cluster, and the `CurrentPositionLiftPercent100ths` attribute should be used to handle the window "level" setting. So I think this change makes sense.
> 
> However, did this originate just from VDA testing or do you know if this has been seen on devices? I just want to make sure this is safe to remove and no devices are actually using levelControl. Based on a quick git blame, it looks like the LevelControl cluster handler has been in the driver since it was originally released, so it seems to me it is just an artifact of an early implementation of the driver/device type that is no longer needed.

This originated from the VDA, which sends the CurrentLevel attribute, always with a value of 0. This would override the value of CurrentPositionLiftPercent100ths after a refresh. This has only been tested with the VDA - do you know of anyone with a physical window covering device that we could test this with?

nickolas-deboom:
After some consideration, this PR will be closed in favor of updating the VDA Window Covering device type to not send the `CurrentLevel` attribute. Implementing this driver change would be a risk because there is a possibility that there is a device out there that uses this attribute. Unless we see a report of a real device with this issue, we shouldn't make this driver change.

PR #1458, opened 2024-06-19
Add device configuration for supporting Routine

HunsupJung:
In the case of mode capability, presentation is implemented in the list display type.
Therefore, device configuration was added to support automation in the device type that supports the mode capability.

### DIFF
--- a/drivers/SmartThings/matter-appliance/profiles/dishwasher-tn.yml
+++ b/drivers/SmartThings/matter-appliance/profiles/dishwasher-tn.yml
@@ -6,10 +6,6 @@ components:
     version: 1
   - id: temperatureSetpoint
     version: 1
-    config:
-      values:
-        - key: "temperatureSetpoint.value"
-          range: [ 0, 100 ]
   - id: mode
     version: 1
   - id: waterFlowAlarm
@@ -26,3 +22,6 @@ components:
     version: 1
   categories:
   - name: Dishwasher
+metadata:
+  mnmn: SmartThingsEdge
+  vid: generic-dishwasher
--- a/drivers/SmartThings/matter-rvc/src/init.lua
+++ b/drivers/SmartThings/matter-rvc/src/init.lua
@@ -155,13 +155,16 @@ local function rvc_run_mode_supported_mode_attr_handler(driver, device, ib, resp
       break
     end
   end
-  local labels_of_supported_modes = get_labels_of_supported_modes_filter_by_current_mode(device,
+  local labels_of_supported_arguments = get_labels_of_supported_modes_filter_by_current_mode(device,
     RVC_RUN_MODE_SUPPORTED_MODES,
     clusters.RvcRunMode.types.ModeTag.IDLE,
     current_mode
   )
+  local labels_of_supported_modes = get_field_labels_of_supported_modes(device, RVC_RUN_MODE_SUPPORTED_MODES)
   local component = device.profile.components["runMode"]
-  local event = capabilities.mode.supportedModes(labels_of_supported_modes, {visibility = {displayed = false}})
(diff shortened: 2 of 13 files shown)


drivers/SmartThings/matter-appliance/profiles/dishwasher-tn.yml
```diff
@@ -6,10 +6,6 @@ components:
     version: 1
   - id: temperatureSetpoint
     version: 1
-    config:
```
ctowns, line 9: @HunsupJung can we not use embedded config values when we have metadata? Do you know if we are still able to set these limits via a function call in the driver? For example, would something like `capabilities.thermostatHeatingSetpoint.heatingSetpointRange()` still work if we are using a profile with metadata?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -162,6 +162,8 @@ local function rvc_run_mode_supported_mode_attr_handler(driver, device, ib, resp
   )
 
   local component = device.profile.components["runMode"]
+  local labels = get_field_labels_of_supported_modes(device, RVC_RUN_MODE_SUPPORTED_MODES)
+  device:emit_component_event(component, capabilities.mode.supportedArguments(labels))
```
ctowns, line 166: Will `supportedArguments` be available for R2?

drivers/SmartThings/matter-appliance/profiles/dishwasher-tn.yml
HunsupJung: > Do you know if we are still able to set these limits via a function call in the driver? For example, would something like capabilities.thermostatHeatingSetpoint.heatingSetpointRange() still work if we are using a profile with metadata?

It doesn't work. If `embedded config` is set, `metadata` setting does not work. and the `temperature range` set in the `embedded config` can be set in `device configuration`. so it is removed from profile.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: It is set to device configuration to use supportedArguments in RVC.
https://github.ecodesamsung.com/iot/st-presentation/blob/master/device-configurations/EdgeGenericRVC.json

ctowns:
Is metadata the only way we can support `supportedArguments` for now? Longterm, could we modify the plugin to utilize the `supportedArguments` value for R3 so we do not need custom metadata?

HunsupJung:
Yes, the metadata is the only way to support `supportedArguments` for now. I will ask Plugin team to apply `supportedArguments` to `mode capability`. (Actually, It also needs to apply dynamic list to `mode capability`). But I'm not sure how long does it take that Plugin team applies my request.
> I think this is okay but in general I think we try to avoid using metadata with custom presentations.

I agree with you

drivers/SmartThings/matter-rvc/src/init.lua
ctowns, line 166: What is the difference between `supportedArguments` and `supportedModes`?

Also, should this be `labels_of_supported_modes` instead of just `labels`?

ctowns (commented):
@HunsupJung can you help me understand what effect this change will have and how I can test it? If I use the Matter VDA to test, what difference should I expect to see with these changes?

Also, do you need to add `supportedArguments` for the appliance drivers also?

HunsupJung:
@ctowns
The list stored in supportedArguments must be available for selection in the Detail view and the list stored in supportedModes must be available for selection in Routine (Automation).

Additionally, The presentation team is preparing patches that the setting that only can be set in `device configuration` can be set in `embedded configuration`. so This patch will be used temporarily and replaced with it when new patch are added.

ctowns (approved):

PR #1459, opened 2024-06-19
New Device (Matter Switch) Zemismart ZM606-3

lelandblue:
This PR is for a WWST Certification request to add the Zemismart ZM606-3 device. According to the DCL it has a  device type of 0x103. 

Here is the VID and the PID for this device

- vendorId: 0x139C
- productId: 0xAB31


ctowns:
Just leaving a note that device type `0x0103` is the ["On Off Light Switch" device type](https://librarian-regionals.smartthingsgdev.com/matter/Matter_1-3/device_library.html#ref_OnOffLightSwitch) which uses client clusters according to the spec, but we are able to provide support in some cases if the clusters are implemented as server instead.

ctowns (approved):

PR #1460, opened 2024-06-19
Add device configuration for matter-fan

HunsupJung:
The matter-fan driver uses airCondtionerFanMode and its label is "Air Condtioner Fan Mode".
This is for changing a label to "Fan Mode"  from "Air Conditioner Fan Mode"


ctowns:
I believe this will solve the issue in the future when we can use this: https://github.ecodesamsung.com/iot/SmartThingsCapabilities/pull/351

@HunsupJung could we wait for the new capability to use here, or do we need to make this change now?

HunsupJung:
We can use Fan Mode capability that you made. Thank you 👍

HunsupJung:
@ctowns  When should we change the airConditionFanMode to the newly created Fan Mode capability?

HunsupJung:
@ctowns 
Can we use this PR until creating new Fan Mode capability?

ctowns (approved):

hcarter-775 (approved):
Yes, I think using this until we release the new capability is good.

PR #1466, opened 2024-06-20
CHAD 13147: Matter Sensor Driver Profiles Cleanup

hcarter-775:
Many profiles in the drivers have outdated names that are unable to be mapped to. This change removes or renames outdated profile names to clean up this problem.

Jira ticket can be found [here](https://smartthings.atlassian.net/browse/CHAD-13417)

And this is the superset query that was used to make these decisions: https://superset.samsungiotanalytics.com/superset/sqllab?id=1155


drivers/SmartThings/matter-switch/profiles/matter-bridge.yml
```diff
@@ -1,4 +1,4 @@
-name: bridge
+name: matter-bridge
```
ctowns, line 1: I would think this profile is definitely being used in the field, so I don't think we can change the name safely

drivers/SmartThings/matter-sensor/profiles/matter-motion-battery-illuminance.yml
```diff
@@ -1,16 +0,0 @@
-name: matter-motion-battery-illuminance
```
ctowns, line 1: Can we safely delete this? How did you confirm it is not being used? Same for the profile below

ctowns: If there is a superset query, you could link it here for posterity :)

hcarter-775:
This is the superset query that was used to make these decisions: https://superset.samsungiotanalytics.com/superset/sqllab
cc: @ctowns

drivers/SmartThings/matter-sensor/profiles/matter-motion-battery-illuminance.yml
hcarter-775: See superset query that is mentioned below

ctowns: That link just goes to the supserset website but not actually the query you used. You'll have to specifically share the query, which you should be able to find next to the `Run` button

hcarter-775: that should be updated now, sorry about that!

ctowns (approved):

PR #1468, opened 2024-06-21
New Device (Matter Switch) Cync Indoor Plug CPLGSTDBLW1MS

lelandblue:
This device is pursuing WWST Certification. It is not published in the DCL, however on the CSA website this[ information was available](https://csa-iot.org/csa_product/cync-indoor-plug-3/). 

When I looked at the PICS documents for this device, in the on/off cluster test plan I see "Level Control for Lighting Matter_App_Clusters_1.5.4. O true". I concluded based on that information that this device supports Level Control and therefor the SwitchLevel capability withen the `deviceProfileName: plug-level`.


ctowns (approved):
This looks good to me. Is the plan to wait for this device to be in the DCL before officially certifying it?

PR #1471, opened 2024-06-24
Busted tests for Hue driver + Bugfix for migrated lights

dljsjr:
We don't run Busted tests in CI, but the Hue driver is now very testable after the recent refactor. The ability to write integration tests for LAN Drivers is still forthcoming, but we're also going to be accumulating unit tests for Hue going forward as well.

We keep the `spec` folder as a sibling to the `src` folder so that it doesn't get included in the driver package, so care should be taken when setting up the LUA PATH to run these tests.

We also fix a bug that was reported by some community users, the presence of which was validated by and is also now covered by the new specs.

## For Reviewers

The PR is broken up in to two commits. The first commits are all of the new Busted work. The second commit is the bugfix.

### DIFF
--- a/drivers/SmartThings/philips-hue/run_specs.sh
+++ b/drivers/SmartThings/philips-hue/run_specs.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+export LUA_PATH="./test/spec/?.lua;./test/spec/?/init.lua;./src/?.lua;./src/?/init.lua;$LUA_LIBS/?.lua;$LUA_LIBS/?/init.lua;;"
+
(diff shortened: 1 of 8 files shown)


varzac (approved):

varzac:
I approved, but actually I think we may want to move the `spec` and other folders under the `test` folder as that will by default be excluded from the archives that get uploaded.  But let me verify that.

varzac:
Yeah, test or tests directory would have the not uploaded and thus not pushed to everyones hub: https://github.com/SmartThingsCommunity/smartthings-cli/blob/0ebd3c72c819fd3681746c4b01740038a7ede910/packages/edge/src/commands/edge/drivers/package.ts#L136

PR #1479, opened 2024-06-26
Matter Thermostat extend Air Quality Sensor

hcarter-775:
Allows for profiles to be created that use any combination of AQS, thermostat, and fan device types. Also adds a new profile that includes much of this extended functionality.


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -212,6 +294,59 @@ local function get_device_type(driver, device)
...
+  {capabilities.veryFineDustHealthConcern.ID,    "-pm1",   clusters.Pm1ConcentrationMeasurement},
+  {capabilities.veryFineDustSensor.ID,           "-pm1",   clusters.Pm1ConcentrationMeasurement},
+}
+
+local function create_air_quality_profile(device)
+  local meas_name, level_name = "", ""
+  local aqs_supported
+
+  local aq_eps = embedded_cluster_utils.get_endpoints(device, clusters.AirQuality.ID)
+  if #aq_eps == 0 then
```
ctowns, line 325: There are devices that can support other concentration measurements but not the generic air quality, so we don't want to skip checking for the other measurements if a device doesn't support air quality.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -225,13 +360,11 @@ local function do_configure(driver, device)
...
   local device_type = get_device_type(driver, device)
   local profile_name = "thermostat"
-  --Note: we have not encountered thermostats with multiple endpoints that support the Thermostat cluster
   if device_type == RAC_DEVICE_TYPE_ID then
     device.log.warn_with({hub_logs=true}, "Room Air Conditioner supports only one profile")
   elseif device_type == FAN_DEVICE_TYPE_ID then
     device.log.warn_with({hub_logs=true}, "Fan supports only one profile")
-  elseif device_type == AP_DEVICE_TYPE_ID then
-    -- currently no profile switching for Air Purifier
+  elseif device_type == AP_DEVICE_TYPE_ID and #thermo_eps < 1 then
```
ctowns, line 367: I think we need to combine both the `AP_DEVICE_TYPE_ID` conditionals so that air purifier devices without thermostats can still get the air measurement support. For example, [this device](https://www.aidot.com/aidot-welov-matter-p200-pro-smart-air-purifier-for-homeup-to-1570-ft-with-air-quality-monitor.html?srsltid=AfmBOoohHvHFESMgWX8FIGGJj3Ij5mfKcH91ipvT9upptW9llLnSUJIKLw8) does not have a thermostat but still has some measurement capabilities.

I think combining them should be pretty simple, you can just add an extra check in within the combined `elseif device_type == AP_DEVICE_TYPE_ID` conditional to check if `#thermo_eps > 0` and then add the appropriate thermo components to the profile name. You can keep the thermo code you already have in the conditional below but just add this extra check before going through the thermo components.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -283,13 +474,153 @@ local function do_configure(driver, device)
...
+  },
+  [units.PPB] = {
+    [units.PPM] = function(value) return utils.round(value/(10^3)) end
+  },
+  [units.PPT] = {
+    [units.PPM] = function(value) return utils.round(value/(10^6)) end
+  },
+  [units.MGM3] = {
+    [units.UGM3] = function(value) return utils.round(value * (10^3)) end,
+    [units.PPM] = function(value) return utils.round((value * 24.45) / (30.031)) end,
```
ctowns, line 549: What is the 24.45 and 30.031 value? These should be defined as constants somewhere

Also, I believe this is the molecular weight of formaldehyde, so these conversion would only work for formaldehyde measurements. We should find a way to generalize this so that this equation can be used for any concentration type. I would propose wrapping the conversion table in a function and then adding special handling for the cases where a molecular weight is required:


Here is some pseudo code:
```
local molecular_weights = {
  [formaldehyde] = 30.031
}

local function convert_unit(current_unit, target_unit, concentration_type)
  if current_unit or target_unit is one that needs to be converted (MGM3, UGM3, etc) then
    local molecular_weight = molecular_weights[concentration_type]
    local value = conversion_tables[current_unit][target_unit]
    value = value / or * molecular_weight depending on the target/current unit
  else
    return conversion_tables[current_unit][target_unit]
end
```

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: handled :+1:

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -245,6 +378,64 @@ local function do_configure(driver, device)
...
+        profile_name = profile_name .. level_name .. "-level"
+      end
+    end
+
+    if profile_name == "air-purifier-hepa-ac-wind-thermostat-humidity-fan-heating-only-nostate-nobattery-aqs-pm10-pm25-ch2o-meas-pm10-pm25-ch2o-no2-tvoc-level" then
+      profile_name = "air-purifier-hepa-ac-wind-thermostat-humidity-fan-heating-only-nostate-nobattery-aqs"
+    end
+    device.log.info_with({hub_logs=true}, string.format("Updating device profile to %s.", profile_name))
+    device:try_update_metadata({profile = profile_name})
+    device.log.info_with({hub_logs=true}, string.format("Updating device profile to %s.", profile_name))
```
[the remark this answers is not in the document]
hcarter-775: done

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: I went a slightly different approach, but it should still work as intended. Let me know what you think.

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: should be working now in line with your suggestion here :+1:

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -283,13 +458,166 @@ local function do_configure(driver, device)
...
+  [units.BQM3] = {
+    [units.PCIL] = function(value) return utils.round(value/37) end
+  },
+}
+
+local function unit_conversion(value, from_unit, to_unit, capability_name)
+  local conversion_function = conversion_tables[from_unit][to_unit]
+  if conversion_function == nil then
+    log.info_with( {hub_logs = true} , string.format("Unsupported unit conversion from %s to %s", unit_strings[from_unit], unit_strings[to_unit]))
+    return 1
```
ctowns, line 556: Why is a return value of 1 being used here? If the value can't be converted we should return an invalid value and handle it appropriately in the calling function so we aren't just emitting random 1's

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -283,13 +458,166 @@ local function do_configure(driver, device)
...
+  if value == nil then
+    log.info_with( {hub_logs = true} , "unit conversion value is nil")
+    return 1
+  end
+
+  -- checks if number of parameters is greater than 1. This supports conversions
+  -- where there is a value other than simple scaling values
+  local conversion_function_info = debug.getinfo(conversion_function, "u")
+  if conversion_function_info.nparams > 1 then
+    return conversion_function(value, molecular_weights[capability_name])
```
ctowns, line 568: You can just call `conversion_function(value, molecular_weights[capability_name])` and any extra params are ignored in lua. However, if this were called for a concentration measurement that isn't `formaldehydeMeasurement` this would result in a driver crash due to a nil value being used for the molecular weight.

You could add the molecular weights for every concentration to the `molecular_weights` table to address this.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -243,6 +370,54 @@ local function do_configure(driver, device)
...
+
+    if meas_name ~= "" then
+      profile_name = profile_name .. meas_name .. "-meas"
+    end
+
+    if level_name ~= "" then
+      profile_name = profile_name .. level_name .. "-level"
+    end
+
+    if profile_name == "air-purifier-hepa-ac-wind-thermostat-humidity-fan-heating-only-nostate-nobattery-aqs-pm10-pm25-ch2o-meas-pm10-pm25-ch2o-no2-tvoc-level" then
```
ctowns, line 417: This should not longer be needed with the dossier profile name change

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: ok, we should be making these changes then in air-quality-sensor then. This part was just directly copied from there.

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: yep, I'm going to remove it in the next update.

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: Would that be necessary, or can we just leave this as is?

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: removed

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: done

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -283,13 +454,168 @@ local function do_configure(driver, device)
...
+local molecular_weights = {
+  [capabilities.carbonDioxideMeasurement.NAME] = 44.010,
+  [capabilities.nitrogenDioxideMeasurement.NAME] = 28.014,
+  [capabilities.ozoneMeasurement.NAME] = 48.0,
+  [capabilities.formaldehydeMeasurement.NAME] = 30.031,
+  [capabilities.veryFineDustSensor.NAME] = "N/A",
+  [capabilities.fineDustSensor.NAME] = "N/A",
+  [capabilities.dustSensor.NAME] = "N/A",
+  [capabilities.radonMeasurement.NAME] = 222.018,
+  [capabilities.tvocMeasurement.NAME] = units.PPM
```
ctowns, line 526: Should this be "N/A" as well?

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: Actually, looking at the problem more closely, I think this should be a separate PR. It should be fixed, but it is distinct enough and large enough that it should not be handled as an add-on here. This is especially true since it is handled equivalently in the air-quality-sensor subdriver

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -283,13 +454,168 @@ local function do_configure(driver, device)
...
+  return function(driver, device, ib, response)
+    local reporting_unit = device:get_field(capability_name.."_unit")
+
+    if reporting_unit == nil then
+      reporting_unit = unit_default[capability_name]
+      device:set_field(capability_name.."_unit", reporting_unit, {persist = true})
+    end
+
+    if reporting_unit then
+      local value = unit_conversion(ib.data.value, reporting_unit, target_unit, capability_name)
```
ctowns, line 582: Add an additional step here to make sure that `value` is not nil before emitting? We could check if value is nil, and if not then emit the value, and if it is then add a log mentioning that the value was not valid and was not emitted.

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: yes, thanks.

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: Yes, like I commented in the earlier review you made, I think this needs a fair amount of refactoring, and since we want this in today or tomorrow, I think this can be handled as a separate PR. Thoughts?

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: FWIW I think we might want to port some of the other functions you have here to the AQS sub driver as well. I think the name building function you have made here is neater and we could modify the AQS driver to follow the same method when we make these changes regarding the return value as well

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -243,6 +370,50 @@ local function do_configure(driver, device)
...
+      if #humidity_eps > 0 and #fan_eps > 0 then
+        profile_name = profile_name .. "-humidity" .. "-fan"
+      elseif #humidity_eps > 0 then
+        profile_name = profile_name .. "-humidity"
+      elseif #fan_eps > 0 then
+        profile_name = profile_name .. "-fan"
+      end
+    end
+
+    if #heat_eps == 0 and #cool_eps == 0 then
```
ctowns, line 385: The heat and cool eps should only be checked if thermo eps > 0, so move these under the conditional on line 374

ctowns: Move everything that is specifc to thermostat under this conditional. This includes the no-state and no-battery check

ctowns:
Tested with 52 and 53 lua libs and confirmed that AiDot Air Purifier will onboard and correctly switch to `air-purifier-hepa-aqs` profile

ctowns:
This is looking good, great work @hcarter-775 !!

One last thing would be to add some unit tests. This change is mostly to onboarding, so I think this would be a good unit test:
1. Create a couple complex devices (i.e. you can recreate the Dyson and AiDot if you want) as mock devices and confirm that they match to the expected profile.
2. Add an additional unit test for one of the unit conversions that requires the molecular weight just to confirm that functionality is working as expected

ctowns (approved):
This looks good to me - great work @hcarter-775 !

An overview of the testing done on this PR can be seen here: https://smartthings.atlassian.net/wiki/spaces/~62de3b63b6b0b70770d7905d/pages/3317399870/Matter+Thermostat+AQS+Support+Testing

Multiple Air Purifer devices were tested and confirmed to onboard to the correct profile and work correctly on both current (v53) and old (v52) lua libs. Additional regression testing was done with the VDA to ensure other devices in the driver continue to work correctly with these changes.

edit: also, don't forget to squash your commits!

PR #1480, opened 2024-06-26
Level attribute handling fixup

hcarter-775:


drivers/SmartThings/matter-pump/src/init.lua
```diff
@@ -176,7 +176,8 @@ end
 
 local function level_attr_handler(driver, device, ib, response)
   if ib.data.value ~= nil then
-    local level = math.floor((ib.data.value / 254.0 * 100) + 0.5)
+    local level = math.floor((ib.data.value / 200 * 100) + 0.5)
```
ctowns, line 179: Could you define these magic numbers at the top of the file?

```
-- Per matter spec, the pump level is in steps of 0.5% and the max level value is 200. Anything above is considered 100%
local MAX_PUMP_ATTR_LEVEL = 200
local MAX_LEVEL = 100
```
And then change this to something like
```suggestion
    local level = math.floor((ib.data.value / MAX_PUMP_ATTR_LEVEL * MAX_LEVEL) + 0.5)
    level = math.max(level, MAX_LEVEL)
```

hcarter-775: done :+1:

drivers/SmartThings/matter-pump/src/init.lua
```diff
@@ -176,7 +180,8 @@ end
 
 local function level_attr_handler(driver, device, ib, response)
   if ib.data.value ~= nil then
-    local level = math.floor((ib.data.value / 254.0 * 100) + 0.5)
+    local level = math.floor((ib.data.value / MAX_PUMP_ATTR_LEVEL * MAX_CAP_SWITCH_LEVEL) + 0.5)
+    level = math.max(level, MAX_CAP_SWITCH_LEVEL)
```
ctowns, line 184: Whoops, sorry this should be math.min :)

hcarter-775: of course, duh! good catch

hcarter-775:
tested with a partner pump device and things are working smoothly

ctowns (approved):

PR #1482, opened 2024-06-27
Remove history of the Supported list

HunsupJung:
Currently, The supportedXXX of mode and temperature level capability is updated to history. but It should be modified not to be updated.


ctowns (approved):

ctowns:
@HunsupJung unit tests were failing on this PR before it was merged. In the future please do not merge changes if the unit tests are failing, thank you!

PR #1483, opened 2024-06-27
Matter Pump: add missing embedded cluster include

ctowns:
This was accidentally missed in the initial commit. This adds the include for the embedded cluster defs in the situation that the hub is running older lua libs (api version < 10)


ctowns:
Tested this and confirmed all values are working on a hub with 53 FW running v52 api 9 lua libs (which don't include the pump cluster definitions).

hcarter-775 (approved):

PR #1490, opened 2024-07-02
HCS5008 - Remove AudioNotification Capability for samsung-audio driver

satyajitanand:
https://smartthings.atlassian.net/browse/HCS-5008: PM wants to exclude audioNotification capability.

Tested this locally. Now, Routine is not allowing options to play message on speaker. Hope this is the desired result as per HCS-5008.
Kindly review.


syamsidhardh (approved):
Changes for Removing the Audio notification

drivers/SmartThings/samsung-audio/src/init.lua
```diff
@@ -213,11 +213,6 @@ local samsungAudio = Driver("samsung-audio", {
       [capabilities.audioVolume.commands.volumeDown.NAME] = handlers.handle_volume_down,
       [capabilities.audioVolume.commands.setVolume.NAME] = handlers.handle_set_volume,
     },
-    [capabilities.audioNotification.ID] = {
-      [capabilities.audioNotification.commands.playTrack.NAME] = handlers.handle_audio_notification,
-      [capabilities.audioNotification.commands.playTrackAndResume.NAME] = handlers.handle_audio_notification,
-      [capabilities.audioNotification.commands.playTrackAndRestore.NAME] = handlers.handle_audio_notification,
```
raj-1219, line 219: We can remove the corresponding handler functions as well as they are no more needed. Do we want to keep them?

satyajitanand: In my opinion, no harm in keeping this if by chance we need handler in future.

satyajitanand:
@greens Hey Steven. Can you pls review.

inasail (approved):
Looks good to me.

raj-1219 (approved):
Thanks for the changes. Looks good.

cjswedes (commented):
We have avoided doing this to profiles, because doing so will automatically remove the capability from every device using this profile, which is not a very good user experience. A better user experience is for a user to try to do an audio notification command and then notice that after it doesnt work that the capability is removed; this also allows any device that is currently functioning with https for audio notifications to continue functioning.

So I think a better approach would be to add a new profile without audio notifications and then modify the error handling for the audio notification handler's [play streaming uri command](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/b50e38710201988589254b36de8f9cb231d095df/drivers/SmartThings/samsung-audio/src/handlers.lua#L144) to update the devices profile to the new profile which will stop it from doing a

greens:
@cjswedes that's actually a way better plan.

Gimmename1004:
@cjswedes - This is the data that the cloud team analyzed `http` used cases.
https://smartthings.atlassian.net/wiki/spaces/AU/pages/3274802589/Audio+notification+HTTP

> All speakers making a HTTP request are using [samsung-audio](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/tree/main/drivers/SmartThings/samsung-audio) edge driver.
> HTTP 호출 중인 스피커는 모두 동일한 [samsung-audio](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/tree/main/drivers/SmartThings/samsung-audio) edge driver를 사용 중에 있음

And in case of `samsung-audio` edge driver, it replaces `https` with `http` when sending the audio uri to all speakers that use this edge driver.
`samsung-audio` profile is only used by `samsung-audio` edge driver so I don't think that it will affect to other edge drivers. If I'm wrong, please correct me.

The concern case where it updates profile only when error happen is that,
if the user creates automation using `samsung-audio` speaker, the user can see audioNotification capability on that device.
The automation will be executed later and then, the profile will be updated with new one (without audioNotification) and the user cannot hear the notification. That will make the user to confuse.

I think that it would be better to update a profile before user tries to use audioNotification capability instead of error handling.
Still it would be possible to replace current `samsung-audio` profile with new `samsung-audio-notification-not-support` profile when driver is updated.

@cjswedes , @greens - Could you please share your opinion how to update `samsung-audio` profile not to support `audioNotification` capability ?

CC : @inasail

sy39ju:
@cjswedes - This is the data that the cloud team analyzed `http` used cases.
https://smartthings.atlassian.net/wiki/spaces/AU/pages/3274802589/Audio+notification+HTTP

> Korean : HTTP 호출 중인 스피커는 모두 동일한 [samsung-audio](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/tree/main/drivers/SmartThings/samsung-audio) edge driver를 사용 중에 있음
> All speakers making a HTTP request are using [samsung-audio](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/tree/main/drivers/SmartThings/samsung-audio) edge driver.

And in case of `samsung-audio` edge driver, it replaces `https` with `http` when sending the audio uri to all speakers that use this edge driver.
`samsung-audio` profile is only used by `samsung-audio` edge driver so I don't think that it will affect to other edge drivers. If I'm wrong, please correct me.

The concern case where it updates profile only when error happen is that,
if the user creates automation using `samsung-audio` speaker, the user can see audioNotification capability on that device.
The automation will be executed later and then, the profile will be updated with new one (without audioNotification) and the user cannot hear the notification. That will make the user to confuse.

I think that it would be better to update a profile before user tries to use audioNotification capability instead of error handling.
Still it would be possible to replace current `samsung-audio` profile with new `samsung-audio-notification-not-support` profile when driver is updated.

@cjswedes , @greens - Could you please share your opinion how to update `samsung-audio` profile not to support `audioNotification` capability ?

CC : @inasail

inasail:
@cjswedes Thank for your opinion. 
@sy39ju Thanks for detailed explanation.

All the speakers(16,000) that using 'samsung-audio' driver can't use https because certificate was already expired.
Among them, only 20 devices are using 'audioNotificaton' functionality for 10 days.(6/7~6/17)
Security team ordered not to use http anymore because of security issue.
So PM deviced not to use this functionality without notice to the user.
If customer make a complain about this, then CS team will handle it.

So, I think we don't need to notify about the removing this capability to user.

cjswedes:
> All the speakers(16,000) that using 'samsung-audio' driver can't use https because certificate was already expired.
Among them, only 20 devices are using 'audioNotificaton' functionality for 10 days.(6/7~6/17)
Security team ordered not to use http anymore because of security issue.
So PM decided not to use this functionality without notice to the users.
If customer make a complain about this, then CS team will handle it.

Thank you for the clarity here @inasail. Given this I am fine with the change as is. Refusing http connections on the server side will also work to just stop this functionality, and IMO we could just do that as soon as possible, and then get the profile updated over the next couple weeks with the normal release cycle.

greens (commented):
LGTM, @cjswedes, Alissa will be out on vacation for a while, but if you want to wait for @varzac 's go-ahead that's alright.

lelandblue (approved):

cjswedes (approved):

PR #1491, opened 2024-07-02
CHAD-12675: Handle Zigbee Min, Max, and Measured Value temperatures

wkhenon:
This pull request updates the base device drivers to handle both min and max measured value attributes of the Temperature Measurement capability

### Developer Testing

- Unit Tests ✅

### DIFF
--- a/drivers/SmartThings/zigbee-thermostat/src/init.lua
+++ b/drivers/SmartThings/zigbee-thermostat/src/init.lua
@@ -13,16 +13,17 @@
 -- limitations under the License.
 
 -- Zigbee Driver utilities
-local ZigbeeDriver      = require "st.zigbee"
-local device_management = require "st.zigbee.device_management"
-local defaults          = require "st.zigbee.defaults"
-local utils             = require "st.utils"
+local ZigbeeDriver          = require "st.zigbee"
+local device_management     = require "st.zigbee.device_management"
+local defaults              = require "st.zigbee.defaults"
+local utils                 = require "st.utils"
 
 -- Zigbee Spec Utils
 local clusters                      = require "st.zigbee.zcl.clusters"
 local PowerConfiguration            = clusters.PowerConfiguration
 local Thermostat                    = clusters.Thermostat
 local FanControl                    = clusters.FanControl
+local TemperatureMeasurement        = clusters.TemperatureMeasurement
 
 local FanMode = FanControl.attributes.FanMode
 local FanModeSequence           = FanControl.attributes.FanModeSequence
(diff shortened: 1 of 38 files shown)


greens (commented):
Only the min/max handling needs to be added to the base drivers. Temperature report handling should be handled by the existing, already-defined defaults in the lua libraries.

Additionally, we probably want to add some querying of the min/max values somewhere. They would typically go in the `added` handler if that already exists. If not, add one.

drivers/SmartThings/zigbee-button/profiles/button-profile.yml
```diff
@@ -6,6 +6,8 @@ components:
     version: 1
   - id: battery
     version: 1
+  - id: temperatureMeasurement
+    version: 1
```
greens, line 10: So we don't want to add this, because, presumably, the devices already using this profile do not support temperature measurement. Instead, you can change the test to use the `one-button-temp-battery` profile that already exists.

drivers/SmartThings/zigbee-thermostat/src/init.lua
```diff
@@ -208,7 +238,9 @@ local do_refresh = function(self, device)
     FanControl.attributes.FanModeSequence,
     FanControl.attributes.FanMode,
     PowerConfiguration.attributes.BatteryVoltage,
-    PowerConfiguration.attributes.BatteryAlarmState
+    PowerConfiguration.attributes.BatteryAlarmState,
+    TemperatureMeasurement.attributes.MinMeasuredValue,
+    TemperatureMeasurement.attributes.MaxMeasuredValue
```
greens, line 243: These two probably should only be read on Add, since we don't expect them to change. This refresh function is already quite chatty.

In the future, we probably want to do the same with ControlSequenceOfOperation and FanModeSequence, but not in this PR.

drivers/SmartThings/zigbee-button/profiles/button-profile.yml
wkhenon: 30ee164ddb4d503fa23b3da7c39ff6fdcb8cd307: removed this change and changed the test to use `one-button-temp-battery`

drivers/SmartThings/zigbee-thermostat/src/init.lua
wkhenon: 5d905d794465a528125c7304cb9454af67f29243: moved the two to the `device_added` and updated the tests

greens (approved):

greens:
great work, Khenon

PR #1492, opened 2024-07-02
Matter Lock: handle COTA even if profile doesn't have lock codes

ctowns:
Fixes regression introduced by: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1413

### DIFF
--- a/drivers/SmartThings/matter-lock/src/init.lua
+++ b/drivers/SmartThings/matter-lock/src/init.lua
@@ -112,17 +112,19 @@ end
 
 local function require_remote_pin_handler(driver, device, ib, response)
   if ib.data.value then
-    --Process after all other info blocks have been dispatched to ensure MaxPINCodeLength has been processed
-    device.thread:call_with_delay(0, function(t)
-      generate_cota_cred_for_device(device)
-      -- delay needed to allow test to override the random credential data
+    if device:get_field(lock_utils.COTA_CRED) ~= nil then
+      --Process after all other info blocks have been dispatched to ensure MaxPINCodeLength has been processed
       device.thread:call_with_delay(0, function(t)
-        -- Attempt to set cota credential at the lowest index
-        set_cota_credential(device, INITIAL_COTA_INDEX)
+        generate_cota_cred_for_device(device)
+        -- delay needed to allow test to override the random credential data
+        device.thread:call_with_delay(0, function(t)
+          -- Attempt to set cota credential at the lowest index
+          set_cota_credential(device, INITIAL_COTA_INDEX)
+        end)
       end)
-    end)
-  else
-    device:set_field(lock_utils.COTA_CRED, false, {persist = true})
+    else
+      device:set_field(lock_utils.COTA_CRED, false, {persist = true})
+    end
   end
 end
 
@@ -519,6 +521,15 @@ end
 local function device_init(driver, device)
   device:set_component_to_endpoint_fn(component_to_endpoint)
   device:subscribe()
+
+  -- Device may require pin for remote operation if it supports COTA and PIN features.
+  local eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.CREDENTIALSOTA | DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
+  if #eps == 0 then
+    device.log.debug("Device will not require PIN for remote operation")
+    device:set_field(lock_utils.COTA_CRED, false, {persist = true})
+  else
+    device:send(DoorLock.attributes.RequirePINforRemoteOperation:read(device, eps[1]))
+  end
  end
 
 local function device_added(driver, device)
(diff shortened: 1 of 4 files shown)


drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -519,6 +521,15 @@ end
...
   device:subscribe()
+
+  -- Device may require pin for remote operation if it supports COTA and PIN features.
+  local eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.CREDENTIALSOTA | DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
+  if #eps == 0 then
+    device.log.debug("Device will not require PIN for remote operation")
+    device:set_field(lock_utils.COTA_CRED, false, {persist = true})
+  else
+    device:send(DoorLock.attributes.RequirePINforRemoteOperation:read(device, eps[1]))
+  end
```
Con-Tejus, line 532: Can we get format this repeat code into a local function and call it in `device_init` and `device_added`?

ctowns: The reason I kept it separate was so that I could merge the read with the other reads in `added` and then I could just send the single request in `init`. I can try to find a way to neatly return a read request if needed, but either way there will be a check to merge it with the existing read req in `added` or just send it as a single read in `init`

ctowns: Or I guess we could just send two requests

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -112,17 +112,19 @@ end
...
+        end)
       end)
-    end)
-  else
-    device:set_field(lock_utils.COTA_CRED, false, {persist = true})
+    else
+      device:set_field(lock_utils.COTA_CRED, false, {persist = true})
+    end
   end
 end
```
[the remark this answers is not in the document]
ctowns: Forgot to add this! Thanks Abdul, I created this function in [ed435f2](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1492/commits/ed435f2f609c41deedad1a6c2a8c4c35ca3d0325)

hcarter-775 (approved):

hcarter-775:
I'm sure you see this, but I think the unit tests will also need to be edited to support the new PIN request in the init function.

Con-Tejus (approved):

drivers/SmartThings/matter-lock/src/test/test_matter_lock_cota_no_lock_codes_profile.lua
```diff
@@ -0,0 +1,331 @@
...
+-- Mock out globals
+local test = require "integration_test"
+local capabilities = require "st.capabilities"
+test.add_package_capability("lockAlarm.yml")
+local t_utils = require "integration_test.utils"
+local clusters = require "st.matter.clusters"
+local DoorLock = clusters.DoorLock
+
+local mock_device_record = {
+  profile = t_utils.get_profile_definition("lock-lockalarm-nobattery.yml"),
```
cjswedes, line 37: Since this is the same profile as used in `test_aqara_matter_lock`, can we swap in `lock-nobattery-notamper` and have the tests still pass?

cjswedes (approved):

drivers/SmartThings/matter-lock/src/test/test_matter_lock_cota_no_lock_codes_profile.lua
cjswedes: I think I am getting at the same thing as others wrt the unit tests, forgot to refresh before my review.

nickolas-deboom (approved):

drivers/SmartThings/matter-lock/src/test/test_matter_lock_cota_no_lock_codes_profile.lua
ctowns: Thanks Carter! I made that change to the unit test in [7f31271](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1492/commits/7f3127131083e1d9bb99689e68f460451e517ee5), so now the test uses the `lock-nocodes-notamper` profile, which does not have the `lockCodes` capability, while the `base-lock` profile that does have the `lockCodes` capability is used in the complimentary `test_matter_lock_cota.lua` test file. So these two files should test COTA for both device that have and devices that don't have `lockCodes` in their profiles.

ctowns:
Tested with a Yale lock and with the Virtual Door Lock on the VDA with help from Jeff.
We were unable to test with the VDA Device with Lock Codes because we were unable to get the pin code to work in the VDA, but the Yale lock should cover this same scenario for when COTA credentials are needed.

ctowns:
This fix was addressed by this PR instead https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1498, which reverts the commits that caused the regression and then moves the Aqara lock functionality to a sub driver.

This PR will still be a good reference to use when when eventually make some of these changes to the base driver, but let's do that in a separate PR after more consideration.

Closing this PR.

PR #1494, opened 2024-07-03
[Aqara/Roller Shade] change the way windowShadeLevel capability updates

seojune79:


drivers/SmartThings/zigbee-window-treatment/src/aqara/roller-shade/init.lua
```diff
@@ -33,6 +33,8 @@ local function window_shade_open_cmd(driver, device, command)
   local initialized = device:get_latest_state("main", initializedStateWithGuide.ID,
     initializedStateWithGuide.initializedStateWithGuide.NAME) or 0
   if initialized == initializedStateWithGuide.initializedStateWithGuide.initialized.NAME then
+    aqara_utils.disable_update_shade_level(device)
+    device:emit_event(capabilities.windowShadeLevel.shadeLevel(100))
```
greens, line 37: There's no guarantee that the shade will actually go to 100, right? What if there are transmission errors?

drivers/SmartThings/zigbee-window-treatment/src/aqara/aqara_utils.lua
```diff
@@ -41,6 +49,7 @@ local function emit_shade_event_by_state(device, value)
   -- update state ui
   if state == SHADE_STATE_STOP or state == 0x04 then
     -- read shade position to update the UI
+    enable_update_shade_level(device)
```
greens, line 52: What happens if the final level event arrives _before_ the device reports it has stopped? Will that level report never be sent?

drivers/SmartThings/zigbee-window-treatment/src/aqara/roller-shade/init.lua
seojune79: Okay. I'll fix it.

drivers/SmartThings/zigbee-window-treatment/src/aqara/aqara_utils.lua
seojune79: The report is ignored. However, when the final level event arrives, it requests the last level again with the READ command.

https://github.com/seojune79/SmartThingsEdgeDrivers/blob/3b6c1109489b8b887f39ac3ab17e50ef1f3b0959/drivers/SmartThings/zigbee-window-treatment/src/aqara/aqara_utils.lua#L51-L53

drivers/SmartThings/zigbee-window-treatment/src/aqara/aqara_utils.lua
```diff
@@ -15,13 +15,23 @@ local SHADE_STATE_STOP = 0
...
 local aqara_utils = {}
 
+local function enable_update_shade_level(device)
+  device:set_field(UPDATE_SHADE_LEVEL, true, { persist = true })
+end
+
+local function disable_update_shade_level(device)
+  device:set_field(UPDATE_SHADE_LEVEL, false, { persist = true })
+end
+
```
greens, line 32: This value doesn't need to be persistent. The useful timeline of this flag is from when the first command is issued until when the stop report is received. Having this value persist during a hub reboot would probably lead to unintended side-effects, especially if the "stop" report is missed.

greens (commented):
It would be nice to see some tests verifying that your flag is working as intended. i.e. a test that shows that reports are ignored after a level command is sent

drivers/SmartThings/zigbee-window-treatment/src/aqara/aqara_utils.lua
seojune79: I see what you mean.

greens (approved):

PR #1495, opened 2024-07-03
Revert matter lock changes and move aqara to sub driver

ctowns:


Con-Tejus:
Sorry I don't know if I missed it during out call, but would we want to also include the fix in recovering already effected devices by adding the read operation for `RequirePINForRemoteOperation` within the `device_init` call for the base matter lock driver?

ctowns:
> Sorry I don't know if I missed it during out call, but would we want to also include the fix in recovering already effected devices by adding the read operation for `RequirePINForRemoteOperation` within the `device_init` call for the base matter lock driver?

Just added this! Let's start testing with Jeff and the VDA so we can verify the changes still work as expected for already joined devices.

drivers/SmartThings/matter-lock/profiles/lock-lockalarm.yml
samadDotDev, line 1: Unintentional permissions change?

ctowns: Fixed!

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -21,6 +21,7 @@ local PowerSource = clusters.PowerSource
 local capabilities = require "st.capabilities"
 local im = require "st.matter.interaction_model"
 local lock_utils = require "lock_utils"
+local COTA_READ_INITIALIZED = "__cota_read_initialized"
```
samadDotDev, line 24: nit: Could we move it to `lock_utils.lua` just to be consistent with other keys, even though temporarily allocated?

samadDotDev (approved):

Con-Tejus (approved):

drivers/SmartThings/matter-lock/src/init.lua
ctowns: Fixed in [7c2a912](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1495/commits/7c2a912bb82a0df160e17df5781c62fc3743a4b4)!

PR #1499, opened 2024-07-04
adding maestro thermostat SMT402AD model

gcavalcantistelpro:
Replacing the stelpro maestro thermostat model to SMT402AD


greens (changes requested):
If the `MaestroStat` model does not (and has never) existed, then the tests in `test_stelpro_thermostat.lua` should be updated to reflect that.

But if it _has_ existed, this should be an _addition_ of a fingerprint, not an amendment.

lelandblue:
Hey @Brianj94 and @gcavalcantistelpro . Lets go the safe route here. 

Could we leave the existing code in, and just only add new code? For example, lets add a whole fringerprint for the  SMT402AD device and add lines in the subdriver for that device verses removing anything that is existing.  

Please review and adjust, and thank you!

drivers/SmartThings/zigbee-thermostat/src/stelpro/init.lua
```diff
@@ -1,4 +1,4 @@
--- Copyright 2022 SmartThings
+-- Copyright 2024 SmartThings
```
greens, line 1: date

drivers/SmartThings/zigbee-thermostat/src/stelpro/stelpro_maestrostat.lua
```diff
@@ -1,4 +1,4 @@
--- Copyright 2022 SmartThings
+-- Copyright 2024 SmartThings
```
greens, line 1: date

drivers/SmartThings/zigbee-thermostat/src/test/test_stelpro_thermostat.lua
```diff
@@ -1,4 +1,4 @@
--- Copyright 2022 SmartThings
+-- Copyright 2024 SmartThings
```
greens, line 1: date

greens (approved):

PR #1500, opened 2024-07-05
Fix an issue that temperature level can't be controlled

HunsupJung:
Fix an issue that driver can't control temperature level.
this issue can occur according to the endpoint order in which the handler is called.

IOTE-4259

### DIFF
--- a/drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
+++ b/drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
@@ -132,7 +132,7 @@ local function temperature_setpoint_attr_handler(driver, device, ib, response)
   if not supports_temperature_number_endpoint(device, ib.endpoint_id) then
     return
   end
-  device.log.info_with({ hub_logs = true }, string.format("temperature_setpoint_attr_handler: %d", ib.data.value))
+  device.log.info(string.format("temperature_setpoint_attr_handler: %d", ib.data.value))
   local min_field = string.format("%s-%d", setpoint_limit_device_field.MIN_TEMP, ib.endpoint_id)
   local max_field = string.format("%s-%d", setpoint_limit_device_field.MAX_TEMP, ib.endpoint_id)
   local min = device:get_field(min_field) or 0
@@ -171,7 +171,7 @@ local function selected_temperature_level_attr_handler(driver, device, ib, respo
   end
   local supportedTemperatureLevels = supportedTemperatureLevelsMap[ib.endpoint_id]
   for i, tempLevel in ipairs(supportedTemperatureLevels) do
-    device.log.info_with({ hub_logs = true }, string.format("selected_temperature_level_attr_handler: %d, %s", i, tempLevel))
+    device.log.info(string.format("selected_temperature_level_attr_handler: %d, %s", i, tempLevel))
     if i - 1 == temperatureLevel then
       device:emit_event_for_endpoint(ib.endpoint_id, capabilities.temperatureLevel.temperatureLevel(tempLevel))
       break
@@ -186,12 +186,12 @@ local function supported_temperature_levels_attr_handler(driver, device, ib, res
   local supportedTemperatureLevelsMap = device:get_field(SUPPORTED_TEMPERATURE_LEVELS_MAP) or {}
   local supportedTemperatureLevels = {}
   for _, tempLevel in ipairs(ib.data.elements) do
-    log.info_with({ hub_logs = true }, string.format("supported_temperature_levels_attr_handler: %s", tempLevel.value))
+    device.log.info(string.format("supported_temperature_levels_attr_handler: %s", tempLevel.value))
     table.insert(supportedTemperatureLevels, tempLevel.value)
   end
   for ep = 1, ib.endpoint_id - 1 do
     if not supportedTemperatureLevelsMap[ep] then
-      log.info_with({ hub_logs = true }, string.format("supportedTemperatureLevelsMap[%d] is nil", ep))
+      device.log.info(string.format("supportedTemperatureLevelsMap[%d] is nil", ep))
       supportedTemperatureLevelsMap[ep] = {"Nothing"}
     end
   end
@@ -255,7 +255,7 @@ end
 
 -- Capability Handlers --
 local function handle_refrigerator_tcc_mode(driver, device, cmd)
-  device.log.info_with({ hub_logs = true }, string.format("handle_refrigerator_tcc_mode mode: %s", cmd.args.mode))
+  device.log.info(string.format("handle_refrigerator_tcc_mode mode: %s", cmd.args.mode))
   local ep = component_to_endpoint(device, cmd.component)
   local supportedRefrigeratorTccModesMap = device:get_field(SUPPORTED_REFRIGERATOR_TCC_MODES_MAP)
   local supportedRefrigeratorTccModes = supportedRefrigeratorTccModesMap[ep] or {}
@@ -272,7 +272,7 @@ local function handle_temperature_setpoint(driver, device, cmd)
   if not supports_temperature_number_endpoint(device, ep) then
     return
   end
-  device.log.info_with({ hub_logs = true }, string.format("handle_temperature_setpoint: %s", cmd.args.setpoint))
+  device.log.info(string.format("handle_temperature_setpoint: %s", cmd.args.setpoint))
 
   local value = cmd.args.setpoint
   local _, temp_setpoint = device:get_latest_state(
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
```diff
@@ -122,9 +150,7 @@ end
 
 local function setpoint_limit_handler(limit_field)
   return function(driver, device, ib, response)
-    local tn_eps = embedded_cluster_utils.get_endpoints(device,clusters.TemperatureControl.ID, {feature_bitmap = clusters.TemperatureControl.types.Feature.TEMPERATURE_NUMBER})
-    if #tn_eps == 0 then
-      device.log.warn_with({ hub_logs = true }, string.format("Device does not support TEMPERATURE_NUMBER feature"))
+    if not supports_temperature_number_endpoint(device, ib.endpoint_id) then
```
ctowns, line 153: I don't think this function is needed, because the `setpoint_limit_handler` function would only be called in the case that the endpoint supports the attribute `MinTemperature` or `MaxTemperature`:
```
    attr = {
      [clusters.TemperatureControl.ID] = {
        [clusters.TemperatureControl.attributes.TemperatureSetpoint.ID] = temperature_setpoint_attr_handler,
        [clusters.TemperatureControl.attributes.MinTemperature.ID] = setpoint_limit_handler(setpoint_limit_device_field.MIN_TEMP),
        [clusters.TemperatureControl.attributes.MaxTemperature.ID] = setpoint_limit_handler(setpoint_limit_device_field.MAX_TEMP),
```

So I think we can assume that value `ib.endpoint` does have temperatureNumber suppport.

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
```diff
@@ -135,17 +161,17 @@ local function setpoint_limit_handler(limit_field)
 end
 
 local function selected_temperature_level_attr_handler(driver, device, ib, response)
-  local tl_eps = embedded_cluster_utils.get_endpoints(device,clusters.TemperatureControl.ID, {feature_bitmap = clusters.TemperatureControl.types.Feature.TEMPERATURE_LEVEL})
-  if #tl_eps == 0 then
-    device.log.warn_with({ hub_logs = true }, string.format("Device does not support TEMPERATURE_LEVEL feature"))
+  if not supports_temperature_level_endpoint(device, ib.endpoint_id) then
```
ctowns, line 164: Similar to the comment above, I don't think we would need to check if the endpoint supports temperatureLevel because this handler would only be called for endpoints that send us an interaction for the `SelectedTemperatureLevel` handler:
```
    attr = {
      [clusters.TemperatureControl.ID] = {
        [clusters.TemperatureControl.attributes.TemperatureSetpoint.ID] = temperature_setpoint_attr_handler,
        [clusters.TemperatureControl.attributes.MinTemperature.ID] = setpoint_limit_handler(setpoint_limit_device_field.MIN_TEMP),
        [clusters.TemperatureControl.attributes.MaxTemperature.ID] = setpoint_limit_handler(setpoint_limit_device_field.MAX_TEMP),
        [clusters.TemperatureControl.attributes.SelectedTemperatureLevel.ID] = selected_temperature_level_attr_handler, -- <- here
```

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
```diff
@@ -96,15 +96,43 @@ local function is_matter_refrigerator(opts, driver, device)
   return false
 end
 
-local function temperature_setpoint_attr_handler(driver, device, ib, response)
-  local tn_eps = embedded_cluster_utils.get_endpoints(device, clusters.TemperatureControl.ID, {feature_bitmap = clusters.TemperatureControl.types.Feature.TEMPERATURE_NUMBER})
+local function supports_temperature_level_endpoint(device, endpoint)
```
ctowns, line 99: I don't think the `supports_temperature_level_endpoint` and `supports_temperature_number_endpoint` functions are needed, see my explanation in the comments below

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
```diff
@@ -96,15 +96,43 @@ local function is_matter_refrigerator(opts, driver, device)
...
+    return false
   end
-  device.log.info_with({ hub_logs = true },
-    string.format("temperature_setpoint_attr_handler: %d", ib.data.value))
+  for i, eps in ipairs(tn_eps) do
+    if eps == endpoint then
+      return true
+    end
+  end
+  device.log.warn_with({ hub_logs = true }, string.format("Endpoint(%d) does not support TEMPERATURE_NUMBER feature", endpoint))
```
ctowns, line 127: There are a lot of cases where we are logging to the hubLogs throughout this file. Could some of these be removed or isolated to the driver logs (by using just `device.log.info()` instead of `device.log.info_with({hub_logs = true}, ..)`) so that we do not clutter the hub logs with unnecessary information?

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
```diff
@@ -239,21 +268,18 @@ local function handle_refrigerator_tcc_mode(driver, device, cmd)
 end
 
 local function handle_temperature_setpoint(driver, device, cmd)
-  local tn_eps = embedded_cluster_utils.get_endpoints(device,clusters.TemperatureControl.ID, {feature_bitmap = clusters.TemperatureControl.types.Feature.TEMPERATURE_NUMBER})
-  if #tn_eps == 0 then
-    device.log.warn_with({ hub_logs = true }, string.format("Device does not support TEMPERATURE_NUMBER feature"))
+  local ep = component_to_endpoint(device, cmd.component)
+  if not supports_temperature_number_endpoint(device, ep) then
```
ctowns, line 272: You can probably remove this also. There shouldn't be a situation where the app is sending a temperature setpoint command for an endpoint that doesn't support the temperature number. The `component_to_endpoint` function will make sure that the component on the app is mapped to the correct endpoint on the device so that if the component on the app (the button the user presses) supports temperatureNumber, then the endpoint that maps to that component should also support temperatureNumber if we have done the endpoint to component mapping correctly

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
ctowns: I see that these checks are used throughout the file, but I don't think they are needed. Have you seen many instances where an attribute handler has been called for an endpoint that doesn't support the correct feature (temperatureNumber or temperatureLevel). That seems like more of an `endpoint_to_component` map issue, but I think that should be addressed with the `COMPONENT_TO_ENDPOINT` field that is used in this driver to map the components to the correct endpoints.

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
HunsupJung: I'm sorry for the late reply.
I tested through the Virtual device app and confirmed that the `Temperature Setpoint` attribute is called even if `Temperature Control` does not support `TN` feature or `Supported Temperature Levels` or `Selected Temperature Levels` are called even if `TL` feature is not supported. So I added this code.

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
HunsupJung: Okay I'll update it :)

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
HunsupJung: We can't know if a manufacturer actually supports Matter correctly. I think we should be prepared even when they are not applying properly. We confirmed such case with Virtual device app.

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
HunsupJung: @ctowns 
I checked that handler for unsupported feature(TN or TL) is called in Virtual Matter App.  I think the server can call the handler for features that are not supported. so we need to prevent this case.

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
HunsupJung: It is the same with [the comment](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1500/files#r1694798795) above

drivers/SmartThings/matter-appliance/src/matter-refrigerator/init.lua
HunsupJung: It is the same with the following [comment](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1500/files#r1694798795)

ctowns (approved):

PR #1501, opened 2024-07-08
Update default profile for smoke-co-alarm device type

HunsupJung:
If the default profile is not super-set with all capabilities and profile is changed by the try_update_metadata function, the driver cannot read the value of the capability that was not in the default profile. 
It seems that these capabilities are not loaded, so the driver cannot find it.


ctowns:
Hey @HunsupJung ! I think this PR can solve the root cause of this issue instead: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1509

HunsupJung:
The issue can be solved with #1509

PR #1508, opened 2024-07-09
Add zigbee-button for Vimar xx591 (Wall-mounted Remote Control)

sciaccom-vimar:
Add support for:

- Vimar Remote Control (xx591, IoT Connected Device Mechanism)

### DIFF
--- a/drivers/SmartThings/zigbee-button/profiles/two-buttons-no-fw-update.yml
+++ b/drivers/SmartThings/zigbee-button/profiles/two-buttons-no-fw-update.yml
@@ -1,0 +1,22 @@
+name: two-buttons-no-fw-update
+components:
+  - id: main
+    capabilities:
+      - id: button
+        version: 1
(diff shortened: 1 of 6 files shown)


wkhenon (approved):

drivers/SmartThings/zigbee-button/profiles/two-buttons-no-fw-update.yml
greens, line 1: the FirmwareUpdate capability is handled completely by us and we'd prefer to have it for all of our zigbee devices

drivers/SmartThings/zigbee-button/src/test/test_vimar_button.lua
```diff
@@ -0,0 +1,215 @@
+-- Copyright 2022 SmartThings
```
greens, line 1: date

PR #1509, opened 2024-07-09
Matter Sensor: resubscribe on profile switch

ctowns:
Resubscribe whenever there is a profile switch so that new attributes can be subscribed to if the new profile includes new capabilities.


nickolas-deboom (approved):

HunsupJung (approved):
I checked your PR with virtual Smoke/CO Alarm app and there's no issue. 👍

PR #1510, opened 2024-07-10
Include embedded config for profiles using temperatureAlarm

hcarter-775:
The temperatureAlarm capability supports four enum values: freeze, cleared, heat, and rateOfRise. Currently, all driver profiles that support this capability exclude rateOfRise since it is unused and will come up in automations if not excluded, and it is the same case for these dishwasher profiles. Since this enum is not used in the dishwasher subdriver, it should be excluded just as is done in other driver profiles.

### DIFF
--- a/drivers/SmartThings/matter-appliance/profiles/dishwasher-tl.yml
+++ b/drivers/SmartThings/matter-appliance/profiles/dishwasher-tl.yml
@@ -14,6 +14,13 @@ components:
     version: 1
   - id: temperatureAlarm
     version: 1
+    config:
+      values:
+        - key: "temperatureAlarm.value"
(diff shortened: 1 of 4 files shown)


tpmanley (approved):

ctowns (approved):

lelandblue:
@hcarter-775 Not urgent, just wanted to check in. Is this PR still needed? I dont see open work for it. Please close if this scope has changed.

hcarter-775:
Closing due to being incompatible with [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1458) per the testing of @HunsupJung

PR #1511, opened 2024-07-11
fix an issue of Fan device type sending command

pInksenberg:
Issue Description: We onboarding Yeelight Fan(bridged device) device to ST App, but we are unable to control device by sending fan mode or speed.

Problem Analysis: Upon reviewing the logs, it was found that the command send to a wrong endpoint (endpoint 0 for the yeelight bridge instead of endpoint 68 for the yeelight fan). It was caused by the local function  "component_to_endpoint" in matter-thermostat/init.lua 158. We revised the edge driver by adding an "elseif" statement for returning Fan device type related endpoint. (As per Matter Device Library Spec, there are 3 device type in HVAC, including Thermostat, Fan and Air Purifier. However, It should be noted that the Fan device is currently absent from the function "component_to_endpoint".)

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -225,8 +225,8 @@ end
 local function component_to_endpoint(device, component_name)
   -- Use the find_default_endpoint function to return the first endpoint that
   -- supports a given cluster.
-  if device:supports_capability(capabilities.airPurifierFanMode) then
-    -- Fan Control is mandatory for the Air Purifier device type
+  if device:supports_capability(capabilities.airPurifierFanMode) or device:supports_capability(capabilities.airConditionerFanMode) then
+    -- Fan Control is mandatory for the Air Purifier device type and Air Conditioner device type
     return find_default_endpoint(device, clusters.FanControl.ID)
   else
     -- Thermostat is mandatory for Thermostat and Room AC device type


wangjinliang12345:
(endpoint 0 for the yeelight bridge instead of endpoint 68 for the yeelight fan) -> (endpoint 1 for the yeelight bridge instead of endpoint 68 for the yeelight fan)

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -161,6 +161,9 @@ local function component_to_endpoint(device, component_name)
   if device:supports_capability(capabilities.airPurifierFanMode) then
     -- Fan Control is mandatory for the Air Purifier device type
     return find_default_endpoint(device, clusters.FanControl.ID)
+  elseif device:supports_capability(capabilities.airConditionerFanMode) then
```
ctowns, line 164: could we combine this conditional with the one above for simplification?
```
if device:supports_capability(capabilities.airPurifierFanMode) or device:supports_capability(capabilities.airConditionerFanMode) then
    -- Fan Control is mandatory for the Air Purifier device type
    return find_default_endpoint(device, clusters.FanControl.ID)
end
```

pInksenberg:
@ctowns could you plz help to review for the updated driver？

lelandblue:
Tagging @z-michel for awareness.

wangjinliang12345:
this is issues from Yeelight WWST

ctowns (approved):

hcarter-775 (approved):

ctowns:
@pInksenberg this has been approved and I think this is ready to merge. Can you confirm that this change has been tested? If it has been tested and is working, it can be merged.

pInksenberg:
@ctowns I attached the test result in https://smartthings.atlassian.net/browse/MTR-780

hcarter-775:
So can this change be merged? It looks like things are working as expected

PR #1517, opened 2024-07-12
Electrical Sensor Integration

hcarter-775:
A prototype for electrical sensor integration. This includes most of the functionality in a first pass form that will be permitted with electrical sensor.


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -501,9 +583,73 @@ local function occupancy_attr_handler(driver, device, ib, response)
   device:emit_event(ib.data.value == 0x01 and capabilities.motionSensor.motion.active() or capabilities.motionSensor.motion.inactive())
 end
 
+local function power_topology_handler(driver, device, ib, response)
```
ctowns, line 586: I believe this should just give us the value of the power topology for the given endpoint that this message was received on, so we should just be able to use the `ib.data` to get the topology type rather than having to check for feature maps on the endpoints.
```
device:set_field(POWER_TOPOLOGY_TYPE, ib.data)
```

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -68,6 +68,87 @@ local device_type_profile_map = {
...
+
+local function delete_poll_schedule(device)
+  local poll_timer = device:get_field(RECURRING_POLL_TIMER)
+  if poll_timer then
+    device.thread:cancel_timer(poll_timer)
+    device:set_field(RECURRING_POLL_TIMER, nil)
+  end
+end
+
+local function send_poll_report(device, LAST_REPORT_TIME, TOTAL_ENERGY)
```
ctowns, line 107: `LAST_REPORT_TIME` and `TOTAL_ENERGY` do not need to be parameters since they are keys on the device object and the only instances in which `send_poll_report` is called, it always uses the same values for these parameters.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -68,6 +68,87 @@ local device_type_profile_map = {
...
+local TOTAL_EXPORTED_ENERGY = "TOTAL_EXPORTED_ENERGY"
+local LATEST_EXPORTED_TIMESTAMP = "LATEST_EXPORTED_TIMESTAMP"
+local RECURRING_EXPORT_REPORT_POLL_TIMER = "RECURRING_EXPORT_REPORT_POLL_TIMER"
+
+local REPORT_SCHEDULE_INIT_CALLED = "REPORT_SCHEDULE_INIT_CALLED"
+local REPORT_POLL_TIMER_IS_SET = "REPORT_POLL_TIMER_IS_SET"
+local RECURRING_POLL_TIMER = "RECURRING_POLL_TIMER"
+local REPORT_TIMEOUT = "REPORT_TIMEOUT"
+local MINIMUM_ST_ENERGY_REPORT_INTERVAL = (15 * 60) -- 15 minutes, reported in seconds
+local SUBSCRIPTIONS_COMPLETE = "SUBSCRIPTIONS_COMPLETE"
```
ctowns, line 83: What is the purpose of this flag? I'm not sure it is needed, but if it is a comment explaining why would be helpful

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -501,9 +583,73 @@ local function occupancy_attr_handler(driver, device, ib, response)
...
+local function energy_report_handler_factory(is_cumulative_report)
+  return function(driver, device, ib, response)
+    if not device:get_field(REPORT_POLL_TIMER_IS_SET) then
+      local cumulative_eps = embedded_cluster_utils.get_endpoints(device, clusters.ElectricalEnergyMeasurement.ID,
+        {feature_bitmap = clusters.ElectricalEnergyMeasurement.types.ElectricalEnergyMeasurementFeature.CUMULATIVEENERGY}
+      )
+      if #cumulative_eps > 1 then
+        device:set_field(CUMULATIVE_ENERGY_IS_REPORTED, true)
+      end
+      if device:get_field(SUBSCRIPTIONS_COMPLETE) then -- with this, we know this must have been called after the subscriptions.
```
ctowns, line 619: I am unsure why we need to wait until after subscriptions are complete, could you help me understand this?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -501,9 +583,73 @@ local function occupancy_attr_handler(driver, device, ib, response)
...
+  local summed_energy_report = latest_energy_report + ib.data.elements["Energy"]
+  device:set_field(TOTAL_EXPORTED_ENERGY, summed_energy_report)
+  device:set_field(LATEST_EXPORTED_TIMESTAMP, ib.data.elements["EndTimestamp"])
+  device:emit_event(capabilities.energyMeter.energy({ value = ib.data.elements["Energy"], unit = "Wh" }))
+end
+
+local function energy_report_handler_factory(is_cumulative_report)
+  return function(driver, device, ib, response)
+    if not device:get_field(REPORT_POLL_TIMER_IS_SET) then
+      local cumulative_eps = embedded_cluster_utils.get_endpoints(device, clusters.ElectricalEnergyMeasurement.ID,
```
ctowns, line 613: I think `is_cumulative_report` should handle all of this. If it is true, then we know we got a cumulative report, otherwise we got a periodic report

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -501,9 +583,73 @@ local function occupancy_attr_handler(driver, device, ib, response)
...
+        {feature_bitmap = clusters.ElectricalEnergyMeasurement.types.ElectricalEnergyMeasurementFeature.CUMULATIVEENERGY}
+      )
+      if #cumulative_eps > 1 then
+        device:set_field(CUMULATIVE_ENERGY_IS_REPORTED, true)
+      end
+      if device:get_field(SUBSCRIPTIONS_COMPLETE) then -- with this, we know this must have been called after the subscriptions.
+        local reportInterval = ib.data.elements["EndTimestamp"] - ib.data.elements["StartTimestamp"]
```
ctowns, line 620: Just a heads up `startTimestamp` is omitted if the energy measurement is cumulative: https://librarian-regionals.smartthingsgdev.com/matter/Matter_1-3/appclusters.html#ref_EnergyMeasurementStartTimestamp

I think we can just use the 15 minute interval, since the device should send an update whenever the value changes since we have a subscription to this attribute, so it wouldn't delay sending the value.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: that's true! it doesn't get used anywhere either haha

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: So here I'm trying to check what the time interval is between energy events from the device, which would be this line `local reportInterval = ib.data.elements["EndTimestamp"] - ib.data.elements["StartTimestamp"].` The `SUBSCRIPTIONS_COMPLETE` gate would make sure this check is generally correct (not coming from the subscription request). 

Then I have the conditional `if reportInterval > MINIMUM_ST_ENERGY_REPORT_INTERVAL then` which checks whether the reporting interval is greater than 15 minutes, the minimum length that we're allowed to report with. And if it is >15 minutes, I want to use that longer interval for the poll report rather than 15 minutes.

The thing is, the matter spec doesn't make the device specify its report interval so I don't think we can just read that somewhere.

Last, I wanted to make sure this logic is only done once, so `REPORT_POLL_TIMER_IS_SET` would get set once we have this interval, and the logic would never be re-entered.

If there is any easier way to check whether we're seeing a subscription read versus a regular event read, I'm super open to suggestions!

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Hm, what would `ib.data` be returning from this def: [clusters.PowerTopology.ID]` = power_topology_handler,`?

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Related to another comment I left :+1:

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Yeah, the thing is I built this when I was under the false impression that we were implementing both imported and exported energy, and this was meant to generalize. I think it should probably be removed for now, but it seems like partners are keen on using the imported energy feature, and so this may be useful again soon.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: I understand that position, though sending the report at the time frame that is more accurate to the device would be better, no? It would also lower cloud costs that were mentioned as a contributing factor to the 15 minute minimum. For example, even a 17 minute timer vs 15 minute timer would bring down sending by 13%.

hcarter-775: And yes :sweat_smile: I realized the StartTimestamp is omitted while writing these responses to your reviews. As can be seen in a few places, this one is definitely a first draft. However, I don't think it would be all that difficult to take the difference between the first two event emissions and call that the time interval. A little more overhead than the current implementation, but not too bad.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -518,6 +664,7 @@ local function device_added(driver, device)
   device_init(driver, device)
 end
 
+
```
HunsupJung, line 667: Remove blank

drivers/SmartThings/matter-switch/src/init.lua
HunsupJung: Is there any case that the reporting interval for energy usage of the device may vary?
If the first report was made in an hour, and it occurs every 15 minutes from then on, the device sends energy usage every 15 minutes, but SmartThings will update it every hour. This reduces the accuracy of energy measurements.
Since the device may send an energy reporting at constant interval or an irregular interval, driver would be better to report energy at regular intervals.
And if there is no change in the value of previousTotalConsumptionWh, it would be nice to add a process that does not send an event.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -68,6 +68,87 @@ local device_type_profile_map = {
...
+  local previousTotalConsumptionWh = device:get_latest_state("main", capabilities.powerConsumptionReport.ID,
+    capabilities.powerConsumptionReport.powerConsumption.NAME)
+
+  local startTime = iso8061Timestamp(last_time)
+  local endTime = iso8061Timestamp(current_time - 1)
+
+  local deltaEnergyWh = 0.0
+  if previousTotalConsumptionWh and previousTotalConsumptionWh.energy then
+      deltaEnergyWh = math.max(latestTotalConsumptionWH - previousTotalConsumptionWh.energy, 0.0)
+  end
```
HunsupJung, line 124: if there is no change in the value of previousTotalConsumptionWh, it would be nice to add a process that does not send an event.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -68,6 +68,87 @@ local device_type_profile_map = {
...
+end
+
+local function delete_poll_schedule(device)
+  local poll_timer = device:get_field(RECURRING_POLL_TIMER)
+  if poll_timer then
+    device.thread:cancel_timer(poll_timer)
+    device:set_field(RECURRING_POLL_TIMER, nil)
+  end
+end
+
```
HunsupJung, line 106: shouldn't the key value where the timer is stored be received as a parameter?
No value is saved in RECURING_POLL_TIMER.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Either way, I don't think we actually need this handling. The way we are using the device type as of now, there is no reason to support power topology.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: In this case, we don't want this, since the event is being sent to ST Energy, and they would still want to know when no change occurs.

PR #1519, opened 2024-07-15
Set state_change to true when receive lockAlarm event

HunsupJung:
LockAlarm event can occur with same value. so it needs to set state_change to true.


hcarter-775 (approved):

ctowns (approved):

lelandblue:
Hello @HunsupJung  Please see my email for additional information that I believe needs to be considered before this PR is merged. Thank you. 

cc: @z-michel

PR #1520, opened 2024-07-15
New Device (Matter Thermostat) Eve Thermostat

lelandblue:
This is a new Matter WWST request for a Eve Thermostat. The product ID for this device is "79 (0x4f)" and has a matter device type id of "769 (0x301)".

### DIFF
--- a/drivers/SmartThings/matter-thermostat/fingerprints.yml
+++ b/drivers/SmartThings/matter-thermostat/fingerprints.yml
@@ -10,6 +10,13 @@ matterManufacturer:
     vendorId: 0x1396
     productId: 0x0043A
     deviceProfileName: air-purifier-hepa
+#Eve
+  - id: "4874/79"
+    deviceLabel: Eve Thermo
+    vendorId: 0x130A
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-thermostat/fingerprints.yml
```diff
@@ -10,6 +10,13 @@ matterManufacturer:
     vendorId: 0x1396
     productId: 0x0043A
     deviceProfileName: air-purifier-hepa
+#Eve
+  - id: "4874/79"
+    deviceLabel: Eve Thermo
+    vendorId: 0x130A
+    productId: 0x004F
+    deviceProfileName: thermostat-heating-only-nostate
```
ctowns, line 18: Just a note that the `matter-thermostat` driver has dynamic profile naming, so when testing this we'll just want to confirm that the driver attempts to switch to this same profile when the device is onboarded. Should be easy to check by confirming that expected profile is still being used after the device is onboarded.

ctowns (approved):

PR #1526, opened 2024-07-15
 CHAD-12676: Update Zigbee Min & Max Value temperatures in necessary device drivers

wkhenon:
This pull request updates the device drivers who have custom temperature range handling to also support the min & max values for the TemperatureMeasurement capability

Developer Testing
Unit Tests ✅


greens (commented):
So, because the handling for this attribute is in the base driver, this addition shouldn't be necessary.

wkhenon:
Closing due to unnecessary changes

PR #1527, opened 2024-07-16
Fix an issue that windowShade value is incorrect

HunsupJung:
Mamaba Curtain controller sometimes sends operational status event twice when it stops moving.
At that time, old code is not working correctly.
So This patch maps each status to each single event.
The mapping is as follows.

- Position ==100 : Open
- Position == 0 : Closed
- 0 < Position < 100 : Partially Open
- Operational Status == Opening : Opening
- Operational Status == Closing : Closing


drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -19,16 +19,7 @@ local log = require "log"
...
-local DEFAULT_LEVEL = 0
 local PROFILE_MATCHED = "__profile_matched"
-local IS_MOVING = "__is_moving"
-local EVENT_STATE = "__event_state"
-
-local WindowCoveringEventEnum = {
-  NO_EVENT = 0x00,
-  CURRENT_POSITION_EVENT = 0x01,
-  OPERATIONAL_STATE_EVENT = 0x02
-}
```
hcarter-775, line 31: nit: maybe just have these as 0, 1, 2 for readability?

hcarter-775:
This change, like you said, maps the status of the window covering to one single event. However, this will remove our "Opening..." and "Closing..." commands, as expected per the change, but is this really what we want? Couldn't we just add extra handling for if an event is sent twice based on time or previous state check logic?

hcarter-775:
One question: what happens in the Mamaba Curtain case when it sends the Operational State twice?

HunsupJung:
> However, this will remove our "Opening..." and "Closing..." commands, as expected per the change.
Almost of window covering send Operational Status Event once at the Operating Start and End. so when window covering starts to open, driver receive moving event and driver can change the window covering status to opening or closing.
Opening and Closing state are not going to be removed.

ctowns (commented):
Since this is changing the base driver behavior, we should regression test this against other existing window shades to confirm that they continue to work as expected. I believe we intentionally added the operational status bit handling for the transition states because that is how other window blind devices have worked, so if we are going to change the default behavior in a way that breaks the base driver, we could consider moving this to a sub driver.
> Mamaba Curtain controller sometimes sends operational status event twice when it stops moving.
This seems like a device-specific issue that could be fixed in a sub driver. 

However, I think the solution you have here makes sense, but I thought we had to do special handling for the operational state in order to prevent the devices from getting stuck saying "opening" or "closing". These changes were introduced in this PR: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1266
This PR seems to specifically address Zemismart. I am wondering which behavior should be considered our "base" behavior and which could be possibly be moved to a sub driver. 

I think the solution you have here makes sense, but we must regression test it to ensure it doesn't break behavior for existing window shade devices.

hcarter-775:
The reason I mention the Opening... and Closing... statements disappearing is because when I ran the VDA device with this PR's changes, those statements disappeared. So I was not sure if this was expected behavior.

HunsupJung:
@ctowns, @hcarter-775 
I don't understand what you said
When cannot we see the opening and closing with this PR?
Could you share the use case?

lelandblue:
> @ctowns, @hcarter-775 I don't understand what you said When cannot we see the opening and closing with this PR? Could you share the use case?

Hey @HunsupJung, @ctowns was expressing that right now the location of this code change is` drivers/SmartThings/matter-window-covering/src/init.lua ` and needs to be changed to a sub-driver. Currently how it is proposed this means that the changes being adding would affect all devices. Our belief is that this change should only be introduced to help support the Mamaba Curtain, which means that a new sub-driver should be introduced which will be a new file path. Please check out this page for reference. https://developer.smartthings.com/docs/edge-device-drivers/driver.html#subdrivers

@ctowns Please feel free to chime in if I am mistaken.

wangjinliang12345:
During CN WWST for YeeLight & OVRIBO,  and also a issues raised from a solution company called Longin.Link

Some  device does not send operational status, some does not send initial operational status or  CurrentPositionLiftPercent100ths, some does not send final operational status.  

For these situations , ST all show incorrect windowShade status now. 

And all the three vendor says their device works fine with Apple or google.

I think this PR will make great help for the issues mentioned above and we will also test here.

ctowns (commented):
Hey @HunsupJung, I understand now why these changes are needed in order to handle cases where the device does not properly set the `OperationalStatus` bit. I understand that these changes will address this issue, however I think this change will result in a change in behavior for the user and we need to decide how to address it.

Here is a sequence of events that I think a user could see:
```
+------+----------------------------------+-------------------+--------------------------------------------+------------------------+
| Step |              Action              | OperationalStatus |        CurrentPositionLiftPercentage       | windowShade capability |
+------+----------------------------------+-------------------+--------------------------------------------+------------------------+
| 1    | None                             | Not moving        | 0%                                         | closed                 |
+------+----------------------------------+-------------------+--------------------------------------------+------------------------+
| 2    | User sets level to 100%          | Not moving        | 0%                                         | closed                 |
+------+----------------------------------+-------------------+--------------------------------------------+------------------------+
| 3    | Receive OperationalStatus update | Opening           | 0% - not yet updated, status updates first | opening                |
+------+----------------------------------+-------------------+--------------------------------------------+------------------------+
| 4    | Receive LiftPercentage update    | Opening           | 33%                                        | partially open         |
+------+----------------------------------+-------------------+--------------------------------------------+------------------------+
| 5    | Receive LiftPercentage update    | Opening           | 67%                                        | partially open         |
+------+----------------------------------+-------------------+--------------------------------------------+------------------------+
| 6    | Lift percent reaches 100%        | Not moving        | 100%                                       | closed                 |
+------+----------------------------------+-------------------+--------------------------------------------+------------------------+
```

With these changes, I think the `CurrentLiftPercentage` will override the `opening/closing` state set by the `OperationalState` handler, which would result in the user seeing `partially_open` instead of seeing the correct operational status of `opening` or `closing`. For example, in steps 4 and 5 above, I would expect the user to see the `windowShade` status as `partially open` instead of as `opening` because the `CurrentPositionLiftPercentage` handler (`current_pos_handler`) will override the `windowShade.windowShade` value to be `partially open`, even while the window covering is still `opening` and has not completed this procedure.

I think this is a UI change and this behavior is probably acceptable because without these changes the blinds do not work at all, but I wanted to hear your opinion on this. Does the behavior I described above occur on the devices you have tested with? I only have the VDA to test window covering with and I know there is a known issue with this device.

Overall, I think this is a good change, I just want to make sure that we don't recreate the issue in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1266 where the wrong status was being displayed when opening and closing.

If you could, please provide a list of the devices that were tested on. Also, could you provide a video/screen recording of what the ST App looks like when one of the window covering devices is being opened or closed?

HunsupJung:
@hcarter-775 
I'm so sorry for the reply late. and thank you for your explanation.
As I said at [slack](https://samsung.slack.com/archives/C038KA8K78S/p1722303803445289?thread_ts=1722302098.890529&cid=C038KA8K78S), the situation which you said above can occurs. but it's only VDA case and a window covering device connected by DIRIGERA Hub (IKEA).
So if you cover those case, we need to separate it to sub driver to support VDA and DIRIGERA.

The following link is the test video.
https://samsung.slack.com/archives/C038KA8K78S/p1722303803445289?thread_ts=1722302098.890529&cid=C038KA8K78S

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -137,75 +126,33 @@ end
...
-        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.windowShade.windowShade.partially_open())
-      else
-        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.windowShade.windowShade.unknown())
-      end
-      device:set_field(EVENT_STATE, WindowCoveringEventEnum.NO_EVENT)
-    end
+  local windowShade = capabilities.windowShade.windowShade
+  local position = 100 - math.floor((ib.data.value / 100))
+  device:emit_event_for_endpoint(ib.endpoint_id, capabilities.windowShadeLevel.shadeLevel(position))
+  if position == 0 then
```
nickolas-deboom, line 135: If we are no longer checking whether the shade is not moving, then what would happen if the shade is opening for example, and the Operational Status is received first, so the windowShade capability is set to "Opening". If the position is then received following this, then wouldn't the capability be updated to "Paritially Open", even if the shade is still moving?

HunsupJung: Are you mean following step?

1. Device send "Opening" value of Operation Status to client.
2. Device send position value to client during it's still moving.

In this case, client displays "Partially Open" for shade status even if the shade is still moving.
However, VDA is the only device that operate like that.
And now I updated sub driver that cover the VDA.

nickolas-deboom: Yes that's the situation I was referring to. I have seen that with the VDA as well, where it sends the updated position many times while moving. So, do the physical window covering devices you've tested with only send the position value after the shade is done moving?

HunsupJung: Yes, I never seen the devices that send the updated position many times while moving except VDA. and when such a device comes out, we can update the VID and PID to operate as a [sub driver](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1527/files#diff-59878b164008097e4769b5821bd5941f862ab2e47b551c0b7b60c7ba97828a25).

HunsupJung:
@lelandblue @ctowns @hcarter-775 @nickolas-deboom 
If there's no problem, could you approve this PR? :)

nickolas-deboom:
Hi @HunsupJung , the changes look good to me. I also tested them with the VDA and everything is working great. I'm not certain but it appears that this even fixes an issue I saw earlier with the VDA Window Covering device where the pause button did not work. I approved the PR, but that said, it would be best to wait to merge this PR until Cooper returns from vacation tomorrow and can verify that his feedback has been addressed as well!

hcarter-775:
Hi @HunsupJung can you handle the warnings that the Luacheck is generating? Otherwise it looks good to me!

hcarter-775:
@nickolas-deboom I don't think this PR is just for the VDA, I think it is for the Mamaba window covering device, and the VDA happens to have similar behaviors to the Mamaba device that we noted.

HunsupJung:
@nickolas-deboom @hcarter-775 
Thank you for your review~! I'll fix the warnings that the Luacheck is generating.

nickolas-deboom:
> @nickolas-deboom I don't think this PR is just for the VDA, I think it is for the Mamaba window covering device, and the VDA happens to have similar behaviors to the Mamaba device that we noted.

Ok gotcha!

hcarter-775:
Hello @HunsupJung, I may have missed something in these exchanges, but if the functionality you want has been moved to a sub-driver, shouldn't the regular init.lua function be reverted to how it used to be? I see there is a lot of removed code and I'm not sure if it should be removed.

HunsupJung:
@hcarter-775 
At first, I updated regular init.lua file for most devices, including Mamaba.
And I made the sub driver for a rare case that works like VDA.

hcarter-775:
Mamaba curtain: shows the previous value when opening/closing until the update is complete. Why is this? Was it doing this before?

Mamaba blinds: When changing values to open, the UI reads "open" for a short moment before switching to "opening..." (1:13). When changing values to closed, the UI reads "partially open" fora short 
moment before switching to closed (1:26). Why does this occur? Was this happening before? Can we fix this? And at (1:40) we see the same thing I noted about Mamaba curtain. Why does this occur? Can we not fix this?

Zemismart Curtain: at (1:57) the UI reads closing... correctly, but then switches to "closed" before it is actually closed. Why does this occur? Was this happening before? Can we fix this? At (2:02) we see the same issue when opening, where it switches to "open" way before it is open. AT (2:18) we see the same issues.	

Zemismart Bridged Curtain: I understand that the UI was not working before the patch, so no comments here. Is there any documentation about this issue though? It seems pretty extreme from a UI standpoint.

Zemismart Blind: At (3:38) we see the same issue I mentioned with the Zemismart Curtain.

So all that said:

1. Why do some devices show the previous value when opening/closing until the update is complete? Was it doing this before, and can this be fixed?
2. Some devices switch to "closed" or "open" before they should. Why is this? Was this happening before, and can this be fixed?
3. Is there anything that can be done about the Zemismart Bridged Curtain for the future?

Were all the issues I'm mentioning happening before?

HunsupJung:
@hcarter-775 
First of all, I'm sorry it took so long to answer. I have to reset the test environment for an answer and it takes half a day and the answer has been delayed due to Matter 1.3 preparation and door lock preparation over the past time.

> Mamaba curtain: shows the previous value when opening/closing until the update is complete. Why is this? Was it doing this before?

When user click open or close, Mamaba send an `opening` event and when this operation is done, Mamaba send a 'closed' event and 0% event. so this is normal operating.

> Mamaba blinds: When changing values to open, the UI reads "open" for a short moment before switching to "opening..." (1:13). 

I think it was device side issue. today I tested it and there's no same issue.

> When changing values to closed, the UI reads "partially open" for a short moment before switching to closed (1:26). Why does this occur? Was this happening before? Can we fix this?

This issue has also been fixed.

> And at (1:40) we see the same thing I noted about Mamaba curtain. Why does this occur? Can we not fix this?

I think this issue was not same with Mamaba curtain. this issue was same with other Mamaba blinds issues and has been fixed.

> Zemismart Curtain: at (1:57) the UI reads closing... correctly, but then switches to "closed" before it is actually closed. Why does this occur? Was this happening before? Can we fix this? At (2:02) we see the same issue when opening, where it switches to "open" way before it is open. AT (2:18) we see the same issues.

Zemismart send 0% value and closed event. So the result value changes even before the operation is over.

```
2024-10-07_13:05:45.367 | I | dev_cmd_bs | [edi] handled device command: 5068ce5b-de84-4be6-bda1-deb14695e65b.main:windowShade.close("[]")
2024-10-07_13:05:46.574 | I | edi        | <Matter Window Covering (6441e3f4)> <MatterDevice: 5068ce5b-de84-4be6-bda1-deb14695e65b [9F68F465BB4B164F-DC49B643D45BD1BF] (Zemismart MT01 Slide Curtain)> emitting event: {"attribute_id":"windowShade","capability_id":"windowShade","component_id":"main","state":{"value":"closing"}}
2024-10-07_13:05:46.779 | I | edi        | <Matter Window Covering (6441e3f4)> <MatterDevice: 5068ce5b-de84-4be6-bda1-deb14695e65b [9F68F465BB4B164F-DC49B643D45BD1BF] (Zemismart MT01 Slide Curtain)> emitting event: {"attribute_id":"shadeLevel","capability_id":"windowShadeLevel","component_id":"main","state":{"value":0}}
2024-10-07_13:05:46.813 | I | edi        | <Matter Window Covering (6441e3f4)> <MatterDevice: 5068ce5b-de84-4be6-bda1-deb14695e65b [9F68F465BB4B164F-DC49B643D45BD1BF] (Zemismart MT01 Slide Curtain)> emitting event: {"attribute_id":"windowShade","capability_id":"windowShade","component_id":"main","state":{"value":"closed"}}
```
(I omitted some part except important log.)

> Zemismart Bridged Curtain: I understand that the UI was not working before the patch, so no comments here. Is there any documentation about this issue though? It seems pretty extreme from a UI standpoint.

I think we need to deal with this issue anew. I don't have any other tickets yet, and I will upload them again after this PR is merged.

> Zemismart Blind: At (3:38) we see the same issue I mentioned with the Zemismart Curtain.

It is the same as the answer to the Zemismart  Curtain.

hcarter-775 (approved):

nickolas-deboom (approved):

drivers/SmartThings/matter-window-covering/src/matter-sub-window-covering/init.lua
```diff
@@ -0,0 +1,148 @@
...
+        [clusters.WindowCovering.attributes.OperationalStatus.ID] = current_status_handler,
+      }
+    }
+  },
+  capability_handlers = {
+  },
+  can_handle = is_matter_sub_window_covering,
+}
+
+return matter_sub_window_covering_handler
```
hcarter-775, line 148: nit: add newline

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -767,17 +785,10 @@ test.register_coroutine_test(
...
-        WindowCovering.attributes.OperationalStatus:build_test_report_data(mock_device, 10, 0),
-      }
-    )
     test.socket.capability:__expect_send(
-      mock_device:generate_test_message(
-        "main", capabilities.windowShade.windowShade.partially_open()
-      )
+        mock_device:generate_test_message(
+            "main", capabilities.windowShade.windowShade.partially_open()
+        )
```
hcarter-775, line 791: nit: remove added whitespace

hcarter-775 (changes requested):
please respond to comments before moving forward. Thank you

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -19,16 +19,9 @@ local log = require "log"
...
 local PROFILE_MATCHED = "__profile_matched"
-local IS_MOVING = "__is_moving"
-local EVENT_STATE = "__event_state"
 
-local WindowCoveringEventEnum = {
-  NO_EVENT = 0x00,
-  CURRENT_POSITION_EVENT = 0x01,
-  OPERATIONAL_STATE_EVENT = 0x02
-}
+local subWindowCovering = require("matter-sub-window-covering")
```
hcarter-775, line 24: remove this line, unecessary with the corresponding comment

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -276,6 +225,9 @@ local matter_driver_template = {
     capabilities.windowShadePreset,
     capabilities.battery,
   },
+  sub_drivers = {
+    subWindowCovering
```
hcarter-775, line 229:
```suggestion
    -- for devices sending multiple positional updates on update
    require("matter-window-covering-position-updates-while-moving")
```
The previous name was unclear, which I think will cause confusion in the future. This is a little wordy, but the case is wordy as well. If anyone can think of a clearer name feel free to throw it out!

Also, be sure to change the name of the driver to follow this new name

hcarter-775:
@HunsupJung Thank you for getting to my (very long) comment!

drivers/SmartThings/matter-window-covering/src/init.lua
HunsupJung: I think it's a clear name.

HunsupJung:
@hcarter-775 @nickolas-deboom 
Thank you for all your reviews. I think all modification is updated. If there are any other problems, please let me know.

drivers/SmartThings/matter-window-covering/src/init.lua
pInksenberg: The Nrf52840 window covering sample (Matter development kit) will update its position every 5% per time while moving, as I have previously tested.

pInksenberg:
I raised a PR that use TargetPosition and CurrentPosition to decide the status of window covering, which avoid considering the order of attribute (current position and operational status) reports from devices, thereby simplifying the code. Please help to check.
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1679/commits/447740d35102bfcb323773e2c028f9f0df87fb3f

hcarter-775:
Hi @HunsupJung would you mind looking at @pInksenberg 's commit? I believe it may be a much simpler solution for the issue @nickolas-deboom mentioned here than creating a new subdriver for this one possibility.

pInksenberg:
There might have some problem creating a new subdriver for special situations, as bridged devices lack a PID and therefore cannot be matched with the subdriver.

HunsupJung:
@pInksenberg 
First of all, thank you for your suggestion.
Considering the correlation between the `target position` value and the `current position` value in Matter Spec, I think it is the perfectly correct way.
However, in SmartThings, the value actually applied to capability is `current position` and the `current position` value may not match the actual device properly due to network error.

For example, some devices continue to send the `current position` value during `opening` / `closing`.
In this device, let's suppose the device is `opening`.
While `shadeLevel` is updated to 96% > 98% > 100%, update to 100% may fail and remain at 98%.
And since the driver has received 100% event, the value of `windowShade` will try to update to `open`.
After that the `shadeLevel` value is 98%, but the `windowShade` value is `open`.
(96% success > 98% success > 100% fail > `open` success)
So the Suwon team conclude to decide that the windowShade only by the `current position` value.
Of course, as an exception, if the operation value changes to opening / closing, we are updating it.

In addition, there is a problem with your commit.
If the `target position` value and the `current position` value come to driver at the same time, the windowShade is not updated.
For example, when `windowShade` is `open`  and driver send a `close` command, the problem that 'shadeLevel` is 0% and `windowShade` is `open` occurs.
This occurs when the `target position` value and the `current position` value come at the same time.
(I confirmed that this commit causes the problem.)

HunsupJung:
> There could be a problem creating a new subdriver for special situations, as bridged devices lack a PID and therefore cannot be matched with the subdriver.

The basic policy is to determine the `windowShade` with only one attribute without comparing the two attributes.
Therefore, it is correct to apply the situation that cannot be solved by the subdriver as a basic policy.

HunsupJung:
@hcarter-775 
Currently, this issue has been raised by users since about three months ago and has kept them waiting for a really long time.
I think there could be a better code than the one I implemented. However, I don't think it's a good idea to keep waiting for users without putting in the current PR.

First of all, I think we need to solve this issue by merging the current PR. And then If there is a better and more efficient code in the future, I don't think it's too late to merge it then. so please approve it.

hcarter-775 (approved):

PR #1548, opened 2024-07-26
Subscribe to Child Device Attributes

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [X] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change adds support for subscribing to the attributes of child devices.

# Summary of Completed Tests

New unit tests were created to verify these changes. Also, it was tested with a VDA virtual switch to ensure all functionality worked as expected and that there were not any unintended effects.


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -65,6 +66,79 @@ local device_type_profile_map = {
   [ON_OFF_SWITCH_ID] = "switch-binary",
   [ON_OFF_DIMMER_SWITCH_ID] = "switch-level",
   [ON_OFF_COLOR_DIMMER_SWITCH_ID] = "switch-color-level",
+  [GENERIC_SWITCH_ID] = "button"
```
tpmanley, line 69: Should this change be in this PR? The `button` profile doesn't exist in this PR or `main`

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -65,6 +66,79 @@ local device_type_profile_map = {
...
+}
+
+local device_type_attribute_map = {
```
tpmanley, line 72: Did you consider mapping capabilities to the different attributes/events instead of device types? Then you would know which attributes to subscribe to based on the profile the device is using which seems more flexible.

nickolas-deboom: Since the child devices are not created at `init` for the parent device, we don't know their profile until they are initialized, so I was doing the subscriptions for child devices in `device_init` using the `get_parent_device` function. But, after talking to Cooper about this, he pointed out that this could lead to race conditions as noted here: https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/00a4fcdba1d29aa9d3b033909a6aaa1df24cc384/lua_libs/st/device.lua#L217

I can't think of another way to do this outside of `device_init`, so I'm thinking it might be best to continue using the device type.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Removed in 7462d040a54b3ca6400798f17f6a279af4cc3c05!

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -121,6 +186,9 @@ local function assign_child_profile(device, child_ep)
         id = math.max(id, dt.device_type_id)
       end
       profile = device_type_profile_map[id]
+      for _, attr in pairs(device_type_attribute_map[id]) do
```
tpmanley, line 189: Is this the best place to add the subscriptions? It looks like we're going to be doing that before the child devices are created since this is the function that determines the profile to use when creating the child device. Why are we not waiting until the child device is created to add the subscription so that it's ready when the device starts sending subscription reports?

nickolas-deboom: I thought this was a logical place for this since we are already determining the device type here. If we add the subscriptions here before the child devices are created, then wouldn't they be ready once the device starts sending the subscription reports?

tpmanley: I could be mistaken here, but I think that if the subscriptions are added before the devices are created then the initial priming report will not result in any capability events being generated since the devices won't exist yet.

nickolas-deboom: Hi Tom, I missed this message. I was under the impression that the subscriptions are all made through the parent device, so would this still be a concern if that is indeed the case?

nickolas-deboom: After discussing, the subscriptions were moved to `device_init` so that they will be subscribed to each time the device is reinitialized.

drivers/SmartThings/matter-switch/src/test/test_matter_bridge.lua
```diff
@@ -70,8 +70,19 @@ local mock_bridge = test.mock_device.build_test_matter_device({
...
+
 local function test_init_mock_bridge()
   test.socket.matter:__set_channel_ordering("relaxed")
+  local subscribe_request = cluster_subscribe_list[1]:subscribe(mock_bridge)
+  for i, cluster in ipairs(cluster_subscribe_list) do
+    if i > 1 then
+      subscribe_request:merge(cluster:subscribe(mock_bridge))
+    end
+  end
+  test.socket.matter:__expect_send({mock_bridge.id, subscribe_request})
```
hcarter-775, line 85: Perhaps I'm reading it wrong, but the section `if i > 1` indicates that that line `subscribe_request:merge(cluster:subscribe(mock_bridge))` never gets called. Since nothing else in this test file is changed, what is the purpose of this change?

hcarter-775:
I'm a little confused to the functionality being added here. So the title states "subscribe to child device attributes", but this subscription logic is being added in init, not where the child profiles are being created in initialize_switch. @nickolas-deboom would you mind explaining the control flow of this logic? Thank you

nickolas-deboom:
> I'm a little confused to the functionality being added here. So the title states "subscribe to child device attributes", but this subscription logic is being added in init, not where the child profiles are being created in initialize_switch. @nickolas-deboom would you mind explaining the control flow of this logic? Thank you

When the child devices are created, init will be ran for each instance and at that point the subscriptions to the attributes for the child devices will occur. This needs to take place in init so that the attributes are re-subscribed to for each re-initialization of the device.

hcarter-775 (approved):
I don't have a lot of experience with the matter switch driver, but from what I can see it lgtm.

drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
```diff
@@ -67,7 +67,7 @@ local mock_device_dimmer = test.mock_device.build_test_matter_device({
   },
   endpoints = {
     {
-      endpoint_id = 5,
+      endpoint_id = 0,
```
ctowns, line 70: Why is this changed to endpoint 0? Endpoint 5 is a valid starting endpoint

drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
```diff
@@ -217,7 +223,18 @@ local function test_init_onoff_client()
   test.mock_device.add_test_device(mock_device_onoff_client)
 end
 
+local cluster_subscribe_list = {
+  clusters.OnOff.attributes.OnOff
+}
+
 local function test_init_parent_client_child_server()
+  local subscribe_request = cluster_subscribe_list[1]:subscribe(mock_device_parent_client_child_server)
```
ctowns, line 231: I am confused why these subscribe requests are needed here when they weren't before? Even before this change, we'd still be sending out a subscribe request for the parent device. Do these tests fail without this added expectation for the subscribe request?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -258,6 +323,18 @@ local function device_init(driver, device)
...
+      if ep.endpoint_id ~= main_endpoint and ep.endpoint_id ~= 0 then
+        local id = 0
+        for _, dt in ipairs(ep.device_types) do
+          id = math.max(id, dt.device_type_id)
+        end
+        for _, attr in pairs(device_type_attribute_map[id] or {}) do
+          device:add_subscribed_attribute(attr)
+        end
+      end
+    end
```
ctowns, line 337: You should create a new mock device that has a parent endpoint that has clusters that are NOT a superset of the child device's clusters and then write a unit test to confirm that the the child device's clusters are still subscribed to.

For example, this mock device could have an `On/Off Plug` on endpoint 1 and an `Extended Color Bulb` on endpoint 2. Previously, only the OnOff cluster would have been subscribed to since that is the only cluster on the parent device, however with these changes we should additionally subscribe to the child device's cluster attributes of ColorControl (hue, sat, etc.)

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -258,6 +323,18 @@ local function device_init(driver, device)
     if device:get_field(IS_PARENT_CHILD_DEVICE) == true then
       device:set_find_child(find_child)
     end
+    local main_endpoint = find_default_endpoint(device)
+    for _, ep in ipairs(device.endpoints) do
+      if ep.endpoint_id ~= main_endpoint and ep.endpoint_id ~= 0 then
+        local id = 0
```
ctowns, line 329: I would also add a unit test for the case where a device type that is not in our device type map is found. In this case, we wouldn't subscribe to the unaccounted for device type's clusters

drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
nickolas-deboom: There isn't a subscribe request being sent out for the parent device, because in these test cases each mock device starts out with the `matter-thing.yml` profile, so there are not any attributes to subscribe to initially. Now with these changes, for the two mock devices with child devices, when the child devices are created (with actual profiles) we subscribe to the appropriate attributes.

drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
nickolas-deboom: This was changed while there were issues in the code previously and I forgot to restore it. Fixed!

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Added a new mock device and test case as suggested!

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Added in 8f9481d2a5eed52608b22e5cab66822305543559!

drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
```diff
@@ -195,7 +195,88 @@ local mock_device_parent_child_switch_types = test.mock_device.build_test_matter
   }
 })
 
+local mock_device_parent_child_different_types = test.mock_device.build_test_matter_device({
+  label = "Matter Switch",
+  profile = t_utils.get_profile_definition("matter-thing.yml"),
```
ctowns, line 200: nit: make this switch binary just to match what this device would join as more realistically

ctowns (commented):
LGTM! I left just one nit to be fixed before merging. Also could you fill out the summary of completed tests section of the PR comment template? Then I think it should be good to go!

drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
nickolas-deboom: Updated as suggested in latest commit

nickolas-deboom:
> LGTM! I left just one nit to be fixed before merging. Also could you fill out the summary of completed tests section of the PR comment template? Then I think it should be good to go!

Fixed the nit and added some info in the test section!

PR #1551, opened 2024-07-29
Production Edge Driver PR for 29-July-2024

lelandblue:
# Description of Change
This is the PR to start the production release for 29-July-2024.


ctowns (commented):
Looks good to me, but I'm not sure why the Delete PR channel check is being shown here. It looks like that CI check was run 4 hours ago for https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1544 but for some reason it is showing up here. It looks like the PR channel for #1544 is still open so I think we'd just need to run the delete channel task again and it should be resolved. @greens have we seen the delete channel task show up on a different PR before?

greens (approved):

greens:
@ctowns nope! Not sure why that happened but I suspect github weirdness.

ctowns (approved):

PR #1554, opened 2024-08-02
Create a series of optional profiles for room ac

hcarter-775:
Edit configure logic to permit a wider array of room ac profiles. Further, I have added several new profiles that can be mapped to.

Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests


drivers/SmartThings/matter-thermostat/profiles/room-air-conditioner-humidity-fan-wind-heating-cooling.yml
```diff
@@ -1,4 +1,4 @@
-name: room-air-conditioner
+name: room-air-conditioner-humidity-fan-wind-heating-cooling
```
ctowns, line 1: We can only safely change a profile name if no device is currently using the profile. We need to either confirm there are no other devices using this profile name or just make a duplicate that uses the new name

You will also need to update all instances where this old profile name was used, for example, in the room ac generic fingerprint

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -328,7 +328,27 @@ local function do_configure(driver, device)
...
+      profile_name = profile_name .. "-fan"
+      if #wind_eps > 0 then
+        profile_name = profile_name .. "-wind"
+      end
+    end
+
+    if #heat_eps > 0 and #cool_eps > 0 then
+      profile_name = profile_name .. "-heating-cooling"
+    else
+      device.log.warn_with({hub_logs=true}, "Thermostat does not support both heating and cooling. No matching profile")
```
ctowns, line 346:
```suggestion
      device.log.warn_with({hub_logs=true}, "Room AC does not support both heating and cooling. No matching profile")
```

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -363,9 +383,6 @@ local function do_configure(driver, device)
         profile_name = profile_name .. "-cooling-only"
       end
 
-      -- TODO remove this in favor of reading Thermostat clusters AttributeList attribute
```
ctowns, line 366: Is this no longer needed? This should be left if we still need to do it, which I think we do

ctowns:
This needs to be regression tested against the room AC device in the Matter VDA

drivers/SmartThings/matter-thermostat/profiles/room-air-conditioner-humidity-fan-wind-heating-cooling.yml
ctowns, line 1: We should have a version of this profile that is exactly the same, except without humidity

hcarter-775: we do have one like that- room-air-conditioner-humidity-fan-wind-heating-cooling

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: There is another place that states the TODO, it looked kinda clunky having it written twice (even if the change applies to both)

drivers/SmartThings/matter-thermostat/profiles/room-air-conditioner-humidity-fan-wind-heating-cooling.yml
ctowns: Oh I see, could you change it so that the capabilities are listed in the same order? It looks like the `thermostatOperatingState` just needs to be moved below the setpoint capabilities like is done here. We'll want the capability order to be internally consistent between the profiles

ctowns: Looks like there are some devices that are currently using the `room-air-conditioner` profile, so we'll need to keep the old profile and essentially make a duplicate profile with the new name. OR we could add a conditional check in the driver that just matches the constructed name `room-air-conditioner-humidity-fan-wind-heating-cooling` to the name `room-air-conditioner`. I think to be consistent with what we did for the air purifier, we'd want to go with option 1 though, would you agree?

hcarter-775: ah, good catch!

hcarter-775: when changing the fingerprint, I run into (another) name length issue, since apparently fingerprint IDs can only be 36 characters long, so I think it would be best to just add two identical profiles- one called room-air-conditioner and one with this new name.

hcarter-775: I don't necessarily like the idea of having two profiles with the same capabilities if we can avoid it, and I don't know exactly what you mean about the air purifier situation? I don't remember needing multiple profiles names being mapped to the same capabilities. 

If anything, I think going with your second option is consistent with how things are handled in matter-sensor, so that's what I prefer.

hcarter-775:
> This needs to be regression tested against the room AC device in the Matter VDA

This has been done. The new configure logic leads to the same functionality as in the main branch.

ctowns (approved):
Looks good! We'd probably want to add another simple unit test for the doConfigure event case like was done for the air purifier that shows that a device matches to the correct profile

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -328,7 +328,31 @@ local function do_configure(driver, device)
...
+    if #heat_eps > 0 and #cool_eps > 0 then
+      profile_name = profile_name .. "-heating-cooling"
+    else
+      device.log.warn_with({hub_logs=true}, "Room AC does not support both heating and cooling. No matching profile")
+      return
+    end
+
+    if profile_name == "room-air-conditioner-humidity-fan-wind-heating-cooling" then
+      profile_name = "room-air-conditioner"
+    end
```
HunsupJung, line 352: @hcarter-775 
The number of profiles that may come out due to the above code is as follows.

- room-air-conditioner
- room-air-conditioner-humidity
- room-air-conditioner-fan
- room-air-conditioner-humidity-fan
- room-air-conditioner-fan-wind
- room-air-conditioner-humidity-fan-wind
- room-air-conditioner-heating-cooling
- room-air-conditioner-humidity-heating-cooling
- room-air-conditioner-fan-heating-cooling
- room-air-conditioner-humidity-fan-heating-cooling
- room-air-conditioner-fan-wind-heating-cooling

You didn't create some profiles that support only HEAT or COOL feature. Is that the code you intended?

hcarter-775: The logic I provided does not permit heat or cool to be created- only heating and cooling together. Since this is a room ac, I think we can safely assume that both heating and cooling will be provided for now.

hcarter-775: Further, because it is a room ac, even though heating and cooling are optional, I only created new profiles that include them both since these are in some ways fundamental to a room ac.

HunsupJung (approved):

PR #1557, opened 2024-08-05
New Device (Matter Lock) Aqara U300

lelandblue:
Check all that apply

# Type of Change

- [ X ] WWST Certification Request
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

- [ X ] I have performed a self-review of my code

# Description of Change
- This PR is a fingerprint only code change to add a fingerprint for the Aqara U300 device. I anticipate that this will use the new sub-driver created for the U200 device. 

# Summary of Completed Tests
- The testing of this device certification will be handled by the WWST process.

### DIFF
--- a/drivers/SmartThings/matter-lock/src/aqara-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/aqara-lock/init.lua
@@ -18,15 +18,11 @@ local device_lib = require "st.device"
 
 local DoorLock = clusters.DoorLock
 local AQARA_MANUFACTURER_ID = 0x115f
-local U200_PRODUCT_ID = 0x2802
-local U300_PRODUCT_ID = 0x2801
 
 local function is_aqara_products(opts, driver, device)
   if device.network_type == device_lib.NETWORK_TYPE_MATTER and
-      device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID and
-      (device.manufacturer_info.product_id == U200_PRODUCT_ID or
-      device.manufacturer_info.product_id == U300_PRODUCT_ID) then
-        return true
+      device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID then
+    return true
   end
   return false
 end
--- a/drivers/SmartThings/matter-lock/src/aqara-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/aqara-lock/init.lua
@@ -19,11 +19,13 @@ local device_lib = require "st.device"
 local DoorLock = clusters.DoorLock
 local AQARA_MANUFACTURER_ID = 0x115f
 local U200_PRODUCT_ID = 0x2802
+local U300_PRODUCT_ID = 0x2801
 
 local function is_aqara_products(opts, driver, device)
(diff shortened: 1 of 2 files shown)


HunsupJung (approved):
👍

hcarter-775 (approved):

nickolas-deboom (approved):

HunsupJung:
I missed that we need to update the function which is checking the aqara U300 product. 
The [function ](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/production/drivers/SmartThings/matter-lock/src/aqara-lock/init.lua#L23C1-L30C4) have to return `true` when product id is  `U300(0x2801)`.

drivers/SmartThings/matter-lock/src/aqara-lock/init.lua
```diff
@@ -19,11 +19,13 @@ local device_lib = require "st.device"
 local DoorLock = clusters.DoorLock
 local AQARA_MANUFACTURER_ID = 0x115f
 local U200_PRODUCT_ID = 0x2802
+local U300_PRODUCT_ID = 0x2801
 
 local function is_aqara_products(opts, driver, device)
   if device.network_type == device_lib.NETWORK_TYPE_MATTER and
       device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID and
-      device.manufacturer_info.product_id == U200_PRODUCT_ID then
+      (device.manufacturer_info.product_id == U200_PRODUCT_ID or
```
ctowns, line 27: @HunsupJung if *all Aqara locks* are going to be using this sub driver, we can remove the check for the product ID and just check for the vendor ID. This would prevent us from needing to modify the sub driver in the future whenever we have a new Aqara fingerprint to add.

For example, in the eve-energy sub driver, we only check for the vendor ID because all Eve products will use the sub driver: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/731d2aebdd763d72d9f5951d9456be5b04e16ffc/drivers/SmartThings/matter-switch/src/eve-energy/init.lua#L54

Can you confirm if all Aqara locks will use this sub driver? If yes, then I think we can remove the check for the product_id and just use vendor_id

HunsupJung: @ctowns 
I agree with your opinion. Ultimately, the goal is to apply it without a sub driver. But now, it would be better to let all Aqara door locks use the sub driver for stability. we can remove the check for the product_id and just check vendor_id.

lelandblue:
@hcarter-775 or @nickolas-deboom Please re-review the latest changes when you have a moment.

drivers/SmartThings/matter-lock/src/aqara-lock/init.lua
ctowns: I made this change here: 728b21afc4f90cd0f0e776a82d52f93177802d00

This should simplify further additions of Aqara lock devices so that they can be fingerprint-only changes.

z-michel (approved):

PR #1563, opened 2024-08-07
Update drivers for Matter 1.3

ctowns:
Update drivers for Matter 1.3.

Release checklist

- [x] Resolve luacheck errors
- [ ] Add unit tests for all new device types

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests


drivers/SmartThings/matter-switch/src/embedded-cluster-utils.lua
```diff
@@ -0,0 +1,51 @@
+local clusters = require "st.matter.clusters"
+local utils = require "st.utils"
+
+-- Include driver-side definitions when lua libs api version is < 10
+local version = require "version"
+if version.api < 10 then
```
nickolas-deboom, line 6: This should be "< 11" right?

nickolas-deboom:
Unrelated to the changes in this PR but all instances of `OFF_ON_AUTO` in `matter-thermostat/src/init.lua` should be changed to `OFF_HIGH_AUTO`

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -69,6 +69,12 @@ local WIND_MODE_MAP = {
   [1]		= capabilities.windMode.windMode.naturalWind
 }
 
+local ROCK_MODE_MAP = {
+  [0] = capabilities.fanOscillationMode.fanOscillationMode.horizontal,
```
ctowns, line 73: I think this would be cleaner if this index should be the actual bitmask that corresponds to the given rock mode, that way you can use a value that's defined in our lua libs. Plus, that removes the need to do the bit shifting check below in the rock support functions.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -456,6 +470,7 @@ local function device_added(driver, device)
   device:send(clusters.Thermostat.attributes.ControlSequenceOfOperation:read(device))
   device:send(clusters.FanControl.attributes.FanModeSequence:read(device))
   device:send(clusters.FanControl.attributes.WindSupport:read(device))
+  device:send(clusters.FanControl.attributes.RockSupport:read(device))
```
ctowns, line 473: I know this existed before your changes, but one possible improvement is to merge these reads into 1 send request and then just send one request instead of 4
```
local req = im.InteractionRequest(im.InteractionRequest.RequestType.READ, {})
req:merge(clusters.Thermostat.attributes.ControlSequenceOfOperation:read(device))
reg:merge(clusters.FanControl.attributes.FanModeSequence:read(device))
reg:merge(clusters.FanControl.attributes.WindSupport:read(device))
reg:merge(clusters.FanControl.attributes.RockSupport:read(device))
device:send(req)
```

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: I also see this is how it is done elsewhere in the file, so it might be a general improvement we can make outside of the Matter 1.3 changes as well. We can do that in a separate PR and just keep this consistent

drivers/SmartThings/matter-switch/src/embedded-cluster-utils.lua
ctowns: Yes!

drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -2320,6 +2320,11 @@ matterGeneric:
     deviceTypes:
       - id: 0x010B # Dimmable Plug-in Unit
     deviceProfileName: plug-level
+  - id: "matter/electrical-sensor"
+    deviceLabel: Matter Electrical Sensor
+    deviceTypes:
+      - id: 0x0510 # Electrical Sensor
+    deviceProfileName: power-energy-powerConsumption
```
ctowns, line 2327: @nickolas-deboom we will want to add a default fingerprint for the Valve device type as well. You can just have it match to the `water-valve` profile for now, and the logic you have in the `init` function should handle adding the `level` capability as needed

drivers/SmartThings/matter-switch/profiles/electrical-power.yml
```diff
@@ -0,0 +1,14 @@
+name: electrical-power
+components:
+- id: main
+  capabilities:
+  - id: switch
```
ctowns, line 5: Let's define a naming convention for profiles for electrical sensor devices. I think in most cases, we would expect to see the electrical sensor on a plug, so if we include the switch capability and the `SmartPlug` category, what if we started the profile name with `plug`? There isn't a convention yet, so I think we should start one in order to make profile name building in the future easier if we need to expand on it. Maybe something like this, where we start with a switch/plug type depending on the device type and that is used to match to the correct category in a profile, and then the following additions are optional checks based on the device functionality which we would append in this order:
switch/plug-(electrical)-(energy)-(power)-(powerConsumption)

What do you think about establishing this convention?

drivers/SmartThings/matter-switch/fingerprints.yml
nickolas-deboom: Added a default fingerprint for water valve!

nickolas-deboom:
> Unrelated to the changes in this PR but all instances of `OFF_ON_AUTO` in `matter-thermostat/src/init.lua` should be changed to `OFF_HIGH_AUTO`

This has been fixed.

drivers/SmartThings/matter-switch/src/embedded-cluster-utils.lua
nickolas-deboom: Fixed!

drivers/SmartThings/matter-appliance/src/matter-dishwasher/init.lua
```diff
@@ -125,27 +125,27 @@ local function dishwasher_alarm_attr_handler(driver, device, ib, response)
   local isWaterFlowVolumeAlarm = false
 
   local state = ib.data.value
-  if state & clusters.DishwasherAlarm.types.AlarmMap.INFLOW_ERROR > 0 then
+  if state & clusters.DishwasherAlarm.types.AlarmBitmap.INFLOW_ERROR > 0 then
```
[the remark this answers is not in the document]
ctowns: Thank you for pointing this out @SathvikGatti ! I think we have two options then for moving forward:
1) Add the alias support to the embedded cluster definitions. This way, we can add an alias that maps `AlarmMap -> Alarm Bitmap` so that the name `AlarmMap` can still be used in the driver when the lua libs api is <= 10. This would require adding the `alias` files to the driver embedded cluster definitions.
2) We could keep the name updates in the embedded cluster definition to match the updates made in Matter 1.3, and then we change the `require` statement to from `if version.api < 10 then` to `if version.api < 11 then`. This means that we will need to use the embedded cluster definition for longer (i.e. until all platforms are updated to use the Matter 1.3 lua libs), but I think it is a cleaner solution overall. Keeping the embedded cluster definitions means the driver package will be larger from a memory perspective, but the driver memory when running will actually be either the same or less than when using  the lua libs definitions. Since we are not close to running out of space in the driver package yet, I think option 2 could be the preferred option.

I think that option 2 is probably the easiest, it will just mean that we will have to keep all embedded clusters up to date with each matter specification, and then updating the drivers to use the latest name for embedded cluster values. @HunsupJung do you have a preference between these options? If not, I think we should go with option 2.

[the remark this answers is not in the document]
ctowns: Actually, on further thought, I think we should actually not update the embedded clusters at all. I think they should remain as they are with the name changes for Matter 1.3. Basically, this is if we continue to use the original name of the attributes, it will allow the same driver to work across all different lua libs versions and avoid the issue that @SathvikGatti originally mentioned. I have a little bit of background on this here: https://smartthings.atlassian.net/wiki/spaces/~62de3b63b6b0b70770d7905d/pages/3406629893/Handling+Cluster+Changes+Across+Matter+Specs+and+FW+Versions

So, I think I will revert [5c0495a](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1563/commits/5c0495ad5dc9a519cc89824cc8d752c349e3dbcf) and we can leave the embedded cluster definitions as they are.

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: I made a ticket describing your proposal [here](https://smartthings.atlassian.net/browse/CHAD-13925)

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: Good idea! That change is being handled [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1578) in a fanOscillationMode specific PR. All fanOscillationMode changes have been removed from this PR.

drivers/SmartThings/matter-switch/profiles/electrical-power.yml
hcarter-775: I like that idea! First off, I think we can scrap the `electrical` keyword since that doesn't really map to any capabilities we'd want to specify. Second, the proposed convention would already by broken a little with the eve profile called `power-energy-powerConsumption` , which using this naming convention should be called `plug-energy-power-powerConsumption`. Personally, I propose we make a second profile with the new name and deprecate the old one, especially since the eve-specific functionality may merge with the electrical sensor functionality someday. Further, since we got some backing for allowing these one-time driver updates, I think this is exactly the kind of thing we're looking to be able to re-organize if we want. 

But other than this little snag, I like this. What do you think about making a new profile to fit the new naming convention, and then just sticking by it?

hcarter-775: We can revert it if we want, but for now I just bit the bullet and made a new profile with the same functionality as already existed in the Eve-specific profile. I haven't touched the Eve stuff though, only this new stuff.

drivers/SmartThings/matter-appliance/src/embedded-cluster-utils.lua
```diff
@@ -13,17 +15,27 @@ if version.api < 10 then
   clusters.TemperatureControl = require "TemperatureControl"
 end
 
+--these clusters are not available in v10 and v11
+clusters.MicrowaveOvenControl = require "MicrowaveOvenControl"
+clusters.MicrowaveOvenMode = require "MicrowaveOvenMode"
+clusters.OvenMode = require "OvenMode"
+
```
hcarter-775, line 22: @HunsupJung I believe these clusters are available in v11, right? So maybe this should be changed to 
```suggestion
if version.api < 11 then
  clusters.MicrowaveOvenControl = require "MicrowaveOvenControl" 
  clusters.MicrowaveOvenMode = require "MicrowaveOvenMode"
  clusters.OvenMode = require "OvenMode"
end
```

drivers/SmartThings/matter-appliance/profiles/oven-with-cook-top.yml
hcarter-775, line 1: I believe this should have a name that fits more with the other new appliance profile names. I think it will be useful to create a formal naming convention right from the start. For example, this device could be called:

oven-cabinet-one-tm-cabinet-two-tm-cook-top-cook-surface-one-tl-cook-surface-two-tl.

Following the order of the components, as well as the optional clusters each component supports.

drivers/SmartThings/matter-appliance/src/embedded-cluster-utils.lua
hcarter-775: I see now that OvenMode is not available in the lua libs version 11. The others are though, correct?
Perhaps it would make sense then to do hat I suggested above, but to leave the OvenMode cluster out with a comment specific to it in particular, since it's not all of them.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -346,7 +346,6 @@ end
 local function find_default_endpoint(device, component)
   local switch_eps = device:get_endpoints(clusters.OnOff.ID)
   local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.Feature.MOMENTARY_SWITCH})
-  local valve_eps = embedded_cluster_utils.get_endpoints(device, clusters.ValveConfigurationAndControl.ID)
```
ctowns: Why was this removed?

drivers/SmartThings/matter-appliance/profiles/oven-with-cook-top.yml
HunsupJung: I think that's a good idea. It would be better to set rules for creating profiles until the function of dynamically configuring profiles is supported.

hcarter-775:
@HunsupJung I believe we should have a unit test for both matter-oven and matter-cook-top to show functionality of `temperatureMeasurement` for *F. Right now all of the tests use *C, and I think it would be good to make sure it works for *F as well.

drivers/SmartThings/matter-appliance/src/matter-cook-top/init.lua
```diff
@@ -0,0 +1,233 @@
...
+  local event = capabilities.temperatureLevel.supportedTemperatureLevels(supportedTemperatureLevels,
+    { visibility = { displayed = false } })
+  device:emit_event_for_endpoint(ib.endpoint_id, event)
+end
+
+local function temp_event_handler(driver, device, ib, response)
+  device.log.info(string.format("temp_event_handler: %s", ib.data.value))
+
+  local temp
+  local unit = "C"
```
hcarter-775, line 181: So we can only handle `temperatureMeasurement` in Celsius? Is this acceptable?

ctowns: Good question! The measured value from the TemperatureMeasurement cluster will always be in degC because it has the temperature data type: https://librarian-regionals.smartthingsgdev.com/matter/Matter_1-3/index.html#ref_DataTypeTemperature

Therefore, this value will always be degC from the driver perspective because this is handling the cluster attribute value, which is always in C. 

The other temp conversion consideration you may be thinking of is due to receiving a capability command, not a cluster attribute. The capability command does not have a clear unit associated with the value, which is where we run into the issues.

So I think this is appropriate for the to use degC to emit the capability event because that is the unit of the value. The hubcore temperature conversion should then localize the unit to degF if needed before sending the value to the cloud to be displayed to the user.

CC: @nickolas-deboom

hcarter-775:
@HunsupJung I think it would be good to add a unit test to make sure the distinction in oven between C and F is working correctly?

samadDotDev:
Closing in favor of https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1640

PR #1578, opened 2024-08-13
CHAD-13561: Add support for FanControl's Rocking functionality

hcarter-775:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change
Originally a 1.3 addition, seen here:
[CHAD-13561](https://smartthings.atlassian.net/browse/CHAD-13561)
but was mentioned in this ticket:
[CHAD-13781](https://smartthings.atlassian.net/browse/CHAD-13781)

and was pushed ahead of the 1.3 launch, since it is not directly related to the 1.3 launch.

* Functionality has been added to matter thermostat to support FanControl's Rock feature using the fanOscillationMode capability.

* Further, a series of new profiles were built to support this feature. They are the same as previously made air purifier profiles, with the addition of the rocking capability.

* Implementation is almost identical to what was built for FanControl's Wind feature support.

New: 

* A series of Fan profiles has been added for more variation
* Logic has been included to support these
Note: 
"fan" should be "fan-wind" but is already used in production. I have created a new "fan-wind" profile with the same capabilities as "fan", and I have made a new "fan-generic" profile that is what "fan" probably should have been. Therefore,
when a device profiles as "fan", it will switch to "fan-generic". Hopefully, after enough driver re-initializations, "fan" will be
able to be deprecated, but until then no *new* profiles should ever be mapped to it again.    

# Summary of Completed Tests
* A new series of test cases were built for this feature, following the wind mode tests structure.
* More, an on-hand Air Purifier supports rocking and the updates were tested successfully on-device.

[CHAD-13561]: https://smartthings.atlassian.net/browse/CHAD-13561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CHAD-13781]: https://smartthings.atlassian.net/browse/CHAD-13781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -323,6 +333,7 @@ local function do_configure(driver, device)
   local thermo_eps = device:get_endpoints(clusters.Thermostat.ID)
   local fan_eps = device:get_endpoints(clusters.FanControl.ID)
   local wind_eps = device:get_endpoints(clusters.FanControl.ID, {feature_bitmap = clusters.FanControl.types.FanControlFeature.WIND})
+  local rock_eps = device:get_endpoints(clusters.FanControl.ID, {feature_bitmap = clusters.FanControl.types.FanControlFeature.ROCK})
```
nickolas-deboom, line 336: I looked at FanControl/types/Feature.lua in the lua libs, and it looks the feature name should be ROCKING rather than ROCK

hcarter-775: huh. you're so right. Weirdly, it onboarded correctly on a real device as well as passing the unit tests. should probably be changed though. good catch!

drivers/SmartThings/matter-thermostat/profiles/fan-generic.yml
```diff
@@ -0,0 +1,14 @@
+name: fan-generic
```
tpmanley, line 1: I think we should switch the generic fingerprint to this profile since not all fans will support wind

hcarter-775: agreed

hcarter-775: though with this dynamic profiling setup, any profile would be updated anyway. Still a good addition for clarity

drivers/SmartThings/matter-thermostat/profiles/air-purifier-ac-rock.yml
```diff
@@ -0,0 +1,26 @@
+name: air-purifier-ac-rock
```
tpmanley, line 1: Is it too late to change ac to something like carbon or cf? I just think it's going to be confusing to see "ac" and for it not to mean air conditioner

drivers/SmartThings/matter-thermostat/profiles/air-purifier-hepa-wind-rock.yml
```diff
@@ -0,0 +1,28 @@
+name: air-purifier-hepa-wind-rock
+components:
+- id: main
+  label: Main
+  capabilities:
+  - id: airPurifierFanMode
+    version: 1
+  - id: fanSpeedPercent
+    version: 1
+  - id: windMode
```
tpmanley, line 10: We should make the order of `windMode` and `fanOscillationMode` consistent in all the profiles that have both. It's different here than in `fan-wind-rock` for example. Unless we have other guidance, I'd say put `fanOscillationMode` before `windMode`

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -358,7 +369,19 @@ local function do_configure(driver, device)
...
   elseif device_type == FAN_DEVICE_TYPE_ID then
-    device.log.warn_with({hub_logs=true}, "Fan supports only one profile")
+    profile_name = "fan"
+    if #wind_eps > 0 then
+      profile_name = profile_name .. "-wind"
+    end
+    if #rock_eps > 0 then
+      profile_name = profile_name .. "-rock"
+    end
+    if profile_name == "fan" then
```
tpmanley, line 379: Why not default `profile_name` to "fan-generic" up above and then you can remove this `if`

drivers/SmartThings/matter-thermostat/profiles/air-purifier-ac-rock.yml
hcarter-775: it is too late I think. I agree, but this is what someone decided

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: If we did this, we'd be creating profiles like "fan-generic-wind", which I guess isn't bad sounding now that I think about it

drivers/SmartThings/matter-thermostat/profiles/air-purifier-hepa-wind-rock.yml
hcarter-775: Agreed, I will re-comb to make sure everything is consistent

drivers/SmartThings/matter-thermostat/src/init.lua
tpmanley: 🤦 didn't think of that. No, it's better as is

drivers/SmartThings/matter-thermostat/profiles/air-purifier-ac-rock.yml
hcarter-775: actually, I negate my statement. In [this superset query](https://superset.samsungiotanalytics.com/superset/sqllab?id=1155), I do not actually see any profiles using -ac yet, so we can change this.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -323,6 +333,7 @@ local function do_configure(driver, device)
   local thermo_eps = device:get_endpoints(clusters.Thermostat.ID)
   local fan_eps = device:get_endpoints(clusters.FanControl.ID)
   local wind_eps = device:get_endpoints(clusters.FanControl.ID, {feature_bitmap = clusters.FanControl.types.FanControlFeature.WIND})
```
nickolas-deboom, line 335: Does `FanControlFeature` need to be changed to `Feature`, or does aliasing take care of this?

hcarter-775: Aliasing will take care of this. However, I don't think it's the worst idea to go ahead and update all the places where this old feature name is used. Thermostat and Switch also use this old naming convention.

tpmanley (approved):

PR #1591, opened 2024-08-19
Matter-Switch: Improve color temp conversion rounding

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

[CHAD-13828](https://smartthings.atlassian.net/browse/CHAD-13828)

This change adjusts the way that the color temperature bounds are rounded while converting them from mireds into Kelvin. Previously, they were rounded to the nearest hundreds value and this sometimes resulted in the range displayed in the ST app being larger than the range supported by the device. This change will prevent this from occurring by shortening the range, if needed, to always be within the range supported by the device.

See [here](https://smartthings.atlassian.net/wiki/spaces/CHIP/pages/3386376304/Color+Temperature+Range+Selection+Proposal) for a comparison of the current rounding method to the method proposed by this PR.

Note that the `mired_to_kelvin` function is not used anywhere else in the matter-switch driver, so making these changes in this function won't affect any other functionality.

# Summary of Completed Tests

The unit test added by this PR tests a case where the minimum and maximum mired values get converted to Kelvin and are send out in the colorTemperatureRange capability. In the main branch, this test will fail because 165 mireds, which is equal to 6060.6K, would be rounded to 6100K, and 365 mireds, equal to 2739.7K would be rounded to 2700K. With the changes in this PR, 165 mireds is converted to 6000K and 365 mireds is converted to 2800K, a range that lies within the supported range in this scenario.

[CHAD-13828]: https://smartthings.atlassian.net/browse/CHAD-13828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -105,7 +105,7 @@ local function mired_to_kelvin(value, minOrMax)
   elseif minOrMax == COLOR_TEMP_MAX then
     return utils.round(MIRED_KELVIN_CONVERSION_CONSTANT / (kelvin_step_size * (value - 1)) - rounding_value) * kelvin_step_size
   else
-    log.warn("Attempted to convert temperature unit for an undefined value")
+    log.warn_with({hub_logs = true}, "Attempted to convert temperature unit for an undefined value")
   end
 end
 
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -96,11 +96,16 @@ local function mired_to_kelvin(value, minOrMax)
   -- bounds to be multiples of 100. For the maximum mired value (minimum K value),
   -- add 1 before converting and round up to nearest hundreds. For the minimum mired
   -- (maximum K value) value, subtract 1 before converting and round down to nearest
-  -- hundreds.
+  -- hundreds. Note that 1 is added/subtracted from the mired value in order to avoid
+  -- rounding errors from the conversion of Kelvin to mireds.
+  local kelvin_step_size = 100
+  local rounding_value = 0.5
   if minOrMax == COLOR_TEMP_MIN then
-    return utils.round(((MIRED_KELVIN_CONVERSION_CONSTANT / (value + 1)) / 100) + 0.5) * 100
+    return utils.round(MIRED_KELVIN_CONVERSION_CONSTANT / (kelvin_step_size * (value + 1)) + rounding_value) * kelvin_step_size
+  elseif minOrMax == COLOR_TEMP_MAX then
+    return utils.round(MIRED_KELVIN_CONVERSION_CONSTANT / (kelvin_step_size * (value - 1)) - rounding_value) * kelvin_step_size
   else
-    return utils.round(((MIRED_KELVIN_CONVERSION_CONSTANT / (value - 1)) / 100) - 0.5) * 100
+    log.warn("Attempted to convert temperature unit for an undefined value")
   end
 end
 
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -87,14 +87,21 @@ local function convert_huesat_st_to_matter(val)
   return utils.clamp_value(math.floor((val * 0xFE) / 100.0 + 0.5), CURRENT_HUESAT_ATTR_MIN, CURRENT_HUESAT_ATTR_MAX)
 end
 
-local function mired_to_kelvin(value)
+local function mired_to_kelvin(value, minOrMax)
   if value == 0 then -- shouldn't happen, but has
     value = 1
     log.warn(string.format("Received a color temperature of 0 mireds. Using a color temperature of 1 mired to avoid divide by zero"))
   end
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -87,14 +87,21 @@ local function convert_huesat_st_to_matter(val)
...
-  -- bounds to be multiples of 100
-  return utils.round((MIRED_KELVIN_CONVERSION_CONSTANT / value) / 100) * 100
+  -- We divide inside the rounding and multiply outside of it because we expect these
+  -- bounds to be multiples of 100. For the maximum mired value (minimum K value),
+  -- add 1 before converting and round up to nearest hundreds. For the minimum mired
+  -- (maximum K value) value, subtract 1 before converting and round down to nearest
+  -- hundreds.
+  if minOrMax == COLOR_TEMP_MIN then
+    return utils.round(((MIRED_KELVIN_CONVERSION_CONSTANT / (value + 1)) / 100) + 0.5) * 100
+  else
```
hcarter-775, line 102:
```suggestion
  elseif minOrMax == COLOR_TEMP_MAX then
  ...
  else
  -- add an error message here
  end
```
I think this makes it more clear what the expected values are.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -87,14 +87,21 @@ local function convert_huesat_st_to_matter(val)
...
-  -- we divide inside the rounding and multiply outside of it because we expect these
-  -- bounds to be multiples of 100
-  return utils.round((MIRED_KELVIN_CONVERSION_CONSTANT / value) / 100) * 100
+  -- We divide inside the rounding and multiply outside of it because we expect these
+  -- bounds to be multiples of 100. For the maximum mired value (minimum K value),
+  -- add 1 before converting and round up to nearest hundreds. For the minimum mired
+  -- (maximum K value) value, subtract 1 before converting and round down to nearest
+  -- hundreds.
+  if minOrMax == COLOR_TEMP_MIN then
+    return utils.round(((MIRED_KELVIN_CONVERSION_CONSTANT / (value + 1)) / 100) + 0.5) * 100
```
hcarter-775, line 101:
```suggestion
    return utils.round((100 * MIRED_KELVIN_CONVERSION_CONSTANT) / (value + 1) ) + 0.5) * 100
```

hcarter-775: and then I suggest doing a similar thing for the other case

hcarter-775: also, this comment does not note where the 1 comes from. would you mind adding this to the explanation you provided?

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated as suggested in the latest commit!

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated as suggested, and also added a comment explaining that the 1 is needed to avoid rounding errors. Consider the case of a value of 357 mireds, which is equal to 2801 K. Mired values are received from the device as integers and so if we simply round up to the nearest hundreds this would be 2900K. However, 357 mireds is the best representation a manufacturer could provide for a bulb that supports a minimum temp of 2800K so it should of course be rounded to 2800K Adding one to it first ensures that we avoid this rounding error (358 mireds is equal to 2793K which we would then round up to 2800K).

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -87,14 +87,26 @@ local function convert_huesat_st_to_matter(val)
...
+  -- hundreds. Note that 1 is added/subtracted from the mired value in order to avoid
+  -- rounding errors from the conversion of Kelvin to mireds.
+  local kelvin_step_size = 100
+  local rounding_value = 0.5
+  if minOrMax == COLOR_TEMP_MIN then
+    return utils.round(MIRED_KELVIN_CONVERSION_CONSTANT / (kelvin_step_size * (value + 1)) + rounding_value) * kelvin_step_size
+  elseif minOrMax == COLOR_TEMP_MAX then
+    return utils.round(MIRED_KELVIN_CONVERSION_CONSTANT / (kelvin_step_size * (value - 1)) - rounding_value) * kelvin_step_size
+  else
+    log.warn("Attempted to convert temperature unit for an undefined value")
```
hcarter-775, line 108: maybe use log.warn_with({hub_logs=true}, ...)?

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: great, thanks nick!

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated!

hcarter-775 (approved):

tpmanley (approved):

PR #1594, opened 2024-08-21
Matter Sensor: Streamline do_configure function in air quality sensor subdriver

hcarter-775:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
Refactor the air quality sensor configure logic for readability and clarity.

# Summary of Completed Tests
Unit test added and old unit tests continue to pass. Tested with the VDA air quality sensor as well.

### DIFF
--- a/drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
+++ b/drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
@@ -164,39 +164,64 @@ local supported_profiles =
   "aqs-temp-humidity-tvoc-level-pm25-meas",
 }
 
+local AIR_QUALITY_MAP = {
+  {capabilities.carbonMonoxideMeasurement.ID,    "-co",    clusters.CarbonMonoxideConcentrationMeasurement},
+  {capabilities.carbonMonoxideHealthConcern.ID,  "-co",    clusters.CarbonMonoxideConcentrationMeasurement},
+  {capabilities.carbonDioxideMeasurement.ID,     "-co2",   clusters.CarbonDioxideConcentrationMeasurement},
+  {capabilities.carbonDioxideHealthConcern.ID,   "-co2",   clusters.CarbonDioxideConcentrationMeasurement},
+  {capabilities.nitrogenDioxideHealthConcern.ID, "-no2",   clusters.NitrogenDioxideConcentrationMeasurement},
+  {capabilities.nitrogenDioxideMeasurement.ID,   "-no2",   clusters.NitrogenDioxideConcentrationMeasurement},
+  {capabilities.ozoneHealthConcern.ID,           "-ozone", clusters.OzoneConcentrationMeasurement},
+  {capabilities.ozoneMeasurement.ID,             "-ozone", clusters.OzoneConcentrationMeasurement},
+  {capabilities.formaldehydeMeasurement.ID,      "-ch2o",  clusters.FormaldehydeConcentrationMeasurement},
+  {capabilities.formaldehydeHealthConcern.ID,    "-ch2o",  clusters.FormaldehydeConcentrationMeasurement},
+  {capabilities.veryFineDustHealthConcern.ID,    "-pm1",   clusters.Pm1ConcentrationMeasurement},
+  {capabilities.veryFineDustSensor.ID,           "-pm1",   clusters.Pm1ConcentrationMeasurement},
+  {capabilities.fineDustSensor.ID,               "-pm25",  clusters.Pm25ConcentrationMeasurement},
+  {capabilities.fineDustHealthConcern.ID,        "-pm25",  clusters.Pm25ConcentrationMeasurement},
+  {capabilities.dustSensor.ID,                   "-pm10",  clusters.Pm10ConcentrationMeasurement},
+  {capabilities.dustHealthConcern.ID,            "-pm10",  clusters.Pm10ConcentrationMeasurement},
+  {capabilities.radonHealthConcern.ID,           "-radon", clusters.RadonConcentrationMeasurement},
+  {capabilities.radonMeasurement.ID,             "-radon", clusters.RadonConcentrationMeasurement},
+  {capabilities.tvocHealthConcern.ID,            "-tvoc",  clusters.TotalVolatileOrganicCompoundsConcentrationMeasurement},
+  {capabilities.tvocMeasurement.ID,              "-tvoc",  clusters.TotalVolatileOrganicCompoundsConcentrationMeasurement},
+}
+
+local function create_level_measurement_profile(device)
+  local meas_name, level_name = "", ""
+  for _, details in ipairs(AIR_QUALITY_MAP) do
+    local cap_id  = details[1]
+    local cluster = details[3]
+    -- capability describes either a HealthConcern or Measurement/Sensor
+    if (cap_id:match("HealthConcern$")) then
+      local attr_eps = embedded_cluster_utils.get_endpoints(device, cluster.ID, { feature_bitmap = cluster.types.Feature.LEVEL_INDICATION })
+      if #attr_eps > 0 then
+        level_name = level_name .. details[2]
+      end
+    elseif (cap_id:match("Measurement$") or cap_id:match("Sensor$")) then
+      local attr_eps = embedded_cluster_utils.get_endpoints(device, cluster.ID, { feature_bitmap = cluster.types.Feature.NUMERIC_MEASUREMENT })
+      if #attr_eps > 0 then
+        meas_name = meas_name .. details[2]
+      end
+    end
+  end
+  return meas_name, level_name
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
```diff
@@ -164,130 +164,104 @@ local supported_profiles =
...
+      local attr_eps = embedded_cluster_utils.get_endpoints(device, cluster.ID, { feature_bitmap = cluster.types.Feature.NUMERIC_MEASUREMENT })
+      if #attr_eps > 0 then
+        meas_name = meas_name .. details[2]
+      end
+    end
+  end
+  return meas_name, level_name
+end
+
+-- local function match(_string, _sub_string)
```
nickolas-deboom, line 211: Is this function needed?

hcarter-775: definitely not

drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
```diff
@@ -164,130 +164,100 @@ local supported_profiles =
...
     device.log.warn_with({hub_logs=true}, string.format("No matching profile for device. Tried to use profile %s", profile_name))
-    if #co_meas_eps > 0 or #no2_meas_eps > 0 or #ozone_meas_eps > 0 or #formaldehyde_meas_eps > 0 or
-        #pm1_meas_eps > 0 or #pm10_meas_eps > 0 or #radon_meas_eps > 0 then
+
+    local function meas_find(sub_name)
+      return string.match(meas_name, sub_name)
+    end
+
+    -- try to best match to existing profiles
+    if profile_name == "aqs-co" or meas_find("co-") or meas_find("no2") or meas_find("ozone") or meas_find("ch2o") or
```
ctowns, line 257: Should this just be `meas_find("co")` instead of `meas_find("co-")`?

```suggestion
    if profile_name == "aqs-co" or meas_find("co") or meas_find("no2") or meas_find("ozone") or meas_find("ch2o") or
```

hcarter-775: No, but I can leave a comment to explain that decision. If it were just `meas_find("co")` that would catch `co2` as well as `co`, which we don't want if we want to copy the current logic,

ctowns: Also, could you help me understand how the profile name `aqs-co` could be constructed from above? I'm not sure I understand why that profile is called out specifically in this conditional

hcarter-775: The one case where `co` wouldn't be caught with a - following it is when it's the only addition,aka`aqs-co`, which is why that is also checked.

hcarter-775: Yes! My second comment explained that, but I definitely think this should include a comment to describe this rare case

hcarter-775: @ctowns I have updated the explanation per our conversation, and I've edited the logic a little. I removed this line `profile_name == "aqs-co"` in place of `meas_name:match("co$")`, since the original conditional was actually a little wrong :sweat_smile: since it would've also needed to check the case swhere `profile_name` included the permutations `-humidity` as well as `-temp`.

drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
```diff
@@ -164,130 +164,103 @@ local supported_profiles =
...
+    profile_name = profile_name .. meas_name .. "-meas"
   end
 
   if not tbl_contains(supported_profiles, profile_name) then
     device.log.warn_with({hub_logs=true}, string.format("No matching profile for device. Tried to use profile %s", profile_name))
-    if #co_meas_eps > 0 or #no2_meas_eps > 0 or #ozone_meas_eps > 0 or #formaldehyde_meas_eps > 0 or
-        #pm1_meas_eps > 0 or #pm10_meas_eps > 0 or #radon_meas_eps > 0 then
+
+    local function meas_find(sub_name)
+      return string.match(meas_name, sub_name)
```
ctowns, line 253: You could optimize this to be
```
return string.match(meas_name, sub_name) ~= nil
```

Since we really only want to know if the substring is present, there's no need to return the found sub string.

drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
```diff
@@ -164,130 +164,103 @@ local supported_profiles =
...
+    end
+
+    -- try to best match to existing profiles
+    -- the check meas_find("co-") includes a dash to distinguish the "co" case from the "co2" case.
+    -- However, the one case where there wouldn't be a dash is when meas_name ends with "co".
+    -- To catch this edge case, we use this check- meas_name:match("co$").
+    if profile_name == meas_name:match("co$") or meas_find("co-") or meas_find("no2") or meas_find("ozone") or meas_find("ch2o") or
```
ctowns, line 260: I believe you could simplify this further by string matching this expression: `"co[^%d]?"`

This _should_ mean that we want to match any case where `co` appears and it is not (^) followed by a digit (%d), with the additional `?` at the end to account for cases where co is at the end of a string.

Lua doesn't support standard regular expressions, but it does have some functionality that you can read up on here: https://www.lua.org/manual/5.4/manual.html#6.4.1

I think distilling this to one search directly for "just co, not co2" would be more clear, along with a short comment explaining why. However, if you disagree I leave it to your discretion whichever is easier to understand.

drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
hcarter-775: I like it :+1:

drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
hcarter-775: I don't feel super strongly about it, and it would definitely shorten the commentary needed, so let's do it :+1: thanks for the lua resource!

drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
```diff
@@ -164,130 +164,101 @@ local supported_profiles =
...
     device.log.warn_with({hub_logs=true}, string.format("No matching profile for device. Tried to use profile %s", profile_name))
-    if #co_meas_eps > 0 or #no2_meas_eps > 0 or #ozone_meas_eps > 0 or #formaldehyde_meas_eps > 0 or
-        #pm1_meas_eps > 0 or #pm10_meas_eps > 0 or #radon_meas_eps > 0 then
+
+    local function meas_find(sub_name)
+      return string.match(meas_name, sub_name) ~= nil
+    end
+
+    -- try to best match to existing profiles
+    -- meas_find("co[^%d]?") matches the string to co and NOT co2.
```
ctowns, line 257: I know this isn't necessarily the scope of this PR, but since we are calling this out directly in the code it would be good to add a test case as an example of this working.

ctowns (approved):
This looks good to me! Only comment is related to adding a unit test case just to validate behavior that is called out in the code.

Also, just to be sure I would test with an air purifier or the VDA so we have a real device test case before merging. I know this is strictly a refactor, but it's still good to just do a smoke test to be sure!

nickolas-deboom (approved):

drivers/SmartThings/matter-sensor/src/air-quality-sensor/init.lua
hcarter-775: Update: that suggested check `meas_find("co[^%d]?")` doesn't work unfortunately, since the `not` part doesn't reject something that's there, it just only returns what would be returned minus the part that is ^ marked. Thus, matching `co2-` would return `co`. I was playing with using `".*co[^%d].*"` as a match, but it fails when the string ends with co. And adding a ? after [^%d] in the second version makes -co2 pass. So basically lua matching functions are not complex enough to permit co but not co2 in all cases. Extra logic is needed.

hcarter-775: This pair of checks does work though to catch the case we're discussing: `meas_find("co%-") or meas_find("co$")`

hcarter-775: I'm glad we made some unit tests- good nose on a possible bug :+1:

PR #1595, opened 2024-08-23
matter-lock: Expose 'not fully locked' when reported by the lock

tpmanley:
Previously we were mapping the 'not fully locked' to 'unknown' but now that the lock attribute has the 'not fully locked' variant we can expose that directly.

https://smartthings.atlassian.net/browse/CHAD-11259

# Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor

# Checklist

- [X] I have performed a self-review of my code
- [X] I have commented my code in hard-to-understand areas
- [X] I have verified my changes by testing with a device or have communicated a plan for testing

# Summary of Completed Tests

Tested with a Yale Matter lock

PR #1602, opened 2024-08-28
matter-lock: Add 'not fully locked' to the embedded device config

tpmanley:
Without this change the 'not fully locked' was being displayed but without the correct translations.

https://smartthings.atlassian.net/browse/CHAD-11259

# Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Summary of Completed Tests

Tested with a Yale lock

### DIFF
--- a/drivers/SmartThings/matter-lock/profiles/base-lock.yml
+++ b/drivers/SmartThings/matter-lock/profiles/base-lock.yml
@@ -10,7 +10,7 @@ components:
         enabledValues:
         - locked
         - unlocked
-        - unknown
+        - not fully locked
   - id: lockCodes
     version: 1
   - id: tamperAlert
--- a/drivers/SmartThings/matter-lock/profiles/lock-lockalarm-nobattery.yml
+++ b/drivers/SmartThings/matter-lock/profiles/lock-lockalarm-nobattery.yml
@@ -10,7 +10,7 @@ components:
         enabledValues:
         - locked
         - unlocked
-        - unknown
+        - not fully locked
   - id: lockAlarm
     version: 1
   - id: firmwareUpdate
--- a/drivers/SmartThings/matter-lock/profiles/lock-lockalarm.yml
+++ b/drivers/SmartThings/matter-lock/profiles/lock-lockalarm.yml
@@ -10,7 +10,7 @@ components:
         enabledValues:
         - locked
         - unlocked
-        - unknown
+        - not fully locked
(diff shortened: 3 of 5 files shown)


z-michel (approved):

Kwang-Hui (approved):
It works for U300

drivers/SmartThings/matter-lock/profiles/base-lock.yml
```diff
@@ -10,7 +10,7 @@ components:
         enabledValues:
         - locked
         - unlocked
-        - unknown
+        - not fully locked
```
ctowns, line 13: Is the expectation now that `unknown` events will just be ignored and not result in a change in the attribute state?

ctowns (commented):
This is currently set to be merged to the `beta` branch. I think this should be merged to `main` first and then cherry-picked to beta as needed if we choose to expedite it

Also, can you help me understand how this affects the translation used?

drivers/SmartThings/matter-lock/profiles/base-lock.yml
tpmanley: It's more that the devices will no longer generate the `unknown` event

ctowns: Oh I see, this is because of the change you had made here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1595

Thanks for the clarification!

tpmanley:
> This is currently set to be merged to the beta branch. I think this should be merged to main first and then cherry-picked to beta as needed if we choose to expedite it

I did it this way because I want it to go to beta since the change to start creating not fully locked events is already in beta. After merging I would merge beta back down to main.

tpmanley:
> Also, can you help me understand how this affects the translation used?

@ctowns the `enabledValues` in the embedded device config drive the `alternatives` in the presentation and those are what are used to find the translations. When `not fully locked` was missing from `enabledValues` it was also not added to `alternatives` so the value was directly displayed instead of the translation.

ctowns (approved):
Thank you for the explanation Tom! Looks good to me 👍 

There should be a post in #edge-driver-development mentioning that this is being expedited to beta. I would also clarify the merging beta -> main strategy in the message since this is the first time we have done it

hcarter-775 (approved):

PR #1603, opened 2024-08-28
[Aqara] K100 - add supportedAlarmValues

seojune79:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [o] New feature
- [ ] Refactor

# Checklist

- [o] I have performed a self-review of my code

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/Aqara/aqara-lock/src/init.lua
+++ b/drivers/Aqara/aqara-lock/src/init.lua
@@ -53,6 +53,7 @@ local function remoteControlShow(device)
 end
 
 local function comp_supported_alarm_values(last_alarm_values)
+  if not last_alarm_values then return false end
   if #last_alarm_values~=#SUPPORTED_ALARM_VALUES then return false end
   for k, v in pairs(last_alarm_values) do
     if SUPPORTED_ALARM_VALUES[k]~=v then return false end
@@ -61,8 +62,8 @@ local function comp_supported_alarm_values(last_alarm_values)
 end
 
 local function device_init(self, device)
-  local last_alarm_values = device:get_latest_state("main", LockAlarm.ID, LockAlarm.supportedAlarmValues.NAME) or nil
-  if last_alarm_values == nil or not comp_supported_alarm_values(last_alarm_values) then
+  local last_alarm_values = device:get_latest_state("main", LockAlarm.ID, LockAlarm.supportedAlarmValues.NAME) or {}
+  if not comp_supported_alarm_values(last_alarm_values) then
     device:emit_event(
       LockAlarm.supportedAlarmValues(SUPPORTED_ALARM_VALUES, { visibility = { displayed = false } })
     )
@@ -259,6 +260,7 @@ local aqara_locks_handler = {
   NAME = "Aqara Doorlock K100",
   supported_capabilities = {
     Lock,
+    LockAlarm,
     Battery,
     lockCredentialInfo,
     capabilities.refresh,
--- a/drivers/Aqara/aqara-lock/src/init.lua
+++ b/drivers/Aqara/aqara-lock/src/init.lua
@@ -23,6 +23,7 @@ local seq_num = 0
 
 local SHARED_KEY = "__shared_key"
 local CLOUD_PUBLIC_KEY = "__cloud_public_key"
+local SUPPORTED_ALARM_VALUES = { "damaged", "forcedOpeningAttempt", "unableToLockTheDoor", "notClosedForALongTime", "highTemperature", "attemptsExceeded" }
 
 local function my_secret_data_handler(driver, device, secret_info)
   if secret_info.secret_kind ~= "aqara" then return end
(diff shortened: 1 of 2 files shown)


drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -51,6 +52,23 @@ local function remoteControlShow(device)
   end
 end
 
+local function comp_supported_alarm_values(last_alarm_values)
+  if #last_alarm_values~=#SUPPORTED_ALARM_VALUES then return false end
```
cjswedes, line 56: `#last_alarm_values` will error the device init handler if the device does not have the field set.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -51,6 +52,23 @@ local function remoteControlShow(device)
...
+  for k, v in pairs(last_alarm_values) do
+    if SUPPORTED_ALARM_VALUES[k]~=v then return false end
+  end
+  return true
+end
+
+local function device_init(self, device)
+  local last_alarm_values = device:get_latest_state("main", LockAlarm.ID, LockAlarm.supportedAlarmValues.NAME) or nil
```
cjswedes, line 64:
```suggestion
  local last_alarm_values = device:get_latest_state("main", LockAlarm.ID, LockAlarm.supportedAlarmValues.NAME) or {}
```

drivers/Aqara/aqara-lock/src/init.lua
cjswedes: A nil check here is probably warranted.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -78,11 +108,19 @@ end
...
+    id = nil
+    label = nil
+  elseif evt_value == 0x80020000 then -- one-time password
     id = "OTP_STANDALONE"
     label = nil
   else
-  id, label = credential_utils.find_userLabel(driver, device, evt_value)
+    id, label = credential_utils.find_userLabel(driver, device, evt_value)
   end
   device:emit_event(Lock.lock.unlocked({ data = { method = evt_name, codeId = id, codeName = label } }))
```
tpmanley, line 125: Do you still want to emit the event even when `evt_name` is nil? Or maybe still emit the event but without `data`?

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -23,6 +23,7 @@ local seq_num = 0
 
 local SHARED_KEY = "__shared_key"
 local CLOUD_PUBLIC_KEY = "__cloud_public_key"
+local SUPPORTED_ALARM_VALUES = { "damaged", "forcedOpeningAttempt", "unableToLockTheDoor", "notClosedForALongTime", "highTemperature", "attemptsExceeded" }
```
tpmanley, line 26: Should "clear" be in here as well?

Kwang-Hui: actually the `clear` is hidden intentionally. lockAlarm capability doesn't represent as state. - e.g. highTemerapture, damaged can be occurs together - so this attribute works as an event.
and in some cases, the `clear` event is could be ambiguous for users - the damage fixed?, no one trying forced opening attempt anymore?.

So the `clear` won't generate push notification. and hidden at history for this device.

drivers/Aqara/aqara-lock/src/init.lua
seojune79: Okay, I'll add a nil check.

drivers/Aqara/aqara-lock/src/init.lua
seojune79: I'll fix it.

drivers/Aqara/aqara-lock/src/init.lua
seojune79: There are no cases where `evt_name` is nil in the code, but should I add code to double-check it?

tpmanley: Lines 113 and 117 are setting `evt_name` to `nil` so yes I think there should be code to protect against it

seojune79: I set that part to “nil” at Samsung's request.

Kwang-Hui: @tpmanley, Actually I asked Aqara team not to send `codeId`, `codeName`, `method` in case of BLUETOOTH, RF447 which was intended "emit the event but without `data`" just like you mentioned.
Do you think we need to seprate `emit_event` implementation?

Kwang-Hui: @tpmanley , we have updated the `lock` capability to have more `method` types. So I think it is better to fill the method as `proprietaryRemote` with `nil` of `codeId`, `codeName` in case of bluetooth and RF447 use cases.
Could you let me know your opinion?

tpmanley (approved):

PR #1620, opened 2024-09-06
fix: Avoid a nil index crash when data loss occurs

dljsjr:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

This fixes an edge case where we were assuming that a persistent field on a device was always present; while this should more or less hold true, there are times when it won't. In those cases, the driver would crash.

Drivers in this state would already be nonfunctional, but we want to avoid chewing up resources with a hot crash loop.

# Summary of Completed Tests

I tested this locally by onboarding my Hue setup and then deliberaly taking actions that would lead to the persisted fields getting blown away; the driver is nonfunctional, but no longer hits a hot crash loop.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
+++ b/drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
@@ -270,7 +270,7 @@ end
 
 local function noop_event_emitter(device, ...)
   local label = (device and device.label) or "Unknown Device Name"
-  local device_type = (device and device:get_field(Fields.DEVICE_TYPE)) or "Unknown Device Type"
+  local device_type = (device and utils.determine_device_type(device)) or "Unknown Device Type"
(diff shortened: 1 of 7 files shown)


varzac:
It's not obvious to me where the assumed key was causing a npe.  Can you clarify where exactly that was happening?

dljsjr:
@varzac It was a missing IP address, actually. This line was the cause of the original crash report: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1620/files#diff-5a04f9b727d9d403a969e8b24dff2d9162b47cd213254faf3f910be2697c24c4L252

The Application Key isn't the only "important" piece of data that can be lost, but it's the only one that's not easy to recover; the other fixes in this PR address other locations where we were trying to access Fields that could have been lost; they were additional crashes or pathological/resource consuming cases that showed up after I fixed the initial crash.

varzac (approved):

PR #1622, opened 2024-09-09
Matter Bridge Aqara Cube Support with Custom Capability

DongHoon-Ryu:
Currently, the Generic EdgeDriver applied to the Matter Bridge Aqara Cube is implemented by mapping six faces into individual components, making it difficult to check the event occurrence of individual faces on one screen.

This commit wants to improve the problem by mapping the event from Matter Bridge with Aqara Cube Custom Capability developed by Zigbee driver. This means that there is a limitation that the Action Event of the Cube that the Matter Bridge does not give can not be processed.

REQ-15926, REQ-16285, IOTE-4217, IOTE-4266

Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [x] Refactor

# Checklist

# Description of Change
This commit wants to improve the problem by mapping the event from Matter Bridge with Aqara Cube Custom Capability developed by Zigbee driver. This means that there is a limitation that the Action Event of the Cube that the Matter Bridge does not give can not be processed.

# Summary of Completed Tests
- Successful test for Plugin UI update and Dash Board's Device Card status update according to the selection of each face
- Successful Automation Routine Test for Cube Face
- Even though the Cube Face is changed, the status is sometimes not updated because the Aqara Matter Bride does not send the event.


drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,217 @@
...
+    for _, ep in ipairs(MS) do -- for each momentary switch endpoint (including main)
+      device.log.debug("Configuring endpoint "..ep)
+      -- build the mapping of endpoints to components
+      component_map[string.format("%d", current_component_number)] = ep
+      current_component_number = current_component_number + 1
+    end
+
+    device:set_field(COMPONENT_TO_ENDPOINT_MAP, component_map, {persist = true})
+    device:try_update_metadata({profile = "cube-t1-pro"})
+    configure_buttons(device)
```
nickolas-deboom, line 149: If we are updating the profile name and deferring configuration of the buttons until the profile is updated, then shouldn't `configure_buttons` not be called here? In the main driver, if a new profile is going to be applied, `configure_buttons` is not called until `info_changed` is called after the profile update.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,217 @@
...
+    local map = device:get_field(COMPONENT_TO_ENDPOINT_MAP) or {}
+    local face = 1
+    for component, ep in pairs(map) do
+      if map[component] == ib.endpoint_id then
+        face = component
+        break
+      end
+    end
+
+    reset_thread(device)
```
nickolas-deboom, line 182: What is the purpose of `reset_thread` exactly? I can't find any other instances of it being used in the drivers so I might be missing something

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,217 @@
+local capabilities = require "st.capabilities"
+local clusters = require "st.matter.generated.zap_clusters"
+local device_lib = require "st.device"
+
+local cubeAction = capabilities["stse.cubeAction"]
+local cubeFace = capabilities["stse.cubeFace"]
+
+local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
```
nickolas-deboom, line 8: When the button driver was brought into the switch driver, a new field was created for the component to endpoint map for button devices. `__component_to_endpoint_map` still exists for legacy switch devices that still use MCD, but will not be used for new switch devices. Could you please update this to use the new field introduced for button MCD devices, `__component_to_endpoint_map_button`?

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,217 @@
+local capabilities = require "st.capabilities"
+local clusters = require "st.matter.generated.zap_clusters"
+local device_lib = require "st.device"
+
+local cubeAction = capabilities["stse.cubeAction"]
```
nickolas-deboom, line 5: Will these capabilities remain as custom, or will they be implemented into https://github.ecodesamsung.com/iot/SmartThingsCapabilities ?

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,217 @@
...
+          end
+        end
+        idx1 = idx1 + 1
+        previous = ep
+      end
+
+      local start = #ept2 + 1
+      idx1 = 1
+      idx2 = start
+      for i=start, 6 do
```
nickolas-deboom, line 93: I think I understand most of this function but I am a little confused about this section. In an example where the given endpoints are [250, 251, 2, 3, 4, 5], then I believe before this loop, `ept1` would contain [250, 251] and `ept2` would contain [2, 3, 4, 5]. Then, this loop would append the contents of `ept1` to `ept2` so that the function will return the final list of [2, 3, 4, 5, 250, 251]. Is my understanding correct?

nickolas-deboom:
Hi @DongHoon-Ryu I went over the changes and left some comments. Outside of that, the changes look good to me!

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: When the hub is restarted, `info_changed` is not called and `configure_buttons` cannot be called, so it is also added here. Is there anything I checked wrong?

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: The purpose is to display the `No action` status on the plugin or device card 3 seconds after the `cubeAction `occurs.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: It is already implemented in EdgeDriver for Zigbee.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: Sorry for causing confusion. [250, 251, 2, 3, 4, 5] is the order of endpoint numbers that are actually mapped to the Aqara Cube from Face1 to Face6. And, since the logic works after the `table.sort (MS)`, the `ept1` includes [2, 3, 4, 5] and the `ep2` includes [250, 251]. In fact, regardless of whether the `table.sort (MS)` is called or not, this function was required because the result of `device:get_endpoints` is already [2, 3, 4, 5, 250, 251].
In my opinion, if 'device:get_endpoints' is improved for this case, the `local function get_reordered_endpoints (driver, device)` function is not necessary.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: Thanks, I updated it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
nickolas-deboom: I see now, makes sense!

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
nickolas-deboom: My understanding is that `configure_buttons` only needs to be called the first time a device is initialized and `info_changed` is called after the profile is joined, so I don't understand the need to call it in `device_added`. Is there a reason it would need to be called when the hub restarts?

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
nickolas-deboom: Understood, thank you

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
nickolas-deboom: Ok thank you! I understand now, I think I had it backwards before.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
nickolas-deboom: Ok I see. I glanced at that driver and was wondering if this driver should also include the following capability events in `device_added`?
`device:emit_event(cubeAction.cubeAction("noAction"))
device:emit_event(cubeFace.cubeFace("face1Up"))`

DongHoon-Ryu: Yes, it is simply to set the initial state.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I checked again and it seems to be a little different from matter-button's lifecycle event. In conclusion, you were right, so I deleted the `configure_buttons` call from `device_added`. Thank you very much~

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
nickolas-deboom: Sorry, I should have been more clear. I meant should those capability events be added in `device_added`? In order to set the initial state as you said. Or is that not needed in this case?

DongHoon-Ryu: Oh, I see. I think it only needs to be in one of `device_added` or `info_changed`. Because even when adding a device, after DEFERRED_CONFIGURE anyway, it will be a real operation, so I decided that this place would be better. If it has already been added, plugin may be updated again depending on the actual last status.

drivers/SmartThings/matter-switch/capabilities/cubeAction.yaml
```diff
@@ -0,0 +1,29 @@
+id: stse.cubeAction
```
hcarter-775, line 1: nit: can you make these .yml files

drivers/SmartThings/matter-switch/capabilities/cubeFace.yaml
```diff
@@ -0,0 +1,25 @@
+id: stse.cubeFace
```
hcarter-775, line 1: nit: can you make these .yml files

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,216 @@
+local capabilities = require "st.capabilities"
+local clusters = require "st.matter.generated.zap_clusters"
```
hcarter-775, line 2:
```suggestion
local clusters = require "st.matter.clusters"
```

drivers/SmartThings/matter-switch/capabilities/cubeAction.yaml
hcarter-775: Actually, what is the plan on releasing the plugins for [these capabilties](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/tree/main/drivers/Aqara/aqara-cube/capabilities)? Further, since we are now using an [updated lua_libs version](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/releases/tag/apiv11_beta) for our CI, these files _should_ no longer be necessary to include

drivers/SmartThings/matter-switch/profiles/cube-t1-pro.yml
hcarter-775, line 1: nit: please make a .yml file

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,216 @@
...
+local cubeFace = capabilities["stse.cubeFace"]
+
+local COMPONENT_TO_ENDPOINT_MAP_BUTTON = "__component_to_endpoint_map_button"
+local DEFERRED_CONFIGURE = "__DEFERRED_CONFIGURE"
+local INITIAL_PRESS_ONLY = "__initial_press_only" -- for devices that support MS (MomentarySwitch), but not MSR (MomentarySwitchRelease)
+
+local CUBEACTION_TIMER = "cubeAction_timer"
+local CUBEACTION_TIME = 3
+
+local function is_aqara_cube(opts, driver, device)
```
hcarter-775, line 15: will there be other Matter aqara cubes in the future that may use this driver?

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,216 @@
...
+
+local function battery_percent_remaining_attr_handler(driver, device, ib, response)
+  if ib.data.value then
+    device:emit_event(capabilities.battery.battery(math.floor(ib.data.value / 2.0 + 0.5)))
+  end
+end
+
+-- [[ register ]]
+local aqara_cube_t1_pro_handler = {
+  NAME = "Aqara Cube T1 Pro",
```
hcarter-775, line 195: since the function `is_aqara_cube` is so general, as well as the aqara-cube subdriver name, should these also be more general?
```suggestion
local aqara_cube_handler = {
  NAME = "Aqara Cube Handler",
```

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,216 @@
...
+  end
+end
+
+local function battery_percent_remaining_attr_handler(driver, device, ib, response)
+  if ib.data.value then
+    device:emit_event(capabilities.battery.battery(math.floor(ib.data.value / 2.0 + 0.5)))
+  end
+end
+
+-- [[ register ]]
```
hcarter-775, line 193: nit: remove comment

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,216 @@
...
+    reset_thread(device)
+    device:emit_event(cubeFace.cubeFace("face1Up"))
+
+    -- profile has changed, and we deferred setting up our buttons, so do that now
+    configure_buttons(device)
+    device:set_field(DEFERRED_CONFIGURE, nil)
+  end
+end
+
+-- initial press
```
hcarter-775, line 169: nit: remove, function name describes this well

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,216 @@
...
+local device_lib = require "st.device"
+
+local cubeAction = capabilities["stse.cubeAction"]
+local cubeFace = capabilities["stse.cubeFace"]
+
+local COMPONENT_TO_ENDPOINT_MAP_BUTTON = "__component_to_endpoint_map_button"
+local DEFERRED_CONFIGURE = "__DEFERRED_CONFIGURE"
+local INITIAL_PRESS_ONLY = "__initial_press_only" -- for devices that support MS (MomentarySwitch), but not MSR (MomentarySwitchRelease)
+
+local CUBEACTION_TIMER = "cubeAction_timer"
```
hcarter-775, line 12:
```suggestion
local CUBEACTION_TIMER = "__cubeAction_timer"
```
This is generally our admittedly arbitrary syntax

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,216 @@
...
+local CUBEACTION_TIME = 3
```
hcarter-775, line 13: would you mind leaving a comment explaining this functionality?

drivers/SmartThings/matter-switch/capabilities/cubeAction.yaml
DongHoon-Ryu: The file extension has been modified. And the capsabilities have already been registered in Zigbee.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I changed it.

drivers/SmartThings/matter-switch/profiles/cube-t1-pro.yml
DongHoon-Ryu: The yml file is already registered, so I don't understand exactly what it means. Please give me an additional explanation.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I added comments. This is added to use the same UI as Aqara Cube for Zigbee. That is, after 3 seconds of `cubeAction`, to automatically change the action status of Plugin UI or Device Card to `noAction`. As you know, we do not receive the `cubeAction` event with Matter, but when the side of the cube is changed, I have changed the `cubeAction` status.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I changed it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I removed it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I removed it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: Okay, I changed it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: i think it will be enough to be `info_changed` only.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,224 @@
...
+    -- At the time of device_added, there is an error that the corresponding capability cannot be found
+    -- because the profile has not been changed from the generic profile of fingerprint to the Aqara Cube.
+    reset_thread(device)
+    device:emit_event(cubeFace.cubeFace("face1Up"))
+
+    device:set_field(DEFERRED_CONFIGURE, nil)
+
+    -- In the current test framework, the values of device.profile.id and args.old_st_store.profile.id are always the same,
+    -- so the test coverage cannot be increased. This is why I added DOING_CONFIGURE flag.
+    device:set_field(DOING_CONFIGURE, true)
```
DongHoon-Ryu, line 169: @nickolas-deboom @hcarter-775 Hi, 
Although already described in the inline comments, I could not find a way to test the code in case of DEFERRED_CONFIGURE in the current test framework. This is because the values of `device.profile.id` and `args.old_st_store.profile.id` are always equal to `00000000-1111-2222-3333-0000000002`. Therefore, I added a variable called `DOING_CONFIGURE`, which can be changed immediately in the test file, but the test case is intermittently failed because the capability information is not updated in time when calling emit_event . Is there any way to solve it?

hcarter-775: Yes, we have a function to solve this called `test.wait_for_events()` that should fix the intermittent failing. Simply place this function before the event that would need the capability information to be updated and this _should_ fix it, based on your problem description.

DongHoon-Ryu: Yes, I've already used that method. However, as you said, I can always succeed if I call as below, but I can not satisfy test coverage. This is because the `received lifecycle event: infoChanged` has already ended, so the `mock_device:set_field (DOING_CONFIGURE, true)` that is called after that has no meaning.
>test.socket.device_lifecycle:__queue_receive(mock_device:generate_info_changed({value = "face1Up"}))
>test.wait_for_events()

Or when I call as below, the test case still fails intermittently as mentioned above.
>test.socket.device_lifecycle:__queue_receive(mock_device:generate_info_changed({value = "face1Up"}))
>mock_device:set_field(DOING_CONFIGURE, true)
>test.wait_for_events()

DongHoon-Ryu: @hcarter-775 Thank you very much. According to your guide, I add the contents below,
>test.add_package_capability("cubeAction.yml")
>test.add_package_capability("cubeFace.yml")
And I was able to raise the test coverage to 91% by `setting_field` before `info_changed` as below.
>mock_device:set_field(DOING_CONFIGURE, true)
>test.socket.device_lifecycle:__queue_receive(mock_device:generate_info_changed({value = "face1Up"}))
> test.wait_for_events()

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,224 @@
...
+    event = {
+      [clusters.Switch.ID] = {
+        [clusters.Switch.events.InitialPress.ID] = initial_press_event_handler
+      }
+    },
+  },
+  can_handle = is_aqara_cube
+}
+
+return aqara_cube_handler
```
hcarter-775, line 224: nit: can you please add a newline here?

drivers/SmartThings/matter-switch/profiles/cube-t1-pro.yml
```diff
@@ -0,0 +1,21 @@
...
+        version: 1
+      - id: firmwareUpdate
+        version: 1
+      - id: refresh
+        version: 1
+    categories:
+      - name: RemoteController
+metadata:
+  mnmn: SolutionsEngineering
+  vid: SmartThings-smartthings-Aqara_CubeT1Pro
```
hcarter-775, line 21: nit: please add a newline here

DongHoon-Ryu: I added it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I added it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -0,0 +1,225 @@
...
+    -- so the test coverage cannot be increased. This is why I added DOING_CONFIGURE flag.
+    device:set_field(DOING_CONFIGURE, true)
+  end
+
+  if device:get_field(DOING_CONFIGURE) then
+    -- profile has changed, and we deferred setting up our buttons, so do that now
+    configure_buttons(device)
+    device:set_field(DOING_CONFIGURE, nil)
+  end
+end
```
hcarter-775, line 177:
```suggestion
-- used in unit testing, since device.profile.id and args.old_st_store.profile.id are always the same 
local TEST_CONFIGURE = "__test_configure"
... 

local function info_changed(driver, device, event, args)
  if (device.profile.id ~= args.old_st_store.profile.id or device:get_field(TEST_CONFIGURE))
    and device:get_field(DEFERRED_CONFIGURE)
    and device.network_type ~= device_lib.NETWORK_TYPE_CHILD then

    -- At the time of device_added, there is an error that the corresponding capability cannot be found
    -- because the profile has not been changed from the generic profile of fingerprint to the Aqara Cube.
    reset_thread(device)
    device:emit_event(cubeFace.cubeFace("face1Up"))

     -- profile has changed, and we deferred setting up our buttons, so do that now
    configure_buttons(device)
    
    device:set_field(DEFERRED_CONFIGURE, nil)
  end
end
```
This may add a little bit more clarity to what is going on, without the need for as much explanation.

hcarter-775:
Just a reminder @DongHoon-Ryu, please squash your commits before merging into main. Thank you very much.

PR #1624, opened 2024-09-09
Add new-matter-lock driver

HunsupJung:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change
Adds a new sub-driver with updates to the door lock functionality. Leverages new capabilities and some new design architecture for the Aqara door locks.

# Summary of Completed Tests
[Completed Testing 9/27](https://smartthings.atlassian.net/wiki/spaces/CHIP/pages/3439755284/Matter+Lock+Driver+Regression+Test?atlOrigin=eyJpIjoiNDBmOGFhODY4ZGUzNDIzZGE3OTQ3MTZkNDZmNmRmNzciLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ)
[Completed Testing 10/18](https://smartthings.atlassian.net/wiki/spaces/CHIP/pages/3509092641/Matter+Lock+Driver+Regression+Test+Result+18-Oct-2024)


drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1801 @@
...
+    device:get_field(lock_utils.COTA_CRED),
+    nil, -- nil user_index creates a new user
+    DoorLock.types.DlUserStatus.OCCUPIED_ENABLED,
+    DoorLock.types.DlUserType.REMOTE_ONLY_USER
+  ))
+end
+
+local function generate_cota_cred_for_device(device)
+  local len = device:get_latest_state("main", capabilities.lockCredentials.ID, capabilities.lockCredentials.maxPinCodeLen.NAME) or 6
+  local cred_data = math.floor(math.random() * (10 ^ len))
```
tpmanley, line 280: We should seed the PRNG with `math.randomseed(os.time())` like the other drivers that use `math.random()` although it looks like that wasn't done in the original lock driver either.

FreeMasen: random is pre-seeded at startup, no need to reseed here

drivers/SmartThings/matter-lock/src/aqara-lock/init.lua
FreeMasen, line 1: in the main `init.lua` this needs to be removed from `sub_drivers` because `require` cannot resolve `aquara-lock`. I am also not seeing any references to the new code

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1801 @@
...
+}
+
+local subscribed_events = {
+  [capabilities.lock.ID] = {
+    DoorLock.events.LockOperation
+  },
+  [capabilities.lockAlarm.ID] = {
+    DoorLock.events.DoorLockAlarm
+  },
+  [capabilities.lockUser.ID] = {
```
FreeMasen, line 68:
```suggestion
  [capabilities.lockUsers.ID] = {
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
tpmanley: I'm observing that sometimes math.random() returns the same value on subsequent driver runs. I'll file that under a different ticket though.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1615 @@
...
+        state_change = true,
+        visibility = {displayed = false}
+      }
+    )
+    device:emit_event(event)
+    device:set_field(lock_utils.BUSY_STATE, false, {persist = true})
+    return
+  end
+
+  if elements.next_credential_index.value ~= nil then
```
tpmanley, line 1044: This isn't handling the COTA PIN correctly in at least a couple of ways:
1. The user type should be `DlUserType.REMOTE_ONLY_USER`
2. `credData` is `nil` because it is getting it from the `CRED_DATA` field instead of `COTA_CRED`

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1615 @@
...
+    device:set_field(lock_utils.BUSY_STATE, false, {persist = true})
+    return
+  end
+
+  -- Update commandResult
+  status = "occupied"
+  if elements.status.value == DoorLock.types.DlStatus.FAILURE then
+    status = "failure"
+  elseif elements.status.value == DoorLock.types.DlStatus.DUPLICATE then
+    status = "duplicate"
```
tpmanley, line 1019: We should add handling for `duplicate` codes, at least for COTA PINs. If the lock returns the duplicate error code we should create a new randomized COTA PIN and use that instead. I think it's OK if that doesn't land in this initial PR but if it doesn't please create a ticket to follow up on it.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1615 @@
...
+-- Matter Handler
+----------------
+-- Lock State --
+----------------
+local function lock_state_handler(driver, device, ib, response)
+  local LockState = DoorLock.attributes.LockState
+  local attr = capabilities.lock.lock
+  local LOCK_STATE = {
+    [LockState.NOT_FULLY_LOCKED] = attr.not_fully_locked(),
+    [LockState.LOCKED] = attr.locked({visibility = {displayed = false}}),
```
tpmanley, line 152: Why is `displayed = false` here?

drivers/SmartThings/matter-lock/profiles/lock-user-pin-schedule.yml
```diff
@@ -0,0 +1,27 @@
...
+    version: 1
+  - id: remoteControlStatus
+    version: 1
+  - id: lockUsers
+    version: 1
+  - id: lockCredentials
+    version: 1
+  - id: lockSchedules
+    version: 1
+  - id: refresh
```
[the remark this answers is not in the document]
HunsupJung: I'll fix it

drivers/SmartThings/matter-lock/profiles/lock-user-pin.yml
```diff
@@ -0,0 +1,25 @@
...
+        - not fully locked
+  - id: lockAlarm
+    version: 1
+  - id: remoteControlStatus
+    version: 1
+  - id: lockUsers
+    version: 1
+  - id: lockCredentials
+    version: 1
+  - id: refresh
```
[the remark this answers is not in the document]
HunsupJung: I'll fix it

drivers/SmartThings/matter-lock/profiles/lock-user-schedule.yml
```diff
@@ -0,0 +1,25 @@
...
+        - not fully locked
+  - id: lockAlarm
+    version: 1
+  - id: remoteControlStatus
+    version: 1
+  - id: lockUsers
+    version: 1
+  - id: lockSchedules
+    version: 1
+  - id: refresh
```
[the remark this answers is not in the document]
HunsupJung: I'll fix it

drivers/SmartThings/matter-lock/profiles/lock-user.yml
```diff
@@ -0,0 +1,23 @@
...
+        - locked
+        - unlocked
+        - not fully locked
+  - id: lockAlarm
+    version: 1
+  - id: remoteControlStatus
+    version: 1
+  - id: lockUsers
+    version: 1
+  - id: refresh
```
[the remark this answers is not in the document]
HunsupJung: I'll fix it

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: the change of lock state occurs twice from lock_state_handler and lock_op_event_handler. so I wanted to add "displayed is false". but now I updated it to add call_with_delay to prevent duplicated event occurs.

lelandblue (dismissed):
Hello @HunsupJung before merging this code , please provide tested results as communicated via slack.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+    end
+  else
+    profile_name = "base-lock"
+  end
+  device.log.info(string.format("Updating device profile to %s.", profile_name))
+  device:try_update_metadata({profile = profile_name})
+end
+
+local function info_changed(driver, device, event, args)
+  for cap_id, attributes in pairs(subscribed_attributes) do
```
ctowns, line 132: This behavior for checking the driver subscribed attributes and adding them if the device supports the corresponding capability is already supported by the [populate_subscribed_attributes_from_driver_defaults() function](https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/main/lua_libs/st/matter/device.lua#L224) that is called each time `device:subscribe()` is called, so just calling `device:subscribe()` in this case should be enough to handle this.

Also, you should check first that the profile changed before resubscribing, or else we'd end up just resubscribing to the same attributes/events as before:
```
  device:subscribe()
end
```

HunsupJung: @ctowns 
Thank you for your explanation~! I'll fix this part

HunsupJung: Thank you for your explanation. but when the above patch was applied, the driver did not work.
I think it is because main driver does not have some capabilities used by new-matter-lock driver. the subscribed attributes and event list in main driver is different with the list in sub driver.

HunsupJung:
@lelandblue @ctowns 
I would like to merge this patch before the next release. I hope you can help me to meet the schedule.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+  }
+
+  -- The lock state is usually updated in lock_state_handler and lock_op_event_handler, respectively.
+  -- In this case, two events occur. To prevent this, when both functions are called,
+  -- it send the event after 1 second so that no event occurs in the lock_state_handler.
+  device.thread:call_with_delay(1, function ()
+    if ib.data.value ~= nil then
+      device:emit_event(LOCK_STATE[ib.data.value])
+    else
+      device:emit_event(attr.unknown())
```
tpmanley, line 169: If `unknown` is going to be used here then it also needs to be added to `enabledValues` in the profiles. Otherwise it won't be translated correctly when that event is emitted.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+      }
+    )
+    device:emit_event(event)
+    return
+  end
+
+  -- Save values to field
+  device:set_field(lock_utils.BUSY_STATE, true, {persist = true})
+  device:set_field(lock_utils.COMMAND_NAME, cmdName, {persist = true})
+  -- device:set_field(lock_utils.USER_INDEX, userIdx, {persist = true}) -- needs to find empty index
```
tpmanley, line 643: Is this needed or can it be removed?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+-----------------------
+-- Set User Response --
+-----------------------
+local function set_user_response_handler(driver, device, ib, response)
+  -- Get result
+  local cmdName = device:get_field(lock_utils.COMMAND_NAME)
+  local userIdx = device:get_field(lock_utils.USER_INDEX)
+  local userType = device:get_field(lock_utils.USER_TYPE)
+  local status = "success"
+  if ib.status == DoorLock.types.DlStatus.FAILURE then
```
tpmanley, line 724: It'd be good to log these error cases

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+
+-------------------------
+-- Clear User Response --
+-------------------------
+local function clear_user_response_handler(driver, device, ib, response)
+  -- Get result
+  local cmdName = device:get_field(lock_utils.COMMAND_NAME)
+  local userIdx = device:get_field(lock_utils.USER_INDEX)
+  local status = "success"
+  if ib.status == DoorLock.types.DlStatus.FAILURE then
```
tpmanley, line 837: These error cases should be logged too

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: Okay I'll check it and modify.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+--- If a device needs a cota credential this function attempts to set the credential
+--- at the index provided. The set_credential_response_handler handles all failures
+--- and retries with the appropriate index when necessary.
+local function set_cota_credential(device, credential_index)
+  local eps = device:get_endpoints(DoorLock.ID)
+  local cota_cred = device:get_field(lock_utils.COTA_CRED)
+  if cota_cred == nil then
+    -- Shouldn't happen but defensive to try to figure out if we need the cota cred and set it.
+    device:send(DoorLock.attributes.RequirePINforRemoteOperation:read(device, #eps > 0 and eps[1] or 1))
+    device.thread:call_with_delay(2, function(t) set_cota_credential(device, credential_index) end)
```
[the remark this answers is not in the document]
HunsupJung: I referred to main driver. I will check and modify it.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I got it. I will add a log.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: It can be removed.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I got it~!

drivers/SmartThings/matter-lock/fingerprints.yml
```diff
@@ -32,4 +32,4 @@ matterGeneric:
     deviceLabel: Matter Door Lock
     deviceTypes:
       - id: 0x000A # Door Lock
-    deviceProfileName: base-lock
+    deviceProfileName: base-lock
```
hcarter-775, line 35: please add a newline, thanks!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+        state_change = true,
+        visibility = {displayed = false}
+      }
+    )
+    device:emit_event(event)
+    return
+  end
+
+  -- Save values to field
+  device:set_field(lock_utils.BUSY_STATE, true, {persist = true})
```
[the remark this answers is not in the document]
HunsupJung: I agree with you. I'll add that process

drivers/SmartThings/matter-lock/fingerprints.yml
HunsupJung: I got it~!

HunsupJung:
@tpmanley @hcarter-775 
I'll go to work in about 6 hours and fix the parts you reviewed ~! Thank you so much.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
ctowns: Thanks for trying that Hunsup! On further investigation it looks like the issue is that the `populate_subscribed_attributes_from_driver_defaults()` function only uses the subscribed attributes that are listed in the parent driver, not the sub driver. Since this is a sub driver, then the the updated `subscribed_attributes` that are defined in the `new-matter-lock` sub driver are not used. 

So there are two ways to fix this:
1. We can use the code you have added here and it will make sure that the new subscribed attributes in the sub driver are checked and added to device's subscribed attribute list when `info_changed` is run.
2. We expand the `subscribed_attributes` list in the parent driver to include the new attributes that are in this driver. I think this is the better long term solution because the `device_subscribe()` call utilizes the `subscribed_attributes` defined in the parent driver to determine which attributes a device should subscribe to. Therefore, the parent driver's `subscribed_attributes` list should be the "master list" for all attributes that could be subscribed to in the driver, including in the sub drivers. Therefore, all you would need to do is combine the subscribed attributes and subscribed events lists in the sub driver with the lists that already exist in the parent driver. This would look like this:
```
subscribed_attributes = {
     -- attributes already defined in parent driver
    [capabilities.lock.ID] = {
      DoorLock.attributes.LockState
    },
    [capabilities.battery.ID] = {
      PowerSource.attributes.BatPercentRemaining
    },
    [capabilities.batteryLevel.ID] = {
      PowerSource.attributes.BatChargeLevel
    },
    -- new attributes from new-matter-lock subdriver
    [capabilities.remoteControlStatus.ID] = {
      DoorLock.attributes.OperatingMode
    },
    [capabilities.lockUsers.ID] = {
      DoorLock.attributes.NumberOfTotalUsersSupported
    },
      DoorLock.attributes.MinPINCodeLength,
      DoorLock.attributes.RequirePINforRemoteOperation
    },
    [capabilities.lockSchedules.ID] = {
      DoorLock.attributes.NumberOfWeekDaySchedulesSupportedPerUser,
      DoorLock.attributes.NumberOfYearDaySchedulesSupportedPerUser
    }
  },
 subscribed_events = {
    -- events already defined in parent driver
    [capabilities.tamperAlert.ID] = {
      DoorLock.events.DoorLockAlarm,
      DoorLock.events.LockOperation
    },
    [capabilities.lockAlarm.ID] = {
      DoorLock.events.DoorLockAlarm
    },
    [capabilities.lockCodes.ID] = {
      DoorLock.events.LockUserChange
    },
    -- new events from new-matter-lock subdriver
    [capabilities.lock.ID] = {
      DoorLock.events.LockOperation
    },
    [capabilities.lockUsers.ID] = {
      DoorLock.events.LockUserChange
    }
  },
```

Either way should work in this case, but I think long term we should move to having all subscribed attributes and events listed in the parent driver so that they do not need to be redefined in the sub drivers. The parent driver's list would serve as a "master list" of all the subscribed attributes and events that could be used in that driver, and the sub drivers can inherit those attributes accordingly!

samadDotDev: Is there a way we can append to the parent driver's `subscribed_attributes` list from a subdriver? I'm thinking that the pattern for the parent driver to contain all attributes/events that a subdriver may be interested in isn't sustainable and can also be problematic if a subdriver only contains manufacturer specific cluster attributes/events that general devices driven from other subdrivers or parent one may not be able to service. 

Also to note that the `device::subscribe()` is idempotent if the same attributes/events are resubscribed so solution (1) should be fine even if it ends up running multiple times.

hcarter-775: I think then that it may be a good idea to add the new capabilities / attributes into the main driver's subscribed attributes table? That way, this behavior can be handled by `device:subscribe()`.

hcarter-775: However, even if you elect not to do this, I do believe that this check: `if device.profile.id ~= args.old_st_store.profile.id then` should be addded around the handling, since it is unrelated to the issue you mentioned and it stops us from attempting to resubscribe on every `info_changed` event.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+end
+-------------------------------------
+-- Number Of Total Users Supported --
+-------------------------------------
+local function total_users_supported_handler(driver, device, ib, response)
+  device:emit_event(capabilities.lockUsers.totalUsersSupported(ib.data.value, {visibility = {displayed = false}}))
+end
+
+----------------------------------
+-- Number Of PIN User Supported --
```
hcarter-775, line 204:
```suggestion
-- Number Of PIN Users Supported --
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+    table.insert(new_cred_table, entry)
+  end
+
+  -- Add new entry to table
+  table.insert(new_cred_table, {userIndex = userIdx, credentialIndex = credIdx, credentialType = credType})
+  device:emit_event(capabilities.lockCredentials.credentials(new_cred_table, {visibility = {displayed = false}}))
+end
+
+local function delete_credential_from_table(device, credIdx)
+  -- If Credential Index is ALL_INDEX, remove all entry from the table
```
hcarter-775, line 436:
```suggestion
  -- If Credential Index is ALL_INDEX, remove all entries from the table
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: @samadDotDev we do not currently support the functionality you suggested, but I do think that's a good idea!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+  local new_user_table = {}
+
+  -- Recreate user table
+  local i = 0
+  for index, entry in pairs(user_table) do
+    if entry.userIndex == userIdx then
+      i = index
+    end
+    table.insert(new_user_table, entry)
+  end
```
hcarter-775, line 380: why are we recreating the table here? Could we not just handle it like so:
```suggestion
  local i = 0
  for index, entry in pairs(user_table) do
    if entry.userIndex == userIdx then
      i = index
      break
    end
  end
 
new_user_table = user_table
```

Where I'm not even sure we would to reinitialize `user_table` like this: `new_user_table = user_table`. Could we not just keep using `user_table`?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+    capabilities.lockUsers.users.NAME
+  ) or {}
+  local new_user_table = {}
+
+  -- Recreate user table
+  for index, entry in pairs(user_table) do
+    if entry.userIndex ~= userIdx then
+      table.insert(new_user_table, entry)
+    end
+  end
```
hcarter-775, line 409: Similar to above, would it not be more efficient to call this:
```suggestion
  -- Recreate user table
  local i = 0
  for index, entry in pairs(user_table) do
    if entry.userIndex ~= userIdx then
      i = index
      break
    end
  end

new_user_table = table.remove(user_table, i)
```
or something like that?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
ctowns: > Is there a way we can append to the parent driver's subscribed_attributes list from a subdriver? 
Good question! I think this can be done in a few ways. We can add subscribed attributes directly to the driver's subscribed attributes list as done above through the `add_subscribed_attribute/event()` function mentioned above. Also, we could access the driver's subscribed attribute/event list directly through the devices table via `self.driver.subscribed_attributes` or `self.driver.subscribed_events`. This is already done [here](https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/889c2977afe7a3b4821eddbf6907518a4ac9a7fb/lua_libs/st/matter/device.lua#L243). However, since this is in the device table this should only change the list on a device-by-device basis. At that point, I think just directly calling the `add_subscribed_attribute` call as needed would be easier.

FWIW, in the matter-appliance driver this style of including a "master list" in the parent driver is already in place: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/40c1b690507068caf852ce0a82c36ca5d778f0c0/drivers/SmartThings/matter-appliance/src/init.lua#L59. Each device type is broken into its own sub driver with handlers for each of the relevant attributes implemented in the sub driver. Also, it should be noted that even if all these attributes are listed in on master list, only the attributes that match to a capability that the device has on it's profile will actually be subscribed to. For example, a refrigerator would completely skip over the attributes related to the `laundryWasherRiseMode` since it would not have that capability in its profile.

Also, [in our public docs is it specifically stated that the `subscribed_attributes` and `subscribed_events` are not supported in sub drivers](https://developer.smartthings.com/docs/edge-device-drivers/matter/driver.html#subscribed-events). I think this support could be added via a lua libs change to add in "official" support, but either of the options proposed above also work for our purposes. Even in the case of a manufacturer-specific cluster, we can just leave that cluster out of the main list and instead directly call `device:add_subscribed_attribute()` in the sub driver to set up the subscription when `device:subscribe()` is called. However, I think additional support for these cases in the lua libs could help prevent some of the inefficiency of sending subscirbe requests for attributes that don't exist on the device in the case that the device is using a capbality that is overloaded by multiple different mutually exclusive attributes, such as the `mode` cluster in `matter-appliance`. The subscribes still work correctly and only subscribe to the relevant attributes on the report, but adding support defining subscribed attributes/events specific to a sub driver could prevent small inconsistencies like this.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+    "main",
+    capabilities.lockCredentials.ID,
+    capabilities.lockCredentials.credentials.NAME
+  ) or {}
+  local new_cred_table = {}
+
+  -- Recreat credential table
+  for index, entry in pairs(cred_table) do
+    table.insert(new_cred_table, entry)
+  end
```
hcarter-775, line 428: Last one- why are we rebuilding the table? Thank you!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+  supported_capabilities = {
+    capabilities.lock,
+    capabilities.lockUsers,
+    capabilities.lockCredentials,
+    capabilities.lockSchedules
+  },
+  can_handle = is_new_matter_lock_products
+}
+
+return new_matter_lock_handler
```
hcarter-775, line 1580: nit: newline please!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+  -- Set the credential to a code
+  device:set_field(lock_utils.BUSY_STATE, true, {persist = true})
+  device:set_field(lock_utils.COMMAND_NAME, "addCota")
+  device:set_field(lock_utils.CRED_INDEX, credential_index)
+  device:set_field(lock_utils.SET_CREDENTIAL, credential_index)
+  device:set_field(lock_utils.USER_TYPE, "remote")
+  device.log.info(string.format("Attempting to set COTA credential at index %s", credential_index))
+  device:send(DoorLock.server.commands.SetCredential(
+    device,
+    #eps > 0 and eps[1] or 1,
```
nickolas-deboom, line 261: Should `eps` be sorted before this like we do in `find_default_endpoint`?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+
+  -- Get latest week day schedule table
+  local week_schedule_table = device:get_latest_state(
+    "main",
+    capabilities.lockSchedules.ID,
+    capabilities.lockSchedules.weekDaySchedules.NAME
+  ) or {}
+  local new_week_schedule_table = {}
+
+  -- Find shcedule list
```
nickolas-deboom, line 510:
```suggestion
  -- Find schedule list
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: If the device has two or more lock clusters, it seems that we need to find an endpoint that supports COTA. However, I think it will be okay as it is because there is no possibility that such a device will come out in my personal opinion.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
ctowns: To summarize, I think we can use the solution you already have here @HunsupJung . This way we can avoid adding the attributes/events to the parent driver as @samadDotDev  had suggested. However, I think you should still add `if device.profile.id ~= args.old_st_store.profile.id then` before calling subscribe since we wouldn't need to resubscribe unless there was a profile change.

Also, if we are going to use this solution rather than adding the attributes to the parent driver, then you will also need to add the code that adds the new subscribed_attributes and subscribed_events every time before `device:subscribe()` is called. Since subscribe is also called in `device_init() `on line 102, you will also need to add the same code there:
```
  device:set_component_to_endpoint_fn(component_to_endpoint)
  for cap_id, attributes in pairs(subscribed_attributes) do
    if device:supports_capability_by_id(cap_id) then
      for _, attr in ipairs(attributes) do
        device:add_subscribed_attribute(attr)
      end
    end
  end
  for cap_id, events in pairs(subscribed_events) do
    if device:supports_capability_by_id(cap_id) then
      for _, e in ipairs(events) do
        device:add_subscribed_event(e)
      end
    end
  end
  device:subscribe()
 end
```

I have created a ticket to add more official support for subscribed attributes in sub drivers here: https://smartthings.atlassian.net/browse/CHAD-14110

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1580 @@
...
+  end
+  device.log.warn(string.format("Did not find default endpoint, will use endpoint %d instead", device.MATTER_DEFAULT_ENDPOINT))
+  return res
+end
+
+local function component_to_endpoint(device, component_name)
+  return find_default_endpoint(device, clusters.DoorLock.ID)
+end
+
+local function device_init(driver, device)
```
ctowns, line 100: You'll need to make this change as I suggested in the comment here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1624#discussion_r1791044889

```suggestion
local function device_init(driver, device)
local function device_init(driver, device)
  device:set_component_to_endpoint_fn(component_to_endpoint)
  for cap_id, attributes in pairs(subscribed_attributes) do
    if device:supports_capability_by_id(cap_id) then
      for _, attr in ipairs(attributes) do
        device:add_subscribed_attribute(attr)
      end
    end
  end
  for cap_id, events in pairs(subscribed_events) do
    if device:supports_capability_by_id(cap_id) then
      for _, e in ipairs(events) do
        device:add_subscribed_event(e)
      end
    end
  end
  device:subscribe()
 end
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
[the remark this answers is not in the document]
HunsupJung: @tpmanley 
I brought the code from main driver without thinking.
I reviewed it again, and I think we can delete `thread:call_with_delay` part and put return as like below.
```
  if cota_cred == nil then
    -- Shouldn't happen but defensive to try to figure out if we need the cota cred and set it.
    device:send(DoorLock.attributes.RequirePINforRemoteOperation:read(device, #eps > 0 and eps[1] or 1))
    return
```
And I think it needs to be modified in the main driver. Please check what you think.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I applied it. please confirm the updating

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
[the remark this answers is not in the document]
tpmanley: That makes sense to me. @ctowns @hcarter-775 any concerns?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1549,7 +1507,6 @@ local new_matter_lock_handler = {
       [capabilities.lock.commands.unlock.NAME] = handle_unlock,
     },
     [capabilities.lockUsers.ID] = {
-      [capabilities.lockUsers.commands.addUser.NAME] = handle_add_user,
```
tpmanley: Why was this removed again? `addUser` is a command in the capability so I think we should have a command handler for it

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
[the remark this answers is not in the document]
ctowns: I agree, this change makes sense to me. the `require_remote_pin_handler()` will end up calling `set_cota_credential()` to handle the response for the `RequirePINforRemoteOperation` attribute.

drivers/SmartThings/matter-lock/fingerprints.yml
```diff
@@ -33,3 +33,4 @@ matterGeneric:
     deviceTypes:
       - id: 0x000A # Door Lock
     deviceProfileName: base-lock
+
```
hcarter-775, line 36: nit: remove this newline please!

hcarter-775, line 36: nit: please remove the trailing whitespace here- thank you!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1558 @@
...
+  local eps = device:get_endpoints(DoorLock.ID)
+  local cota_cred = device:get_field(lock_utils.COTA_CRED)
+  if cota_cred == nil then
+    -- Shouldn't happen but defensive to try to figure out if we need the cota cred and set it.
+    device:send(DoorLock.attributes.RequirePINforRemoteOperation:read(device, #eps > 0 and eps[1] or 1))
+    device.thread:call_with_delay(2, function(t) set_cota_credential(device, credential_index) end)
+  elseif not cota_cred then
+    device.log.debug("Device does not require PIN for remote operation. Not setting COTA credential")
+    return
+  end
```
hcarter-775, line 269: I think this can be removed since `not cota_cred` and `cota_cred == nil` are basically equivalent here.

hcarter-775: Or, if you are trying to catch the case `cota_cred == false`, I would make that explicit here.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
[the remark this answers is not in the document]
hcarter-775: Yes I agree with Cooper. To make his point more explicit, `require_remote_pin_handler` calls `apply_cota_credentials_if_absent` which then calls `set_cota_credential`. Therefore, either way, `set_cota_credential` is getting called one way or another.

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -426,19 +426,22 @@ end
...
     device.log.info(string.format("Delaying scanning since currently checking credential %d", device:get_field(lock_utils.CHECKING_CREDENTIAL)))
     device.thread:call_with_delay(2, function(t) handle_reload_all_codes(driver, device, command) end)
     return
   end
-  device:emit_event(capabilities.lockCodes.scanCodes("Scanning"))
-  device:send(
-    clusters.DoorLock.server.commands.GetCredentialStatus(
-      device, device:component_to_endpoint(command.component),
-      {credential_type = DoorLock.types.DlCredentialType.PIN, credential_index = device:get_field(lock_utils.CHECKING_CREDENTIAL)}
+  device.thread:call_with_delay(5, function(t)
```
ctowns, line 436: For my understanding, why was this delay added here? Was it discovered during testing that locks need additional time to respond to reloading the codes?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I brought the code from main driver without thinking. but I think this handles the case where `cota_credit == false` and logs that COTA is not set. I think we can keep this code.

drivers/SmartThings/matter-lock/src/init.lua
HunsupJung: When the driver is changed, all field values for COTA are initialized. (lock_utils.COTA_CRED)
So the driver creates a new pin for COTA. and I added 5 seconds because a new PIN had to be created and reloaded.

When proceeding without delay, there is an issue that the pin for COTA is duplicated because reloading is performed before the pin for COTA is created.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: In fact, the `adduser` command is not used when creating a new user. All use `addCredential` to add a new user. And there is no `userindex` parameter in the `adduser` command, so driver needs to find and create an empty user index. this process is complicated.
If you think I have to maintain the function, please let me know.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I did not know how to use table api, so I implemented it as above.
If a new entry is added to the `cred_table` as an insert, there is an issue that the existing entries information disappears, so I made it as above. I implemented it as below, could you tell me which part is the problem?

```
local function add_credential_to_table(device, userIdx, credIdx, credType)
  -- Get latest credential table
  local cred_table = device:get_latest_state(
    "main",
    capabilities.lockCredentials.ID,
    capabilities.lockCredentials.credentials.NAME
  ) or {}

  -- Add new entry to table
  table.insert(cred_table, {userIndex = userIdx, credentialIndex = credIdx, credentialType = credType})
  device:emit_event(capabilities.lockCredentials.credentials(cred_table, {visibility = {displayed = false}}))
end
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: Ok that makes sense. Can you change this line from `not cota_cred` to `cota_cred == false`? I think it would be much more clear what is expected.

drivers/SmartThings/matter-lock/fingerprints.yml
hcarter-775: Sorry @HunsupJung I know it's a small thing, but I think there is still some whitespace here. You can see if there is or is not based on whether the red circle with a minus shows up at the bottom of the file.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: This is true of all relevant credential functions. Though I don't know all of the details, the call to `device:get_latest_state` is pulling something out of memory that is not mutable in-place, which the function `table.insert()` attempts to do. Without getting into the specifics of _why_ this is the case, I'm gonna propose a couple alternatives.

First, I looked into the [lockCredentials capability definition](https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/main/capabilities/lock-credentials.src/lock-credentials.yml) and it seems to supports a series of commands that directly handle this memory such as `addCredential`, `deleteCredential`, etc. I believe these could be leveraged to handle these systems.

In the case that these cannot be leveraged for whatever reason, I believe we can use the `deep_copy` function like so: `utils.deep_copy(device:get_latest_state())` which will do the same functionality as you describe here- creating a deep copy. This is currently used in the Aqara lock credential handler, see [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/Aqara/aqara-lock/src/credential_utils.lua#L49).

Again, I think the capability commands SHOULD do all of this for us, but in the case that it does not, we should leverage `deep_copy` here. Thanks!

cc: @ctowns so he knows this is still an open item

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: Hi Hunsup, out of curiosity do you know where `addCredential` is being leveraged?

hcarter-775: Please ignore this- I was checked out on the wrong branch. Apologies for the confusion.

tpmanley: @HunsupJung I do think we need a handler for all the commands in the capability since other things besides the app can call capability commands and since `addUser` is a command on this capability it should be implemented.

ctowns:
@HunsupJung could you add more unit tests for the new-matter-lock driver? Currently the code coverage for this driver is very low, so we should add additional test cases to increase the code coverage. You should be able to copy the test cases from the `test_matter_lock_cota.lua` and `test_matter_lock_codes.lua` files as a example, and make the necessary changes to account for the new capabilities (such as using `lockAlarm` instead of `tamperAlert`, and including the new capabilities such as `lockCredentials'). The structure should be similar to what is done in the other test files, but you will need to swap in the new capabilities. If you need additional help with this, please let me know and I can help!

drivers/SmartThings/matter-lock/fingerprints.yml
HunsupJung: Oh, I'm sorry. I thought it was modified, but I didn't merge properly. I will modify it carefully again.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I got it~!

drivers/SmartThings/matter-lock/fingerprints.yml
HunsupJung: I checked here and there's no whitespace. could you check it again by cherry-pick?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: Thank you for helping me~! I modified these parts related to the table api and it works well.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I modified it. Thank you~!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I modified it. Thank you~!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I updated addUser command and some sequence.

drivers/SmartThings/matter-lock/fingerprints.yml
hcarter-775: all good, thanks! Sorry about that

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: I'm sorry, did you make this change?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1670 @@
...
+    capabilities.lockUsers.ID,
+    capabilities.lockUsers.users.NAME,
+    {}
+  ))
+
+  -- Re-create user table
+  for index, entry in pairs(user_table) do
+    if entry.userIndex == userIdx then
+      table.remove(user_table, index)
+      break
```
hcarter-775, line 431: Is it possible that a userIdx could be accessed multiple times? I guess, to be clear, should we be `break`ing here?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1670 @@
...
+
+  -- Get latest user table
+  local user_table = utils.deep_copy(device:get_latest_state(
+    "main",
+    capabilities.lockUsers.ID,
+    capabilities.lockUsers.users.NAME,
+    {}
+  ))
+
+  -- Re-create user table
```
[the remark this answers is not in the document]
hcarter-775: Or just whatever the behavior here should be- which is no longer re-creating the table (:

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1670 @@
...
+  ) or {}
+  local new_cred_table = {}
+
+  -- Re-create credential table
+  for index, entry in pairs(cred_table) do
+    if entry.userIndex ~= userIdx then
+      table.insert(new_cred_table, entry)
+    end
+  end
+
```
hcarter-775, line 502: we can also use deep_copy here!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: oh I missed it. I will fix it.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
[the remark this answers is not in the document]
HunsupJung: Okay~! I will modify it.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I don't understand what you mean. Do you mean we should check all entries and remove the entries that has same user index?
we don't need to check all entries. the all user indexes are unique so if we find the entry with same user index, we don't need to check rest entries anymore.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I didn't use deep_copy on purpose here. I think the above way is good because we need to check all entries and exclude credentials with the same user index.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: ah, perfect. you answered my question, thank you!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: I see, good point

hcarter-775 (approved):

hcarter-775 (approved):

hcarter-775:
@lelandblue I think I added the testing you requested that's blocking the merge. Was there anything else you needed for this?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1670 @@
...
+    -- Shouldn't happen but defensive to try to figure out if we need the cota cred and set it.
+    device:send(DoorLock.attributes.RequirePINforRemoteOperation:read(device, #eps > 0 and eps[1] or 1))
+    return
+  elseif cota_cred == false then
+    device.log.debug("Device does not require PIN for remote operation. Not setting COTA credential")
+    return
+  end
+
+  if device:get_field(lock_utils.SET_CREDENTIAL) ~= nil then
+    device.log.debug("delaying setting COTA credential since a credential is currently being set")
```
tpmanley, line 271: When I test with a lock that uses COTA and the initial attempt to set the COTA PIN fails with a "duplicate" error, then it gets stuck in a loop logging this message and failing to ever retry setting the COTA PIN.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1670 @@
...
+  device:send(DoorLock.server.commands.ClearUser(device, ep, ALL_INDEX))
+end
+
+-------------------------
+-- Clear User Response --
+-------------------------
+local function clear_user_response_handler(driver, device, ib, response)
+  -- Get result
+  local cmdName = device:get_field(lock_utils.COMMAND_NAME)
+  local userIdx = device:get_field(lock_utils.USER_INDEX)
```
ctowns, line 920: I think you should use the userIdx provided in the clearUser command args instead of this device field. https://librarian-regionals.smartthingsgdev.com/matter/Matter_1-3/appclusters.html#_clearuser_command

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -0,0 +1,1670 @@
...
+        state_change = true,
+        visibility = {displayed = false}
+      }
+    )
+    device:emit_event(event)
+    return
+  end
+
+  -- Save values to field
+  device:set_field(lock_utils.COMMAND_NAME, cmdName, {persist = true})
```
ctowns, line 992: @HunsupJung I understand how you are using these fields to save information to be used in a different function, but could you help me understand why we need to do it this way? Do the commands themselves not have the relevant information that we can use in the handler? It seems this is the case for things like the ClearUser command, but do we really need to save all of this information or can be pull some of it from the command args and ib elements?

Also, it seems like there could be issues if two commands that rely on a field like `USER_INDEX` are processed at the same time. For example, if one command sets `USER_INDEX` to 1 and then before we handle that response, another command sets `USER_INDEX` to 2 , the original value would be overwritten and could cause problems in handling the first command that expected `USER_INDEX` to be 1. Are there cases where this could happen?

HunsupJung: > Also, it seems like there could be issues if two commands that rely on a field like USER_INDEX are processed at the same time. For example, if one command sets USER_INDEX to 1 and then before we handle that response, another command sets USER_INDEX to 2 , the original value would be overwritten and could cause problems in handling the first command that expected USER_INDEX to be 1. Are there cases where this could happen?

Therefore, I used busy_state to proceed with only one command at a time.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1065,6 +1068,15 @@ end
 -----------------------------
 -- Set Credential Response --
 -----------------------------
+local RESPONSE_STATUS_MAP = {
+  [DoorLock.types.DlStatus.FAILURE] = "failure",
+  [DoorLock.types.DlStatus.DUPLICATE] = "duplicate",
+  [DoorLock.types.DlStatus.OCCUPIED] = "occupied",
+  [DoorLock.types.DlStatus.INVALID_FIELD] = "invalidCommand",
+  [DoorLock.types.DlStatus.RESOURCE_EXHAUSTED] = "resourceExhausted",
+  [DoorLock.types.DlStatus.NOT_FOUND] = "failure"
```
tpmanley, line 1077: Should this be "notFound" instead of "failure"?

tpmanley: I guess not because "notFound" is not a valid value for the `statusCode` property in the capability

HunsupJung: Yes, there's no `notFound` value for capability

ctowns:
Thank you for writing the new unit tests @HunsupJung ! I think this looks like it is ready to go.

One last thing - were you able to run the regression tests listed [here](https://smartthings.atlassian.net/wiki/spaces/CHIP/pages/3439755284/Matter+Lock+Driver+Regression+Test) with the new code changes? I know you have been doing a lot of testing over the past few days, so I think many of these cases have been covered but I just want to make sure the testing is good to go.

Otherwise, I think this PR is ready to go. Thank you for your hard work on this 💪 !

HunsupJung:
@ctowns 
Thank you for checking. I updated the regression tests [here](https://smartthings.atlassian.net/wiki/spaces/CHIP/pages/3509092641/Matter+Lock+Driver+Regression+Test+Result+18-Oct-2024).
It's all thanks to the help of you and the US team.

ctowns (approved):
Thank you for your hard work on this @HunsupJung !

ctowns:
In the future, we should squash these commits before merging. This will make cherry-picking easier and also improve our git history.

cc: @HunsupJung

PR #1630, opened 2024-09-12
Update Github Actions Artifacts version

hcarter-775:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -154,7 +154,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Upload event file artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v3
         with:
           name: event-file
           path: ${{ github.event_path }}
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -154,7 +154,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Upload event file artifact
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         with:
           name: event-file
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -154,7 +154,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Upload event file artifact
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
```
ctowns, line 157: Should this be v4? That seems to be the latest version: https://github.com/actions/upload-artifact

However, v3 should be fine as well, just maybe less future proof. FWIW it looks like there are breaking change in v4 that might need to be consdiered if we upgrade that far: https://github.com/actions/upload-artifact/tree/main?tab=readme-ov-file#breaking-changes

Whereas no breaking changes are mentioned for v3: https://github.com/actions/upload-artifact/tree/v3

Also could you try temporarily making this change in your PR and see if it addresses the CI issue?

greens: I don't _think_ it will cause an issue, depending on what the scope of "same name" is for artifacts.

greens (approved):

.github/workflows/run-tests.yml
hcarter-775: Yeah v4 seems like it would be a bigger lift, and we use v3 all over the place already so this is more consistent. Also, see that the change fixed the PR [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1563).

tpmanley (approved):

ctowns (approved):

PR #1646, opened 2024-09-24
CI: Increase timeout for getting all channels

greens:
As the number of channels has grown, it sometimes takes the API longer than 5s to return all the channels. This should be a little bandage on that issue.


greens:
ping

ctowns (approved):

PR #1660, opened 2024-10-01
Add Dynamic Battery Support for Base-Lock Profiled Door Lock Devices 

hcarter-775:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change
Jira ticket showing mis-profiling of a bridged door lock: https://smartthings.atlassian.net/browse/MTR-821
Further, there is a general ticket on adding dynamic profiling to this driver: https://smartthings.atlassian.net/browse/CHAD-13948

This logic will permit dynamic profiling on the base-lock profile depending on the Power Source cluster, and will ignore devices that do not auto-profile as base-lock.

# Summary of Completed Tests
Unit tests
Bridged Switchbot device - Passed
VDA Door Lock - Passed
VDA Door Lock (Lock Codes) - Passed
Yale Assure Matter Lock - Passed

### DIFF
--- a/drivers/SmartThings/matter-lock/src/init.lua
+++ b/drivers/SmartThings/matter-lock/src/init.lua
@@ -197,9 +197,6 @@ local function set_credential_response_handler(driver, device, ib, response)
     if device:get_field(lock_utils.NONFUNCTIONAL) and cota_cred_index == credential_index then
       device.log.info("Successfully set COTA credential after being non-functional")
       device:set_field(lock_utils.NONFUNCTIONAL, false, {persist = true})
-
-      -- TODO: Add check here for battery
-
       device:try_update_metadata({profile = "base-lock", provisioning_state = "PROVISIONED"})
     end
   elseif device:get_field(lock_utils.COTA_CRED) and credential_index == device:get_field(lock_utils.COTA_CRED_INDEX) then
@@ -523,19 +520,28 @@ local function component_to_endpoint(device, component_name)
 end
 
 local function do_configure(driver, device)
+  local fingerprinted_devices = {
+    { vendor_id = 0x115F, product_id = 0x2802 }, -- Aqara Smart Lock U200
+    { vendor_id = 0x115F, product_id = 0x2801 }, -- Aqara Smart Lock U300
+    { vendor_id = 0x129F, product_id = 0x0001 }, -- Level Lock Plus (Matter)
+    { vendor_id = 0x135d, product_id = 0xb1 }, -- Nuki Smart Lock Pro
+    { vendor_id = 0x135d, product_id = 0xb0 }, -- Nuki Smart Lock
+  }
+
+  -- check if device has a wwst fingerprint
+  if device.manufacturer_info ~= nil then
+    for _, fingerprint in ipairs(fingerprinted_devices) do
+      if device.manufacturer_info.vendor_id == fingerprint.vendor_id and
+        device.manufacturer_info.product_id == fingerprint.product_id then
+        return
+      end
+    end
+  end
+
+  -- if not  fingerprinted, dynamically configure base-lock profile based on Power Source cluster checks
   local power_source_eps = device:get_endpoints(clusters.PowerSource.ID)
   local battery_feature_eps = device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
-  local pin_eps = device:get_endpoints(DoorLock.ID, {feature_bitmap = DoorLock.types.DoorLockFeature.CREDENTIALSOTA | DoorLock.types.DoorLockFeature.PIN_CREDENTIALS})
-
-  local profile_name = "lock"
-
-  -- we need to check if a device is already fingerprinted.
-
-
-  -- check for lock codes
-  if #pin_eps == 0 then
-    profile_name = profile_name .. "nocodes"
-  end
+  local profile_name = "base-lock"
 
   -- check for battery type
   if #power_source_eps == 0 then
@@ -543,6 +549,9 @@ local function do_configure(driver, device)
   elseif #battery_feature_eps == 0 then
     profile_name = profile_name .. "-batteryLevel"
   end
+
+  device.log.info_with({hub_logs=true}, string.format("Updating device profile to %s.", profile_name))
+  device:try_update_metadata({profile = profile_name})
 end
 
 local function device_init(driver, device)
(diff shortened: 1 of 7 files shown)


drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -516,6 +516,41 @@ local function component_to_endpoint(device, component_name)
   return find_default_endpoint(device, clusters.DoorLock.ID)
 end
 
+local function do_configure(driver, device)
+  local fingerprinted_devices = {
```
nickolas-deboom, line 520: A new fingerprint was added yesterday and should be added to this list: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1650

nickolas-deboom:
This looks good to me, but I was just thinking that one potential drawback with this approach is that each PR for new WWST device fingerprints would also have to include an update to the list of fingerprinted devices in `do_configure`. Maybe that is no big deal but I thought I'd mention it

hcarter-775:
Yeah @nickolas-deboom I agree, that's definitely a downside. Unfortunately, there's not really any better method in place of solving this issue atm. I figured we would inform Alissa of this in the next Risk Analysis meeting and then you and I can help monitor this as well until we put a better solution in place.

hcarter-775:
@nickolas-deboom I also think this example might get some more proactive desire to fix this situation in a larger way

drivers/SmartThings/matter-lock/src/init.lua
ctowns: I think Nick makes a very good point about this requiring an update to this list every time we add a WWST fingerprint. I don't think this would be very difficult, but we have received push back on this in the past due to the way it complicates the WWST process. I think if we communicate with @lelandblue and make sure this is coordinated with the WWST team, this could be an extra step that is added to the WWST process and we can maintain this list here ourselves.

However, I think there is a different way to think about issue as well. Currently, we are trying to filter out all devices that have a fingerprint because these already have assigned profiles that we do not want to change. However, the assumption we are making here is that all non-fingerprinted devices are using the `base-lock` profile, which is true in this case.

So, what if instead of filtering out fingerprinted devices, we instead just specifically targeted devices using the `base-lock` profile? This would be a different way to get the same subset of devices, and it would prevent the need to maintain a list of fingerprinted devices in the driver. 

One way we can do this is by using the `device:supports_capability()` function to determine if a device supports a given capability, and we can call this function multiple times for each of the capabilities in the `base-lock` profile to see if the device supports all the capabilities in the `base-lock` profile. If it does, we can assume that it is using the `base-lock` profile, because based on my initial review of the profiles, the `base-lock` profile does not contain capabilities that are a subset of any other profile, so therefore if a device supports all the capabilities in `base-lock`, it is definitely using the `base-lock` profile. I have done something similar to this for the `matter-sensor` driver here, where we essentially reconstructed a profile based on the capabilities that the device supports and then we append `-battery` on the end if the device also supports battery: 
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/967/files#diff-8a122f65a4419a8fe00de96e7444d9d71df5609075dbb217531dc6d125197755R26-R36

I think we could do something similar here where we do something like this:
```
  if device:supports_capability(capabilities.lock) and
    device:supports_capability(capabilities.lockCodes) and
    device:supports_capability(capabilities.tamperAlert) and
    device:supports_capability(capabilities.battery) then
    -- we can assume device is using base-lock profile
    -- handle switching to no battery profile as needed
  end
```

Now, we know there will be changes coming to the `matter-lock` driver that will make this assumption of "non-fingerprinted devices use `base-lock`" a little more complicated, but I think we could use the same method of essentially constructing the profile from it's capabilities that is used above to handle things more granularly once we need to consider the profiles that support `lockAlarm` with the new matter lock driver.

hcarter-775: Good catch!

hcarter-775:
And I just thought of this, but another point of note here is that there is a new and improved matter Lock driver coming soon, so this issue may be extremely temporary.

nickolas-deboom:
Those are good points, and I can't think of a better way to handle this either, so I think this is a good solution for now, especially given it might be temporary since there are improvements coming to this driver as you mentioned.

nickolas-deboom (approved):

drivers/SmartThings/matter-lock/src/init.lua
hcarter-775: Yeah @ctowns I think this seems like a very reasonable idea that does the same thing without needing an extra table!

ctowns (commented):
This looks good, however, there are multiple other places in the `matter-lock` driver that update the device's profile, for example:
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/44087fd33c74c98dcfae8acd5e2dc5b7cc498aaa/drivers/SmartThings/matter-lock/src/init.lua#L573
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/44087fd33c74c98dcfae8acd5e2dc5b7cc498aaa/drivers/SmartThings/matter-lock/src/init.lua#L200
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/44087fd33c74c98dcfae8acd5e2dc5b7cc498aaa/drivers/SmartThings/matter-lock/src/init.lua#L213

Two of the instances have to do with setting a profile for a non-functional lock, however in the `device_added` function, I think we'd also need to do something similar in the case that the device is going to be switched to the `lock-without-codes` profile. Essentially, you'd have to just add a check that the device would switch to `lock-without-codes-batteryLevel/nobattery` depending on what it supports like you have done above.

Also, when recovering from a non-functional state, we'd probably want to copy what you've done and make sure the device joins to `base-lock` plus whatever is appropriate for it's battery situation.

ctowns:
Also before merging, this change should be tested with both the new bridged the device as well as regression tested against an existing lock.

hcarter-775:
Yes, I agreee Cooper. I will add the dynamic battery profiling logic to those places as well. And I have already reached out on the affected ticket to get some testing with the new device.

ctowns (approved):
Looks good! I would suggest waiting for the new matter lock PR to merge first and then rebasing these changes on top of it. This PR should be easier to rebase and we want the new matter lock PR to be as frictionless as possible a this point.

Huangxiangjie (commented):
If device Power Source 0x002F Features is BAT, But do not support 0x000D BatTimeRemaining

PR #1547, opened 2024-07-26
Merge Switch and Button Drivers

nickolas-deboom:
# Type of Change

- [ ] WWST Certification Request
     - If this is your first time contributing code:
          - [ ] I have reviewed the README.md file
          - [ ] I have reviewed the CODE_OF_CONDUCT.md file
          - [ ] I have signed the CLA
     - [ ] I plan on entering a WWST Certification Request or have entered a request through the WWST Certification console at developer.smartthings.com
- [ ] Bug fix
- [ ] New feature
- [X] Refactor

# Checklist

- [X] I have performed a self-review of my code
- [X] I have commented my code in hard-to-understand areas
- [X] I have verified my changes by testing with a device or have communicated a plan for testing
- [x] I am adding new behavior, such as adding a sub-driver, and have added and run new unit tests to cover the new behavior

# Description of Change

This PR addresses [CHAD-13270](https://smartthings.atlassian.net/browse/CHAD-13270) to integrate the functionality from matter-button into the matter-switch driver, so that we can deprecate the matter-button driver.

Note that these changes were originally located in [PR 1423](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1423), but that PR was split into multiple PRs in order to simplify the amount of changes. This PR only includes support for pure-button and pure-switch devices. Once more discussion has been had on how combination button-switch devices should be configured, [PR 1596](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1596) will be updated to include support for these combination devices.

# Summary of Completed Tests

Testing was performed on several switch and button devices as documented here: https://smartthings.atlassian.net/wiki/spaces/CHIP/pages/3340304693/Matter-Switch+Matter-Button+Integration+Testing+Status

[CHAD-13270]: https://smartthings.atlassian.net/browse/CHAD-13270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1663, opened 2024-10-01
Support combination switch/button devices

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

[CHAD-13993](https://smartthings.atlassian.net/browse/CHAD-13993)

This change adds support for devices that contain both switch and button endpoints. The button driver was initially merged into the driver in [PR 1547](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1547), but until now only pure-button and pure-switch devices were supported.

The first switch endpoint and all button endpoints will join as a multi-component device, and additional switch endpoints will join as child devices.

# Summary of Completed Tests

See new unit tests and also [here](https://smartthings.atlassian.net/wiki/spaces/~7120201ec2fa4a1b0548dd86df78b89d08422e/pages/3449815291/Supporting+Combination+Button+Switch+Devices#Testing) for an overview of testing. Additional regression testing is documented [here](https://smartthings.atlassian.net/wiki/spaces/CHIP/pages/3340304693/Matter-Switch+Matter-Button+Integration+Testing+Status).

[CHAD-13993]: https://smartthings.atlassian.net/browse/CHAD-13993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -140,6 +140,13 @@ local device_type_attribute_map = {
     clusters.ColorControl.attributes.CurrentSaturation,
     clusters.ColorControl.attributes.CurrentX,
     clusters.ColorControl.attributes.CurrentY
+  },
+  [GENERIC_SWITCH_ID] = {
```
hcarter-775, line 144: Should this be generic Button ID?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -250,17 +257,14 @@ end
 local function find_default_endpoint(device, component)
   local switch_eps = device:get_endpoints(clusters.OnOff.ID)
   local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.Feature.MOMENTARY_SWITCH})
-  local all_eps = {}
 
-  for _,ep in ipairs(switch_eps) do
-    table.insert(all_eps, ep)
-  end
+  -- use first button endpoint as main endpoint if one exists
```
ctowns, line 261: I think it might be worth it to add just a little more context on why the first button endpoint would be used as the main endpoint. You've mentioned some of the reasoning in your confluence doc, I think it would just be helpful to have a bit more context why this is the case in the code.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -250,17 +257,14 @@ end
...
   for _,ep in ipairs(button_eps) do
-    table.insert(all_eps, ep)
+    if ep ~= 0 then --0 is the matter RootNode endpoint
```
ctowns, line 263: I think you would need to sort the button endpoints first. This was done in like 261 in the previous code after all eps were combined into one table FYI. In this case, we may need to sort the button and switch eps separately and then return the relevant ep.

drivers/SmartThings/matter-switch/src/test_matter_button_switch_mcd.lua
```diff
@@ -0,0 +1,199 @@
...
+
+local clusters = require "st.matter.clusters"
+
+local button1_ep = 10
+local button2_ep = 20
+local light_ep = 30
+
+local mock_device = test.mock_device.build_test_matter_device({
+  label = "Matter Switch",
+  profile = t_utils.get_profile_definition("2-button-battery-switch.yml"),
```
ctowns, line 29: Is this a real profile? I don't see it here or on main

ctowns (commented):
Just a reminder to rebase as well so that the latest changes are present and so that unit tests can run :)

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: The button device type is called "Generic Switch"

drivers/SmartThings/matter-switch/src/test_matter_button_switch_mcd.lua
nickolas-deboom: This test file and that profile are leftovers from experimenting with other types of configurations, and so this file should be removed.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Good call - I made this change in the latest commit.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Added a comment with an explanation in the latest commit!

nickolas-deboom:
> Just a reminder to rebase as well so that the latest changes are present and so that unit tests can run :)

Done!

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -559,7 +559,7 @@ local function initialize_switch(driver, device)
     else
       configure_buttons(device)
     end
-  elseif #switch_eps > 0 then
+  else
```
ctowns, line 562: I would still recommend explicitly making this case check for the switch_eps, just that is it clear that this is meant to handle the switch case rather than just being a catch-all case. Theoretically, this logic should behave the same either way with how our driver is currently setup, but I think still trying it explicitly to the switch eps makes it less prone to edge cases.

nickolas-deboom: Sure thing, added it back!

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -341,17 +348,18 @@ end
 local function find_default_endpoint(device, component)
   local switch_eps = device:get_endpoints(clusters.OnOff.ID)
   local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.Feature.MOMENTARY_SWITCH})
-  local all_eps = {}
+  table.sort(switch_eps)
+  table.sort(button_eps)
```
hcarter-775, line 352: Why are we doing this work twice, since this same sorting is accomplished in `initialize_switch`? 

Further, the `component` field is not used here- we could remove it and replace it with an opts field that includes the presorted switch and button endpoint table.

nickolas-deboom: I removed the component field, but since `initialize_switch` is not guaranteed to run before `find_default_endpoint`, for existing device for example, I think we should leave this sorting here.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -477,6 +485,21 @@ local function initialize_switch(driver, device)
   -- support for bindings.
   local num_switch_server_eps = 0
   local main_endpoint = find_default_endpoint(device)
+
+  if #button_eps > 0 then
```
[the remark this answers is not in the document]
nickolas-deboom: Agreed - removed!

ctowns:
One thing we will want to do before merging these changes is complete a set of regression tests for the various device types that share this driver since we are touching a critical code path. I think you could potentially add the test cases to the great confluence page you already put together, listing which cases are covered by unit tests and real devices, or both: https://smartthings.atlassian.net/wiki/spaces/~7120201ec2fa4a1b0548dd86df78b89d08422e/pages/3449815291/Supporting+Combination+Button+Switch+Devices#Testing.

Let's think of some device compositions that are commonly used and also add in some that might be corner cases. For example:
1. Single switch device (bulb, plug, etc.)
2. Single button device
3. Multi-switch only device
4. Multi-button only device
5. Switch/Button combo device with a single switch and multiple buttons
6. Switch/Button combo with a multiple switches and a single button
7. Switch/Button combo with multiple switches and multiple buttons
8. Switch/Button combo with multiple switches and multiple buttons where the switch has lower endpoints than the buttons 

A good chunk of these are already covered by existing unit tests and the ones you have added, I just wanted to list all the ones I could think of to be complete. Where we can, we should do smoke testing with real devices in addition to the unit tests. Some of these will be impossible to test with real devices (test 8 for example, I'm not sure we have a real device like that), so they can be covered by a unit test case (which it looks like you've already done for test 8). If you can think of any more edge cases, please add them!

nickolas-deboom:
> One thing we will want to do before merging these changes is complete a set of regression tests for the various device types that share this driver since we are touching a critical code path. I think you could potentially add the test cases to the great confluence page you already put together, listing which cases are covered by unit tests and real devices, or both: https://smartthings.atlassian.net/wiki/spaces/~7120201ec2fa4a1b0548dd86df78b89d08422e/pages/3449815291/Supporting+Combination+Button+Switch+Devices#Testing.
> 
> Let's think of some device compositions that are commonly used and also add in some that might be corner cases. For example:
> 
>     1. Single switch device (bulb, plug, etc.)
> 
>     2. Single button device
> 
>     3. Multi-switch only device
> 
>     4. Multi-button only device
> 
>     5. Switch/Button combo device with a single switch and multiple buttons
> 
>     6. Switch/Button combo with a multiple switches and a single button
> 
>     7. Switch/Button combo with multiple switches and multiple buttons
> 
>     8. Switch/Button combo with multiple switches and multiple buttons where the switch has lower endpoints than the buttons
> 
> 
> A good chunk of these are already covered by existing unit tests and the ones you have added, I just wanted to list all the ones I could think of to be complete. Where we can, we should do smoke testing with real devices in addition to the unit tests. Some of these will be impossible to test with real devices (test 8 for example, I'm not sure we have a real device like that), so they can be covered by a unit test case (which it looks like you've already done for test 8). If you can think of any more edge cases, please add them!

Good point! I added section [Test coverage for other device compositions](https://smartthings.atlassian.net/wiki/spaces/~7120201ec2fa4a1b0548dd86df78b89d08422e/pages/3449815291/Supporting+Combination+Button+Switch+Devices#Test-coverage-for-other-device-compositions) to the confluence page. I created a table with the endpoint configurations you listed and linked to the test case files that cover each configuration. I tested out a few cases with physical devices I have, and I also expanded `test_matter_multi_button_child_switch.lua` to include test cases to cover cases 5, 7, and 8. Please let me know if you think there is anything else I should add!

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -464,51 +470,48 @@ local function initialize_switch(driver, device)
...
-              profile = child_profile,
-              parent_device_id = device.id,
-              parent_assigned_child_key = string.format("%d", ep),
-              vendor_provided_label = name
-            }
-          )
-          parent_child_device = true
-        end
+
+  -- If button endpoints are present, use MCD for the main endpoint and all button
```
ctowns, line 483: I would update this comment to be a little clearer. Just make sure to explain that if a switch endpoint is present, then it will be used as the main endpoint and therefore the main component. If a button endpoints are present, then will be combined into an MCD profile attached to the main switch endpoint

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -338,20 +345,19 @@ end
...
 --- BRIDGED_NODE_DEVICE_TYPE on bridged devices.
-local function find_default_endpoint(device, component)
+local function find_default_endpoint(device)
   local switch_eps = device:get_endpoints(clusters.OnOff.ID)
   local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.Feature.MOMENTARY_SWITCH})
-  local all_eps = {}
+  table.sort(switch_eps)
+  table.sort(button_eps)
 
+  -- Use the first switch endpoint as the main endpoint if one is present.
```
ctowns, line 354: One thing to mention about this strategy is that if a switch endpoint is going to be used first, that means anytime we encounter a device that is a switch+button combination that we don't currently support in our profiles, then we would have to create a new profile to cover that switch+button combination. For example, right now we have light level switch + N buttons supported in our profiles, but let's say another device comes out that is just on/off switch + buttons. Then we would have to create new permutations of the existing button profiles.

We can continue to make profiles, but again this is something where we will have to know what products are on the market and create new profiles as needed to cover the capabilities of those new switch+button combos. This is something we can do with profiles, but dynamic devices would greatly reduce this burden when that is available.

ctowns: @nickolas-deboom we are making the switch endpoint the main component for UX reasons, right?

If we didn't do that, we could decouple the switch endpoints from the button endpoints in the profiles entirely. However, I understand this could provide a bad user experience for customers if they have all their buttons in one device card and then an on/off switch in a separate one.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -526,15 +529,37 @@ local function initialize_switch(driver, device)
...
   end
 
-  if num_switch_server_eps > 0 then
+  if #switch_eps > 0 and #button_eps > 0 then
+    if tbl_contains(STATIC_BUTTON_PROFILE_SUPPORTED, #button_eps) then
+      local dimmable_light = false
+      for _, ep in ipairs(device.endpoints) do
+        for _, dt in ipairs(ep.device_types) do
+          if dt.device_type_id == DIMMABLE_LIGHT_DEVICE_TYPE_ID then
+            dimmable_light = true
```
ctowns, line 538:
```suggestion
            profile_name = "light-level"
```

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -526,15 +529,37 @@ local function initialize_switch(driver, device)
     device:set_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON, component_map, {persist = true})
   end
 
-  if num_switch_server_eps > 0 then
+  if #switch_eps > 0 and #button_eps > 0 then
+    if tbl_contains(STATIC_BUTTON_PROFILE_SUPPORTED, #button_eps) then
+      local dimmable_light = false
+      for _, ep in ipairs(device.endpoints) do
+        for _, dt in ipairs(ep.device_types) do
+          if dt.device_type_id == DIMMABLE_LIGHT_DEVICE_TYPE_ID then
```
ctowns, line 537: This is essentially saying we are only going to support swithc/button combos where the switch component is a dimmable light. We could add handling for the other switch device types pending what devices are on the market. Are dimmable light + buttons the only switch/button combo devices we have tested/know about?

If we do expand this, we will need to create more profiles as I mentioned above because this design couples a switch component with the button components in a single MCD profile.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -464,51 +470,48 @@ local function initialize_switch(driver, device)
...
-            }
-          )
-          parent_child_device = true
-        end
+
+  -- If button endpoints are present, use MCD for the main endpoint and all button
+  -- endpoints. Note that if switch endpoints are present, the first switch
+  -- endpoint will be considered the main endpoint. Otherwise, the first button
+  -- endpoint will be considered the main endpoint.
+  if #button_eps == 1 or tbl_contains(STATIC_BUTTON_PROFILE_SUPPORTED, #button_eps) then
```
ctowns, line 487: You could just add `1` to the list of supported button profiles in `STATIC_BUTTON_PROFILE_SUPPORTED`

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -526,15 +529,37 @@ local function initialize_switch(driver, device)
...
+    end
+
+    if profile_name then
+      profile_name = profile_name .. string.format("-%d-button", #button_eps)
+      device:try_update_metadata({profile = profile_name})
+      device:set_field(DEFERRED_CONFIGURE, true)
+    else
+      configure_buttons(device)
+    end
+  elseif #switch_eps > 0 then
```
ctowns, line 554: I believe this could just be
```suggestion
  elseif num_switch_server_eps > 0 then
```

And then you can remove the switch server eps check below on line 562.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I decided to do it this way because `STATIC_BUTTON_PROFILE_SUPPORTED` is used later to determine if a profile change is needed for pure button devices (and doesn't include `1` on purpose). See https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/41c5e8529faba5d7c6efce0a6a0981c46c2857bd/drivers/SmartThings/matter-switch/src/init.lua#L563

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Yes, dimmable light + buttons is the only combination we are aware of so far. The logic here could easily be expanded to cover more device types in the future as needed. And, you are right that new profiles would also be needed in that case.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated in 96e319121530b6b2ad5ff7114b02531588b3bdcd!

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated as suggested in 96e319121530b6b2ad5ff7114b02531588b3bdcd

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated as suggested in 96e319121530b6b2ad5ff7114b02531588b3bdcd!

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Unfortunately you are right, that new profiles would be needed for new products if they use different device types than what is currently supported.

nickolas-deboom: And yes that's correct - after some conversations with Tom, it would be best from a UX perspective to keep the buttons with the light device in the same device card. This also matches the way that zigbee implements a similar Inovelli device.

drivers/SmartThings/matter-switch/src/init.lua
ctowns: So if a switch/button combo device type we do not support is onboarded, then how would it look?

It looks like this would go down the path of trying to configure the buttons immediately, but we can't know if the device joined to a button or switch profile, so it could start configuring button that don't even exist in the profile. Even if it joins as a button profile, it wouldn't neccessarily have all the buttons represented because the fingerprint defaults to `button-battery`, a single component device.

This is because this wouldn't detect a match for the `DIMMABLE_LIGHT_DEVICE_TYPE_ID` type, which would mean `profile_name` would be `nil`. We should define a default behavior for unsupported devices in which we can provide at least partial support. We'd essentially have to choose between partially supporting the switch half of the device, or the button half. I would probably choose the button half if we are going to provide partial support. If profile name is nil, then I think we need to set the profile name correctly to be just buttons:
```
profile_name = string.format("%d-button", #button_eps)
```

Also one other thing to point out is that if the main endpoint is a switch device type that is unsupported, this logic would also fail. So if we commit to just supporting the button half of unsupported combos, then I'd add an additional check at that top of this function to confirm that we only handle the switch as the main endpoint if we plan to support it. Does that makes sense?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -527,7 +529,25 @@ local function initialize_switch(driver, device)
     device:set_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON, component_map, {persist = true})
   end
 
-  if num_switch_server_eps > 0 then
+  if #switch_eps > 0 and #button_eps > 0 then
+    if tbl_contains(STATIC_BUTTON_PROFILE_SUPPORTED, #button_eps) then
```
ctowns, line 533: I would set `profile_name = nil` here for clarity since this conditional block makes an assumption that `profile_name` is nil for this logic to work.

ctowns:
I think this confluence page needs to be updated with the latest screenshots with the new component layout: [Test coverage for other device compositions](https://smartthings.atlassian.net/wiki/spaces/~7120201ec2fa4a1b0548dd86df78b89d08422e/pages/3449815291/Supporting+Combination+Button+Switch+Devices#Test-coverage-for-other-device-compositions)

I would also note on that page what switch/button combos are supported, and what expected behavior is for an unsupported combination device.

Also, I think another round of device regression testing might be good to verify the latest changes with the new layout. Fortunately, a lot of this is covered with unit tests, so good job on that test coverage!

nickolas-deboom:
> I think this confluence page needs to be updated with the latest screenshots with the new component layout: [Test coverage for other device compositions](https://smartthings.atlassian.net/wiki/spaces/~7120201ec2fa4a1b0548dd86df78b89d08422e/pages/3449815291/Supporting+Combination+Button+Switch+Devices#Test-coverage-for-other-device-compositions)
> 
> I would also note on that page what switch/button combos are supported, and what expected behavior is for an unsupported combination device.
> 
> Also, I think another round of device regression testing might be good to verify the latest changes with the new layout. Fortunately, a lot of this is covered with unit tests, so good job on that test coverage!

I updated the confluence page with new details about testing, for both the Inovelli and the GE/Cync device. I added some notes on what the behavior would be for unsupported device types as well in the [Testing section](https://smartthings.atlassian.net/wiki/spaces/~7120201ec2fa4a1b0548dd86df78b89d08422e/pages/3449815291/Supporting+Combination+Button+Switch+Devices#Testing).

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I actually decided to make this change because it simplified logic in other places that use this table.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: In 25cf9d3d30e4bfd062ee4b21b26303b624bc0682 I updated `find_default_endpoint` to use the first switch endpoint as the default endpoint unless (1) there are only button endpoints, or (2) the first switch endpoint does not contain a supported device type (currently only `DIMMABLE_LIGHT_DEVICE_TYPE_ID`) for a combination button/switch device. I think this solves this issue and it also simplified the logic in `initialize_switch` quite a bit. I added a new test case in the subsequent commit to verify that for a combo button/switch where an unsupported device type is used, the default endpoint is the first button endpoint, the buttons join a MCD profile, and the "unsupported" switch device type joins as a child device.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I have refactored this section which makes this change unnecessary now I believe!

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -334,29 +341,61 @@ local function mired_to_kelvin(value, minOrMax)
...
-      return ep
+  -- If both switch and button endpoints are present, check the device type on the main switch
+  -- endpoint. If it is not a supported device type, return the first button endpoint as the
+  -- default endpoint.
+  if #switch_eps > 0 and #button_eps > 0 then
+    local main_endpoint = get_first_non_zero_endpoint(switch_eps)
+    if is_supported_combination_button_switch_device_type(device, main_endpoint) then
+      return main_endpoint
+    else
+      return get_first_non_zero_endpoint(button_eps)
```
tpmanley, line 395: I think this worth logging here because the device has both OnOff and Switch clusters but we're only going to expose the Switch (aka button) in this case. This is likely not what is expected and we should at least log so it's easier to debug.

tpmanley (approved):
Few minor comments but otherwise looks good to me. Need to make sure to do a thorough regression test with devices that would use the affected code paths.

drivers/SmartThings/matter-switch/profiles/light-level-button.yml
```diff
@@ -0,0 +1,24 @@
...
+          values:
+            - key: "level.value"
+              range: [1, 100]
+      - id: firmwareUpdate
+        version: 1
+      - id: refresh
+        version: 1
+    categories:
+      - name: Light
+  - id: button1
```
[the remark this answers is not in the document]
nickolas-deboom: I had it this way to make it easier like you said but I made the change you suggested because I think it's better this way.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Added a log as suggested!

nickolas-deboom:
> Few minor comments but otherwise looks good to me. Need to make sure to do a thorough regression test with devices that would use the affected code paths.

I did regression testing on all of the devices listed on this page, which includes lights, plugs, and buttons as well as the GE/Cync and Inovelli switch/button combo devices. I also created unit tests to verify the functionality for a device that would join a MCD profile containing switch and button components as well as a device that has button endpoints and a switch endpoint that would not be supported by a MCD profile and uses parent-child for the switch endpoint instead.

PR #1667, opened 2024-10-07
New Device IM Pushbutton Module

cjswedes:
Cherry pick https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1661 so it makes it to production today.


lelandblue (approved):
I have confirmed that this is the only necessary change that needs to go to Beta in advance of the Production default release.

PR #1671, opened 2024-10-07
Remove state cache usage from devices to reduce datastore size

cjswedes:
There are some current usages of the state cache in the driver, but the impact seems minimal.

Similar to #1670 

@dljsjr it would be great if you could do a test with this on your setup prior to merging. Also, looking for your input on the existing state cache optimizations that will be lost with this change. All usages should be in this diff.


drivers/SmartThings/philips-hue/src/handlers/commands.lua
```diff
@@ -92,6 +92,7 @@ local function do_switch_level_action(driver, device, args)
     return
   end
 
+  --Note: this will always be 0 due to state cache changes
```
varzac, line 95: comment is not quite right since it won't be 0.  I assume it will be `nil`

drivers/SmartThings/philips-hue/src/handlers/commands.lua
```diff
@@ -177,6 +178,7 @@ end
 ---@param args table
 local function do_setHue_action(driver, device, args)
 
+  --Note: this will always be 0 due to state cache changes
```
varzac, line 181: This actually seems like it could be a problem, because using a 0 for the other value will be a different color.  So I think we need to identify another solution than just not storing.

dljsjr: Same issue here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1671/files#diff-ddccea1fd0c65608f45092f0f576c125a2afaebd37e7b9bb78855eece84037e5R197

My pitch is to capture the values in a field [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua#L26), either vol or nonvol

drivers/SmartThings/philips-hue/src/handlers/commands.lua
dljsjr: This change will lead to us sending an `on` to the light with every switch level capability command. I'd like to think about storing this piece of state in a field as well to avoid that extra round trip. It could add up for automations with lots of lights.

cjswedes: I changed this to be a persisted field, set when we emit on/off events.

drivers/SmartThings/philips-hue/src/handlers/commands.lua
cjswedes: Captured in a field.

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
```diff
@@ -35,8 +35,10 @@ local function _emit_light_events_inner(light_device, light_repr)
 
     if light_repr.on and light_repr.on.on then
       light_device:emit_event(capabilities.switch.switch.on())
+      light_device:set_field(Fields.SWTICH_STATE, "on", {persist = true})
     elseif light_repr.on and not light_repr.on.on then
       light_device:emit_event(capabilities.switch.switch.off())
+      light_device:set_field(Fields.SWTICH_STATE, "off", {persist = true})
```
dljsjr, line 41:
```suggestion
      light_device:set_field(Fields.SWITCH_STATE, "on", {persist = true})
    elseif light_repr.on and not light_repr.on.on then
      light_device:emit_event(capabilities.switch.switch.off())
      light_device:set_field(Fields.SWITCH_STATE, "off", {persist = true})
```

dljsjr (approved):
Tested the typo fix and everything is working well. Approved contingent on those fixes.

PR #1685, opened 2024-10-14
fp2: Change in Credential Management Approach

hdlee27:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/Aqara/aqara-presence-sensor/src/discovery.lua
+++ b/drivers/Aqara/aqara-presence-sensor/src/discovery.lua
@@ -4,47 +4,59 @@ local fields = require "fields"
 local discovery_mdns = require "discovery_mdns"
 local socket = require "cosock.socket"
 
--- mapping from device DNI to info needed at discovery/init time
-local device_discovery_cache = {}
+local processing_devices = {}
 
-local function set_device_field(driver, device)
-  local device_cache_value = device_discovery_cache[device.device_network_id]
+function discovery.set_device_field(driver, device)
+  local device_cache_value = driver.datastore.discovery_cache[device.device_network_id]
 
   -- persistent fields
   if device_cache_value ~= nil then
+    device:set_field(fields.CREDENTIAL, device_cache_value.credential, { persist = true })
     device:set_field(fields.DEVICE_IPV4, device_cache_value.ip, { persist = true })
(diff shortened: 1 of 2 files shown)


drivers/Aqara/aqara-presence-sensor/src/discovery.lua
```diff
@@ -4,47 +4,59 @@ local fields = require "fields"
...
 local socket = require "cosock.socket"
 
--- mapping from device DNI to info needed at discovery/init time
-local device_discovery_cache = {}
+local processing_devices = {}
 
-local function set_device_field(driver, device)
-  local device_cache_value = device_discovery_cache[device.device_network_id]
+function discovery.set_device_field(driver, device)
+  local device_cache_value = driver.datastore.discovery_cache[device.device_network_id]
```
cjswedes, line 10: What is the reason for trying to persist the discovery cache in the datastore? Please explain the use case where this is needed and not covered by the local discovery cache.

hdlee27: During Near by Search, it was found that Discovery stopped when user exit the page by pressing Back button.
At this time, drivers without registered devices appear to be killed, and the local cache disappeared when the driver was executed again.
Datastore was used to keep the cache even when running the driver again after the driver was killed.

cjswedes: > it was found that Discovery stopped when user exit the page by pressing Back button.
At this time, drivers without registered devices appear to be killed, and the local cache disappeared when the driver was executed again

This is how the system is intended to work.

It seems like a nice feature for a user who is going in and out of the scan near by search wondering why a device isn't being discovered, but is that something we need to handle? If they are backing out of scan nearby, I assume they want to stop scanning and do not expect to join a device. There is also a small concern that the cache may become outdated with lots of time between scan attempts.

cjswedes (approved):
This seems like a low risk change, so I am approving. But the optimization it's making to discovery in situations where the user is starting/stopping scan nearby quickly doesn't seem like a use case we need to optimize IMO.

PR #1686, opened 2024-10-14
Prevent multiple reads of MinSetpointDeadBand

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

[CHAD-14135](https://smartthings.atlassian.net/browse/CHAD-14135)

The driver currently attempts to only read `MinSetpointDeadBand` once during the first time `device_init` is called, but the `MIN_SETPOINT_DEADBAND_CHECKED` field is not persisted, so the read request is sent during each device init. This is contributing to a spike in `InteractionQueueFull` exceptions. This change ensures that the field is persisted so that the read request is only sent once.

# Summary of Completed Tests

Added unit test to verify that the `MIN_SETPOINT_DEADBAND_CHECKED` flag is set after the `MinSetpointDeadBand` attribute is received to ensure that this value is only read one time.

Tested with a VDA bridge consisting of 20 thermostats and 20 color temperature lights to ensure that no regressions were introduced. Note that the issue is still present when dealing with a large number of bridged devices because the queue can still reach capacity even with this change. This is meant to help alleviate the issue rather than provide a full solution, which would require optimizations to the read request logic in the firmware.

[CHAD-14135]: https://smartthings.atlassian.net/browse/CHAD-14135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -280,7 +280,7 @@ local function device_init(driver, device)
       deadband_read:merge(clusters.Thermostat.attributes.MinSetpointDeadBand:read())
       device:send(deadband_read)
     end
-    device:set_field(setpoint_limit_device_field.MIN_SETPOINT_DEADBAND_CHECKED, true)
+    device:set_field(setpoint_limit_device_field.MIN_SETPOINT_DEADBAND_CHECKED, true, {persist = true})
   end
 end
 
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -271,7 +271,6 @@ end
 local function device_init(driver, device)
   device:subscribe()
   device:set_component_to_endpoint_fn(component_to_endpoint)
-
   if not device:get_field(setpoint_limit_device_field.MIN_SETPOINT_DEADBAND_CHECKED) then
     local auto_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.AUTOMODE})
     --Query min setpoint deadband if needed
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -280,7 +280,7 @@ local function device_init(driver, device)
       deadband_read:merge(clusters.Thermostat.attributes.MinSetpointDeadBand:read())
       device:send(deadband_read)
     end
-    device:set_field(setpoint_limit_device_field.MIN_SETPOINT_DEADBAND_CHECKED, true)
+    device:set_field(setpoint_limit_device_field.MIN_SETPOINT_DEADBAND_CHECKED, true, {persist = true})
```
ctowns, line 283: I think we could move this read back to the configure or added function unless there is a good reason to have the deadband read once on init. Typically, does this kind of `once on init` handling is to provide some cleanup for devices that are already in the field and onboarded. However, in this case, devices in the field that are already onboarded would have already had the deadband read during either doConfigure (prior to this change) or in init (recurring on each init with this change). So, I think it makes more sense for this to be in one of the one-time events anyway, such as added or configure

However, given the tight deadline this can be a patch for now that we can move to configure later as needed.

ctowns: Also, it probably makes more sense for this flag to be set once we actually read the value, not just when we send the request.

nickolas-deboom: I think I will just move this to do_configure to simplify things. This is where it was in the past so it will be a simple change to move it back. I'm doing a test to confirm that this fixes the issue right now and then I will push up the changes.

nickolas-deboom: After discussion, we will leave this read request in `device_init` but move setting the `MIN_SETPOINT_DEADBAND_CHECKED` field to the `MinSetpointDeadBand` attribute handler.

ctowns (approved):

PR #1691, opened 2024-10-15
Convert Power and Energy reads from mW(h) to W(h). 

hcarter-775:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
The Electrical Energy Measurement cluster and Electrical Power Measurement cluster report readings in mWh and mW respectively, but SmartThings uses W and Wh for our capabilities. This change converts the read values into those of the correct unit. 

# Summary of Completed Tests
Unit tests were converted to match the new system.
See user testing reports in [this Slack page](https://smartthings.slack.com/archives/CDNPD1P8Q/p1728941844831519):

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -158,7 +158,7 @@ local LAST_EXPORTED_REPORT_TIMESTAMP = "__last_exported_report_timestamp"
 local RECURRING_EXPORT_REPORT_POLL_TIMER = "__recurring_export_report_poll_timer"
 local MINIMUM_ST_ENERGY_REPORT_INTERVAL = (15 * 60) -- 15 minutes, reported in seconds
 local SUBSCRIPTION_REPORT_OCCURRED = "__subscription_report_occurred"
-local CONVERSION_MILLIWATTS_TO_WATTS = 0.001 -- A milliwatt is 1/1000th of a watt
+local CONVERSION_CONST_MILLIWATT_TO_WATT = 1000 -- A milliwatt is 1/1000th of a watt
 
 local embedded_cluster_utils = require "embedded-cluster-utils"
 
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -923,15 +924,21 @@ local function occupancy_attr_handler(driver, device, ib, response)
 end
 
 local function cumul_energy_exported_handler(driver, device, ib, response)
-  device:set_field(TOTAL_EXPORTED_ENERGY, ib.data.elements.energy.value)
-  device:emit_event(capabilities.energyMeter.energy({ value = ib.data.elements.energy.value, unit = "Wh" }))
+  if ib.data.elements.energy then
+    local watt_hour_value = ib.data.elements.energy.value * CONVERSION_MILLIWATTS_TO_WATTS
```
tpmanley, line 928: Maybe a nit, but I think it'd be more clear to just `/ 1000` instead of using the global. The constant itself isn't going to change, it's a bit of a misnomer since energy is in mWh not mW, and also I think it's easier to read if it is more explicitly dividing by 1000 than multiplying by a decimal.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -923,15 +924,21 @@ local function occupancy_attr_handler(driver, device, ib, response)
 end
 
 local function cumul_energy_exported_handler(driver, device, ib, response)
-  device:set_field(TOTAL_EXPORTED_ENERGY, ib.data.elements.energy.value)
-  device:emit_event(capabilities.energyMeter.energy({ value = ib.data.elements.energy.value, unit = "Wh" }))
+  if ib.data.elements.energy then
```
tpmanley, line 927: Is this just intended to make sure it is not `nil` for some reason?

hcarter-775: Yes, since nil + float results an error. In the periodic case it's clear that this may occur since it uses summations, but it may also occur in the cumulative case when `TOTAL_EXPORTED_ENERGY` is used in the PowerConsumptionReport update.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: I think I disagree here honestly- I personally like this better since it makes it clear why this is needed without adding a comment every time, since the global is kinda a comment in itself. It is a misnomer technically, though a small and clear one imo. 

I do conceed it may be good make it more like:
 ```suggestion
    local watt_hour_value = ib.data.elements.energy.value / CONVERSION_MILLIWATTS_TO_WATTS
```
and turn `CONVERSION_MILLIWATTS_TO_WATTS` into 1000- I just liked the math-y feeling of it being a factor product.

nickolas-deboom (approved):

ctowns (approved):

PR #1693, opened 2024-10-16
Update matter-switch for Aqara Cube

DongHoon-Ryu:
- Set configuration in the device_init because only the device_init is called when the driver is updated.
- Improve the Aqara Cube identification method to refer to the read only name of the device

REQ-15926, REQ-16285, IOTE-4217, IOTE-4266

Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [X] Bug fix
- [ ] New feature
- [X] Refactor

# Checklist

# Description of Change
- Set configuration in the device_init because only the device_init is called when the driver is updated.
- Improve the Aqara Cube identification method to refer to the read only name of the device

# Summary of Completed Tests
- Successful test for Plugin UI update and Dash Board's Device Card status update according to the selection of each face
- Successful Automation Routine Test for Cube Face


hcarter-775:
I still think the change I suggested in [this comment](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1622/files#r1783771164) on the last PR would be a good addition, to add to the clarity of this subdriver.

hcarter-775:
I suggest pulling the logic in `device_added` out of `device_added`, and moving it directly into `init`. Currently, this logic will be called twice now when the device is first created (aka on `added` and then `init`, which may cause multiple `try_update_metadata` calls, something we don't want). Further, I think some gating should be added to keep the `try_update_metadata` from being called multiple times like this anyway.

ctowns:
For reference @DongHoon-Ryu , the driver lifecycle events occur at different times for different situations. However, the simple explanation is this:

1. driver init - this will occur every time the driver restarts. This means that any set up/initialization that needs to occur for the device each time the driver restarts needs to occur here. This is commonly where we set up our subscribe requests.
2. added/doConfigure - these events only happen once in the lifetime of the device (there are exceptions for doConfigure but they are not relevant in this case). These lifecycle handlers will only run once when a device is onboarded, and then they will likely not be run again during the devices lifetime. These lifecycle events are usually used for handling one-time configurations, such as profile-switching when onboarding a device, or creating component-to-endpoint maps.

In this case, it seems like there is some logic in the `device_added` handler that should be in the `device_init` function. Any code that needs to run each time the driver is initialized needs to be in the device_init function. The driver is initialized on each restart, and the driver restarts each time we deploy an update to the driver. In the future, we should make sure that we test the driver restart/driver switch test case so we can catch issues like this.


You can see our developer documentation on these events here: https://developer.smartthings.com/docs/edge-device-drivers/driver.html#lifecycle-event-handlers

DongHoon-Ryu:
@hcarter-775 I'm sorry. I missed the proposal regarding TEST_CONFIGURE last time. I think it's a good idea. First of all, I improved it with the method you suggested, and if I `try_update_metadata` at the `device_init` stage, test case crash occurs in the test framework, so I used TEST_CONFIGURE as a way to avoid it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -7,15 +7,20 @@ local cubeFace = capabilities["stse.cubeFace"]
...
 local INITIAL_PRESS_ONLY = "__initial_press_only" -- for devices that support MS (MomentarySwitch), but not MSR (MomentarySwitchRelease)
 
 -- after 3 seconds of cubeAction, to automatically change the action status of Plugin UI or Device Card to noAction
 local CUBEACTION_TIMER = "__cubeAction_timer"
 local CUBEACTION_TIME = 3
 
 local function is_aqara_cube(opts, driver, device)
-  local name = string.format("%s", device.label)
+  -- device.label is a value that can be changed by the user,
+  -- so it is changed to read only value(device.manufacturer_info.product_name).
```
hcarter-775, line 22: nit: remove comment. Useful for us in PR, but not in the prod version

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -126,53 +124,69 @@ local function configure_buttons(device)
...
+  if device:get_field(TEST_CONFIGURE) then
+    set_configure(driver, device)
+  end
+
+  if (device.profile.id ~= args.old_st_store.profile.id or device:get_field(TEST_CONFIGURE))
     and device:get_field(DEFERRED_CONFIGURE)
     and device.network_type ~= device_lib.NETWORK_TYPE_CHILD then
 
     -- At the time of device_added, there is an error that the corresponding capability cannot be found
     -- because the profile has not been changed from the generic profile of fingerprint to the Aqara Cube.
```
hcarter-775, line 183: You can remove this comment; this behavior is a standard part of the `info_changed` event

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -126,53 +124,69 @@ local function configure_buttons(device)
...
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    device:subscribe()
+    device:set_endpoint_to_component_fn(endpoint_to_component)
+
+    -- during driver update, only device_init is called, so configuration including profile change should be set here.
+    -- vid_0x115f_Cube is a label used in the unit test. This is used to avoid profile updates because the test framework
+    -- does not allow profile updates in the device_init stage. And it must be executed unconditionally in the user environment.
+    local name = string.format("%s", device.label)
+    if string.find(name, "vid_0x115f_Cube") == nil then
+      set_configure(driver, device)
```
hcarter-775, line 161: What are you seeing in the test environment that is forcing you to add this handling? Does expect_metadata_update not work for you?

hcarter-775:
```suggestion
    -- when unit testing, call set_configure elsewhere
    if not device:get_field(TEST_CONFIGURE) then
      set_configure(driver, device)
   end
```
however, if we cannot find a solution to your unit test problem, I suggest writing something more like the above. It reuses the TEST_CONFIGURE variable, which keeps things neater.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -126,53 +124,69 @@ local function configure_buttons(device)
...
     and device:get_field(DEFERRED_CONFIGURE)
     and device.network_type ~= device_lib.NETWORK_TYPE_CHILD then
 
     -- At the time of device_added, there is an error that the corresponding capability cannot be found
     -- because the profile has not been changed from the generic profile of fingerprint to the Aqara Cube.
     reset_thread(device)
+    device:emit_event(cubeAction.cubeAction("flipToSide1"))
     device:emit_event(cubeFace.cubeFace("face1Up"))
 
+    -- the profile change has been delayed.
```
hcarter-775, line 188: nit: remove comment, this behavior is clear in `info_changed`

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -126,53 +124,69 @@ local function configure_buttons(device)
...
-      component_map[string.format("%d", current_component_number)] = ep
-      current_component_number = current_component_number + 1
+local function set_configure(driver, device)
+  local MS = get_reordered_endpoints(driver, device)
+  local main_endpoint = device.MATTER_DEFAULT_ENDPOINT
+  if #MS > 0 then
+    main_endpoint = MS[1] -- the endpoint matching to the non-child device
+    if MS[1] == 0 then main_endpoint = MS[2] end -- we shouldn't hit this, but just in case
+  end
+  device.log.debug("main button endpoint is "..main_endpoint)
```
hcarter-775, line 134:
```suggestion
  device.log.debug("The main button endpoint for the Aqara T1 Pro is " .. main_endpoint)
```

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -126,53 +124,69 @@ local function configure_buttons(device)
...
+
+  -- At the moment, we're taking it for granted that all momentary switches only have 2 positions
+  local current_component_number = 1
+  local component_map = {}
+  for _, ep in ipairs(MS) do -- for each momentary switch endpoint (including main)
+    device.log.debug("Configuring endpoint "..ep)
```
hcarter-775, line 140:
```suggestion
    device.log.debug_with({hub_logs = true}, "Configuring endpoint: " .. ep)
```

hcarter-775: As in, with the `require` statement, you can remove `device.` from all of the `log` statements.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -126,53 +124,69 @@ local function configure_buttons(device)
...
-      -- build the mapping of endpoints to components
-      component_map[string.format("%d", current_component_number)] = ep
-      current_component_number = current_component_number + 1
+local function set_configure(driver, device)
+  local MS = get_reordered_endpoints(driver, device)
+  local main_endpoint = device.MATTER_DEFAULT_ENDPOINT
+  if #MS > 0 then
+    main_endpoint = MS[1] -- the endpoint matching to the non-child device
+    if MS[1] == 0 then main_endpoint = MS[2] end -- we shouldn't hit this, but just in case
+  end
```
hcarter-775, line 133:
```suggestion
  local main_endpoint
  if #MS > 0 and MS[1] == 0 then -- we shouldn't hit this, but just in case
      main_endpoint = MS[2] 
  elseif #MS > 0 then
      main_endpoint = MS[1] -- matches to the non-child device
  else
    main_endpoint = device.MATTER_DEFAULT_ENDPOINT
  end
```
Hi, I think the current layout of this is difficult to understand. I made a suggestion to attempt to clarify what you are intending here by creating less layers. Let me know what you think.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I applied it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I think it's much better. I applied it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I applied it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I applied it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: There was a problem in my test case file. I resolved the problem and applied your suggestion. Thanks~

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I applied it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: I applied it.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
DongHoon-Ryu: > What are you seeing in the test environment that is forcing you to add this handling? Does expect_metadata_update not work for you?

There is no problem calling `set_configure` in normal use cases. However, if there is a `metadata` update in the `init` during the unit test, an assertion occurs during the test case `init`. That's why I added it to avoid this situation during the unit test.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -126,53 +123,66 @@ local function configure_buttons(device)
...
-
-    -- At the moment, we're taking it for granted that all momentary switches only have 2 positions
-    local current_component_number = 1
-    local component_map = {}
-    for _, ep in ipairs(MS) do -- for each momentary switch endpoint (including main)
-      device.log.debug("Configuring endpoint "..ep)
-      -- build the mapping of endpoints to components
-      component_map[string.format("%d", current_component_number)] = ep
-      current_component_number = current_component_number + 1
+local function set_configure(driver, device)
```
nickolas-deboom, line 126: Hi, I was wondering which part of `device_added` is specifically needed to run on each init (is it the endpoint configuration or the profile update for example). Could we put those parts in `set_configure` and leave the rest in `device_added`? Or is everything needed to run on each init?

DongHoon-Ryu: @nickolas-deboom This Aqara Cube update requires both endpoint configuration and the profile update. In particular, when the driver is updated, only the `init` is called, so I think both of the above should be moved to the `init`.

hcarter-775 (approved):

PR #1721, opened 2024-10-24
Fix wallhero translation strings

greens:


greens:
@liuyd96 small capitalization error I didn't notice during review prevented the string replace from happening

The change to the fingerprints file is simply so that the changes to the csv file get deployed.

liuyd96:
> @liuyd96 small capitalization error I didn't notice during review prevented the string replace from happening
> 
> The change to the fingerprints file is simply so that the changes to the csv file get deployed.

Oops, sorry for the typo...

lelandblue (approved):

PR #1728, opened 2024-11-04
New Devices (Matter Switch and Matter Sensor) 4 Elko devices

lelandblue:
Check all that apply

# Type of Change

- [ X ] WWST Certification Request
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

- [ X ] I have performed a self-review of my code

# Description of Change
This PR is in support of 3 new WWST Certification Request(s)

1.   deviceLabel: RFGB-40B(W)/MT, by Elko that has a Matter Device Type ID of (0xf)
2.  deviceLabel: RFWD-100/MT, by Elko, that has a Matter Device Type ID of (0x15)
3.   deviceLabel: RFSAI-62B-SL/MT, by Elko, that has a Matter Device Type ID of (0x100) according to the DCL

Please review the profile assignments. 

These devices may have to be broken up in to multiple PRs later but for now I kept them (to avoid merge conflicts) as one but am open to breaking them up if needed.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -17,6 +17,17 @@ matterManufacturer:
     vendorId: 0x1426
     productId: 0x0003
     deviceProfileName: plug-binary
+#Elko
+  - id: "5170/4096"
+    deviceLabel: RFSAI-62B-SL/MT
+    vendorId: 0x1432
+    productId: 0x1000
+    deviceProfileName: switch-2
+  - id: "5170/4097"
+    deviceLabel: RFGB-40B(W)/MT
+    vendorId: 0x1432
+    productId: 0x1001
+    deviceProfileName: 4-button-battery
 #Eve
   - id: "Eve/Energy/US"
     deviceLabel: Eve Energy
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -17,6 +17,17 @@ matterManufacturer:
     vendorId: 0x1426
     productId: 0x0003
     deviceProfileName: plug-binary
+#Elko
+  - id: "5170/4096"
+    deviceLabel: RFSAI-62B-SL/MT
+    vendorId: 0x1432
+    productId: 0x1000
+    deviceProfileName: switch-2
```
ctowns, line 25: Matter-switch devices that are multi-component should use the `switch-binary` in the fingerprint, and then depend on the driver to create a child device for each additional switch. For example, this device would join with `switch-binary` with the parent device covering the main endpoint (in this case, the lowest numbered endpoint that supports the OnOff cluster), and then an additional child device would be made for the additional child endpoint. This device would use parent-child and appear as two device cards.

With the current state of the matter-switch driver, I would not expect this profile to work as expected by design. These profiles are kept around though so we still have support for switch devices that were onboarded before the `matter-switch` driver was transitioned to a primarily parent-child driver.

lelandblue: I will update this PR to point to the `switch-binary` profile for this device.

lelandblue: @ctowns 👍 Updated.

PR #1754, opened 2024-11-14
Add support for Inovelli VTM31-SN Dimmer Switch

nickolas-deboom:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change adds support for the Inovelli VTM31-SN Dimmer Switch. Initial support for this device was added by [PR 1663](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1663). This PR includes a profile for this device with embedded preferences as well as handling for these preferences.

# Summary of Completed Tests

See here for screenshots from testing with the device.
Also see [test_matter_multi_button_switch_mcd.lua](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua) which tests a mock device with a similar endpoint structure as the Inovelli device.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -19,6 +19,7 @@ local MatterDriver = require "st.matter.driver"
 local lua_socket = require "socket"
 local utils = require "st.utils"
 local device_lib = require "st.device"
+local data_types = require "st.matter.data_types"
 
 local MOST_RECENT_TEMP = "mostRecentTemp"
 local RECEIVED_X = "receivedX"
@@ -154,6 +155,21 @@ local child_device_profile_overrides = {
   { vendor_id = 0x1321, product_id = 0x000D,  child_profile = "switch-binary" },
 }
 
+local fingerprint_profile_overrides = {
+  { vendor_id = 0x1361, product_id = 0x0001 }, -- Inovelli VTM31-SN
+}
+
+local LATEST_CLOCK_SET_TIMESTAMP = "latest_clock_set_timestamp"
+
+local preference_map_inovelli_vtm31sn = {
+  switchMode = {parameter_number = 1, size = data_types.Uint8},
+  smartBulbMode = {parameter_number = 2, size = data_types.Uint8},
+  dimmingEdge = {parameter_number = 3, size = data_types.Uint8},
+  dimmingSpeed = {parameter_number = 4, size = data_types.Uint8},
+  relayClick = {parameter_number = 5, size = data_types.Uint8},
+  ledIndicatorColor = {parameter_number = 6, size = data_types.Uint8},
+}
+
 local detect_matter_thing
 
 local CUMULATIVE_REPORTS_NOT_SUPPORTED = "__cumulative_reports_not_supported"
@@ -246,6 +262,14 @@ local function set_poll_report_timer_and_schedule(device, is_cumulative_report)
     end
     device:set_field(EXPORT_POLL_TIMER_SETTING_ATTEMPTED, true)
   end
+end
+
+local preferences_to_numeric_value = function(new_value)
+  local numeric = tonumber(new_value)
+  if numeric == nil then -- in case the value is Boolean
+    numeric = new_value and 1 or 0
+  end
+  return numeric
 end
 
 local START_BUTTON_PRESS = "__start_button_press"
@@ -499,6 +523,16 @@ local function find_child(parent, ep_id)
   return parent:get_child_by_parent_assigned_key(string.format("%d", ep_id))
 end
 
+local function check_fingerprint_profile_overrides(device)
+  for _, fingerprint in ipairs(fingerprint_profile_overrides) do
+    if device.manufacturer_info.vendor_id == fingerprint.vendor_id and
+      device.manufacturer_info.product_id == fingerprint.product_id then
+      return true
+    end
+  end
+  return false
+end
+
 local function initialize_switch(driver, device)
   local switch_eps = device:get_endpoints(clusters.OnOff.ID)
   local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
@@ -568,6 +602,15 @@ local function initialize_switch(driver, device)
 
   if component_map_used then
     device:set_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON, component_map, {persist = true})
+  end
+
+  -- If there is a custom static profile for the device, configure buttons if needed and
+  -- then return to prevent profile from being updated.
+  if check_fingerprint_profile_overrides(device) then
+    if #button_eps > 0 then
+      configure_buttons(device)
+    end
+    return
   end
 
   if #button_eps > 0 and is_supported_combination_button_switch_device_type(device, main_endpoint) then
@@ -1112,6 +1155,28 @@ local function info_changed(driver, device, event, args)
       -- profile has changed, and we deferred setting up our buttons, so do that now
       configure_buttons(device)
       device:set_field(DEFERRED_CONFIGURE, nil)
+    end
+  end
+
+  if device.network_type ~= device_lib.NETWORK_TYPE_CHILD then
+    if device.manufacturer_info.vendor_id == fingerprint_profile_overrides[1].vendor_id and
+      device.manufacturer_info.product_id == fingerprint_profile_overrides[1].product_id then
+      local time_diff = 3
+      local last_clock_set_time = device:get_field(LATEST_CLOCK_SET_TIMESTAMP)
+      if last_clock_set_time ~= nil then
+        time_diff = os.difftime(os.time(), last_clock_set_time)
+      end
+      device:set_field(LATEST_CLOCK_SET_TIMESTAMP, os.time(), {persist = true})
+      if time_diff > 2 then
+        local preferences = preference_map_inovelli_vtm31sn
+        for id, value in pairs(device.preferences) do
+          if args.old_st_store.preferences[id] ~= value and preferences and preferences[id] then
+            local new_parameter_value = preferences_to_numeric_value(device.preferences[id])
+            local req = clusters.ModeSelect.server.commands.ChangeToMode(device, preferences[id].parameter_number, new_parameter_value)
+            device:send(req)
+          end
+        end
+      end
     end
   end
 end
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -261,6 +261,12 @@ matterManufacturer:
     vendorId: 0x1372
     productId: 0x0002
     deviceProfileName: switch-4
+# Inovelli
+  - id: "4961/1"
+    deviceLabel: Inovelli Dimmer Switch
+    vendorId: 0x1361
+    productId: 0x0001
+    deviceProfileName: inovelli-vtm31-sn
 #Legrand
   - id: "4129/3"
     deviceLabel: Smart Lights Smart Plug
--- a/drivers/SmartThings/matter-switch/profiles/inovelli-vtm31-sn.yml
+++ b/drivers/SmartThings/matter-switch/profiles/inovelli-vtm31-sn.yml
@@ -1,0 +1,127 @@
+name: inovelli-vtm31-sn
+components:
+  - id: main
+    capabilities:
+      - id: switch
+        version: 1
+      - id: switchLevel
+        version: 1
+      - id: refresh
+        version: 1
+      - id: configuration
+        version: 1
+      - id: firmwareUpdate
+        version: 1
+    categories:
+      - name: Switch
+  - id: button1
+    label: Up Button
+    capabilities:
+      - id: button
+        version: 1
+    categories:
+      - name: RemoteController
+  - id: button2
+    label: Down Button
+    capabilities:
+      - id: button
+        version: 1
+    categories:
+      - name: RemoteController
+  - id: button3
+    label: Config Button
+    capabilities:
+      - id: button
+        version: 1
+    categories:
+      - name: RemoteController
+preferences:
+  - name: "switchMode"
+    title: "1. Switch Mode"
+    description: "Use as a Dimmer or an On/Off switch"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        "0": "On/Off + Single (default)"
+        "1": "On/OFf + Dumb"
+        "2": "On/Off + Aux"
+        "3": "On/OFf + Full Wave"
+        "4": "Dimmer + Single"
+        "5": "Dimmer + Dumb"
+        "6": "Dimmer + Aux"
+      default: 0
+  - name: "smartBulbMode"
+    title: "2. Smart Bulb Mode"
+    description: "For use with Smart Bulbs that need constant power and are controlled via commands rather than power. Smart Bulb Mode does not work in Sumb 3-Way Switch mode"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        "0": "Disabled (default)"
+        "1": "Smart Bulb Mode"
+      default: 0
+  - name: "dimmingEdge"
+    title: "3. Dimming Edge"
+    description: "Change dimming type to leading edge (default) or trailing edge for better bulb compatibility"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        "0": "Leading (default)"
+        "1": "Trailing"
+      default: 0
+  - name: "dimmingSpeed"
+    title: "4. Dimming Speed"
+    description: "This changes the speed that the light dims. A setting of '0' turns the light immediately on. Increasing the value slows down the transition speed. Default=25 (2500ms or 2.5s)"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        "0": "Instant"
+        "1": "500ms"
+        "2": "800ms"
+        "3": "1s"
+        "4": "1.5s"
+        "5": "2s"
+        "6": "2.5s (default)"
+        "7": "3s"
+        "8": "3.5s"
+        "9": "4s"
+        "10": "5s"
+        "11": "6s"
+        "12": "7s"
+        "13": "8s"
+        "14": "10s"
+      default: 6
+  - name: "relayClick"
+    title: "5. Relay Click"
+    description: "Audible Click in On/Off mode"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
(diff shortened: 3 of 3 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -154,6 +155,21 @@ local child_device_profile_overrides = {
   { vendor_id = 0x1321, product_id = 0x000D,  child_profile = "switch-binary" },
 }
 
+local fingerprint_profile_overrides = {
+  { vendor_id = 0x1361, product_id = 0x0001 }, -- Inovelli VTM31-SN
+}
+
+local LATEST_CLOCK_SET_TIMESTAMP = "latest_clock_set_timestamp"
+
+local preference_map_inovelli_vtm31sn = {
```
ctowns, line 164: It would make sense to move preferences that are specific to a certain device to a sub driver if we are able to. Or these configs could be moved to their own file at least to keep it separate from the base driver where we should try to handle things as generically as possible

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1114,6 +1157,30 @@ local function info_changed(driver, device, event, args)
...
+    if last_clock_set_time ~= nil then
+      time_diff = os.difftime(os.time(), last_clock_set_time)
+    end
+    device:set_field(LATEST_CLOCK_SET_TIMESTAMP, os.time(), {persist = true})
+    if time_diff > 2 then
+      local preferences = preference_map_inovelli_vtm31sn
+      for id, value in pairs(device.preferences) do
+        if args.old_st_store.preferences[id] ~= value and preferences and preferences[id] then
+          local new_parameter_value = preferences_to_numeric_value(device.preferences[id])
+          local req = clusters.ModeSelect.server.commands.ChangeToMode(device, preferences[id].parameter_number, new_parameter_value)
```
ctowns, line 1178: If the mode cluster is going to be used here, this would be the first use of it in the `matter-switch` driver. This means we'd need to add it as an embedded cluster so it could be used on older FW that doesn't support the mode cluster via the lua libs definitions yet.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1114,6 +1157,30 @@ local function info_changed(driver, device, event, args)
       device:set_field(DEFERRED_CONFIGURE, nil)
     end
   end
+
+  if not device.preferences or device.network_type == device_lib.NETWORK_TYPE_CHILD then
+    return
+  end
+
+  if device.manufacturer_info.vendor_id == fingerprint_profile_overrides[1].vendor_id and
```
ctowns, line 1165: This code is somewhat confusing and I think it could use some comments or documentation somewhere since there is some device specific knowledge here. For example, why is the time diff here and what is the significance of the values?

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: It looks like ModeSelect has been been in the spec since matter 1.0. Is there a way to check when it was first added to the lua libs?

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I believe this code is used to prevent preferences from being changed faster than every two seconds. This was the same way that Inovelli implemented this for the zigbee driver, so it could be a device limitation. I will add some comments to make things a little more clear.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I moved the inovelli-specific handling to a subdriver in 2e8b4b52da67567d53df71878a9515205ce5b46e

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -571,6 +585,15 @@ local function initialize_switch(driver, device)
     device:set_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON, component_map, {persist = true})
   end
 
+  -- If there is a custom static profile for the device, configure buttons if needed and
```
ctowns, line 588: Since the inovelli switch already has a sub driver, I would say we should remove this override from the main driver and just add the special handling for the `init` event in the inovelli sub driver. This way we can try to keep the base driver as generic as possible (although there are already exceptions to this).

So, I would just copy whatever functionality is needed from the base driver and then put it in the sub driver for the init function, and then trim whatever is not needed. Better yet, you could move the `configure_buttons` function to its own file so that it can be shared between the base and sub driver, similar to what is done with the `embedded-cluster-utils` module.

ctowns (commented):
Could you update these changes with the latest from main? I believe there are some code changes in main related to the overrides in the `matter-switch` driver that will impact this PR

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: looks like the ModeSelect was first introduced in [this commit](https://github.ecodesamsung.com/iot-hub/scripting-engine/commit/b158ef313b03901acaaa50737802b35c6e5551c5) with the Matter 1.2 stuff.

nickolas-deboom: Ah, thanks for checking that! I'm currently trying to test whether we can set the device preferences directly, which is what is done in the zigbee driver. If that doesn't end up working in matter (in which case we would continue using ModeSelect for this), I'll add ModeSelect as an embedded cluster for lua libs < 1.2.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -500,6 +504,16 @@ local function find_child(parent, ep_id)
   return parent:get_child_by_parent_assigned_key(string.format("%d", ep_id))
 end
 
+local function check_fingerprint_profile_overrides(device)
```
hcarter-775, line 507: what is the reason we still implement this as so with the sub driver in place?

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Updated in latest commit! I moved the init to the subdriver and also moved `configure-buttons` to its own file as suggested. See 34873ca9bb46350d64dc405e52627d264fcba230

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I moved the initialization for the device to the subdriver in the latest commit: 34873ca9bb46350d64dc405e52627d264fcba230

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I added embedded cluster for ModeSelect in the last commit.

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
```diff
@@ -0,0 +1,301 @@
...
+local DIMMABLE_LIGHT_DEVICE_TYPE_ID = 0x0101
+local EXTENDED_COLOR_LIGHT_DEVICE_TYPE_ID = 0x010D
+local GENERIC_SWITCH_ID = 0x000F
+local ON_OFF_DIMMER_SWITCH_ID = 0x0104
+
+local PRIVATE_CLUSTER_ATTR_ID = 0x122F0000
+local PRIVATE_CLUSTER_ENDPOINT_ID = 0x01
+local PRIVATE_CLUSTER_ID = 0x122FFC31
+
+local device_type_attribute_map = {
```
hcarter-775, line 41: I don't think we need this whole table, since we know what we're subscribing to?

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
```diff
@@ -0,0 +1,301 @@
...
+local function initialize_buttons_and_switches(driver, device, main_endpoint)
+  local switch_eps = device:get_endpoints(clusters.OnOff.ID)
+  local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
+  local component_map = {}
+  local current_component_number = 1
+  -- Since we do not support bindings at the moment, we only want to count clusters
+  -- that have been implemented as server. This can be removed when we have
+  -- support for bindings.
+  local num_switch_server_eps = 0
+  device:try_update_metadata({profile = "inovelli-vtm31-sn"})
```
hcarter-775, line 163: since we have the fingerprint, I'm don't think this is required.

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
```diff
@@ -0,0 +1,301 @@
...
+  local map = device:get_field(COMPONENT_TO_ENDPOINT_MAP) or {}
+  for component, endpoint in pairs(map) do
+    if endpoint == ep then
+      return component
+    end
+  end
+  return "main"
+end
+
+local function device_init(driver, device)
```
hcarter-775, line 210: why is this being overwritten from the main function?

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
nickolas-deboom: True, do you know if it would be needed during a driverSwitched event? Or would it match the fingerprint in that situation? I could just move this line to that handler if not

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
nickolas-deboom: You're right that it's not currently needed, however I realized that 
```
  device:set_field(LATEST_CLOCK_SET_TIMESTAMP, os.time())
```
should be set in `device_init` like it is for the zigbee version of this driver. This value is used in `info_changed` to prevent preferences from being updated too quickly. So, I think we can continue to overload this function with that line included

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
```diff
@@ -0,0 +1,301 @@
...
+  -- The first switch endpoint will be the main component and the three buttons
+  -- will be added as additional components in a MCD profile.
+  component_map["main"] = main_endpoint
+  table.sort(button_eps)
+  for _, ep in ipairs(button_eps) do
+    component_map[string.format("button%d", current_component_number)] = ep
+    current_component_number = current_component_number + 1
+  end
+  device:set_field(COMPONENT_TO_ENDPOINT_MAP, component_map, {persist = true})
+  configure_buttons.configure_buttons(device)
```
hcarter-775, line 173: can we rename things such that we end with:
```suggestion
  button_capability.configure(device)
```
Where we perhaps rename the .lua file itself? This could be better for extending the generic button functionality into there as well, which I think could be a good idea in general.

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
hcarter-775: No, that's still not required I don't think, since the device will still go through the fingerprinting process on a driver switch. Could be good to double check that works how I think it does though.

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
hcarter-775: can we set that in `device_added`?

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
nickolas-deboom: Sounds good, I'll double check :+1:

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
```diff
@@ -0,0 +1,301 @@
...
+  end
+  device.log.warn(string.format("Did not find default endpoint, will use endpoint %d instead", device.MATTER_DEFAULT_ENDPOINT))
+  return device.MATTER_DEFAULT_ENDPOINT
+end
+
+local function find_child(parent, ep_id)
+  return parent:get_child_by_parent_assigned_key(string.format("%d", ep_id))
+end
+
+local function initialize_buttons_and_switches(driver, device, main_endpoint)
```
hcarter-775, line 154: I think we should name this function, since the handling is much less generic than the version in the main file.

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
hcarter-775: or maybe `buttons` instead of `button_capability`. At first I thought button_capability might be clearer, but I'm not too worried about that, and it's a bit of a mouthful.

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
hcarter-775: Actually, I'm still curious what you find but I think I was wrong. I do not think the device fingerprints again on a driver Switched, so I do think we should put that metadata update attempt in the driverSwitched handler

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
nickolas-deboom: I changed it to the list of attributes needed for the child device, to make sure that they get subscribed to

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
nickolas-deboom: Yeah it looks like it's needed after all.

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
nickolas-deboom: Sure thing - I changed it to `buttons` in the latest commit

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
nickolas-deboom: Agreed, renamed it to `initialize_vtm31_sn`

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
nickolas-deboom: I'm not sure that would work since we still need to call `device_init` from `device_added`, so if we put this statement in `device_added`, we would still need to have a `device_init` function to call

nickolas-deboom: Plus, we still need to make sure the attributes specific to the child device get subscribed to.

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
```diff
@@ -0,0 +1,238 @@
...
+  clusters.ColorControl.attributes.CurrentHue,
+  clusters.ColorControl.attributes.CurrentSaturation,
+  clusters.ColorControl.attributes.CurrentX,
+  clusters.ColorControl.attributes.CurrentY,
+  clusters.ColorControl.attributes.ColorTemperatureMireds,
+  clusters.ColorControl.attributes.ColorTempPhysicalMaxMireds,
+  clusters.ColorControl.attributes.ColorTempPhysicalMinMireds
+}
+
+local preference_map = {
```
ctowns, line 47: Do these parameters have any associated labels, or are they just numbers? Without familiarity with this driver/device, this list of parameters is somewhat confusing. Would it be helpful to include a comment for what each of these corresponds to? If it isn't something we easily have access to or wouldn't be helpful, it probably isn't worth worrying about.

nickolas-deboom: I think a comment would be helpful, although this information is the same as what is defined in the profile. Other than that they are just the attribute ID from the private cluster (well, the lower byte of the full attribute, the upper half being 0x122F).

drivers/SmartThings/matter-switch/src/embedded-cluster-utils.lua
```diff
@@ -1,8 +1,22 @@
+-- Copyright 2025 SmartThings
```
nickolas-deboom, line 1: I only re-formatted this file, no actual code change

nickolas-deboom:
@hcarter-775 , I decided to move the remaining button functionality (helper functions, handlers, etc) to `buttons.lua` because I thought it was a bit confusing to have a `buttons.lua` but only with `configure_buttons` implemented there, so why not move everything else related to buttons over there too. Let me know what you think, I believe we have talked about doing this at one point. I can also move it to a separate PR and keep this one focused on the Inovelli device if you prefer (but note that no functionality should have changed after moving the code)

drivers/SmartThings/matter-switch/src/inovelli-vtm31-sn/init.lua
hcarter-775: why do we need to call `device_init` in `device_added`? I think we discussed this on a separate PR before and that extra call is used specifically for child device handling.

nickolas-deboom: The init.lua version of `device_init` wouldn't be called since it's being overloaded, and we need to overload this function to set the clock field. Also, I don't think we can use `require "init"` here because that would mean that both files are including each other

hcarter-775: that's true.... I forgot that the sub-drivers are supersets of the main driver and are `required` in the main driver. Something about this feels weird, but I agree that we can just do it like this for now :+1:

nickolas-deboom: Similar to your idea though, I have been playing around with separating code into more modules, for example see `common-utils` in [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2142/files), which contains any code that's needed by multiple modules

nickolas-deboom:
Closing for the time being due to PR age

PR #1763, opened 2024-11-19
CHAD-14369 Add ezex valve sub-driver

greens:
This valve uses the IAS Zone Status to report low battery and does not otherwise support the PowerConfiguration cluster, which is usually used for reporting the battery. The porting work done for the DTH missed this behavior.

Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [X] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-valve/src/ezex/init.lua
+++ b/drivers/SmartThings/zigbee-valve/src/ezex/init.lua
@@ -1,0 +1,80 @@
+-- Copyright 2024 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local clusters = require "st.zigbee.zcl.clusters"
+local capabilities = require "st.capabilities"
+
+local IASZone = clusters.IASZone
+local Basic = clusters.Basic
+local OnOff = clusters.OnOff
+
+local configuration = {
+  {
+    cluster = IASZone.ID,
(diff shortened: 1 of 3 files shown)


greens:
See the Jira ticket for a detailed explanation.

drivers/SmartThings/zigbee-valve/src/ezex/init.lua
```diff
@@ -0,0 +1,80 @@
...
+  {
+    cluster = OnOff.ID,
+    attribute = OnOff.attributes.OnOff.ID,
+    minimum_interval = 0,
+    maximum_interval = 600,
+    data_type = OnOff.attributes.OnOff.base_type
+  }
+}
+
+local function ias_zone_status_attr_handler(driver, device, zone_status, zb_rx)
```
inasail, line 47: Don't we need the initial battery value for this device?
The battery level for this device are only 3 levels?(100%, 50% 5%)

greens: I went into this a bit more in the Jira ticket, but very _old_ versions of this device only ever reported their battery status using the `IAS Zone Status`, which can report "low battery" or "normal battery". On the DTH, for a while, we interpreted these as 5% and 50%, but the device used a custom metadata that just reported "low" and "normal".

Eventually, it seems they did a firmware update and started using the `PowerConfiguration` cluster. (and the behavior was removed from the DTH: https://github.com/SmartThingsCommunity/SmartThingsPublic/commit/1fdbf086796c87cd32bb90dc5a35667e4eafa4e1)

So only _old_ versions (note how the `can_handle` function is restricted to devices that do not support the `PowerConfiguration` cluster) will use this handling, which will only ever report 5% or 50%, but this matches the old DTH handling of the same.

I expect that _most_ customers in the field have the updated firmware that uses the standard Zigbee cluster for reporting battery level, so this should only ever be a corner case.

greens: as for initial state, for these specific old devices, a read of the IAS Zone Status will occur on add or refresh that will result in a battery value of 5% or 50%.

inasail: Okay, I got it. Thank you.

wkhenon (approved):
Great job, Steven! 🔥

PR #1764, opened 2024-11-19
Fp2:Stabilization of connection and addition of convenience functions

hdlee27:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
**fp2: add get_credential failure handling** 
When 403 error occurs after get_credential during device registration, the driver is waiting for socket timeout to occur while calling fill_body().
In the event of a 403 error, there is no need to fill the body, so in this case, the fill_body() is not called and treated as an error so that a quick return can be made.
This will help users with multiple devices quickly register their devices.
In addition, the log was changed more clearly when get_creditial failed.

**fp2: add nil check in eventsource**
add nil check for sock.
As shown below, there was a driver restart due to sock nil, so there was a request for modification.
2024-11-14_05:01:00.094 | E | edi | <Aqara Presence Sensor (01976eca)> runtime error: [string "cosock.lua"]:313: [string "lunchbox/sse/eventsource.lua"]:346: attempt to index a nil value (field '_sock')

**fp2: Label change for easy distinction** 
if a user with two or more FP2 registers the device, it is registered only under the name 'Aqara FP2', making it difficult to distinguish devices.
I would like to make it easier to distinguish by adding the last four letters of mac to the device name when adding fp2 in the Aqara app.

**Offline event after reconnection attempt to prevent fake offline**
If the socket connection is lost due to the router settings, etc., an offline event has been raised.
This causes confusion by posting a fake offline event in a situation where reconnection is possible.
Therefore, after checking if reconnection is possible, it modifies offline to be raised only when reconnection is not possible.

**fp2: remove socket connection for fp2 resource** 
the socket connection for the rest call was maintained, but it was removed to reduce the resource of fp2

**fp2: update status when sse reconnect** 
modified to update to the current value because status mismatch may occur after sse reconnect and before sse event occurs.

# Summary of Completed Tests

complete fp2 registration test on V3

### DIFF
--- a/drivers/Aqara/aqara-presence-sensor/src/fp2/discovery_helper.lua
+++ b/drivers/Aqara/aqara-presence-sensor/src/fp2/discovery_helper.lua
@@ -33,10 +33,17 @@ function discovery_helper.get_device_create_msg(driver, device_dni, device_ip)
     return nil
   end
 
+  local device_label = device_info.label or "Aqara-FP2"
+  if device_dni then
+    -- To make it easier to distinguish devices, add the last four letters of dni to the label
+    -- for example, if device_info.label is "Aqara-FP2" and device_dni is "00:11:22:33:44:55", then device_label will be "Aqara-FP2 (4455)"
+    device_label = string.format("%s (%s)", device_label, string.sub(string.gsub(tostring(device_dni), ":", ""), -4))
+  end
+
   local create_device_msg = {
     type = "LAN",
     device_network_id = device_dni,
-    label = device_info.label,
+    label = device_label,
     profile = "aqara-fp2-zoneDetection",
     manufacturer = device_info.manufacturerName,
     model = device_info.modelName,
--- a/drivers/Aqara/aqara-presence-sensor/src/init.lua
+++ b/drivers/Aqara/aqara-presence-sensor/src/init.lua
@@ -20,29 +20,27 @@ local function status_update(driver, device)
   local conn_info = device:get_field(fields.CONN_INFO)
   if not conn_info then
     log.warn(string.format("refresh : failed to find conn_info, dni = %s", device.device_network_id))
+    return "failed to find conn_info"
   else
     local resp, err, status = conn_info:get_attr()
 
     if err or status ~= 200 then
       log.error(string.format("refresh : failed to get attr, dni= %s, err= %s, status= %s", device.device_network_id, err,
         status))
+      return "failed to get attr"
     else
       driver.device_manager.handle_status(driver, device, resp)
     end
   end
+  return nil
 end
 
 local function create_sse(driver, device, credential)
   local conn_info = device:get_field(fields.CONN_INFO)
 
-  if not driver.device_manager.is_valid_connection(driver, device, conn_info) then
-    log.warn("create_sse : invalid connection")
-    return
-  end
-
   local sse_url = driver.device_manager.get_sse_url(driver, device, conn_info)
   if not sse_url then
-    log.error("failed to get sse_url")
+    log.error_with({ hub_logs = true }, "failed to get sse_url")
   else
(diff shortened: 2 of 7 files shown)


drivers/Aqara/aqara-presence-sensor/src/fp2/discovery_helper.lua
```diff
@@ -33,10 +33,17 @@ function discovery_helper.get_device_create_msg(driver, device_dni, device_ip)
     return nil
   end
 
+  local device_label = device_info.label or "Aqara-FP2"
+  if device_dni then
+    -- To make it easier to distinguish devices, add the last four letters of dni to the label
+    -- for example, if device_info.label is "Aqara-FP2" and device_dni is "00:11:22:33:44:55", then device_label will be "Aqara-FP2 (4455)"
+    device_label = string.format("%s (%s)", device_label, string.sub(string.gsub(tostring(device_dni), ":", ""), -4))
```
dljsjr, line 40: Can `device_info.label` ever be a user-set value? I think it makes sense to add the four digits in the case that the device name is just `Aqara-FP2`. But if the user sets a label I think we should avoid appending those numbers, since they'll show up on the device card in the mobile app.

drivers/Aqara/aqara-presence-sensor/src/init.lua
```diff
@@ -20,32 +20,27 @@ local function status_update(driver, device)
   local conn_info = device:get_field(fields.CONN_INFO)
   if not conn_info then
     log.warn(string.format("refresh : failed to find conn_info, dni = %s", device.device_network_id))
+    return "failed to find conn_info"
```
dljsjr, line 23: There is no issue with the logic here, but for style purposes, I think that it would be better if this was something more like `return false, "error message"`. The Lua convention that we prefer to use is to return an error in the 2nd return position for fallible functions. And then instead of `return nil` on line 35, you would `return true`.

drivers/Aqara/aqara-presence-sensor/src/init.lua
```diff
@@ -64,7 +59,9 @@ local function create_sse(driver, device, credential)
     end
 
     eventsource.onopen = function()
+      log.info_with({ hub_logs = true }, string.format("Eventsource open: dni= %s", device.device_network_id))
       device:online()
+      status_update(driver, device)
```
dljsjr, line 64: Should the error case be checked and logged or handled here?

cjswedes (approved):
LGTM, but I agree with Doug's style changes.

drivers/Aqara/aqara-presence-sensor/src/fp2/discovery_helper.lua
hdlee27: device_info.label is a value passed from firmware and is fixed to “Aqara-FP2” unless a derivative model of Aqara-FP2 appears. This is not a value that the user can set.

drivers/Aqara/aqara-presence-sensor/src/init.lua
hdlee27: Thank you for letting me know. I modified the code as you said.

drivers/Aqara/aqara-presence-sensor/src/init.lua
hdlee27: Thank you for letting me know. We added error handling logic for this.
I called status_update again in case it failed due to a simple timing issue.

hdlee27:
@dljsjr Thank you for your review. I have modified what you told me.
I tested the entire operation again through hubV3, hubV2, Station, and M5 SmartMonitor Hub with the drivers after modification.
Default Join via Scan Nearby
Remove device and rejoin
Change Zone name
LAN Protocol - Presence Sensor
LAN Protocol - Illuminance Sensor
Reconnection after disconnection by FP2
Reconnection after hub reboot

PR #1768, opened 2024-11-20
Matter Window Covering: Add preference to reverse device operation

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

Add the reverse preference for the window covering device type, so that a user may select to toggle the polarity of the device. This will not change the motor direction of the window covering device, but rather will invert the 0-100% scale in the driver, such that a `shadeLevel` of 100% would present as Closed rather than Open in the `windowShadeLevel` capability, and a `shadeLevel` of 0% would present as Open. Additionally, while the shade level is being increased, the capability would display as "Closing", and while it is being decreased the capability would display as "Opening".

Note that the default value of the preference is false, which will match the previous behavior.

# Summary of Completed Tests

This change was tested with the VDA to ensure that when the preference is not set, the window shade behaves as normal, and when the preference is set, the window shade behaves with its polarity inverted as described above. Also see new unit test.


hcarter-775:
We should merge the recent window covering changes and include reverse in those new profiles as well

PR #1772, opened 2024-11-21
Add support for the color mode attribute

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

In certain situations such as after a driver restart, the hue and saturation attributes of the `colorControl` capability may be overwritten by device attributes that are not currently controlling the color of the device. This change reads the `ColorMode` attribute during init to check the current color mode and then reads the appropriate attributes based on the result. The attribute handlers for the Hue, Saturation, X, and Y attributes wait to emit the hue and saturation capabilities until they are reached for the second time, whether for the read request returning from the device or from the subscription report. This way, we ensure that the hue and saturation capabilities are only emitted based on the attributes controlling the color of the device. Also, any further attribute reports will result in the hue and saturation capabilities being updated, even if the color mode has changed, because all of the attribute handlers have ran at least one time.

# Summary of Completed Tests

Tested using Zemismart, Nanoleaf, and Cync color bulbs. Without these changes, restarting hubcore results in the `CurrentX` and `CurrentY` attributes overriding the `colorControl` capability which results in the ST app displaying a different color than what the device shows.

With the changes, the hue and saturation capabilities are not emitted until the read request for the `CurrentHue` and `CurrentSaturation` attributes are received. The attribute handlers for X and Y return early because those values are only received one time (from the subscription report).

[CHAD-13851]: https://smartthings.atlassian.net/browse/CHAD-13851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -840,14 +841,26 @@ end
 local function hue_attr_handler(driver, device, ib, response)
   if ib.data.value ~= nil then
     local hue = math.floor((ib.data.value / 0xFE * 100) + 0.5)
-    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.hue(hue))
+    local current_color_mode = get_field_for_endpoint(device, COLOR_MODE, ib.endpoint_id)
+    -- don't send capability events if the color of the device is being determined by CurrentX and CurrentY (1) or ColorTemperatureMireds (2)
+    if current_color_mode == 1 or current_color_mode == 2 then
```
ctowns, line 846: This might be more clear just changing this to
```suggestion
    if current_color_mode != 0 then
```

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -955,12 +974,24 @@ local function y_attr_handler(driver, device, ib, response)
...
+      log.info_with({hub_logs=true}, string.format("CurrentX and CurrentY are not currently determining the color of the device. ColorMode is %d", current_color_mode))
+    else
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.hue(h))
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.saturation(s))
+    end
     device:set_field(RECEIVED_X, nil)
   end
 end
 
+local function color_mode_attr_handler(driver, device, ib, response)
```
ctowns, line 989: Are we guaranteed to receive the color mode attribute update before the actually x/y or hue/sat attribute update if it is changed on the device? Do you see this attribute actually change on the device?

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I did it this way in case a device doesn't implement the ColorMode attribute. In that case, `COLOR_MODE` will be nil in the driver, so doing the logic this way maintains the same behavior (we only want to prevent capabilities from being emitted if ColorMode is actually implemented)

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: That is a really good point, and no that is not guaranteed. The ColorMode attribute is always sent by the device if it changes however.

I added new logic to `color_mode_attr_handler` to account for ColorMode not necessarily being received first. When the attribute handlers for Hue, Sat, X, and Y run, they now store the computed Hue and Saturation values, and then emit the hue and sat capabilities depending on the ColorMode. The capabilities are not emitted if ColorMode is set to one of the other 3 possible values. In that case, either the ColorMode is currently set to one of those other values, or it was just changed to a new value but the ColorMode attribute handler hasn't run yet. In the latter case, the ColorMode attribute handler will emit the capabilities using the stored values.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -955,12 +977,47 @@ local function y_attr_handler(driver, device, ib, response)
...
+          set_field_for_endpoint(device, CURRENT_SAT, ib.endpoint_id, nil, {persist = true})
+        end
+      elseif ib.data.value == 1 then
+        if hue ~= nil and sat ~= nil then
+          device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.hue(hue))
+          device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.saturation(sat))
+          set_field_for_endpoint(device, CURRENT_HUE, ib.endpoint_id, nil, {persist = true})
+          set_field_for_endpoint(device, CURRENT_SAT, ib.endpoint_id, nil, {persist = true})
+        end
+      end
```
tpmanley, line 1015: I think we may want a handler for 2 (`ColorTemperatureMireds`) that clears out the saved hue/sat so that later when it switches back to 0 or 1 we don't emit the old hue/sat values

hcarter-775:
also, we still need to update the unit tests to handle this read in init

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -944,7 +946,6 @@ local function x_attr_handler(driver, device, ib, response)
     local h, s, _ = color_utils.safe_xy_to_hsv(x, y)
     device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.hue(h))
     device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.saturation(s))
-    device:set_field(RECEIVED_Y, nil)
```
hcarter-775, line 947: we don't want to get rid of these I think

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -957,7 +958,26 @@ local function y_attr_handler(driver, device, ib, response)
     local h, s, _ = color_utils.safe_xy_to_hsv(x, y)
     device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.hue(h))
     device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.saturation(s))
-    device:set_field(RECEIVED_X, nil)
```
hcarter-775, line 960: including this one

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Oh whoops, I thought I added those in an earlier commit :sweat_smile:

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -957,7 +958,26 @@ local function y_attr_handler(driver, device, ib, response)
...
+    device:send(read_req)
+  elseif color_mode == clusters.ColorControl.types.ColorMode.CURRENT_X_AND_CURRENT_Y then
+    read_req:merge(clusters.ColorControl.attributes.CurrentX:read(device))
+    read_req:merge(clusters.ColorControl.attributes.CurrentY:read(device))
+    device:send(read_req)
+  else -- color_mode = clusters.ColorControl.types.ColorMode.COLOR_TEMPERATURE
+    read_req:merge(clusters.ColorControl.attributes.ColorTemperatureMireds:read(device))
+    device:send(read_req)
   end
 end
```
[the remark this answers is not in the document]
nickolas-deboom: Updated as suggested in latest commit :+1:

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -727,6 +733,11 @@ local function device_init(driver, device)
       end
     end
   end
+
+  if device:supports_capability(capabilities.colorControl) then
+    device:set_field(IGNORE_INITIAL_COLOR_ATTRS, IGNORE_INITIAL_HUE | IGNORE_INITIAL_SAT | IGNORE_INITIAL_X | IGNORE_INITIAL_Y)
```
hcarter-775, line 738: shouldn't this be using & instead of | ?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -65,6 +65,12 @@ local COLOR_TEMP_MAX = "__color_temp_max"
 local LEVEL_BOUND_RECEIVED = "__level_bound_received"
 local LEVEL_MIN = "__level_min"
 local LEVEL_MAX = "__level_max"
+local IGNORE_INITIAL_COLOR_ATTRS = "__ignore_initial_color_attrs"
+local IGNORE_INITIAL_HUE = 0x0001
+local IGNORE_INITIAL_SAT = 0x0002
+local IGNORE_INITIAL_X = 0x0004
+local IGNORE_INITIAL_Y = 0x0008
```
hcarter-775, line 72: nit-ish:
I don't think these need to be globally defined, since we can define them in the handler itself when needed, as well as in init/info_changed.

drivers/SmartThings/matter-switch/src/init.lua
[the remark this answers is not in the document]
hcarter-775: actually, also nit / my personal preference, but I like the name `COLOR_MODE_ATTRS_BITMAP` better, since there's no need to specify the use case in the name if the functions themselves are clear

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -879,17 +889,23 @@ local function level_attr_handler(driver, device, ib, response)
 end
 
 local function hue_attr_handler(driver, device, ib, response)
-  if ib.data.value ~= nil then
-    local hue = math.floor((ib.data.value / 0xFE * 100) + 0.5)
-    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.hue(hue))
+  local ignore_initial_color_attrs = device:get_field(IGNORE_INITIAL_COLOR_ATTRS)
+  if ib.data.value == nil or (ignore_initial_color_attrs ~= nil and ignore_initial_color_attrs & IGNORE_INITIAL_HUE == 0) then
+    device:set_field(IGNORE_INITIAL_COLOR_ATTRS, ignore_initial_color_attrs & ~IGNORE_INITIAL_HUE)
+    return
```
hcarter-775, line 895: once again, I think a helper function would clarify this and cut down repeated code, like:
 ```suggestion
local function ignore_initial_color_read(device)
  local color_attr_bitmap = device:get_field(IGNORE_INITIAL_COLOR_ATTRS)
  if (color_attr_bitmap >> IGNORE_INITIAL_HUE) & 1 > 0 then
    device:set_field(IGNORE_INITIAL_COLOR_ATTRS, color_attr_bitmap | ~IGNORE_INITIAL_HUE)
    return true
  end
  return false
end

local function hue_attr_handler(driver, device, ib, response)
  if ignore_initial_color_read(device) then 
    return
  end
  if ib.data.value ~= nil then
    local hue = math.floor((ib.data.value / 0xFE * 100) + 0.5)
    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.hue(hue))
  end
end
```

This does a coupele things- first, it makes a helper. second, I think there was a bug in the initial bit manip. also, I don't think we should make the nil check part of this, so I pulled it out from that logic.

File: drivers/SmartThings/matter-switch/src/init.lua (hunk 7 above)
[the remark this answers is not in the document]
Comment (nickolas-deboom) [reply drivers/SmartThings/matter-switch/src/init.lua]: Updated the name as suggested but I would prefer to leave the bit names here to make it clear exactly which attributes are being enabled

File: drivers/SmartThings/matter-switch/src/init.lua (hunk 9 above)
Comment (hcarter-775) [reply drivers/SmartThings/matter-switch/src/init.lua]: yes it definitely should, that was a typo ! fixing it for clarity

File: drivers/SmartThings/matter-switch/src/init.lua (hunk 8 above)
Comment (nickolas-deboom) [reply drivers/SmartThings/matter-switch/src/init.lua]: I like these as a global so the bit values are standardized throughout. I did put them together in a table to clean it up a bit

File: drivers/SmartThings/matter-switch/src/init.lua (hunk 9 above)
Comment (nickolas-deboom) [reply drivers/SmartThings/matter-switch/src/init.lua]: I updated it to use a helper function as suggested

File: drivers/SmartThings/matter-switch/src/init.lua (hunk 7 above)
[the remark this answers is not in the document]
Comment (hcarter-775) [reply drivers/SmartThings/matter-switch/src/init.lua]: like it :+1:

File: drivers/SmartThings/matter-switch/src/init.lua [hunk 10]
```diff
@@ -882,17 +909,19 @@ local function level_attr_handler(driver, device, ib, response)
 end
 
 local function hue_attr_handler(driver, device, ib, response)
-  if ib.data.value ~= nil then
-    local hue = math.floor((ib.data.value / 0xFE * 100) + 0.5)
-    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorControl.hue(hue))
+  if ib.data.value == nil or ignore_initial_color_read(device, color_mode_attr_bits.HUE) then
```
hcarter-775, line 912: Maybe I'm overthinking, but I think it would be good to change the order of these statements in all cases: 
```suggestion
  if ignore_initial_color_read(device, color_mode_attr_bits.HUE) or ib.data.value == nil then
```
to ensure we _always_ ignore the first read and only the first read, irrelevant to whether we receive a nil value or not.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -65,6 +65,14 @@ local COLOR_TEMP_MAX = "__color_temp_max"
 local LEVEL_BOUND_RECEIVED = "__level_bound_received"
 local LEVEL_MIN = "__level_min"
 local LEVEL_MAX = "__level_max"
+local COLOR_MODE_ATTRS_BITMAP = "__color_mode_attrs_bitmap"
+local color_mode_attr_bits = {
```
ctowns, line 69: Does color temp also need to be tracked here? Without it, could the intial color be set based on the colorTemp value on the device rather than the H-S or X/Y?

nickolas-deboom: It doesn't need to be tracked because the colorTemp value would result in the colorTemperature capability being emitted, which doesn't affect the colorControl capability.

ctowns (approved):
Good work Nick! The last thing I would suggest is just updating the PR description and completed tests as needed. I know there were some specifics about the testing that needed to be sorted out, so if you think it'd be helpful I'd just put it in the description.

PR #1796, opened 2024-12-04
Add battery level attribute support to matter drivers

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [x] Refactor

# Checklist

# Description of Change

[CHAD-12161](https://smartthings.atlassian.net/browse/CHAD-12161)

matter-sensor, matter-thermostat, matter-lock, matter-switch, and matter-window-covering currently assume that `BatPercentRemaining` is available if the `BAT` feature is supported, but this attribute is only optionally required if this feature is present. The changes in this PR improve the profile selection logic in these drivers by reading the `AttributeList`, checking if `BatPercentRemaining` is available, and then re-profiling the device to use a profile with the `battery` capability rather than `batteryLevel`.

# Summary of Completed Tests

See new test cases.

[CHAD-12161]: https://smartthings.atlassian.net/browse/CHAD-12161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -529,6 +557,13 @@ end
 
 local function info_changed(driver, device, event, args)
   if device.profile.id ~= args.old_st_store.profile.id then
+    for cap_id, attributes in pairs(subscribed_attributes) do
```
ctowns, line 560: This should all be handled by the implementation of the `subscribe` function in the lua libs. See the `populate_subscribed_attributes_from_driver_defaults` function that is called in the `subscribe` definition in the lua libs.

Therefore, I think this can be removed and just the `subscribe()` call is needed.

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -544,20 +579,23 @@ local function do_configure(driver, device)
...
   local profile_name = "base-lock"
-
-  -- check for battery type
-  if #power_source_eps == 0 then
+  device:set_field(PROFILE_BASE_NAME, profile_name, {persist = true})
+  local battery_feature_eps = device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
+  if #battery_feature_eps == 0 then
     profile_name = profile_name .. "-nobattery"
-  elseif #battery_feature_eps == 0 then
+  elseif not device:get_field(SUPPORT_BATTERY_PERCENTAGE) then
```
ctowns, line 588: I would not expect this value to have been set at the time of `doConfigure` since the `doConfigure` event itself sends the `read` request for the AttributeList below.

If #battery feature eps is 0, we could move forward with the metadata update to change the profile to the `no-battery` variety here, but if the #battery eps is > 0, then we could hold off on the metadata update so that the proper profile can be determined and then set in the AttributeList handler. That way, we could avoid setting the profile if we know it is likely to change when we receive an AttributeList report. What do you think?

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -587,14 +625,15 @@ local function device_added(driver, device)
...
-      local power_source_eps = device:get_endpoints(clusters.PowerSource.ID)
-      local battery_feature_eps = device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
       local profile_name = "lock-without-codes"
-      if #power_source_eps == 0 then
+      device:set_field(PROFILE_BASE_NAME, profile_name, {persist = true})
+      local battery_feature_eps = device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
+      if #battery_feature_eps == 0 then
         profile_name = profile_name .. "-nobattery"
-      elseif #battery_feature_eps == 0 then
+      elseif not device:get_field(SUPPORT_BATTERY_PERCENTAGE) then
```
ctowns, line 633: Similar to above, since this is being handled in the `added` handler, I wouldn't expect this value to have been updated yet since it relies on a read request that is sent during the initial onboarding. This will always be false then, and the device would always join to a `-batteryLevel` profile. I would suggest again that we count on the AttributeList handler to determine what battery support needs to be added to the profile, as mentioned above. I think this would have the added benefit of centralizing this logic in once place too!

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -107,6 +116,23 @@ local function handle_battery_charge_level(driver, device, ib, response)
...
+    if attr.value == 0x0C then
+      support_battery_percentage = true
+    end
+  end
+  device:set_field(SUPPORT_BATTERY_PERCENTAGE, support_battery_percentage, {persist = true})
+  local profile_name = device:get_field(PROFILE_BASE_NAME)
+  if not device:get_field(SUPPORT_BATTERY_PERCENTAGE) then
+    profile_name = profile_name .. "-batteryLevel"
+  end
+  device.log.info_with({hub_logs=true}, string.format("Updating device profile to %s.", profile_name))
```
ctowns, line 132: I don't think this needs to be included in the hub logs personally, but could be left in the driver logs.

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -175,15 +303,26 @@ local function do_configure(driver, device)
...
 local function device_init(driver, device)
   log.info("device init")
   device:subscribe()
 end
 
 local function info_changed(driver, device, event, args)
   if device.profile.id ~= args.old_st_store.profile.id then
-    device:subscribe()
     set_boolean_device_type_per_endpoint(driver, device)
+    for cap_id, attributes in pairs(subscribed_attributes) do
```
ctowns, line 318: Similar to my message above, this shouldn't be needed. This same loop is included in the base definition of the `subscribe` function: https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/c401bb4dfece0158aa0903e3843e81e383ff09e7/lua_libs/st/matter/device.lua#L242

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -157,8 +280,13 @@ local function do_configure(driver, device)
     profile_name = profile_name .. "-leak"
   end
 
-  if supports_battery_percentage_remaining(device) then
-    profile_name = profile_name .. "-battery"
+  local battery_feature_eps = device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
+  if #battery_feature_eps > 0 then
+    if device:get_field(SUPPORT_BATTERY_PERCENTAGE) then
```
ctowns, line 285: Same issue as mentioned above, give the order in which these lifecycle events occur, I wouldn't expect this field to have been populated at this point since we'd still need to rely on a read response from the device. Even if the read request is sent and received before this function is called via `doConfigure`, there is still no guarantee of this order of events which could lead to a race condition. Therefore, I would rely on distinguishing between `battery` vs `battery-level` support in the AttributeList handler itself.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -436,10 +441,10 @@ local function create_thermostat_modes_profile(device)
   return thermostat_modes
 end
 
-local function do_configure(driver, device)
+local function match_profile(driver, device)
```
ctowns, line 444: As a blanket statement, the same concerns I mentioned in regards to the timing and order of the lifecycle events on the other two drivers apply to this driver as well.

ctowns (commented):
55 PROFILES 55 BURGERS 55 FRIES

drivers/SmartThings/matter-lock/src/init.lua
nickolas-deboom: Good call! I changed it so that it would only update the profile if it finds that the battery feature is not available. The AttributeList handler can update it again once it runs.

drivers/SmartThings/matter-lock/src/init.lua
nickolas-deboom: Like in the other comment, I changed this to only update the profile if the battery feature is not available!

drivers/SmartThings/matter-lock/src/init.lua
nickolas-deboom: Updated as suggested!

drivers/SmartThings/matter-sensor/src/init.lua
nickolas-deboom: Updated in latest commit!

drivers/SmartThings/matter-sensor/src/init.lua
nickolas-deboom: In 061a1d0fa55282f8b120ebad74cd6a78d865f366, I moved the read request to `do_configure` after `match_profile` runs in order to avoid the race condition. Do you agree with this approach?

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: I made the same change here as I did for matter-sensor (moved the read request from `device_added` to `do_configure` to ensure that it occurs after `match_profile` runs for the first time.

drivers/SmartThings/matter-lock/src/init.lua
```diff
@@ -107,6 +116,23 @@ local function handle_battery_charge_level(driver, device, ib, response)
   end
 end
 
+local function handle_power_source_attribute_list(driver, device, ib, response)
```
ctowns, line 119: I think this is a good idea to use this method of calling `device:try_update_metadata({profile = profile_name})` in the attribute list handler itself, which will ideally avoid any of the race conditions I mentioned.

I think you'll want to use this same idea in the handlers for the other drivers as well, as it looks like it's missing in a few of them.

drivers/SmartThings/matter-sensor/src/init.lua
ctowns: I would update the attribute list handler in this driver and other to match what you did in the matter-lock driver!

drivers/SmartThings/matter-lock/src/init.lua
ctowns: Can you also add an explicit check for the battery level in this function too? I know that attribute is mandatory IF the battery feature is supported, but it'd still be good to be explicit about it here to prevent any bugs in the event that this attribute list handler would be used in the future. Something like this:
```
if attr.value == 0x0C then
  support_battery_percentage = true
end
if attr.value == 0x0E then
    support_battery_level = true
end
------then later add this :-----------
  if support_battery_percentage then
    profile_name = profile_name .. "-battery"
  elseif support_battery_level
  end

Comment (nickolas-deboom) [reply drivers/SmartThings/matter-lock/src/init.lua:118]: I made a change similar to this in cd23b27bffda02948c16d41c293cbe99d46004be. However, unlike the new lock sub driver, the profiles in the main driver don't use "-battery" for devices that support the battery percentage (they use either  "-nobattery", "-batteryLevel", or nothing), so I used slightly different logic.

File: drivers/SmartThings/matter-sensor/src/init.lua (hunk 6 above)
Comment (nickolas-deboom) [reply drivers/SmartThings/matter-sensor/src/init.lua]: I made similar updates in matter-sensor and matter-thermostat, however they are slightly different because they put "-battery" or "-batteryLevel" in the middle of profile names rather than at the end like in matter-lock. So for that reason I set `SUPPORT_BATTERY_PERCENTAGE` or `SUPPORT_BATTERY_LEVEL` in the AttributeList handler and then call `match_profile` rather than updating the profile name directly in the AttributeList handler.

**Review by ctowns — COMMENTED** [review]
We will want to run these changes on real devices if available. These changes are complex and have to do with runtime configuration across multiple driver lifecycle events, so I think this warrants extra testing with real devices to confirm things are working as expected.

File: drivers/SmartThings/matter-sensor/src/init.lua [hunk 9]
```diff
@@ -175,15 +294,25 @@ local function do_configure(driver, device)
   device:try_update_metadata({profile = profile_name})
 end
 
+local function do_configure(driver, device)
+  match_profile(driver, device)
```
ctowns, line 298: This `match_profile` function is essentially just rebuilding the profile name of the device using the `supports_capability` function, and then appending the correct battery support tag on the end of the profile name. The "base" profile name without the battery flag would just be whatever profile the device initially onboarded to. I wouldn't expect the battery support to have been determined by this point (it will be determined in the power source attribute list handler), so you shouldn't need to call `match_profile` at this point since theoretically it would just assign the device to the same profile it already has since the battery configuration has not yet been determined.

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -314,6 +443,32 @@ local function battery_percent_remaining_attr_handler(driver, device, ib, respon
...
+  end
+end
+
+local function power_source_attribute_list_handler(driver, device, ib, response)
+  for _, attr in ipairs(ib.data.elements) do
+    -- Re-profile the device if BatPercentRemaining (Attribute ID 0x0C) or
+    -- BatChargeLevel (Attribute ID 0x0E) is present.
+    if attr.value == 0x0C then
+      device:set_field(SUPPORT_BATTERY_PERCENTAGE, true)
+      match_profile(driver, device)
```
ctowns, line 462: Since this `match_profile` function is really just a "let's switch to a profile for this device based on the battery support it has" function, I would rename it to something like `update_profile_for_battery_support` and just add an additional parameter for the type of battery support. It is only used in this one place, so we shouldn't need to save it on the device object and can just share the value directly as a parameter for one time use here.

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -157,8 +274,10 @@ local function do_configure(driver, device)
     profile_name = profile_name .. "-leak"
   end
 
-  if supports_battery_percentage_remaining(device) then
+  if device:get_field(SUPPORT_BATTERY_PERCENTAGE) then
```
ctowns, line 277: I believe there is a confluence page somewhere that indicates the expected order of capabilities in a profile. I cannot seem to find it, but I believe @hcarter-775 might have updated it somewhat recently.

Anyway, in a similar vein, we should document the expected order of our profile name tags so that when we are creating new profiles, we can ensure that their name are internally consistent. For example, this function essentially outlines the order I am talking about (motion is first, contact is second...preferences are last). This way it is clear the order of precedence when using all these combinations of flags to create the profile name.

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -367,137 +524,15 @@ local matter_driver_template = {
...
-        clusters.BooleanState.attributes.StateValue,
-    },
-    [capabilities.temperatureAlarm.ID] = {
-        clusters.BooleanState.attributes.StateValue,
-    },
-    [capabilities.rainSensor.ID] = {
-        clusters.BooleanState.attributes.StateValue,
-    },
-  },
+  subscribed_attributes = subscribed_attributes,
```
ctowns, line 527: Just wondering if there was a reason for moving these elsewhere other than just a stylistic preference? I am fine either way

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1092,6 +1099,31 @@ local function battery_percent_remaining_attr_handler(driver, device, ib, respon
...
+  elseif ib.data.value == clusters.PowerSource.types.BatChargeLevelEnum.WARNING then
+    device:emit_event(capabilities.batteryLevel.battery.warning())
+  elseif ib.data.value == clusters.PowerSource.types.BatChargeLevelEnum.CRITICAL then
+    device:emit_event(capabilities.batteryLevel.battery.critical())
+  end
+end
+
+local function power_source_attribute_list_handler(driver, device, ib, response)
+  for _, attr in ipairs(ib.data.elements) do
+    -- Re-profile the device if BatPercentRemaining (Attribute ID 0x0C) is present.
```
ctowns, line 1114: I would just add a comment that if the device supports `batteryLevel` only, it would have already been profiled to that profile in the initialization.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -542,12 +555,17 @@ local function do_configure(driver, device)
   end
 end
 
-local function device_added(driver, device)
+local function do_configure(driver, device)
```
ctowns, line 558: I think this should still be the added event handler function, not `do_Configure`. Is there a reason for this change?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1327,10 +1345,35 @@ local function battery_percent_remaining_attr_handler(driver, device, ib, respon
...
+  end
+end
+
+local function power_source_attribute_list_handler(driver, device, ib, response)
+  for _, attr in ipairs(ib.data.elements) do
+    -- Re-profile the device if BatPercentRemaining (Attribute ID 0x0C) or
+    -- BatChargeLevel (Attribute ID 0x0E) is present.
+    if attr.value == 0x0C then
+      device:set_field(SUPPORT_BATTERY_PERCENTAGE, true)
+      match_profile(driver, device)
```
ctowns, line 1364: Same comment I made above with passing the type of battery support as a parameter rather than needing to use variables saved on the device object (that should be one-time use anyway)

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -169,8 +178,36 @@ local function battery_percent_remaining_attr_handler(driver, device, ib, respon
...
+  elseif ib.data.value == clusters.PowerSource.types.BatChargeLevelEnum.CRITICAL then
+    device:emit_event(capabilities.batteryLevel.battery.critical())
+  end
+end
+
+local function power_source_attribute_list_handler(driver, device, ib, response)
+  for _, attr in ipairs(ib.data.elements) do
+    -- Re-profile the device if BatPercentRemaining (Attribute ID 0x0C) is present.
+    if attr.value == 0x0C then
+      device:set_field(SUPPORT_BATTERY_PERCENTAGE, true, {persist = true})
```
ctowns, line 195: I don't think this needs to persist. I would also recommend the same `match_profile` logic I mentioned above by adding an optional parameter for the battery support

ctowns (commented):
One other note for large changes like this that are spread across multiple drivers, you can try to break the change up in to multiple PRs split either by driver or other logical groupings (for example, if matter-sensor and matter-thermostat have a similar implementation, you can group them into the same PR, and then have the other drivers in separate PRs as appropriate).

Creating multiple PRs where things can be logically divided can reduce the burden for each review, and also it can prevent the churn needed on your end to address changes on each driver since some drivers need changes while others don't. We can't always split large PRs into reasonable chunks, but when it makes sense it can help to speed up the overall process so that we can make single-driver-sized changes and get them reviewed, changed, and merged sooner while more complex changes can be worked on in parallel. I think this is fine since we've already gone through a few review cycles on it, but just some food for thought :)

drivers/SmartThings/matter-sensor/src/init.lua
nickolas-deboom: This was a leftover from a previous commit where I moved the definition to the top so that it could be used to subscribe to attributes in `info_changed`, which you pointed out previously wasn't needed! (see [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1796/commits/e49b41dfcead31b6f5e0914a216da59d5047ffa4#r1872194562). I think I'll change it back, if only to reduce the diff of this PR. Speaking of this, the thermostat driver still does the re-calculation of subscribed attributes in `info_changed`. We could remove that as well since it's not needed. [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/f4cefe9d36b90e873f63a1e0ae0a9f8299982553/drivers/SmartThings/matter-thermostat/src/init.lua#L297)

drivers/SmartThings/matter-sensor/src/init.lua
nickolas-deboom: Could you clarify what you mean by the profile that the device initially onboarded to? I thought that was happening right here.

drivers/SmartThings/matter-sensor/src/init.lua
nickolas-deboom: I moved this back btw.

drivers/SmartThings/matter-sensor/src/init.lua
hcarter-775: Not to speak for him, but to respond quickly, he's talking about the profile the device was fingerprinted to, which is how it is initially onboarded before running do_configure.

nickolas-deboom: Ohh gotcha thanks. Is the reason why this profile building exists in matter sensor just for the preference support then?

drivers/SmartThings/matter-sensor/src/init.lua
nickolas-deboom: `match_profile` does the same thing as `do_configure` used to do, and so I put a call to `match_profile` in `do_configure` so that the behavior would be the same. Are you saying that that isn't necessary, and that we should only be updating the profile after the AttributeList is received? If so, changing the name makes sense, but if we want to keep the call to `match_profile` in `do_configure` then maybe the name is ok

drivers/SmartThings/matter-sensor/src/init.lua
ctowns: I did not realize that adding preferences to the profile name was also being handled on `do_configure/match_profile` now, my bad. Originally that function had just been adding battery support, but I did not check closely enough with the changes in the 1.3 update. Therefore, I think you are right, we would still want to call this in do_configure. However, we could probably just call it in `do_configure` if we know there is no battery support, since if there is battery support we would expect that it would be called again once we get the `AttributeList` update, so it would essentially be called twice in that case. Thanks for helping to clarify this!

nickolas-deboom: I think that's a good solution!

hcarter-775: maybe I'm just confused, but what is the solution being proposed here? How do we know if there is or isn't battery support? And if there is, how do we keep try_update_metadata from being called?

nickolas-deboom:
Closing this PR in favor of one PR for each driver:
Matter Lock: [PR 1864](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1864)
Matter Sensor: [PR 1863](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1863)
Matter Switch: [PR 1862](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1862)
Matter Thermostat: [PR 1861](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1861)
Matter Window Covering: [PR 1860](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1860)

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: I changed this back to how it was, see https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1861/files#diff-6ce21995583b1055011482e027d29317199d37c8d3f1fe2b2c03c08bb78cb750 (I originally changed it to reduce the number of read requests, but I agree that it shoudl stay in `device_added`

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I updated this logic to behave the same as the other drivers, so that if the device has battery support, it won't profile during init but will instead send a read request for AttributeList and wait to profile the device until the attribute handler runs. See in the new matter switch PR: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1862/files#diff-7ab55ddbc6a9760ea5b6f928abee244c3142c74389743ce05b895b938ad81f10

drivers/SmartThings/matter-sensor/src/init.lua
nickolas-deboom: As discussed offline, match_profile is called during do_configure if battery is not supported. Otherwise, a read request is sent for AttributeList and the device will be re-profiled during the attribute handler.

PR #1799, opened 2024-12-05
Update profile selection logic to support plug/button devices

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

[CHAD-14507](https://smartthings.atlassian.net/browse/CHAD-14507)

Add a new profile and handling in the matter-switch driver for NEO plug + button device (see WWSTCERT-4757).

# Summary of Completed Tests

[CHAD-14507]: https://smartthings.atlassian.net/browse/CHAD-14507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -571,6 +585,15 @@ local function initialize_switch(driver, device)
...
   end
 
+  -- If there is a custom static profile for the device, configure buttons if needed and
+  -- then return to prevent profile from being updated.
+  if check_fingerprint_profile_overrides(device) then
+    if #button_eps > 0 then
+      configure_buttons(device)
+    end
+    return
+  end
```
hcarter-775, line 595: this check is to stop the device from configuring to light-level-button? It seems like this should be handled more dynamically, since I can see many more devices being added that use a SmartPlug category rather than a Light, no?

hcarter-775: and more than that, devices that don't use switch level. sorry, missed that other difference on my first pass

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -571,6 +585,15 @@ local function initialize_switch(driver, device)
     device:set_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON, component_map, {persist = true})
   end
 
+  -- If there is a custom static profile for the device, configure buttons if needed and
+  -- then return to prevent profile from being updated.
+  if check_fingerprint_profile_overrides(device) then
+    if #button_eps > 0 then
+      configure_buttons(device)
+    end
+    return
```
ctowns, line 594: could we make this an if/else block instead? For readability, I would recommend avoiding early returns in large functions like this if they are not at the top of the function.

nickolas-deboom: I had it this way so that it would configure the buttons and then return right afterwards, to skip the profile selection logic that follows.

ctowns: To clarify, I meant could we put this entire override profile section be in an `if` statement, and then the `else` case will be the part that handles the logic in the rest of the function?
```
if profile_overide then
    config buttons and skip profile selection
else
    do the rest of the function, including profile selection
end
```

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I was debating adding this device to `is_supported_combination_button_switch_device_type`, and then adding a bit of logic to use the `plug-button` profile rather than `light-level-button`, but since there is going to be a specific fingerprint added for this device, I decided to add it to `fingerprint_profile_overrides` and just skip the profile selection logic instead. `fingerprint_profile_overrides` was planning to be added by a different PR anyway so I thought it might be easier this way, but I would be ok with either option if anyone has an opinion on this!

Adding it to `is_supported_combination_button_switch_device_type` would have the benefit of supporting devices with a plug and any number of buttons as well.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: I agree with Cooper here- this may be a good time to refactor this function a little bit.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: yeah, I believe the general take on these kinds of questions has been to create the correct profile dynamically (even if there's a fingerprint) since this will future proof this work, will allow non-wwst devices to profile correctly, and will add some more error-prone logic for wwst devices.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: As I mentioned in another comment, I modified `is_supported_combination_button_switch_device_type` to include the plug device type, and also to return the device type ID so that it can be used in `initialize_switch` to set the correct profile. I think this simplifies the logic quite a bit. Could you take another look and see if it looks alright?

PR #1803, opened 2024-12-09
Update supported mode for every operation

HunsupJung:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
Spec has the following items.

>Cleaning can only be started when the RVC Run Mode cluster’s CurrentMode attribute is set to a
mode that has the Idle mode tag associated with it, and the RVC Operational State cluster’s
OperationalState attribute is set to the Stopped, Paused, Docked or Charging state. #endif::[]

So If it is set to a mode that does not have an idle tag, it should be prevented from switching to a mode that does not have an idle tag because it is not possible.

# Summary of Completed Tests


drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -28,213 +25,232 @@ if version.api < 10 then
...
+        tag = clusters.RvcRunMode.types.ModeTag.IDLE
+        break
+      elseif t.elements.value.value == clusters.RvcRunMode.types.ModeTag.CLEANING then
+        tag = clusters.RvcRunMode.types.ModeTag.CLEANING
+        break
+      elseif t.elements.value.value == clusters.RvcRunMode.types.ModeTag.MAPPING then
+        tag = clusters.RvcRunMode.types.ModeTag.MAPPING
+        break
+      end
+    end
```
nickolas-deboom, line 176: Do we want to return here if the tag isn't valid?
```suggestion
    end
    if tag == 0xFFFF then
      return
    end
```

drivers/SmartThings/matter-rvc/profiles/rvc.yml
```diff
@@ -18,13 +18,59 @@ components:
...
-metadata:
-  mnmn: SmartThingsEdge
-  vid: generic-rvc
+deviceConfig:
+  dashboard:
+    states:
+    - component: main
+      capability: robotCleanerOperatingState
+      version: 1
+  detailView:
```
hcarter-775, line 27: Hi @HunsupJung I'm not super familiar with this profile handling. what is this supposed to be doing for the device card?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -28,213 +25,232 @@ if version.api < 10 then
...
-  return device:get_field(field)
+local function info_changed(driver, device, event, args)
+  for cap_id, attributes in pairs(subscribed_attributes) do
+    if device:supports_capability_by_id(cap_id) then
+      for _, attr in ipairs(attributes) do
+        device:add_subscribed_attribute(attr)
+      end
+    end
+  end
+  device:subscribe()
```
hcarter-775, line 86:
```suggestion
if device.profile.id ~= args.old_st_store.profile.id then
    device:subscribe()
end
```
a little cleaner and does the same thing

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -28,213 +25,232 @@ if version.api < 10 then
...
 end
 
-local function get_labels_of_supported_modes(device, field_prefix, cluster_id, attribute_id, response)
-  local supported_modes = get_field_labels_of_supported_modes(device, field_prefix)
-  if supported_modes ~= nil then
-    return supported_modes
+-- Helper functions --
+local function update_supported_arguments(device, current_run_mode, operating_state)
+  device.log.info(string.format("update_supported_arguments: %s, %s", current_run_mode, operating_state))
```
hcarter-775, line 91: this log should be moved, since it's possible, per the early return, that either current_run_mode or operating_state is nil.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -28,213 +25,232 @@ if version.api < 10 then
...
-    end
+  -- Error state
+  if operating_state == "Error" then
+    -- Set runMode to empty
+    local component = device.profile.components["runMode"]
+    local event = capabilities.mode.supportedArguments({}, {visibility = {displayed = false}})
+    device:emit_component_event(component, event)
+    -- Set cleanMode to empty
+    component = device.profile.components["cleanMode"]
+    device:emit_component_event(component, event)
```
hcarter-775, line 104:
```suggestion
    local empty_event = capabilities.mode.supportedArguments({}, {visibility = {displayed = false}})
    -- Set runMode to empty
   device:emit_component_event(device.profile.components["runMode"], event)
    -- Set cleanMode to empty
   device:emit_component_event(device.profile.components["cleanMode"], event)
```
maybe we could write it more like this? Sorry, the original wording confused me a lot when I first read it.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -28,213 +25,232 @@ if version.api < 10 then
...
+  local current_tag = 0xFFFF
+  local supported_run_modes = device:get_field(RUN_MODE_SUPPORTED_MODES) or {}
+  for i, mode in ipairs(supported_run_modes) do
+    if mode[1] == current_run_mode then
+      current_tag = mode[2]
       break
     end
   end
+  if current_tag == 0xFFFF then
+    return
```
hcarter-775, line 118: maybe add a failure log here?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -28,213 +25,232 @@ if version.api < 10 then
...
 
-local function get_labels_of_supported_modes(device, field_prefix, cluster_id, attribute_id, response)
-  local supported_modes = get_field_labels_of_supported_modes(device, field_prefix)
-  if supported_modes ~= nil then
-    return supported_modes
+-- Helper functions --
+local function update_supported_arguments(device, current_run_mode, operating_state)
+  device.log.info(string.format("update_supported_arguments: %s, %s", current_run_mode, operating_state))
+  if current_run_mode == nil or operating_state == nil then
+    return
```
hcarter-775, line 93: maybe add a failure log here?

nickolas-deboom:
I tested this with the VDA and it looks good except the operational state is not updating (it always says Stopped). I think this is an issue with the VDA because it is not sending an update for Operational State after changing the run mode

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -28,213 +25,232 @@ if version.api < 10 then
...
   end
+  if current_tag == 0xFFFF then
+    return
+  end
 
-  if is_idle_current_mode then
-    return labels
+  -- Check whether non-idle mode can be selected or not
+  local op_state = capabilities.robotCleanerOperatingState.operatingState
+  local can_be_selected_as_non_idle = 0
```
hcarter-775, line 123: nit: 
```suggestion
  local can_be_non_idle = false
```
and then true in the other places where that makes sense

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -321,45 +345,46 @@ local function handle_robot_cleaner_mode(driver, device, cmd)
   end
 end
 
-local matter_driver_template = {
+local function handle_refresh(driver, device, command)
+  local req = clusters.RvcRunMode.attributes.CurrentMode:read(device)
+  req:merge(clusters.RvcCleanMode.attributes.CurrentMode:read(device))
+  req:merge(clusters.RvcOperationalState.attributes.OperationalState:read(device))
+  device:send(req)
+end
```
hcarter-775, line 353: Isn't this the refresh handling by default? I may be mistaken, but I want to note it.

hcarter-775:
> I tested this with the VDA and it looks good except the operational state is not updating (it always says Stopped). I think this is an issue with the VDA because it is not sending an update for Operational State after changing the run mode

Can a unit test ensure this works as expected, or does one already exist? Thank you :smile:

drivers/SmartThings/matter-rvc/profiles/rvc.yml
HunsupJung: this is `device configuration`. Previously, `device configuration` was defined elsewhere and linked to a profile. But now we can specify the details of `device configuration` directly in the profile.
`device configuration` is used for mode capability. it's applied to change the `display type` of `mode` capability to `dynamicList`.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Thank you. I will modify it.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: By putting the log in front of the `if` statement, we can check whether only one of `current_run_mode` and `operating_state` is null or both are null. (If null is outputted as %s, it is outputted as nil.) If it is not null, it is possible to check whether the value is properly transmitted.
Do you want to add logs inside `if` statement and after `if` statement?

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: I will modify it~!

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Okay, I will add failure log

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Yes, we have to insert the supported_mode when tag isn't 0xFFFF. I will modify it. 
```lua
if tag ~= 0xFFFF then
    table.insert(supported_modes, mode.elements.label.value)
    table.insert(supported_modes_with_tag, {mode.elements.label.value, tag})
end
```

drivers/SmartThings/matter-rvc/src/init.lua
hcarter-775: ah, I see. Yes, this is fine as-is.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -28,213 +25,232 @@ if version.api < 10 then
...
+  local supported_clean_modes = device:get_field(CLEAN_MODE_SUPPORTED_MODES) or {}
+  supported_arguments = {}
+  for i, mode in ipairs(supported_clean_modes) do
+    table.insert(supported_arguments, mode)
+  end
+
+  -- Send event to set supported clean modes
+  local component = device.profile.components["cleanMode"]
+  local event = capabilities.mode.supportedArguments(supported_arguments, {visibility = {displayed = false}})
+  device:emit_component_event(component, event)
```
nickolas-deboom, line 147:
```suggestion
  local component = device.profile.components["cleanMode"]
  if component ~= nil then
    local event = capabilities.mode.supportedArguments(supported_arguments, {visibility = {displayed = false}})
    device:emit_component_event(component, event)
  end
```
In case cleanMode isn't supported and the device is using the `rvc.yml` profile.

drivers/SmartThings/matter-rvc/src/init.lua
nickolas-deboom:
```suggestion
    if component ~= nil then
      device:emit_component_event(component, event)
    end
```
Similar thing here.

hcarter-775 (approved):
lgtm! remember to squash your commits before merging :+1:

PR #1812, opened 2024-12-10
Update matter-lock driver for U-tec

HunsupJung:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
https://smartthings.atlassian.net/browse/IOTE-4705

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-lock/profiles/lock-user-pin-schedule-battery.yml
+++ b/drivers/SmartThings/matter-lock/profiles/lock-user-pin-schedule-battery.yml
@@ -1,0 +1,31 @@
+name: lock-user-pin-schedule-battery
+components:
+- id: main
+  capabilities:
+  - id: lock
+    version: 1
+    config:
+      values:
+      - key: "lock.value"
+        enabledValues:
+        - locked
+        - unlocked
+        - not fully locked
+  - id: lockAlarm
+    version: 1
+  - id: remoteControlStatus
+    version: 1
+  - id: lockUsers
+    version: 1
+  - id: lockCredentials
+    version: 1
+  - id: lockSchedules
+    version: 1
+  - id: battery
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: SmartLock
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -33,7 +33,7 @@ local NEW_MATTER_LOCK_PRODUCTS = {
   {0x115f, 0x2802}, -- AQARA, U200
   {0x115f, 0x2801}, -- AQARA, U300
   {0x147F, 0x0001}, -- U-tec
-  {0x10E1, 0x2002} -- VDA
+  {0x10E1, 0x2002}  -- VDA
 }
 
 local PROFILE_BASE_NAME = "__profile_base_name"
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -376,20 +376,26 @@ end
 ---------------------------------
 local function handle_power_source_attribute_list(driver, device, ib, response)
   local support_battery_percentage = false
+  local support_battery_level = false
   for _, attr in ipairs(ib.data.elements) do
     -- Re-profile the device if BatPercentRemaining (Attribute ID 0x0C) is present.
     if attr.value == 0x0C then
       support_battery_percentage = true
     end
+    if attr.value == 0x0E then
+      support_battery_level = true
+    end
   end
   local profile_name = device:get_field(PROFILE_BASE_NAME)
-  if support_battery_percentage then
-    profile_name = profile_name .. "-battery"
-  else
-    profile_name = profile_name .. "-batteryLevel"
+  if profile_name ~= nil then
+    if support_battery_percentage then
+      profile_name = profile_name .. "-battery"
+    elseif support_battery_level then
+      profile_name = profile_name .. "-batteryLevel"
+    end
+    device.log.info(string.format("Updating device profile to %s.", profile_name))
+    device:try_update_metadata({profile = profile_name})
   end
-  device.log.info(string.format("Updating device profile to %s.", profile_name))
-  device:try_update_metadata({profile = profile_name})
 end
 
 -------------------------------
@@ -1790,7 +1796,6 @@ local new_matter_lock_handler = {
     capabilities.lockUsers,
     capabilities.lockCredentials,
     capabilities.lockSchedules,
-    capabilities.remoteControlStatus,
     capabilities.battery,
     capabilities.batteryLevel
   },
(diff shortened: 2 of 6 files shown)


nickolas-deboom:
After an initial review, the changes look good to me overall. I have two comments - (1) are there additional profiles needed for the possible combinations of lock/user/pin/schedule + battery/batteryLevel? Or are we just providing profiles for known configurations at this time? (2) Would you be able to create a few test cases similar to what I have here? https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1796/files#diff-188bbf519a245b9967a0a86552cf7b20bef0a2b5ce6598196cca7e6591acc56d

In order to show that the `BatPercentRemaining` attribute being present results in the correct profile being chosen.

HunsupJung:
>(1) are there additional profiles needed for the possible combinations of lock/user/pin/schedule + battery/batteryLevel? Or are we just providing profiles for known configurations at this time?

There may be other possible combinations. But I added lock-user-pin-batteryLevel, lock-user-pin-battery and lock-user-pin-scheduled-battery which are highly likely. For U-tec, the lock-user-pin-batteryLevel profile is enough. 
Do you want to consider all edge cases?

>(2) Would you be able to create a few test cases similar to what I have here?

Okay, I will update it.

nickolas-deboom:
> > (1) are there additional profiles needed for the possible combinations of lock/user/pin/schedule + battery/batteryLevel? Or are we just providing profiles for known configurations at this time?
> 
> There may be other possible combinations. But I added lock-user-pin-batteryLevel, lock-user-pin-battery and lock-user-pin-scheduled-battery which are highly likely. For U-tec, the lock-user-pin-batteryLevel profile is enough. Do you want to consider all edge cases?
> 
> > (2) Would you be able to create a few test cases similar to what I have here?
> 
> Okay, I will update it.

I think it's ok to include just the likely ones for now rather than every combination.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1685,7 +1752,10 @@ local new_matter_lock_handler = {
     capabilities.lock,
     capabilities.lockUsers,
     capabilities.lockCredentials,
-    capabilities.lockSchedules
+    capabilities.lockSchedules,
+    capabilities.remoteControlStatus,
```
ctowns, line 1756: There is no handler for `remoteControlStatus`, should there be or should this be removed?

drivers/SmartThings/matter-lock/src/test/test_new_matter_lock_battery.lua
```diff
@@ -0,0 +1,144 @@
...
+  subscribe_request:merge(DoorLock.attributes.RequirePINforRemoteOperation:subscribe(mock_device))
+  subscribe_request:merge(DoorLock.events.LockOperation:subscribe(mock_device))
+  subscribe_request:merge(DoorLock.events.DoorLockAlarm:subscribe(mock_device))
+  subscribe_request:merge(DoorLock.events.LockUserChange:subscribe(mock_device))
+  test.socket["matter"]:__expect_send({mock_device.id, subscribe_request})
+  test.mock_device.add_test_device(mock_device)
+end
+test.set_test_init_function(test_init)
+
+test.register_coroutine_test(
```
ctowns, line 77: Could you create a test case for the following cases also?
1) battery percent remaining is available ✅ 
2) battery charge level only is available ❌ 
3) battery charge level and battery percent remaining are available ❌ 

You will just need to modify the attributes list in the `AttributeList:build_test_report_data` function you are using to test these different combinations of attributes in the new test cases.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -340,6 +359,49 @@ local function max_year_schedule_of_user_handler(driver, device, ib, response)
...
 end
 
+---------------------------------
+-- Power Source Attribute List --
+---------------------------------
+local function handle_power_source_attribute_list(driver, device, ib, response)
+  local support_battery_percentage = false
+  for _, attr in ipairs(ib.data.elements) do
+    -- Re-profile the device if BatPercentRemaining (Attribute ID 0x0C) is present.
+    if attr.value == 0x0C then
```
ctowns, line 369: Can we add an explicit check for the `BatChargeLevel` attribute here?
```
if attr.value == 0x0C then
  support_battery_percentage = true
end
if attr.value == 0x0E then
    support_battery_level = true
end
------then later add this :-----------
  if support_battery_percentage then
    profile_name = profile_name .. "-battery"
  elseif support_battery_level
  end

**Review by ctowns — COMMENTED** [review]
This looks pretty good to me! After adding the additional unit tests and making the small change in the handler, I think this will be good to go - thank you @HunsupJung !

Also, has this been regression tested with the other locks that use this driver?

**Comment by Kwang-Hui** [comment]
I tested this PR with Aqara U200/U300 door lock - it was fine. Thank you.

File: drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua (hunk 1 above)
Comment (HunsupJung) [reply drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua:1756]: I don't think it's needed. I will remove it.

File: drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua (hunk 3 above)
Comment (HunsupJung) [reply drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua:369]: Okay, I will

File: drivers/SmartThings/matter-lock/src/test/test_new_matter_lock_battery.lua (hunk 2 above)
Comment (HunsupJung) [reply drivers/SmartThings/matter-lock/src/test/test_new_matter_lock_battery.lua:77]: I didn't understand your explanation.
There are first cases and second one below. And there's no different between first one and third one.
I do not know which test case to upload additionally.

Comment (HunsupJung) [reply drivers/SmartThings/matter-lock/src/test/test_new_matter_lock_battery.lua:77]: I thought the second case was added, but it wasn't. I will modify it.

Comment (HunsupJung) [reply drivers/SmartThings/matter-lock/src/test/test_new_matter_lock_battery.lua:77]: For reference, it is impossible to support only `Battery percent remaining`. Because if server supports BAT feature, `Battery change level` is mandatory. I will create as follows.

1. Attributes related to BAT feature is not available.
2. battery charge level only is available.
3. battery charge level and battery percent remaining are available.

**Comment by Kwang-Hui** [comment]
@ctowns could you review this pr again?

File: drivers/SmartThings/matter-lock/profiles/lock-user-pin-schedule-battery.yml [hunk 4]
```diff
@@ -0,0 +1,31 @@
+name: lock-user-pin-schedule-battery
```
hcarter-775, line 1: for completion, could we also add this profile with -batteryLevel?

HunsupJung: Sure. I will add the lock-user-pin-schedule-batteryLevel profile.

Kwang-Hui:
@hcarter-775 Could you please review again?

hcarter-775 (approved):
looks good! just rebase and resolve conflicts :+1:

PR #1817, opened 2024-12-10
Matter Window Covering: Support the tilt feature

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

[CHAD-14539](https://smartthings.atlassian.net/browse/CHAD-14539)

This change adds support for the tilt feature in the `WindowCovering` cluster, to subscribe to the `CurrentPositionTiltPercent100ths` attribute and map this value to the the `windowShadeTiltLevel` capability.

# Summary of Completed Tests

Tested with the VDA Window Covering device. This has recently been updated to support tilt in addition to lift. Also see new unit tests.

[CHAD-14539]: https://smartthings.atlassian.net/browse/CHAD-14539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-only-battery.yml
+++ b/drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-only-battery.yml
@@ -1,0 +1,16 @@
+name: window-covering-tilt-only-battery
+components:
+- id: main
+  capabilities:
+  - id: windowShade
+    version: 1
+  - id: windowShadeTiltLevel
+    version: 1
+  - id: battery
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Blind
--- a/drivers/SmartThings/matter-window-covering/src/init.lua
+++ b/drivers/SmartThings/matter-window-covering/src/init.lua
@@ -20,6 +20,8 @@ local log = require "log"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
 
+local CURRENT_LIFT = "__current_lift"
+local CURRENT_TILT = "__current_tilt"
 local battery_support = {
   NO_BATTERY = "NO_BATTERY",
   BATTERY_LEVEL = "BATTERY_LEVEL",
@@ -46,7 +48,16 @@ local function component_to_endpoint(device, component_name)
 end
 
 local function match_profile(device, battery_supported)
+  local lift_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.LIFT})
+  local tilt_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.TILT})
   local profile_name = "window-covering"
+  if #tilt_eps > 0 then
+    profile_name = profile_name .. "-tilt"
+    if #lift_eps == 0 then
+      profile_name = profile_name .. "-only"
+    end
+  end
+
   if battery_supported == battery_support.BATTERY_PERCENTAGE then
     profile_name = profile_name .. "-battery"
   elseif battery_supported == battery_support.BATTERY_LEVEL then
@@ -100,13 +111,23 @@ local function device_removed(driver, device) log.info("device removed") end
 -- capability handlers
 local function handle_preset(driver, device, cmd)
   local endpoint_id = device:component_to_endpoint(cmd.component)
-  local lift_value = 100 - device.preferences.presetPosition
-  local hundredths_lift_percent = lift_value * 100
-  local req = clusters.WindowCovering.server.commands.GoToLiftPercentage(
-                device, endpoint_id, hundredths_lift_percent
-              )
-
-  device:send(req)
+  local lift_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.LIFT})
+  local tilt_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.TILT})
+  if #lift_eps > 0 then
+    local lift_value = 100 - device.preferences.presetPosition
+    local hundredths_lift_percent = lift_value * 100
+    local req = clusters.WindowCovering.server.commands.GoToLiftPercentage(
+      device, endpoint_id, hundredths_lift_percent
+    )
+    device:send(req)
+  end
+  if #tilt_eps > 0 then
+    -- Use default preset tilt percentage to 50 until a canonical preference is created for preset tilt position
+    local req = clusters.WindowCovering.server.commands.GoToTiltPercentage(
+      device, endpoint_id, 50 * 100
+    )
+    device:send(req)
+  end
 end
 
 -- close covering
@@ -130,8 +151,7 @@ local function handle_pause(driver, device, cmd)
   device:send(req)
 end
 
--- move to shade level
--- beteween 0-100
+-- move to shade level between 0-100
 local function handle_shade_level(driver, device, cmd)
   local endpoint_id = device:component_to_endpoint(cmd.component)
   local lift_percentage_value = 100 - cmd.args.shadeLevel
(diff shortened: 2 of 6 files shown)


drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-only-battery.yml
```diff
@@ -0,0 +1,21 @@
+name: window-covering-tilt-only-battery
+components:
+- id: main
+  capabilities:
+  - id: windowShade
+    version: 1
+  - id: windowShadePreset
```
tpmanley, line 7: This capability is currently implemented using `GoToLiftPercentage`, so I think we need to do one of two things:

1)  Only support `windowShadePreset` and the associated `presetPosition` preference if the `windowShadeLevel` capability is supported
2) Allow `windowShadePreset` to be used to set the tilt to a preset value

If we go with Option 2, we'd also have the case where a device supports both `windowShadeLevel` and `windowShadeTiltLevel `. That would get a bit tricky to deal with the setting of the preset position for both level and tilt. We may need a new preference for that case.

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -123,29 +131,42 @@ local function handle_shade_level(driver, device, cmd)
...
-  if position == 0 then
-    device:emit_event_for_endpoint(ib.endpoint_id, windowShade.closed())
-  elseif position == 100 then
-    device:emit_event_for_endpoint(ib.endpoint_id, windowShade.open())
-  elseif position > 0 and position < 100 then
-    device:emit_event_for_endpoint(ib.endpoint_id, windowShade.partially_open())
-  else
-    device:emit_event_for_endpoint(ib.endpoint_id, windowShade.unknown())
+-- move to shade tilt level between 0-100
+local function handle_shade_tilt_level(driver, device, cmd)
```
tpmanley, line 135: What events are generated for the `windowShade` capability for a tilt-only device? It looks like they are only generated when the lift position changes.

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -41,6 +41,15 @@ end
...
   local profile_name = "window-covering"
+  local lift_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.LIFT})
+  local tilt_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.TILT})
+  if #tilt_eps > 0 then
+    if #lift_eps > 0 then
+      profile_name = profile_name .. "-tilt"
+    else
+      profile_name = profile_name .. "-tilt-only"
+    end
+  end
```
hcarter-775, line 52:
```suggestion
  local lift_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.LIFT})
  local tilt_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.TILT})
  
  if #tilt_eps > 0 then
    profile_name = profile_name .. "-tilt"
    if #[lift_eps](device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.LIFT})) == 0 then
      profile_name = profile_name .. "-only"
    end
  end
```

would this be more future-proof against further configure logic changes? not sure

drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-only-battery.yml
nickolas-deboom: My feeling is that it might make sense to go with Option 2 and create a new preference for the tilt preset position.

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: If I understand your question correctly, the `windowShade` capability is now updated when either the lift position or the tilt level changes. See the changes made to `current_pos_handler`.

drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-only-battery.yml
nickolas-deboom: I made this change in d5424d31248d3d81549e817624f8905565ec15da, let me know if that looks ok or not. Also, I created it as an embedded preference, is that ok or would we want this to be created as an 'official' preference?

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: Good idea! I made this change in d5424d31248d3d81549e817624f8905565ec15da. I also realized that the battery check was going to overwrite the profile name rather than append to it so I fixed that as well.

drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-only-battery.yml
nickolas-deboom: After discussion, I removed the embedded preference and used a default value of 50% tilt until a canonical preference is created for preset window tilt.

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -123,29 +139,42 @@ local function handle_shade_level(driver, device, cmd)
...
+end
+
+-- current lift/tilt percentage, changed to 100ths percent
+local current_pos_handler = function(attribute)
+  return function(driver, device, ib, response)
+    if ib.data.value == nil then
+      return
+    end
+    local windowShade = capabilities.windowShade.windowShade
+    local position = 100 - math.floor((ib.data.value / 100))
```
SathvikGatti, line 160: @nickolas-deboom I think the position values of both tilt and lift must be considered on every report when deciding the window shade's status as currently the shade's status is based solely on the attribute that was last reported.
For example: 
1. Lift is at 100%. Window Shade says "Open".
2. Move Tilt to 0%. Window Shade would now be shown as "Closed".

Please let me know if I'm overlooking anything, or if this is intentional.

nickolas-deboom: Hi @SathvikGatti , I think you are correct, and I made an update to consider both lift and tilt on each position update. The lift/tilt values are captured each time `current_pos_handler` runs, and new logic is introduced with a few considerations for shades that support both lift and tilt. For example, if the lift is at 100%, the tilt value does not matter, the shade will be considered open. Similarly, if the lift is at 0%, the shade is only considered closed if the tilt value is also 0% (it would be partially open for any other tilt value).

SathvikGatti: @nickolas-deboom, thanks for the update! There is a scenario where the issue persists when a device with both Tilt and Lift features that is also handled by sub-driver is onboarded.

This issue is seen as the lift attribute handling is done in the sub-driver, while the tilt reports are handled in the base/main driver. Due to this, the window shade status shown is based on the attribute that was last reported.

Currently, only the Virtual Device App is listed in the sub-driver's can_handle. Should this be handled as well?

CC: @HunsupJung

HunsupJung: @SathvikGatti 
I'm against modifying to sub driver. VDA's role is to test drivers before actual products are released. Virtual window covering is using a sub driver, which is only used to test edge cases. I don't think it's a good direction. VDA should use the main driver used by most real devices. There are two ways to do that.

1. Remove the VDA from the sub-driver and modify the VDA to operate like a device that uses a main driver.
2. Create a new virtual window covering using the main driver.

I think we should decide between the two.

nickolas-deboom: @HunsupJung I agree with you and think the VDA should probably be updated to be compatible with the main driver.

nickolas-deboom: Besides, the VDA window covering device doesn't support tilt so no update is needed there.

HunsupJung: @nickolas-deboom 
Could you remove vid/pid of VDA on sub-driver? We will modify VDA to operate in main driver.

hcarter-775: @HunsupJung if you are going to put up a PR to remove the VDA from sub driver, the vid/pid should be removed in that PR whenever it comes out, no?

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -123,29 +141,80 @@ local function handle_shade_level(driver, device, cmd)
...
-  elseif position > 0 and position < 100 then
-    device:emit_event_for_endpoint(ib.endpoint_id, windowShade.partially_open())
-  else
-    device:emit_event_for_endpoint(ib.endpoint_id, windowShade.unknown())
+-- move to shade tilt level between 0-100
+local function handle_shade_tilt_level(driver, device, cmd)
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local tilt_percentage_value = 100 - cmd.args.level
+  local hundredths_tilt_percentage = tilt_percentage_value * 100
+  local req = clusters.WindowCovering.server.commands.GoToTiltPercentage(
```
hcarter-775, line 149: @nickolas-deboom just curious cause it's all over the place, why are we using the req() to send(req) system all over the place here? Why not use emit_event_for_endpoint() in these cases?

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -123,29 +141,80 @@ local function handle_shade_level(driver, device, cmd)
...
+      if tilt_position == 0 then
+        device:emit_event_for_endpoint(ib.endpoint_id, windowShade.closed())
+      elseif tilt_position > 0 and tilt_position < 100 then
+        device:emit_event_for_endpoint(ib.endpoint_id, windowShade.partially_open())
+      elseif tilt_position == 100 then
+        device:emit_event_for_endpoint(ib.endpoint_id, windowShade.open())
+      else
+        device:emit_event_for_endpoint(ib.endpoint_id, windowShade.unknown())
+      end
+    end
```
hcarter-775, line 210:
```suggestion
-- Update the window shade status according to the lift and tilt positions.
    --   LIFT     TILT      Window Shade
    --   100      any       Open
    --   1-99     any       Partially Open
    --   0        1-100     Partially Open
    --   0        nil       Closed
    --   0        0         Closed
    --   nil      100       Open
    --   nil      1-99      Partially Open
    --   nil      0         Closed
    -- Note that lift or tilt may be nil if either the window shade does not
    -- support them or if they haven't been received from a device report yet.

    if lift_position == 100 then
      device:emit_event_for_endpoint(ib.endpoint_id, windowShade.open())
    end

    if lift_position > 0 then
      device:emit_event_for_endpoint(ib.endpoint_id, windowShade.partially_open())
    end

    if lift_position == 0 then
      if tilt_position > 0 then
        device:emit_event_for_endpoint(ib.endpoint_id, windowShade.partially_open())
      else
        device:emit_event_for_endpoint(ib.endpoint_id, windowShade.closed())
      end
    end

    if lift_position == nil then
      if tilt_position == 100 then
        device:emit_event_for_endpoint(ib.endpoint_id, windowShade.open())
      elseif tilt_position > 0 then
        device:emit_event_for_endpoint(ib.endpoint_id, windowShade.partially_open())
      else
        device:emit_event_for_endpoint(ib.endpoint_id, windowShade.closed())
      end
    end

```
I made an attempt to clean up this giant if block. Lmk what you think!

hcarter-775: made a quick edit to the if block structure after testing and realizing the previous version I made was failing nil checks.

nickolas-deboom: I like it, thank you!

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: I agree and that way the VDA will continue to work in the meantime!

drivers/SmartThings/matter-window-covering/src/init.lua
SathvikGatti: @nickolas-deboom sometimes there's an issue seen with window shade status observed with VDA window covering having tilt feature and probably could be seen with real devices.
Currently VDA window covering reports tilt and lift values once the movement is complete. I'll try to explain with an example, pardon me for the lengthy  explanation :)

Assuming lift and tilt are set to 100, on invoking "Close", tilt and lift move towards 0. Depending on what attribute's report is received (tilt/lift) different window shade status can be emitted twice in a short span of time. 
Driver receives tilt's report set to 0 first:
windowShadeTiltLevel="0"
windowShade="\open\" 

 The window shade status set to  "Open" is updated as current tilt "0" is compared with previous lift "100". This is because the latest value of lift is not received yet.

Then on receiving lift's report(0) from device "closed" is emitted
**shadeLevel="0"**
windowShade="\closed\" 

Since there were two window shade statuses emitted within a short span (arround 200ms) sometimes capability events displayed in ST-App's event history are not in order and the first sent status I.e "Open" is shown.

CC: @HunsupJung

drivers/SmartThings/matter-window-covering/src/init.lua
SathvikGatti: @nickolas-deboom sometimes there's an issue seen with window shade status observed with VDA window covering having tilt feature and probably could be seen with real devices.
Currently VDA window covering reports tilt and lift values once the movement is complete. I'll try to explain with an example, pardon me for the lengthy explanation :)

Assuming lift and tilt are set to 100, on invoking "Close", tilt and lift move towards 0. Depending on what attribute's report is received (tilt/lift) different window shade status can be emitted twice in a short span of time.
Driver receives tilt's report set to 0 first:
windowShadeTiltLevel="0"
windowShade="\open"

The window shade status set to "Open" is updated as current tilt "0" is compared with previous lift "100". This is because the latest value of lift is not received yet.

Then on receiving lift's report(0) from device "closed" is emitted
shadeLevel="0"
windowShade="\closed"

Since there were two window shade statuses emitted within a short span (arround 200ms) sometimes capability events displayed in ST-App's event history are not in order and the first sent status I.e "Open" is shown.

CC: @HunsupJung

hcarter-775: Thanks @SathvikGatti, I believe this is a known issue with the VDA, which is why we introduced a sub-driver in [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1527) to handle this case. I think the discussion now that @HunsupJung opened was whether we should modify the functionality of the VDA window covering device so that it does not do this, since it no real devices are handled this way.

hcarter-775: @SathvikGatti so if I'm understanding correctly, there are two orders that these events can be sent in. 

The first:
1. tilt received for 0 - leaving it as "open"
2. lift received for 0 - correctly showing it as "closed"

The second:
1. lift received for 0 - showing it as "partially open"
2. tilt received for 0 - correctly showing it as "closed"

Both of these paths leave the device in the "closed" state in the end, so I am confused what case you are describing.

hcarter-775 (approved):
one discussion is still open, but I think this is good to go since it is a VDA issue we are discussing there.

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: I think they meant that there is no guaranteed order of the capabilities being processed and so the final state of the capability in the app could appear as either open or closed (for scenario 1) after both capabilities are emitted. This is a good point and I think we may need to make an update with this consideration.

SathvikGatti: yes, even though the right window shade status is reported at the end, sometimes the capability events are not processed in the right order and the final status shown is not correct.

nickolas-deboom: The most recent commit contains a change to wait until tilt and lift are both received before updating the capability. This way, we don't need to worry about this race condition. Also, I separated the handling for devices that support lift and tilt, only lift, and only tilt, to make the logic easier to follow. Test case updates are still needed.

nickolas-deboom:
After testing with the VDA, which now supports tilt and also works with the main driver now, I was not able to reproduce the issue where capabilities are processed out of order on the cloud side. This is possible, but probably very unlikely with a real window covering device, since they wouldn't be likely to have lift and tilt finish updating at exactly the same time like the VDA does. Therefore, I would like to move forward with this PR and re-address the potential side-affect noted in this PR if we see this occur with a real device.

nickolas-deboom:
Also, the subdriver can now be removed by a separate PR, since the VDA was updated to work with the main driver.

PR #1818, opened 2024-12-11
Chore/fixup unit tests for 0.55 api v12 release

cjswedes:
This fixes unit tests to reflect new native handlers released in 0.55 FW/api v12 lua libs (Note we still need to enable colorTemperature and colorControl cmd registration in the matter-switch driver).

There was also an integration test framework bug that was fixed in api v12, that caused some tests to fail that should have never been passing (related to native handler registration for parent child/ mcd device)

One other test started failing with the integration test framework bugfix that needs analysis by someone more familiar with the matter profile switching

### DIFF
--- a/drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
@@ -345,7 +345,9 @@ end
 
 local function test_init_water_valve()
   test.mock_device.add_test_device(mock_device_water_valve)
+  test.socket.device_lifecycle:__queue_receive({ mock_device_water_valve.id, "doConfigure" })
   mock_device_water_valve:expect_metadata_update({ profile = "water-valve-level" })
+  mock_device_water_valve:expect_metadata_update({ provisioning_state = "PROVISIONED" })
 end
 
 local function test_init_parent_child_different_types()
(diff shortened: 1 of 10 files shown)


drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
```diff
@@ -425,6 +426,8 @@ test.register_coroutine_test(
 test.register_coroutine_test(
     "Test profile change on init for water valve parent cluster as server",
     function()
+        test.socket.device_lifecycle:__queue_receive({ mock_device_water_valve.id, "doConfigure" })
+        mock_device_water_valve:expect_metadata_update({ provisioning_state = "PROVISIONED" })
```
cjswedes, line 430: It seems like the test was written to have a profile switch occur during the `device_init`/`device_added` event, but that was not happening, and the profile switch only occurs during doConfigure. @hcarter-775 what is the correct behavior?

hcarter-775: correct behavior is that this profiling should be happening during doConfigure.

hcarter-775: looks like this was a copy-paste error since the 1.3 update added some new logic to this driver

PR #1822, opened 2024-12-12
Add Aqara Light Switch H2

DongHoon-Ryu:
This is a new Matter Device with four physical switches.
 - Two of each of the four switchs have On/Off Light(0x0100) and Generic Switch(0x000F) device types.
 - The other two switches have a Generic Switch(0x000F) device type.
 - The root node has Electrical Sensor(0x0510) utility device type.

Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [X] New feature
- [ ] Refactor

# Checklist

# Description of Change
Add new device profile and sub-driver

# Summary of Completed Tests
- Successful test for Plugin UI update and Dashboard's Device Card status update according to the switch operations


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1132,8 +1154,18 @@ local function device_added(driver, device)
     device:emit_event(capabilities.energyMeter.energy({ value = 0.0, unit = "Wh" }))
   end
 
-  -- call device init in case init is not called after added due to device caching
-  device_init(driver, device)
+  -- When device_init is performed in the aqara switch that has EDGE_CHILE,
+  -- a problem related to capability mapping occurs during configuration_buttons(), so this conditions are added.
+  if device.manufacturer_info.vendor_id ~= AQARA_MANUFACTURER_ID then
```
nickolas-deboom, line 1159: Could this be simplified into the following?
`if not (device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID and 
    (device.manufacturer_info.product_id == 0x1008 or device.manufacturer_info.product_id == 0x1009)) then
    -- call device init in case init is not called after added due to device caching
      device_init(driver, device)
 end`

nickolas-deboom, line 1159: Could this be simplified into the following?
```suggestion
  if not (device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID and 
    (device.manufacturer_info.product_id == 0x1008 or device.manufacturer_info.product_id == 0x1009)) then
    -- call device init in case init is not called after added due to device caching
      device_init(driver, device)
 end
```

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1132,8 +1154,18 @@ local function device_added(driver, device)
...
+    -- call device init in case init is not called after added due to device caching
+    device_init(driver, device)
+  else
+    if device.manufacturer_info.product_id ~= 0x1008 and
+       device.manufacturer_info.product_id ~= 0x1009 then
+      -- call device init in case init is not called after added due to device caching
+      device_init(driver, device)
+    end
+  end
```
nickolas-deboom, line 1168: Could this be simplified into the following?
```suggestion
  if not (device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID and 
    (device.manufacturer_info.product_id == 0x1008 or device.manufacturer_info.product_id == 0x1009)) then
    -- call device init in case init is not called after added due to device caching
      device_init(driver, device)
 end
```

nickolas-deboom:
Hi @DongHoon-Ryu these changes look good to me with one comments that I left. I was also wondering if you could create a few test cases to show that this device would join the expected profile and the child devices are created as expected?

nickolas-deboom:
Also, there appears to be a luacheck issue as well as an issue with an existing test case

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: Although it is an Aqara device, there may be cases where device_init() is required, so it was written as above.

DongHoon-Ryu:
> Hi @DongHoon-Ryu these changes look good to me with one comment that I left. I was also wondering if you could create a few test cases to show that this device would join the expected profile and the child devices are created as expected?

DongHoon-Ryu:
> Hi @DongHoon-Ryu these changes look good to me with one comment that I left. I was also wondering if you could create a few test cases to show that this device would join the expected profile and the child devices are created as expected?

I will add test cases by tomorrow.

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: I found the root cause of the problem and modified it so that device_init (driver, device) is always called like the code in main.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -405,15 +417,23 @@ local function find_default_endpoint(device)
...
   -- check if device has an overridden child profile that differs from the profile
   -- that would match the child's device type
   for _, fingerprint in ipairs(child_device_profile_overrides) do
     if device.manufacturer_info.vendor_id == fingerprint.vendor_id and
        device.manufacturer_info.product_id == fingerprint.product_id then
-      return fingerprint.child_profile
+      if device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID then
+        if ep_sequence == 1 then
+          -- To add Electrical Sensor only to the first EDGE_CHILD
+          device:set_field(IS_AQARA_SWITCH_DEVICE, true)
```
hcarter-775, line 431:
```suggestion
          device:set_field(IS_AQARA_SWITCH_DEVICE, true, {persist = true})
```
This persist flag is needed for a note like this

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -405,15 +417,23 @@ local function find_default_endpoint(device)
...
+          return fingerprint.child_profile
+        end
+      else
+        return fingerprint.child_profile
+      end
```
hcarter-775, line 436:
```suggestion
        end
        return fingerprint.child_profile
```
There's no reason to overcomplicate this with multiple returns. There just needs to be extra logic before the return for the aqara devices.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1046,7 +1074,11 @@ end
 local function active_power_handler(driver, device, ib, response)
   if ib.data.value then
     local watt_value = ib.data.value / CONVERSION_CONST_MILLIWATT_TO_WATT
-    device:emit_event(capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
+    if device:get_field(IS_AQARA_SWITCH_DEVICE) then
+      device:emit_event_for_endpoint(1, capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
+    else
```
hcarter-775, line 1079: Remove this- is unnecessary

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -983,7 +1007,11 @@ local function cumul_energy_exported_handler(driver, device, ib, response)
   if ib.data.elements.energy then
     local watt_hour_value = ib.data.elements.energy.value / CONVERSION_CONST_MILLIWATT_TO_WATT
     device:set_field(TOTAL_EXPORTED_ENERGY, watt_hour_value)
-    device:emit_event(capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
+    if device:get_field(IS_AQARA_SWITCH_DEVICE) then
+      device:emit_event_for_endpoint(1, capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
+    else
```
hcarter-775, line 1012: Remove this- is unnecessary

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -680,7 +702,9 @@ local function device_init(driver, device)
     end
     local main_endpoint = find_default_endpoint(device)
     for _, ep in ipairs(device.endpoints) do
-      if ep.endpoint_id ~= main_endpoint and ep.endpoint_id ~= 0 then
+      if ep.endpoint_id ~= main_endpoint and
+         (device:get_field(IS_AQARA_SWITCH_DEVICE) or ep.endpoint_id ~= 0) then
+        -- insert energy management into InteractionRequest list when IS_AQARA_SWITCH_DEVICE
```
hcarter-775, line 707: What attribute are we expecting to track for the Aqara switch at its 0 endpoint?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1174,6 +1209,7 @@ local matter_driver_template = {
         [clusters.ElectricalPowerMeasurement.attributes.ActivePower.ID] = active_power_handler,
       },
       [clusters.ElectricalEnergyMeasurement.ID] = {
+        [clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyImported.ID] = energy_report_handler_factory(true),
```
hcarter-775, line 1212: sorry for the confusion- see [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1830) which will change the system to only use imported and not exported energy. This fixup be merged before this PR.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1132,8 +1164,11 @@ local function device_added(driver, device)
     device:emit_event(capabilities.energyMeter.energy({ value = 0.0, unit = "Wh" }))
   end
 
-  -- call device init in case init is not called after added due to device caching
-  device_init(driver, device)
+  -- when unit testing, call set_configure elsewhere
+  if not device:get_field(TEST_CONFIGURE) then
+    -- call device init in case init is not called after added due to device caching
```
hcarter-775, line 1169: I don't believe TEST_CONFIGURE should be necessary in this case. Please see [this test file](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/57b46bff0a96a52f868bf6c7d84daffc66a7ea36/drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua) for an example how to test without this extra logic inserted.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -454,7 +474,7 @@ local function do_configure(driver, device)
     end
   end
 
-  if profile_name then
+  if not device:get_field(IS_AQARA_SWITCH_DEVICE) and profile_name then
```
hcarter-775, line 477: I think I'd prefer leaving the configure logic as-is, and add a more general SWITCH_DEVICE_PROFILED with an early return to the top of the function. This SWITCH_DEVICE_PROFILED field can be set for all devices using initialize_switch, not just aqara devices.
 
```suggestion
local function do_configure(driver, device)
  if device:get_field(SWITCH_DEVICE_PROFILED) then
     return
  end
 local energy_eps = embedded_cluster_utils.get_endpoints(device, clusters.ElectricalEnergyMeasurement.ID)
  ...
  if profile_name then
```

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: Yes, I resolved it.

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: If two Light On/Off devices are added as EDGE_CHILD, only the first device is defined as `fingerprint.child_profile` (`light-power-energy-powerConsumption` profile), and the profile of the second device is to select the `light-binary` profile by the logic below.
Therefore, we need to keep the existing logic as it is, so I added comments.

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: Since the device card is implemented as below, the event related to the electrical sensor comes to the main profile containing ep4, so it is added to send the event to the first endpoint containing the actual `powerMeter` or `energyMeter` capability. If we do not do this, a crash occurs in the driver.
===========================
endpoint       device type
----------------------------------------------
ep0        Root(0x0016), Electrical sensor(0x0510)
ep1        On/Off Light(0x0100)            <-- EDGE_CHILD(light-power-energy-powerConsumption profile)
ep2        On/Off Light(0x0100)            <-- EDGE_CHILD(light-binary profile)
ep4        Generic Switch(0x000F)         <-- main profile(4-button profile)
ep5        Generic Switch(0x000F)
ep6        Generic Switch(0x000F)
ep7        Generic Switch(0x000F)

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: Since the device card is implemented as below, the event related to the electrical sensor comes to the main profile containing ep4, so it is added to send the event to the first endpoint containing the actual powerMeter or energyMeter capability. If we do not do this, a crash occurs in the driver.

endpoint device type

ep0 : Root(0x0016), Electrical sensor(0x0510)
ep1 : On/Off Light(0x0100) <-- EDGE_CHILD(light-power-energy-powerConsumption profile)
ep2 : On/Off Light(0x0100) <-- EDGE_CHILD(light-binary profile)
ep4 : Generic Switch(0x000F) <-- main profile(4-button profile)
ep5 : Generic Switch(0x000F)
ep6 : Generic Switch(0x000F)
ep7 : Generic Switch(0x000F)

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: This is to add a list related to energy management according to the expected subscribe_list used in the unit test.

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: Yes, that's a good idea. However, when `energy_eps` or `power_eps` have a value greater than `0`, there is always a switch endpoint, so `BUTTON_DEVICE_PROFILED` was used instead of `SWITCH_DEVICE_PROFILED`.

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: Thanks, I resolved it.

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: After that PR is merged, I will merge this PR.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: sorry, what I meant here is that we don't need to do and early return. both paths end in the call `fingerprint.child_profile`, so we don't need and else statement like that. Both paths return the same thing at the end.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: that PR I mentioned has been merged- please merge these newest changes when you have a chance!

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Sorry, I don't see any reference to BUTTON_DEVICE_PROFILED in the PR. Have you made this update?

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: @nickolas-deboom it seems like [this dicussions](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1843#discussion_r1905789472) would apply over here as well

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Resolving the other suggestion, but continuing the general discussion here. I believe, in that case, we should generalize this and save the endpoint pointing to  a `power-energy` profile in a separate field. This can then be referenced in place of 1, and be used for all cases.

nickolas-deboom:
Pending resolutions to the other comments that Harrison left, these changes look good to me!

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: Yes, I did it at the commit below.
`1d680a035ac0c89035afc21fa6ec4fec8b896da2`

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: @hcarter-775 I'm sorry. If there is anything I still misunderstand, please tell me again.
This is because if `ep_sequence` is `2` or more, `fingerprint.child_profile` should not be returned.
In other words, in the second `EDGE_CHILD` device with `ep_sequence` of `2`, the `light-binary` profile must be specified, so it is implemented as above. Is this part wrong?

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: @nickolas-deboom Yes, but I used endpoint 4 as default endpoint. Because `EDGE_CHILD` can specify only one endpoint by default. For reference, I attached some of the ReportData I received from the device.

```shell
2024-12-13_07:34:30.735 | D | matter.sdk | [DMG] 			AttributeDataIB =
2024-12-13_07:34:30.735 | D | matter.sdk | [DMG] 			{
2024-12-13_07:34:30.736 | D | matter.sdk | [DMG] 				DataVersion = 0xcd776a14,
2024-12-13_07:34:30.737 | D | matter.sdk | [DMG] 				AttributePathIB =
2024-12-13_07:34:30.737 | D | matter.sdk | [DMG] 				{
2024-12-13_07:34:30.738 | D | matter.sdk | [DMG] 					Endpoint = 0x0,
2024-12-13_07:34:30.738 | D | matter.sdk | [DMG] 					Cluster = 0x1d,
2024-12-13_07:34:30.738 | D | matter.sdk | [DMG] 					Attribute = 0x0000_0000,
2024-12-13_07:34:30.738 | D | matter.sdk | [DMG] 				}
2024-12-13_07:34:30.738 | D | matter.sdk | [DMG] 					
2024-12-13_07:34:30.738 | D | matter.sdk | [DMG] 				Data = [
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 						
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 					{
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 						0x0 = 1296 (unsigned), 
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 						0x1 = 1 (unsigned), 
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 					},
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 					{
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 						0x0 = 18 (unsigned), 
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 						0x1 = 1 (unsigned), 
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 					},
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 					{
2024-12-13_07:34:30.739 | D | matter.sdk | [DMG] 						0x0 = 22 (unsigned), 
2024-12-13_07:34:30.740 | D | matter.sdk | [DMG] 						0x1 = 3 (unsigned), 
2024-12-13_07:34:30.740 | D | matter.sdk | [DMG] 					},
2024-12-13_07:34:30.742 | D | matter.sdk | [DMG] 				],
2024-12-13_07:34:30.743 | D | matter.sdk | [DMG] 			},
2024-12-13_07:34:30.743 | D | matter.sdk | [DMG] 			
2024-12-13_07:34:30.743 | D | matter.sdk | [DMG] 		},
2024-12-13_07:34:30.744 | D | matter.sdk | [DMG] 		
2024-12-13_07:34:30.744 | D | matter.sdk | [DMG] 		AttributeReportIB =
2024-12-13_07:34:30.744 | D | matter.sdk | [DMG] 		{
2024-12-13_07:34:30.745 | D | matter.sdk | [DMG] 			AttributeDataIB =
2024-12-13_07:34:30.745 | D | matter.sdk | [DMG] 			{
2024-12-13_07:34:30.746 | D | matter.sdk | [DMG] 				DataVersion = 0xad1ac45d,
2024-12-13_07:34:30.746 | D | matter.sdk | [DMG] 				AttributePathIB =
2024-12-13_07:34:30.746 | D | matter.sdk | [DMG] 				{
2024-12-13_07:34:30.746 | D | matter.sdk | [DMG] 					Endpoint = 0x1,
2024-12-13_07:34:30.747 | D | matter.sdk | [DMG] 					Cluster = 0x1d,
2024-12-13_07:34:30.747 | D | matter.sdk | [DMG] 					Attribute = 0x0000_0000,
2024-12-13_07:34:30.747 | D | matter.sdk | [DMG] 				}
2024-12-13_07:34:30.748 | D | matter.sdk | [DMG] 					
2024-12-13_07:34:30.748 | D | matter.sdk | [DMG] 				Data = [
2024-12-13_07:34:30.748 | D | matter.sdk | [DMG] 						
2024-12-13_07:34:30.748 | D | matter.sdk | [DMG] 					{
2024-12-13_07:34:30.749 | D | matter.sdk | [DMG] 						0x0 = 256 (unsigned), 
2024-12-13_07:34:30.749 | D | matter.sdk | [DMG] 						0x1 = 3 (unsigned), 
2024-12-13_07:34:30.749 | D | matter.sdk | [DMG] 					},
2024-12-13_07:34:30.750 | D | matter.sdk | [DMG] 				],
2024-12-13_07:34:30.750 | D | matter.sdk | [DMG] 			},
2024-12-13_07:34:30.751 | D | matter.sdk | [DMG] 			
2024-12-13_07:34:30.751 | D | matter.sdk | [DMG] 		},
```

nickolas-deboom:
One thing I forgot to mention - would it be better to have everything in the same profile? It's been brought up before that for these light devices that also contain buttons that the best user experience is to have the light(s) be considered the 'main' device rather than the buttons. We did something similar for a GE/Cync device that contained multiple light endpoints and multiple button endpoints (see [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1663)).

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -405,15 +417,24 @@ local function find_default_endpoint(device)
...
+      if device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID then
+        if ep_sequence == 1 then
+          -- To add Electrical Sensor only to the first EDGE_CHILD(light-power-energy-powerConsumption)
+          -- The profile of the second EDGE_CHILD is determined in the "for" loop below (e.g., light-binary)
+          device:set_field(IS_AQARA_SWITCH_DEVICE, true, {persist = true})
+          return fingerprint.child_profile
+        end
+      else
+        return fingerprint.child_profile
+      end
```
hcarter-775, line 437:
```suggestion
      if device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID then
        if ep_sequence == 1 then
          -- To add Electrical Sensor only to the first EDGE_CHILD(light-power-energy-powerConsumption)
          -- The profile of the second EDGE_CHILD is determined in the "for" loop below (e.g., light-binary)
          device:set_field(IS_AQARA_SWITCH_DEVICE, true, {persist = true})
        end
        return fingerprint.child_profile
      end
```
sorry, what I've been trying to say is that this edited implementation has the same handling, but is simpler. There's no reason to branch off and do an else statement.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: just noting, since there's been a little bit of update on this PR, that this still needs to be done :smile:

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: please see [this message](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1822/files#r1917200411)

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: yes I see, sorry about that, the commit was not showing up on my UI without toggling something.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -680,7 +708,9 @@ local function device_init(driver, device)
     end
     local main_endpoint = find_default_endpoint(device)
     for _, ep in ipairs(device.endpoints) do
-      if ep.endpoint_id ~= main_endpoint and ep.endpoint_id ~= 0 then
+      if ep.endpoint_id ~= main_endpoint and
+         (device:get_field(IS_AQARA_SWITCH_DEVICE) or ep.endpoint_id ~= 0) then
```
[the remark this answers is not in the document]
nickolas-deboom: Also, an entry for the Electrical Sensor device type with the attributes to subscribe to for this device type should be added to `device_type_attribute_map`.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -405,15 +417,24 @@ local function find_default_endpoint(device)
...
   local profile
 
   -- check if device has an overridden child profile that differs from the profile
   -- that would match the child's device type
   for _, fingerprint in ipairs(child_device_profile_overrides) do
     if device.manufacturer_info.vendor_id == fingerprint.vendor_id and
        device.manufacturer_info.product_id == fingerprint.product_id then
-      return fingerprint.child_profile
+      if device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID then
+        if ep_sequence == 1 then
```
hcarter-775, line 429: we do not need ep_sequence here- we can just use `child_ep.endpoint_id` to achieve the same thing.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: After thinking about this more and talking to @nickolas-deboom about it, I have the following suggestions.
First, in the attribute handlers, I think they can be handled more generally like so:
```
if ib.endpoint_id ~= 0 then
      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
else
      device:emit_event_for_endpoint(1, capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
end
```
This removes the need for a specific  `IS_AQARA_SWITCH_DEVICE` check. I am still unsure how to catch this for the completely general case, but I think this is an improvement at least.

hcarter-775: This and the other comment should remove all need for the `IS_AQARA_SWITCH_DEVICE` flag, so all references to this and the logic used to set it should be removed

drivers/SmartThings/matter-switch/src/init.lua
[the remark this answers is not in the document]
DongHoon-Ryu: @nickolas-deboom

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: That's a great idea. I applied it.

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: I removed `IS_AQARA_SWITCH_DEVICE`. And the first switch endpoint is saved in `ENERGY_MANAGEMENT_ENDPOINT` and modified as follows to use this value as an enpoint.(`device:get_field(ENERGY_MANAGEMENT_ENDPOINT)`)

```lua
local function cumul_energy_exported_handler(driver, device, ib, response)
  if ib.data.elements.energy then
    local watt_hour_value = ib.data.elements.energy.value / CONVERSION_CONST_MILLIWATT_TO_WATT
    device:set_field(TOTAL_EXPORTED_ENERGY, watt_hour_value)
    if ib.endpoint_id ~= 0 then
      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
    else
      -- when energy management is defined in the root endpoint(0), replace it with the first switch endpoint and process it.
      -- In other words, define the capability related to energy management in the first switch endpoint and process it.
      -- ENERGY_MANAGEMENT_ENDPOINT = first switch endpoint
      device:emit_event_for_endpoint(device:get_field(ENERGY_MANAGEMENT_ENDPOINT), capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
    end
  end
end
```

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: First, I would like to modify it as below. Because we have to consider the case below as well.
- if `device.manufacturer_info.vendor_id` is not `AQARA_MANUFACTURER_ID`, `fingerprint.child_profile` should be returned.
- If `ep_sequence == 2`, `fingerprint.child_profile` should not be returned. That is, if ‘ep_sequence == 2’, the ‘light-binary’ profile must be selected according to the for loop routine below.

```lua
    if device.manufacturer_info.vendor_id == fingerprint.vendor_id and
       device.manufacturer_info.product_id == fingerprint.product_id then
      if device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID then
        if child_ep ~= 1 then
          -- To add Electrical Sensor only to the first EDGE_CHILD(light-power-energy-powerConsumption)
          -- The profile of the second EDGE_CHILD is determined in the "for" loop below (e.g., light-binary)
          break
        end
      end
      return fingerprint.child_profile
    end

**Comment by DongHoon-Ryu** [comment]
> One thing I forgot to mention - would it be better to have everything in the same profile? It's been brought up before that for these light devices that also contain buttons that the best user experience is to have the light(s) be considered the 'main' device rather than the buttons. We did something similar for a GE/Cync device that contained multiple light endpoints and multiple button endpoints (see [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1663)).

When I asked some of my colleagues, there were many opinions that the UI I implemented now is good. However, when there is power management at the root endpoint as it is now, it would be better to consider implementing it as a plugin because there are many difficulties while handling related capabilities on the EDGE_CHILD device rather than the main endpoint. I will update the results again tomorrow regarding this.

File: drivers/SmartThings/matter-switch/src/init.lua [hunk 14]
```diff
@@ -983,7 +1010,14 @@ local function cumul_energy_imported_handler(driver, device, ib, response)
   if ib.data.elements.energy then
     local watt_hour_value = ib.data.elements.energy.value / CONVERSION_CONST_MILLIWATT_TO_WATT
     device:set_field(TOTAL_IMPORTED_ENERGY, watt_hour_value)
-    device:emit_event(capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
+    if ib.endpoint_id ~= 0 then
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
+    else
+      -- when energy management is defined in the root endpoint(0), replace it with the first switch endpoint and process it.
+      -- In other words, define the capability related to energy management in the first switch endpoint and process it.
+      -- ENERGY_MANAGEMENT_ENDPOINT = first switch endpoint
```
hcarter-775, line 1018: great implementation :+1: would you mind removing this extra comment though? I think we only need the first line. Thank you

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1046,7 +1080,14 @@ end
 local function active_power_handler(driver, device, ib, response)
   if ib.data.value then
     local watt_value = ib.data.value / CONVERSION_CONST_MILLIWATT_TO_WATT
-    device:emit_event(capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
+    if ib.endpoint_id ~= 0 then
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
+    else
+      -- when energy management is defined in the root endpoint(0), replace it with the first switch endpoint and process it.
+      -- In other words, define the capability related to energy management in the first switch endpoint and process it.
+      -- ENERGY_MANAGEMENT_ENDPOINT = first switch endpoint
```
hcarter-775, line 1088: same here !

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: I deleted it.

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: I deleted it.

DongHoon-Ryu:
> > One thing I forgot to mention - would it be better to have everything in the same profile? It's been brought up before that for these light devices that also contain buttons that the best user experience is to have the light(s) be considered the 'main' device rather than the buttons. We did something similar for a GE/Cync device that contained multiple light endpoints and multiple button endpoints (see [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1663)).
> 
> When I asked some of my colleagues, there were many opinions that the UI I implemented now is good. However, when there is power management at the root endpoint as it is now, it would be better to consider implementing it as a plugin because there are many difficulties while handling related capabilities on the EDGE_CHILD device rather than the main endpoint. I will update the results again tomorrow regarding this.

@nickolas-deboom  I have implemented `ep1` as MCD as `main_endpoint` like GE / Sync device, but the same problem still exists when energy management is in ep0. First of all, it would be better to merge with main branch with the current implementation. Please review it.

hcarter-775:
This looks very good, and I am almost ready to approve it. Last thing though, the matter-switch driver is a very important driver, so would you mind onboarding a couple devices for some regression testing to make sure the flow hasn't been changed? This would need to include 1 or 2 button-switch devices, and 1 or 2 multi-switch devices.

nickolas-deboom:
Agreed, the changes you made look great!

nickolas-deboom:
> > > One thing I forgot to mention - would it be better to have everything in the same profile? It's been brought up before that for these light devices that also contain buttons that the best user experience is to have the light(s) be considered the 'main' device rather than the buttons. We did something similar for a GE/Cync device that contained multiple light endpoints and multiple button endpoints (see [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1663)).
> > 
> > 
> > When I asked some of my colleagues, there were many opinions that the UI I implemented now is good. However, when there is power management at the root endpoint as it is now, it would be better to consider implementing it as a plugin because there are many difficulties while handling related capabilities on the EDGE_CHILD device rather than the main endpoint. I will update the results again tomorrow regarding this.
> 
> @nickolas-deboom I have implemented `ep1` as MCD as `main_endpoint` like GE / Sync device, but the same problem still exists when energy management is in ep0. First of all, it would be better to merge with main branch with the current implementation. Please review it.

Just to be clear, is this how the device will be onboarded?
Device card 1:
On/Off Light
Electrical Sensor
Buttons

Device card 2:
On/Off Light

If so, then you would need to add the button components to the profile you created. Or am I misunderstanding?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -413,6 +424,13 @@ local function assign_child_profile(device, child_ep)
   for _, fingerprint in ipairs(child_device_profile_overrides) do
     if device.manufacturer_info.vendor_id == fingerprint.vendor_id and
        device.manufacturer_info.product_id == fingerprint.product_id then
+      if device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID then
+        if child_ep ~= 1 then
+          -- To add Electrical Sensor only to the first EDGE_CHILD(light-power-energy-powerConsumption)
+          -- The profile of the second EDGE_CHILD is determined in the "for" loop below (e.g., light-binary)
+          break
+        end
+      end
```
nickolas-deboom, line 433: This won't be needed after making the changes I suggested to `is_supported_combination_button_switch_device_type`, since endpoint 1 would be considered the main endpoint now.
```suggestion
```

hcarter-775 (approved):

nickolas-deboom (approved):

PR #1827, opened 2024-12-16
Matter Switch: Improve color temp boundary logic

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

[MTR-849](https://smartthings.atlassian.net/browse/MTR-849)

This change improves edge condition logic when a device sends a color temperature value at or beyond its limits in mireds and also when a capability command is received that is at or beyond the limits in Kelvin.

If a device reports a value at/beyond the limits determined by `ColorTempPhysicalMaxMireds`/`ColorTempPhysicalMinMireds`, the converted Kelvin value would exceed the rounded min/max kelvin values calculated by the driver. The current behavior results in the `colorTemperature` capability being emitted with a value beyond the limits allowed by the plugin, and causes an "Out of range" error in the plugin. Instead, the kelvin value emitted in the capability should be set to the corresponding min/max kelvin value, rather than the converted value.

In the other direction, if the capability is set to the min/max kelvin value in the ST app, we should send the corresponding min/max mireds value to the device rather than doing a conversion.

# Summary of Completed Tests

See new unit tests.

[MTR-849]: https://smartthings.atlassian.net/browse/MTR-849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -48,7 +48,8 @@ local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 -- rather than COMPONENT_TO_ENDPOINT_MAP.
 local COMPONENT_TO_ENDPOINT_MAP_BUTTON = "__component_to_endpoint_map_button"
 local IS_PARENT_CHILD_DEVICE = "__is_parent_child_device"
-local COLOR_TEMP_BOUND_RECEIVED = "__colorTemp_bound_received"
+local COLOR_TEMP_BOUND_RECEIVED_KELVIN = "__colorTemp_bound_received_kelvin"
+local COLOR_TEMP_BOUND_RECEIVED_MIRED = "__colorTemp_bound_received_mired"
 local COLOR_TEMP_MIN = "__color_temp_min"
 local COLOR_TEMP_MAX = "__color_temp_max"
 local LEVEL_BOUND_RECEIVED = "__level_bound_received"
@@ -780,7 +781,16 @@ end
 
 local function handle_set_color_temperature(driver, device, cmd)
   local endpoint_id = device:component_to_endpoint(cmd.component)
-  local temp_in_mired = utils.round(MIRED_KELVIN_CONVERSION_CONSTANT/cmd.args.temperature)
+  local temp_in_kelvin = cmd.args.temperature
+  local min_temp_kelvin = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED_KELVIN..COLOR_TEMP_MIN, endpoint_id)
+  local max_temp_kelvin = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED_KELVIN..COLOR_TEMP_MAX, endpoint_id)
+
+  local temp_in_mired = utils.round(MIRED_KELVIN_CONVERSION_CONSTANT/temp_in_kelvin)
+  if min_temp_kelvin ~= nil and temp_in_kelvin <= min_temp_kelvin then
+    temp_in_mired = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED_MIRED..COLOR_TEMP_MAX, endpoint_id)
+  elseif max_temp_kelvin ~= nil and temp_in_kelvin >= max_temp_kelvin then
+    temp_in_mired = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED_MIRED..COLOR_TEMP_MIN, endpoint_id)
+  end
   local req = clusters.ColorControl.server.commands.MoveToColorTemperature(device, endpoint_id, temp_in_mired, TRANSITION_TIME, OPTIONS_MASK, OPTIONS_OVERRIDE)
   device:set_field(MOST_RECENT_TEMP, cmd.args.temperature)
   device:send(req)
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -780,7 +781,16 @@ end
 
 local function handle_set_color_temperature(driver, device, cmd)
   local endpoint_id = device:component_to_endpoint(cmd.component)
-  local temp_in_mired = utils.round(MIRED_KELVIN_CONVERSION_CONSTANT/cmd.args.temperature)
+  local temp_in_kelvin = cmd.args.temperature
```
ctowns, line 784: Can you help my understanding here - shouldn't our boundaries set for the colorTemperature capability prevent us from setting a colorTemp value outside the device's known range? It seems like if we know the colorTemp range of the device, we should be able to prevent this issue via dynamic constraints, unless I am missing something here.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -852,38 +862,56 @@ local function sat_attr_handler(driver, device, ib, response)
...
-    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.colorTemperature.colorTemperature(temp))
+  local temp_in_mired = ib.data.value
+  if temp_in_mired == nil then
+    return
   end
+  if (temp_in_mired < COLOR_TEMPERATURE_MIRED_MIN or temp_in_mired > COLOR_TEMPERATURE_MIRED_MAX) then
+    device.log.warn_with({hub_logs = true}, string.format("Device reported color temperature %d mired outside of sane range of %.2f-%.2f", temp_in_mired, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX))
+    return
+  end
+  local min_temp_mired = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED_MIRED..COLOR_TEMP_MIN, ib.endpoint_id)
```
ctowns, line 873: Are we storing min and max for both kelvin and mired? It seems to me that we should only need to store one and that can be our source of truth. I don't know if this was introduced in this PR or if it already existed, but I think we'd realistically only want to store one of these values and can do our conversions in the driver as needed unless there's a good reason to store both.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -852,38 +862,56 @@ local function sat_attr_handler(driver, device, ib, response)
...
-    if (ib.data.value < COLOR_TEMPERATURE_MIRED_MIN or ib.data.value > COLOR_TEMPERATURE_MIRED_MAX) then
-      device.log.warn_with({hub_logs = true}, string.format("Device reported a color temperature %d mired outside of sane range of %.2f-%.2f", ib.data.value, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX))
+    if (temp_in_mired < COLOR_TEMPERATURE_MIRED_MIN or temp_in_mired > COLOR_TEMPERATURE_MIRED_MAX) then
+      device.log.warn_with({hub_logs = true}, string.format("Device reported a color temperature %d mired outside of sane range of %.2f-%.2f", temp_in_mired, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX))
       return
     end
-    local temp_in_kelvin = mired_to_kelvin(ib.data.value, minOrMax)
-    set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp_in_kelvin)
+    local temp_in_kelvin = mired_to_kelvin(temp_in_mired, minOrMax)
+    set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp_in_kelvin, {persist = true})
```
ctowns, line 908: Why does this need to persist? Don't we subscribe to this value, so we would get it from the device each start up anyway?

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: The issue is that the dynamic constraints set the min and max values in the plugin using rounded kelvin values (to the nearest 100 I believe). Take the example from the ticket linked to this PR. The temp was set to 6500 in the ST app, which is the maximum of the allowed range. When we convert this to mireds we get 154 (154 mired = 6493 K). However, this max of 6500 was originally calculated from converting 153 mireds to kelvin (153 mireds = 6535 K). So, in order to be able to use the full range allowed by the device, we should send 153 to the device rather than 154 if we set the capability to 6500.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: The reason to store both is that it would allow us to know when we are dealing with a value at the minimum or maximum of the allowed range, either in terms of Kelvin when using the slider in the app, or in terms of mireds when dealing with values received from the device. Like I mentioned in the other comment, if the capability is set to the maximum for example in the plugin, then the driver would just convert this to the nearest value in mireds rather than setting the device to the maximum allowed value in mireds.

Going the other direction, if the device reported that its color temp is equal to the minimum mireds value (153 in the example in the ticket), the driver currently just converts this to kelvin and emits that value, without checking that it is within the allowed kelvin range. 153 mireds = 6535 kelvin, which exceeds the rounded value of 6500 that we calculated for the max kelvin capability. This results in the plugin saying "Out of range"

The changes in this PR would prevent this by just emitting the max kelvin value of 6500 if we receive the minimum allowed mireds value, and if we receive a capability command that is setting the temp to the max of 6500, then we would just send the min allowed mireds value of 153 to the device. (And the same kind of thing if we are dealing with the minimum instead.)

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Yeah it might not be necessary, I was thinking it's needed since I'm using these values in `handle_set_color_temperature`, but you're right that we'd get it on the next start up anyway

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -780,7 +781,16 @@ end
 
 local function handle_set_color_temperature(driver, device, cmd)
   local endpoint_id = device:component_to_endpoint(cmd.component)
-  local temp_in_mired = utils.round(MIRED_KELVIN_CONVERSION_CONSTANT/cmd.args.temperature)
+  local temp_in_kelvin = cmd.args.temperature
+  local min_temp_kelvin = get_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..COLOR_TEMP_MIN, endpoint_id)
```
ctowns, line 785: What happens if these `get_field_for_endpoint` calls are nil for some reason? For example, if we do not get a sane color temp range from the device, those values would not be set, but then we'd run into a similar issue with the devices having out of range issues because they wouldn't have the min/max range saved to handle the cases on the edge. So, in order to be complete, I think we'd also need to add a default value for the min/max in the cases that the device doesn't share a correct value... however, then we run up against the issue of some of the default values being different because they are defined in the profile that is set by the fingerprint, right? Overall, it seems like this solution will fix one particular use case, i.e. when the device provides a min/max value and we can save that to our defaults. However, we would still see this issue on the boundaries in the case that we don't receive the min/max from the device, which we know is a common issue on devices.

Does the device in question in MTR-849 report the min/max values correctly? We will want to test this on the device to confirm. If my understanding is correct, this will only fix the edge case on the min/max mired boundary if the device is correctly reporting it's min/max to begin with.

In addition, we should do some testing with a device that we know _doesn't_ share the correct min/max just so we can confirm the behavior described above.

ctowns (commented):
I left a comment, but I think this particular solution will work in the case that the device shares correct min/max data with us, but otherwise we'd still likely encounter the issue described in MTR-849 on devices that don't properly share the min/max.

Let's test with both the device from MTR-849 (I believe Tom mentioned that he has it) and also some of the bulbs that we know don't share the min/max values properly. It could be the case that we choose to ignore this boundary condition for those bulbs since we just have to make a guess at their ranges anyway, but I think it is worth discussing.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I mentioned this in my other comment, but if they are `nil` then there will be no change in behavior. The device from MTR-849 does report the correct min/max values. I don't have this device on hand but I can find out if we have one available.

Is there a reason we need to fix this issue for devices that don't report a proper min and max value? If they don't define that then I'm not sure how the driver is supposed to work correctly at the min/max values anyway.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I ended up removing `persist=true`.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -852,38 +862,56 @@ local function sat_attr_handler(driver, device, ib, response)
...
     end
-    if (ib.data.value < COLOR_TEMPERATURE_MIRED_MIN or ib.data.value > COLOR_TEMPERATURE_MIRED_MAX) then
-      device.log.warn_with({hub_logs = true}, string.format("Device reported a color temperature %d mired outside of sane range of %.2f-%.2f", ib.data.value, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX))
+    if (temp_in_mired < COLOR_TEMPERATURE_MIRED_MIN or temp_in_mired > COLOR_TEMPERATURE_MIRED_MAX) then
+      device.log.warn_with({hub_logs = true}, string.format("Device reported a color temperature %d mired outside of sane range of %.2f-%.2f", temp_in_mired, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX))
       return
     end
-    local temp_in_kelvin = mired_to_kelvin(ib.data.value, minOrMax)
+    local temp_in_kelvin = mired_to_kelvin(temp_in_mired, minOrMax)
     set_field_for_endpoint(device, COLOR_TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp_in_kelvin)
```
hcarter-775, line 908: kinda nit but not too much: for clarity, can we change `COLOR_TEMP_BOUND_RECEIVED` to `COLOR_TEMP_BOUND_RECEIVED_KELVIN`? This will make it clearer what it is with the `COLOR_TEMP_BOUND_RECEIVED_MIRED` value all over the place

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: If they are nil, they wouldn't be set up to receive any handling like this. The 'CT` feature indicates whether a `ColorControl` cluster uses these max/min settings, and so a device that doesn't support this wouldn't have specific handling at the edges, like Nick said.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Good idea, updated in latest commit!

hcarter-775 (approved):
:+1:

PR #1830, opened 2024-12-17
Change use of the EnergyExported attribute to EnergyImported in matter switch

hcarter-775:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [X] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
According to internal review as well as [MTR-873](https://smartthings.atlassian.net/browse/MTR-873?focusedCommentId=1542288&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1542288), the ElectricalEnergyMeasurement cluster's imported attributes should be used for positive energy values rather than the exported attributes. This change commits that statement to the drivers. 

# Summary of Completed Tests
All of the extensive unit tests have been converted from exported to imported tests, which properly handles all of the cases we're most interested in.

[MTR-873]: https://smartthings.atlassian.net/browse/MTR-873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


drivers/SmartThings/matter-switch/src/ElectricalEnergyMeasurement/server/attributes/CumulativeEnergyImported.lua
```diff
@@ -0,0 +1,68 @@
+local cluster_base = require "st.matter.cluster_base"
```
ctowns, line 1: Would we be able to remove the `CumulativeEnergyExported` and `PeriodicEnergyExported` embedded cluster defs now that we are just using the imported variation?

hcarter-775: We could. I figured why not leave them in, but it's just as well to take them out

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -183,44 +183,44 @@ local function iso8061Timestamp(time)
...
+    device:set_field(RECURRING_IMPORT_REPORT_POLL_TIMER, nil)
+    device:set_field(IMPORT_POLL_TIMER_SETTING_ATTEMPTED, nil)
   end
 end
 
-local function send_export_poll_report(device, latest_total_exported_energy_wh)
+local function send_import_poll_report(device, latest_total_imported_energy_wh)
   local current_time = os.time()
-  local last_time = device:get_field(LAST_EXPORTED_REPORT_TIMESTAMP) or 0
-  device:set_field(LAST_EXPORTED_REPORT_TIMESTAMP, current_time, { persist = true })
```
ctowns, line 198: This flag will have been persisted in the case on devices that have already joined to the platform and run this code. However, it wouldn't be needed anymore so we should set it to nil and remove it.

However, I don't think there are a lot/any devices like this yet, so in this case we are okay to just remove this, but just something to be aware of for the future.

hcarter-775: To note: I have not removed anything- only renamed the word export to import.

hcarter-775: Oh, nevermind I see what you mean. Thanks for pointing that out!

hcarter-775: update on this: from the superset query counting all matter profiles in use, I'm seeing:
` Matter Switch	plug-power-energy-powerConsumption	510 `, so there are 510 of these devices that would have run this code already. 

@ctowns Do you think we should add extra handling to transition these devices?

ctowns: Could you see what device these are? If it's the eve energy plug, then those devices would use the `eve-energy` sub driver, so then we technically wouldn't need to worry about clean up for those since they wouldn't have run this code, they would have just used the sub driver.

hcarter-775: @ctowns These wouldn't be for the eve-energy driver, since those devices still fingerprints to this `power-energy-powerConsumption`. You can see these in the query as well: `Matter Switch	power-energy-powerConsumption	9516`. The 512 would only have come from the 1.3 support

ctowns: thanks for clarifying! So is it reasonable to assume that the 510 devices just aren't working right now? If we are reading the wrong attribute, then I would assume the data we are getting is wrong so I'm not sure what "transition" we'd need to do clean up wise since the data would've just been wrong. I assume we can't change it once it's made it to ST energy. So in that case, I guess the only clean up we could do would be setting the export fields to nil. FWIW, ultimately this would save just a few bytes of memory per device.

I am surprised there are this many devices using this driver already, I would suggest we figure out what these devices are and try to procure one so we can test this functionality with a real device! That could give us some more confidence moving forward.

hcarter-775: Yes, that should be the case. In fact, if these devices are correctly using the 'Imported' attributes, we would not have read anything at all, so nothing would have even been reported to ST Energy in the first place, since the . 

Further, if anyone were incorrectly using the 'exported' attribute like us, there's this logic gate in the report:
`  if previous_imported_report and previous_imported_report.energy then
    energy_delta_wh = math.max(latest_total_imported_energy_wh - previous_imported_report.energy, 0.0)
  end`  
which will make sure it never reported a negative value. 

Therefore, by this logic we shouldn't have reported anything incorrectly so far to ST Energy except sending a lot of nil values from the fact that we're never updating 'TOTAL_IMPORTED_ENERGY', the value that the report reads from.

And as for saving a few bytes like you mentioned, I'm not sure this is worth it either.

drivers/SmartThings/matter-switch/src/ElectricalEnergyMeasurement/server/attributes/CumulativeEnergyImported.lua
hcarter-775: removed in [438b6252f08fba1d](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1830/commits/438b6252f08fba1dce8f2b398bef8a95ca1f05ce)

drivers/SmartThings/matter-switch/src/ElectricalEnergyMeasurement/server/attributes/CumulativeEnergyImported.lua
```diff
@@ -2,44 +2,44 @@ local cluster_base = require "st.matter.cluster_base"
...
 local TLVParser = require "st.matter.TLV.TLVParser"
 
-local PeriodicEnergyExported = {
-  ID = 0x0004,
-  NAME = "PeriodicEnergyExported",
-  base_type = require "ElectricalEnergyMeasurement.types.EnergyMeasurementStruct",
+local CumulativeEnergyImported = {
+  ID = 0x0001,
+  NAME = "CumulativeEnergyImported",
+  base_type = require "st.matter.generated.zap_clusters.ElectricalEnergyMeasurement.types.EnergyMeasurementStruct",
```
ctowns, line 8: This `st.matter.generated.zap_clusters` prefix needs to be removed for this to work properly on hub running older lua libs where these definitions do not exist, same in the file below.

hcarter-775: fixed

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -158,12 +158,12 @@ local child_device_profile_overrides = {
 local detect_matter_thing
 
 local CUMULATIVE_REPORTS_NOT_SUPPORTED = "__cumulative_reports_not_supported"
-local FIRST_EXPORT_REPORT_TIMESTAMP = "__first_export_report_timestamp"
-local EXPORT_POLL_TIMER_SETTING_ATTEMPTED = "__export_poll_timer_setting_attempted"
-local EXPORT_REPORT_TIMEOUT = "__export_report_timeout"
-local TOTAL_EXPORTED_ENERGY = "__total_exported_energy"
-local LAST_EXPORTED_REPORT_TIMESTAMP = "__last_exported_report_timestamp"
-local RECURRING_EXPORT_REPORT_POLL_TIMER = "__recurring_export_report_poll_timer"
+local FIRST_IMPORT_REPORT_TIMESTAMP = "__first_import_report_timestamp"
```
nickolas-deboom, line 161: Will changing the field names have any negative effect for devices already using the old names?

hcarter-775: see this conversation: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1830#discussion_r1889291624

hcarter-775: tldr: not really

nickolas-deboom: Great, thanks!

nickolas-deboom (approved):

ctowns (approved):
Looks good! I would run the unit tests against an older version of the lua libs locally. Also, let's get a physical device as soon as we can

PR #1831, opened 2024-12-18
Hue: Fix multi-component device initialization race condition

### LINKED ISSUE #1674: Philips hue dimmer switch driver
Since adding the Philips Hue accessories in the SmartThings Edge driver, buttons 2,3 and 4 of the Hue dimmer switch, does not report "pressed" or "held" status and as such is not functional in SmartThings. Button status is working in Hue environment.

dljsjr:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

We've identified an issue in the Hue driver related to the way we handled device init on driver startup for child devices that were ahead of the parent bridge in database order.

The race condition primarily manifested in the non-primary component for multi-component devices (such as multi-button remotes) not emitting attribute events or being able to handle capability commands.

The actual bugfix change is here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/e4cc7a4d54608b6dbedf70f6342eaf9e193fb3bd

The other two commits are chore work that we tackled in the process of fixing the bug:

- Mostly updates type hints/annotations
- Fix one place where typechecks caught a missing nil checkCheck all that apply
- Updates to the cheekily named "logjam" module that this driver uses to wrangle its extensive/complicated logging.

Fixes #1674 

# Summary of Completed Tests

Tested with various multi-button remotes.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/hue/api.lua
+++ b/drivers/SmartThings/philips-hue/src/hue/api.lua
@@ -5,11 +5,14 @@ local json = require "st.json"
 local log = require "log"
 local RestClient = require "lunchbox.rest"
 local st_utils = require "st.utils"
+-- trick to fix the VS Code Lua Language Server typechecking
+---@type fun(val: any?, name: string?, multi_line: boolean?): string
+st_utils.stringify_table = st_utils.stringify_table
 
 local HueDeviceTypes = require "hue_device_types"
 
 -- trick to fix the VS Code Lua Language Server typechecking
----@type fun(val: table, name: string?, multi_line: boolean?): string
+---@type fun(val: any?, name: string?, multi_line: boolean?): string
 st_utils.stringify_table = st_utils.stringify_table
 
 local APPLICATION_KEY_HEADER = "hue-application-key"
@@ -82,6 +85,7 @@ end
 ---@return table? tbl the table representation of the JSON response, nil on error
 ---@return string? err the error message, nil on success
 ---@return string? partial the partial response if the response was not complete
+---@return ...
 local function process_rest_response(response, err, partial, err_callback)
   if err == nil and response == nil then
     log.error_with({ hub_logs = true },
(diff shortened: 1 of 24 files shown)


drivers/SmartThings/philips-hue/src/logjam.lua
```diff
@@ -1,9 +1,84 @@
...
+  opts.call_info = nil
+
+  local module_name = nil
+  if type(call_info.source) == "string" then
+    module_name =
+    call_info.source
+    :gsub("%.lua", "")
+    :gsub("/init", "")
+    :gsub("/", ".")
+    :gsub("^init$", "philips-hue")
```
FreeMasen, line 62: nit: indentation here feels a bit off
```suggestion
    module_name = call_info.source
      :gsub("%.lua", "")
      :gsub("/init", "")
      :gsub("/", ".")
      :gsub("^init$", "philips-hue")
```

drivers/SmartThings/philips-hue/src/hue/api.lua
```diff
@@ -266,6 +273,7 @@ end
 ---@return HueApiKeyResponse[]? api_key_response nil on err
 ---@return string? error nil on success
 ---@return string? partial partial response if available, nil otherwise
+---@return ...
```
FreeMasen, line 276: nit: what is this indicating, that the function could return any number of additional `any` typed arguments? What is the advantage of adding this instead of enumerating the actual return types after the two know error values?

FreeMasen (approved):
I really like the `logjam` stuff. I wonder if we could get this into the `lua_libs` as it seems like it would be universally useful drivers.

drivers/SmartThings/philips-hue/src/hue/api.lua
dljsjr: This was to make the typechecker happy. The actual `unpack` here should stop at the three things indicated, but the signature of `table.unpack` in `LuaLS`'s meta typedefs indicates up to 12 return values.

dljsjr:
> I really like the `logjam` stuff. I wonder if we could get this into the `lua_libs` as it seems like it would be universally useful drivers.

I had the same thought. There are a few more refinements I want to make, but I was thinking it would make sense to port this over to the main `log` API eventually.

cjswedes (approved):

NoahCornell (approved):

PR #1843, opened 2024-12-26
Add Aqara Climate Sensor W100

DongHoon-Ryu:
This device has a Temperature Sensor(0x0302), a Humidity Sensor(0x0307), and 3 Generic Switchs(0x000F).

Check all that apply

# Type of Change

          - [ ] I have signed the CLA
     - [X] I plan on entering a WWST Certification Request or have entered a request through the WWST Certification console at developer.smartthings.com
- [ ] Bug fix
- [X] New feature
- [ ] Refactor

# Checklist

# Description of Change
As the first type of device with buttons in the sensor device, button feature is added to the sensor EdgeDriver

# Summary of Completed Tests
Successful test for Plugin UI and Dashboard update according to the change of sensor or button operation

### DIFF
--- a/drivers/SmartThings/matter-switch/profiles/temperature-humidity-3-button-battery.yml
+++ b/drivers/SmartThings/matter-switch/profiles/temperature-humidity-3-button-battery.yml
@@ -1,4 +1,4 @@
-name: temperature-humidity-3button-battery
+name: temperature-humidity-3-button-battery
 components:
 - id: main
   capabilities:
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -280,10 +280,6 @@ local EMULATE_HELD = "__emulate_held" -- for non-MSR (MomentarySwitchRelease) de
 local SUPPORTS_MULTI_PRESS = "__multi_button" -- for MSM devices (MomentarySwitchMultiPress), create an event on receipt of MultiPressComplete
 local INITIAL_PRESS_ONLY = "__initial_press_only" -- for devices that support MS (MomentarySwitch), but not MSR (MomentarySwitchRelease)
 
-local TEMP_BOUND_RECEIVED = "__temp_bound_received"
-local TEMP_MIN = "__temp_min"
-local TEMP_MAX = "__temp_max"
-
 local HUE_MANUFACTURER_ID = 0x100B
 local AQARA_MANUFACTURER_ID = 0x115F
 
@@ -543,13 +539,7 @@ local function initialize_switch(driver, device)
   -- that have been implemented as server. This can be removed when we have
   -- support for bindings.
   local num_switch_server_eps = 0
-  local main_endpoint
-  if device:supports_capability(capabilities.temperatureMeasurement) then
-    -- In case of Aqara Climate Sensor W100, in order to sequentially set the button name to button1, 2, 3
-    main_endpoint = device.MATTER_DEFAULT_ENDPOINT
-  else
-    main_endpoint = find_default_endpoint(device)
-  end
+  local main_endpoint = find_default_endpoint(device)
 
   -- If a switch endpoint is present, it will be the main endpoint and therefore the
   -- main component. If button endpoints are present, they will be added as
@@ -623,10 +613,7 @@ local function initialize_switch(driver, device)
       battery_support = true
     end
     if #button_eps > 1 and tbl_contains(STATIC_BUTTON_PROFILE_SUPPORTED, #button_eps) then
-      if device:supports_capability(capabilities.temperatureMeasurement) then
-        -- to call configure_buttons, keep the profile_name as nil.
-        device.log.debug("So far, it means Aqara Climate Sensor W100.")
-      elseif battery_support then
+      if battery_support then
         profile_name = string.format("%d-button-battery", #button_eps)
       else
         profile_name = string.format("%d-button", #button_eps)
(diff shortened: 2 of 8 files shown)


hcarter-775:
Since the matter-switch driver is the most complete and vast of our drivers, I believe it would be more prudent to include new humidity and temperature support in matter-switch rather than including the whole array of switch logic in matter-sensor.

DongHoon-Ryu:
> Since the matter-switch driver is the most complete and vast of our drivers, I believe it would be more prudent to include new humidity and temperature support in matter-switch rather than including the whole array of switch logic in matter-sensor.

Thanks, I modified to support w100 in matter-switch, not matter-sensor.

drivers/SmartThings/matter-switch/profiles/temperature-humidity-3-button-battery.yml
```diff
@@ -0,0 +1,99 @@
+name: temperature-humidity-3-button-battery
```
hcarter-775, line 1: can we change this name to `3-button-battery-temperature-humidity`? This follows the unofficial naming conventions more closely.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -517,7 +521,13 @@ local function initialize_switch(driver, device)
   -- that have been implemented as server. This can be removed when we have
   -- support for bindings.
   local num_switch_server_eps = 0
-  local main_endpoint = find_default_endpoint(device)
+  local main_endpoint
+  if device:supports_capability(capabilities.temperatureMeasurement) then
+    -- In case of Aqara Climate Sensor W100, in order to sequentially set the button name to button1, 2, 3
```
hcarter-775, line 526: @nickolas-deboom out of curiosity, is this sequential ordering not the default handling? If not, seems like we should change that?

hcarter-775, line 526: nit:
```suggestion
    -- In case of Aqara Climate Sensor W100, in order to sequentially set the button name to button 1, 2, 3
```

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -586,7 +596,10 @@ local function initialize_switch(driver, device)
...
     end
     if #button_eps > 1 and tbl_contains(STATIC_BUTTON_PROFILE_SUPPORTED, #button_eps) then
-      if battery_support then
+      if device:supports_capability(capabilities.temperatureMeasurement) then
+        -- to call configure_buttons, keep the profile_name as nil.
+        device.log.debug("So far, it means Aqara Climate Sensor W100.")
+      elseif battery_support then
         profile_name = string.format("%d-button-battery", #button_eps)
       else
         profile_name = string.format("%d-button", #button_eps)
```
hcarter-775, line 605:
```suggestion
      if device:supports_capability(capabilities.temperatureMeasurement) and device:supports_capability(capabilities.relativeHumidityMeasurement) then
        profile_name = "temperature-humidity"
      end
      if battery_support then
        profile_name = string.format("%d-button-battery", #button_eps) .. profile_name
      else
        profile_name = string.format("%d-button", #button_eps) .. profile_name
```

@nickolas-deboom do you think we should support full dynamicism here (a device:get_endpoints() call) or is leaving this as a fingerprint specific handling ok?

hcarter-775: the `.. profile_name` assumes the profile name has been changed according to my previous comment

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I believe that what is happening is that since this device doesn't contain a 'switch' endpoint (aka an endpoint containing the OnOff cluster), `find_default_endpoint` is returning the first button endpoint as the main endpoint. That would cause the first button component name to be 'main' and the other two would be 'button2' and 'button3'. I think that `find_default_endpoint` should be updated rather than using the workaround here. Either it could be updated to check for an endpoint containing the `TemperatureMeasurement` (or `RelativeHumidityMeasurement`) cluster and return that as the default endpoint if one exists, or it could be updated to just return the first non-button endpoint if one exists (rather than checking for `OnOff`/`TemperatureMeasurement`/`RelativeHumidityMeasurement` clusters)..

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I am fine with it either way for now, especially since we'd just be re-profiling to the same profile it has already joined. If we start supporting more devices like this in the future then I'd definitely support making it more dynamic.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Yes, I agree with this

drivers/SmartThings/matter-switch/profiles/temperature-humidity-3-button-battery.yml
DongHoon-Ryu: I changed `temperature-humidity-3-button-battery` to `3-button-battery-temperature-humidity`.

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: I changed it.

drivers/SmartThings/matter-switch/profiles/temperature-humidity-3-button-battery.yml
Apshakil: > @@ -0,0 +1,99 @@
+name: temperature-humidity-3-button-battery

DongHoon-Ryu: @Apshakil Sorry, Please refer to my latest commit(https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1843/commits/8f513e265d5fc8ba76544d524b97aae50df0c0a1)

drivers/SmartThings/matter-switch/src/init.lua
DongHoon-Ryu: @nickolas-deboom @hcarter-775 If we use fingerprint of `Aqara Climate Sensor W100` device, there is a problem that `configure_buttons(device)` is not called because a value is assigned to `profile_name`. So I modified to use the number of `temperature_eps` and `humidity_eps` obtained using `device:get_endpoints` to specify profile in the more dynamic way you mentioned.

```lua
    if #button_eps > 1 and tbl_contains(STATIC_BUTTON_PROFILE_SUPPORTED, #button_eps) then
      if #temperature_eps > 0 and #humidity_eps > 0 then
        device.log.debug("So far, it means Aqara Climate Sensor W100.")
        profile_name = "-temperature-humidity"
      end
      if battery_support then
        profile_name = string.format("%d-button-battery", #button_eps) .. profile_name
      else
        profile_name = string.format("%d-button", #button_eps) .. profile_name
      end
    elseif not battery_support then
      -- a battery-less button/remote
      profile_name = "button"
    end

    if profile_name ~= "" then
      device:set_field(DEFERRED_CONFIGURE, true)
      device:set_field(BUTTON_DEVICE_PROFILED, true)
    else
      configure_buttons(device)
    end
```

DongHoon-Ryu:
@nickolas-deboom @hcarter-775 Thank you for your comments on many reviews all the time. Please review and if there is no problem, please approve.

nickolas-deboom:
@DongHoon-Ryu this looks good to me! Thank you for making the requested updates.

nickolas-deboom (approved):

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: @nickolas-deboom @DongHoon-Ryu not that this is a huge deal, but I think these changes that Nick suggested were missed in this PR.

PR #1860, opened 2025-01-10
Add batteryLevel to matter window covering

nickolas-deboom:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

[CHAD-12161](https://smartthings.atlassian.net/browse/CHAD-12161)

Matter drivers currently assume that `BatPercentRemaining` is available if the `BAT` feature is supported, but this attribute is only optionally required if this feature is present. The changes in this PR improve the profile selection logic in matter-window-covering by reading the `AttributeList`, checking if `BatPercentRemaining` or `BatChargeLevel` is available, and then profiling the device as needed.

Note that these changes were originally in [PR 1796](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1796) but this PR was split up into 5 separate PRs, one for each affected driver.

# Summary of Completed Tests

Tested with Hunter Douglas blinds to ensure no change in behavior. New unit tests are also included.

[CHAD-12161]: https://smartthings.atlassian.net/browse/CHAD-12161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hcarter-775 (approved):

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -608,7 +636,7 @@ test.register_coroutine_test(
   function()
     test.socket.device_lifecycle:__queue_receive(mock_device:generate_info_changed({}))
     mock_device:expect_metadata_update({
-      profile = "window-covering-battery",
+      profile = "window-covering",
```
ctowns, line 639: Why does this no longer update to `window-covering-battery`? The mock device hasn't changed as far as I can tell. And if it isn't changing, then it seems like this unit test is not longer properly testing the InfoChanged event.

nickolas-deboom: Because in this test case, the `AttributeList` is not being read and so the driver is not aware that battery is supported. I could update it to do this read before the `info_changed` event (aka precede it with `do_configure`), but technically this is how `info_changed` would behave without already knowing if the devices supports a battery or not (which in real life it should know because do_configure would have already ran)

hcarter-775: Yeah, I suppose since this isn't necessarily a real case, I would suggest updating it to include the do_configure logic (since this is what a real device would do as you mentioned).

nickolas-deboom: I updated the `info_changed` event to read the attribute list like in do_configure. That way it will know if the battery feature is supported before attempting to match the profile

hcarter-775 (approved):

ctowns (approved):

PR #1870, opened 2025-01-13
Matter 1.4 Driver Release

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

This PR contains changes to support new device types introduced by the Matter 1.4 specification. These device types include the following:
- Mounted On/Off Control
- Mounted Dimmable Load Control
- Water Heater
- Solar Power
- Battery Storage
- Heat Pump

# Summary of Completed Tests

[CHAD-14600]: https://smartthings.atlassian.net/browse/CHAD-14600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


drivers/SmartThings/matter-hrap/src/ThreadBorderRouterManagement/init.lua
```diff
@@ -0,0 +1,148 @@
...
+  local direction = ThreadBorderRouterManagement.attribute_direction_map[key]
+  if direction == nil then
+    error(string.format("Referenced unknown attribute %s on cluster %s", key, ThreadBorderRouterManagement.NAME))
+  end
+  return ThreadBorderRouterManagement[direction].attributes[key]
+end
+ThreadBorderRouterManagement.attributes = {}
+setmetatable(ThreadBorderRouterManagement.attributes, attribute_helper_mt)
+
+local command_helper_mt = {}
```
ctowns, line 127: For what it's worth, since there are no commands and events included in this embedded definition, you could delete this code and other code related to it to save some driver package space since it would essentially be unused.

drivers/SmartThings/matter-hrap/src/ThreadBorderRouterManagement/server/attributes/init.lua
```diff
@@ -0,0 +1,29 @@
...
+
+function ThreadBorderRouterManagementServerAttributes:set_parent_cluster(cluster)
+  self._cluster = cluster
+  return self
+end
+
+setmetatable(ThreadBorderRouterManagementServerAttributes, attr_mt)
+
+local has_aliases, aliases = pcall(require, "st.matter.clusters.aliases.ThreadBorderRouterManagement.server.attributes")
+if has_aliases then
```
ctowns, line 24: We can also remove all references to the aliases from the embedded cluster defs since these definitions would not use/have access to the aliases anyway

drivers/SmartThings/matter-hrap/src/ThreadBorderRouterManagement/types/init.lua
```diff
@@ -0,0 +1,20 @@
...
+    types_mt.__types_cache[key] = require("ThreadBorderRouterManagement.types." .. key)
+  end
+  return types_mt.__types_cache[key]
+end
+
+local ThreadBorderRouterManagementTypes = {}
+
+setmetatable(ThreadBorderRouterManagementTypes, types_mt)
+
+local status, aliases = pcall(require, "st.matter.clusters.aliases.ThreadBorderRouterManagement.types")
```
ctowns, line 14: Alias code can be removed here and throughout the embedded cluster definitions

drivers/SmartThings/matter-hrap/src/WiFiNetworkManagement/init.lua
```diff
@@ -0,0 +1,125 @@
...
+  local direction = WiFiNetworkManagement.attribute_direction_map[key]
+  if direction == nil then
+    error(string.format("Referenced unknown attribute %s on cluster %s", key, WiFiNetworkManagement.NAME))
+  end
+  return WiFiNetworkManagement[direction].attributes[key]
+end
+WiFiNetworkManagement.attributes = {}
+setmetatable(WiFiNetworkManagement.attributes, attribute_helper_mt)
+
+local command_helper_mt = {}
```
ctowns, line 104: Similar to above, code referencing the commands and events which are not used in the embedded cluster definition can be removed here and throughout the other embedded cluster files to reduce driver package size.

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,129 @@
...
+  clusters.ThreadBorderRouterManagement = require "ThreadBorderRouterManagement"
+  clusters.WifiNetworkMangement = require "WiFiNetworkManagement"
+end
+
+
+--[[ ATTRIBUTE HANDLERS ]]--
+
+local function ssid_attribute_handler(driver, device, ib)
+  if ib.data.value == string.char(0x014) then -- TLV-encoded NULL
+    log.info("Ssid is null. Per the spec, no primary Wi-Fi network is available at the moment.")
```
ctowns, line 39: as a rule of thumb, I would suggest using `device.log` instead of `log` just because its essentially the same thing but it prefixes the log message with the device ID for easier debugging

drivers/SmartThings/matter-hrap/src/ThreadBorderRouterManagement/init.lua
```diff
@@ -0,0 +1,148 @@
+local cluster_base = require "st.matter.cluster_base"
+local ThreadBorderRouterManagementServerAttributes = require "ThreadBorderRouterManagement.server.attributes"
+local ThreadBorderRouterManagementServerCommands = require "ThreadBorderRouterManagement.server.commands"
+local ThreadBorderRouterManagementClientCommands = require "ThreadBorderRouterManagement.client.commands"
```
nickolas-deboom, line 4: These directories don't exist anymore
```suggestion
```

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,129 @@
...
+  local valid_utf8, utf8_err = utils.validate_utf8(ib.data.value)
+  if valid_utf8 then
+    device:emit_event_for_endpoint(ib.endpoint, wifiSsid.ssid({ value = ib.data.value }))
+  else
+    log.info("UTF8 validation of Ssid failed: Error: '"..utf8_err.."'. Raw Ssid data: "..ib.data.value)
+  end
+end
+
+local function border_router_name_attribute_handler(driver, device, ib)
+  local router_name = ib.data.value
```
ctowns, line 51: For my own understanding - why doesn't there need to be as much string verification for the border router name vs. the SSID name above?

drivers/SmartThings/matter-hrap/src/WiFiNetworkManagement/init.lua
```diff
@@ -0,0 +1,125 @@
+local cluster_base = require "st.matter.cluster_base"
+local WiFiNetworkManagementServerAttributes = require "WiFiNetworkManagement.server.attributes"
+local WiFiNetworkManagementServerCommands = require "WiFiNetworkManagement.server.commands"
+local WiFiNetworkManagementClientCommands = require "WiFiNetworkManagement.client.commands"
+local WiFiNetworkManagementTypes = require "WiFiNetworkManagement.types"
```
nickolas-deboom, line 5: Same with these
```suggestion
```

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -661,7 +665,8 @@ local function initialize_switch(driver, device)
         if main_endpoint == ep.endpoint_id then
           for _, dt in ipairs(ep.device_types) do
             -- no device type that is not in the switch subset should be considered.
-            if (ON_OFF_SWITCH_ID <= dt.device_type_id and dt.device_type_id <= ON_OFF_COLOR_DIMMER_SWITCH_ID) then
+            if (ON_OFF_SWITCH_ID <= dt.device_type_id and dt.device_type_id <= ON_OFF_COLOR_DIMMER_SWITCH_ID) or
+              dt.device_type_id == MOUNTED_ON_OFF_CONTROL_ID or dt.device_type_id == MOUNTED_DIMMABLE_LOAD_CONTROL_ID then
```
ctowns, line 669: In theory, a device with one of the new mounted control device types should not fall into this conditional because it would be backed by the generic fingerprints you have added and would not join as a matter-thing. These new device types are essentially a way to remedy the issue around the light switch device types that is explained in the comment above on line 654. Therefore, I don't think you need to add these here.

nickolas-deboom:
I left a few comments but everything else in the HRAP driver looks good to me! I can't remember but was there a file we've updated in the past to prevent github from checking test coverage of the embedded lua libs files? We could update that as part of this PR as well (I just can't remember where it is)

nickolas-deboom:
I get an error when trying to package the hrap driver locally:
```
» smartthings edge:drivers:package                                                                                                                                nickloasdeboom@MN-NDEBOOM-L
    AxiosError: Request failed with status code 422: {"requestId":"1610617032588185407","error":{"code":"ConstraintViolationError","message":"Invalid device profile specification for 
    eb1b3ce7-a434-4cd9-8fab-e5289b628b35","details":[{"code":"NotValidValue","target":"main.capabilities.routerName","message":"Capability does not exist: routerName version 
    1","details":[]},{"code":"NotValidValue","target":"main.capabilities.threadVersion","message":"Capability does not exist: threadVersion version 
    1","details":[]},{"code":"NotValidValue","target":"main.capabilities.routerState","message":"Capability does not exist: routerState version 1","details":[]}]}}
    Code: ERR_BAD_REQUEST
```

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Good point - I removed this in the latest commit and also updated the test cases accordingly.

ctowns:
Just a heads up, the current unit test results posted on the PR include coverage results for the embedded cluster definitions as well, so you'll want to add each of the new embedded cluster names to the [config.luacov](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/tools/config.luacov) so that these cluster definitions are ignored in the code coverage calculations since they can add a lot of bloat to the coverage results and we don't necessarily use all the code in the embedded clusters.

drivers/SmartThings/matter-hrap/src/ThreadBorderRouterManagement/init.lua
hcarter-775: @nickolas-deboom @ctowns I specifically did not go remove extra embedded code, and I stand by that. It seems like a relatively time consuming and error prone process for very little gain. I'm not saying these things can't be removed, but I don't think it's really worth a few bytes, especially since so many other embedded clusters have full attributes, commands, etc that they don't need.

hcarter-775:
> I get an error when trying to package the hrap driver locally:
> 
> ```
> » smartthings edge:drivers:package                                                                                                                                nickloasdeboom@MN-NDEBOOM-L
>     AxiosError: Request failed with status code 422: {"requestId":"1610617032588185407","error":{"code":"ConstraintViolationError","message":"Invalid device profile specification for 
>     eb1b3ce7-a434-4cd9-8fab-e5289b628b35","details":[{"code":"NotValidValue","target":"main.capabilities.routerName","message":"Capability does not exist: routerName version 
>     1","details":[]},{"code":"NotValidValue","target":"main.capabilities.threadVersion","message":"Capability does not exist: threadVersion version 
>     1","details":[]},{"code":"NotValidValue","target":"main.capabilities.routerState","message":"Capability does not exist: routerState version 1","details":[]}]}}
>     Code: ERR_BAD_REQUEST
> ```
> 
> Is this happening due to the embedded capability definitions?

No, this is because the capabilities don't exist yet. I am currently in the process of getting these defined under a good namespace

nickolas-deboom:
> Just a heads up, the current unit test results posted on the PR include coverage results for the embedded cluster definitions as well, so you'll want to add each of the new embedded cluster names to the [config.luacov](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/tools/config.luacov) so that these cluster definitions are ignored in the code coverage calculations since they can add a lot of bloat to the coverage results and we don't necessarily use all the code in the embedded clusters.

I updated config.luacov in the latest commit: ae7066bd9f3b253e2ffa3f8c4f6c9e0450f61556

drivers/SmartThings/matter-hrap/src/WiFiNetworkManagement/init.lua
hcarter-775: removed

drivers/SmartThings/matter-hrap/src/ThreadBorderRouterManagement/init.lua
hcarter-775: removed

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: All of this is per the spec: 
The border router name attribute is of type string, while the ssid is a type octsr. Therefore, border router name can be taken as is, while we have to ensure the Ssid can be displayed in a printable way (per wifi spec, ssid does not have to be written in UTF8). Further, SSID can be TLV-encoded null if a wifi network hasn't been set up yet on the TBR/NIM. So there's just some extra stuff with Ssid to watch out for per the spec.

However, I did add some extra logic for the border router name after a quick review of the spec- I had missed this on the first pass, but the spec specifies that the name is written `<VendorName> <ProductName>._meshcop._udp`. the new logic should remove `._meshcop._udp` if present.

hcarter-775:
@HunsupJung a general update for the embedded cluster definitions you provided would be to remove the unused files (commands, events, unused attributes) to cut down on some size.

drivers/SmartThings/matter-thermostat/profiles/heat-pump.yml
```diff
@@ -0,0 +1,40 @@
+name: heat-pump
```
hcarter-775, line 1: This profile name should be changed to something like `heat-pump-2-thermostat`, so that it is more extensible.

drivers/SmartThings/matter-thermostat/profiles/water-heater-power-energy-powerConsumption.yml
```diff
@@ -0,0 +1,24 @@
...
+- id: main
+  capabilities:
+  - id: temperatureMeasurement
+    version: 1
+  - id: thermostatMode
+    version: 1
+  - id: thermostatHeatingSetpoint
+    version: 1
+  - id: mode
+    version: 1
```
hcarter-775, line 12: this capability ordering doesn't match the ordering in water-heater.yml. Which way would you prefer for it to be portrayed? And then apply it to both, thank you :+1:

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -240,9 +270,105 @@ local subscribed_attributes = {
...
+end
+
+local function remove_timers(device)
+  delete_reporting_timer(device)
+  local poll_timer = device:get_field(RECURRING_POLL_TIMER)
+  if poll_timer ~= nil then
+    device.thread:cancel_timer(poll_timer)
+    device:set_field(RECURRING_POLL_TIMER, nil)
+  end
+end
```
hcarter-775, line 366: This seems like unnecessary breakdown of functionality to me. I would suggest:
```suggestion
local function remove_timers(device)
  -- remove report timer
  local reporting_poll_timer = device:get_field(RECURRING_REPORT_TIMER)
  if reporting_poll_timer ~= nil then
    device.thread:cancel_timer(reporting_poll_timer)
    device:set_field(RECURRING_REPORT_TIMER, nil)
  end
  -- remove poll timer
  local poll_timer = device:get_field(RECURRING_POLL_TIMER)
  if poll_timer ~= nil then
    device.thread:cancel_timer(poll_timer)
    device:set_field(RECURRING_POLL_TIMER, nil)
  end
end
```

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -267,18 +393,82 @@ end
...
+    return component_to_endpoint_map[component_name]
+  end
   if device:supports_capability(capabilities.airPurifierFanMode) then
     -- Fan Control is mandatory for the Air Purifier device type
     return find_default_endpoint(device, clusters.FanControl.ID)
   else
     -- Thermostat is mandatory for Thermostat and Room AC device type
-    return find_default_endpoint(device, clusters.Thermostat.ID)
+    local val = find_default_endpoint(device, clusters.Thermostat.ID)
+    return val
```
hcarter-775, line 406: why was this changed?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -240,9 +270,105 @@ local subscribed_attributes = {
...
+  pcr_interval = utils.clamp_value(pcr_interval, DEFAULT_REPORT_TIME_INTERVAL, MAX_REPORT_TIMEOUT)
+  local timer = device.thread:call_on_schedule(pcr_interval, function()
+    local last_time = device:get_field(LAST_REPORTED_TIME) or 0
+    local current_time = os.time()
+    local total_energy = get_total_cumulative_energy_imported(device)
+    device:set_field(LAST_REPORTED_TIME, current_time, { persist = true })
+
+    -- Calculate the energy consumed between the start and the end time
+    local previousTotalConsumptionWh = device:get_latest_state("main", capabilities.powerConsumptionReport
+      .ID,
```
hcarter-775, line 332:
```suggestion
    local previousTotalConsumptionWh = device:get_latest_state("main", capabilities.powerConsumptionReport.ID,
```

drivers/SmartThings/matter-evse/src/init.lua
```diff
@@ -154,13 +183,33 @@ local EVSE_FAULT_STATE_ENUM_MAP = {
...
+  [clusters.PowerSource.types.BatChargeStateEnum.IS_AT_FULL_CHARGE] = capabilities.chargingState.chargingState.fullyCharged,
+}
+
+-- Matter Handlers
+local function read_cumulative_energy(device)
+  local cumul_eps = embedded_cluster_utils.get_endpoints(device,
+    clusters.ElectricalEnergyMeasurement.ID,
+    {feature_bitmap = clusters.ElectricalEnergyMeasurement.types.Feature.CUMULATIVE_ENERGY })
+  if cumul_eps and #cumul_eps > 0 then
+    local read_req = clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyImported:read(device)
```
nickolas-deboom, line 198: Should there be a loop here similar to below in case there are multiple endpoints containing the ElectricalEnergyMeasurement cluster? Or is it not expected to have multiple endpoints?
```suggestion
    local read_req = clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyImported:read(device)
    for i, ep in ipairs(cumul_eps) do
      if i > 1 then
        read_req:merge( clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyExported:read(device, cumul_eps[i]))
      end
    end
```

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -502,8 +700,14 @@ local function do_configure(driver, device)
...
-
-  elseif #thermostat_eps > 0 then
+  elseif device_type == WATER_HEATER_DEVICE_TYPE_ID then
+    -- If a Water Heater is composed of Electrical Sensor device type, it must support both ElectricalEnergyMeasurement and
+    -- ElectricalPowerMeasurement clusters.
+    local electrical_sensor_eps = device:get_endpoints(clusters.ElectricalEnergyMeasurement.ID)
+    if #electrical_sensor_eps > 0 then
+      profile_name = "water-heater-power-energy-powerConsumption"
+    end
+  elseif #thermostat_eps > 0 and device_type ~= HEAT_PUMP_DEVICE_TYPE_ID then
```
hcarter-775, line 710: I think we should keep this simple and implement it as:
```suggestion
  elseif device_type == HEAT_PUMP_DEVICE_TYPE_ID then
    profile_name = "heat-pump"
  elseif #thermostat_eps > 0 then
```
This is more extensible and keeps the original logical flow.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -542,13 +746,36 @@ local function do_configure(driver, device)
...
   device:send(req)
+  local heat_pump_eps = get_endpoints_for_dt(device, HEAT_PUMP_DEVICE_TYPE_ID)
+  local thermostat_eps = get_endpoints_for_dt(device, THERMOSTAT_DEVICE_TYPE_ID)
+  if #heat_pump_eps > 0 then
+    local component_to_endpoint_map = {
+      ["thermostatOne"] = thermostat_eps[1],
+      ["thermostatTwo"] = thermostat_eps[2],
+    }
+    device:set_field(COMPONENT_TO_ENDPOINT_MAP, component_to_endpoint_map, {persist = true})
+  end
```
hcarter-775, line 778:
```suggestion
  if #heat_pump_eps > 0 then
    local thermostat_eps = get_endpoints_for_dt(device, THERMOSTAT_DEVICE_TYPE_ID)
    local component_to_endpoint_map = {
      ["thermostatOne"] = thermostat_eps[1],
      ["thermostatTwo"] = thermostat_eps[2],
    }
    device:set_field(COMPONENT_TO_ENDPOINT_MAP, component_to_endpoint_map, {persist = true})
  end
```
Though this is pretty un-extendable as-is. I do think we should extend this logic to permit 1, 2, or more thermostats, even if we only have a profile to support the 2 case for now.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1323,18 +1573,126 @@ local function set_rock_mode(driver, device, cmd)
...
+local function periodic_energy_imported_handler(driver, device, ib, response)
+  local endpoint_id = ib.endpoint_id
+  local cumul_eps = embedded_cluster_utils.get_endpoints(device,
+    clusters.ElectricalEnergyMeasurement.ID,
+    { feature_bitmap = clusters.ElectricalEnergyMeasurement.types.Feature.CUMULATIVE_ENERGY })
+
+  if ib.data then
+    if version.api < 11 then
+      clusters.ElectricalEnergyMeasurement.attributes.PeriodicEnergyImported:augment_type(ib.data)
+    end
```
hcarter-775, line 1613: for my understanding, what is this logic doing?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1323,18 +1573,126 @@ local function set_rock_mode(driver, device, cmd)
...
+      device:set_field(DEVICE_POWER_CONSUMPTION_REPORT_TIME_INTERVAL, device_reporting_time_interval, { persist = true })
+      delete_reporting_timer()
+      schedule_energy_report_timer(device)
+    end
+
+    if tbl_contains(cumul_eps, endpoint_id) then
+      -- Since cluster at this endpoint supports both CUME & PERE features, we will prefer
+      -- cumulative_energy_imported_handler to handle the energy report for this endpoint.
+      return
+    end
```
hcarter-775, line 1631: let's move this top the top of the function handler. There's no need to do any further logic if this is true.

drivers/SmartThings/matter-evse/src/init.lua
```diff
@@ -189,105 +259,107 @@ local function create_poll_report_schedule(device)
...
-  local electrical_sensor_eps = get_endpoints_for_dt(device, ELECTRICAL_SENSOR_DEVICE_ID) or {}
-  local device_energy_mgmt_eps = get_endpoints_for_dt(device, DEVICE_ENERGY_MANAGEMENT_DEVICE_ID) or {}
-  local component_to_endpoint_map = {
-    ["main"] = device.MATTER_DEFAULT_ENDPOINT,
-    ["electricalSensor"] = electrical_sensor_eps[1],
-    ["deviceEnergyManagement"] = device_energy_mgmt_eps[1]
-  }
-  log.debug("component_to_endpoint_map " .. utils.stringify_table(component_to_endpoint_map))
-  device:set_field(COMPONENT_TO_ENDPOINT_MAP, component_to_endpoint_map, { persist = true })
+  local evse_eps = get_endpoints_for_dt(device, EVSE_DEVICE_TYPE_ID)
```
nickolas-deboom, line 328:
```suggestion
  local evse_eps = get_endpoints_for_dt(device, EVSE_DEVICE_TYPE_ID) or {}
```
To prevent an error on the next line in the case that `get_endpoints_for_dt` returns nil

drivers/SmartThings/matter-evse/src/init.lua
```diff
@@ -642,9 +780,15 @@ matter_driver_template = {
         [clusters.DeviceEnergyManagementMode.attributes.CurrentMode.ID] = device_energy_mgmt_mode_attr_handler,
       },
       [clusters.ElectricalEnergyMeasurement.ID] = {
-        [clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyImported.ID] = cumulative_energy_imported_handler,
-        [clusters.ElectricalEnergyMeasurement.attributes.PeriodicEnergyImported.ID] = periodic_energy_imported_handler
+        [clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyImported.ID] = cumulative_energy_handler(TOTAL_CUMULATIVE_ENERGY_IMPORTED),
+        [clusters.ElectricalEnergyMeasurement.attributes.PeriodicEnergyImported.ID] = periodic_energy_handler(TOTAL_CUMULATIVE_ENERGY_IMPORTED),
+        [clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyExported.ID] = cumulative_energy_handler(TOTAL_CUMULATIVE_ENERGY_EXPORTED),
+        [clusters.ElectricalEnergyMeasurement.attributes.PeriodicEnergyImported.ID] = periodic_energy_handler(TOTAL_CUMULATIVE_ENERGY_EXPORTED),
```
nickolas-deboom, line 786:
```suggestion
        [clusters.ElectricalEnergyMeasurement.attributes.PeriodicEnergyExported.ID] = periodic_energy_handler(TOTAL_CUMULATIVE_ENERGY_EXPORTED),
```

nickolas-deboom: I think this was meant to be `PeriodicEnergyExported`

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1323,18 +1573,126 @@ local function set_rock_mode(driver, device, cmd)
...
+
+    local energy_imported_mWh = ib.data.elements.energy.value
+    endpoint_id = string.format(ib.endpoint_id)
+    local energy_imported_Wh = utils.round(energy_imported_mWh / 1000)
+    local cumulative_energy_imported = device:get_field(TOTAL_CUMULATIVE_ENERGY_IMPORTED_MAP) or {}
+    cumulative_energy_imported[endpoint_id] = cumulative_energy_imported[endpoint_id] + energy_imported_Wh
+    device:set_field(TOTAL_CUMULATIVE_ENERGY_IMPORTED_MAP, cumulative_energy_imported, { persist = true })
```
hcarter-775, line 1638: why are we adding this map handling? Are we expecting devices that support energy control on multiple endpoints? As far as I see, neither of the two new thermostat device types support this.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1323,18 +1573,126 @@ local function set_rock_mode(driver, device, cmd)
...
+
+local function cumulative_energy_imported_handler(driver, device, ib, response)
+  if ib.data then
+    if version.api < 11 then
+      clusters.ElectricalEnergyMeasurement.server.attributes.CumulativeEnergyImported:augment_type(ib.data)
+    end
+    local endpoint_id = string.format(ib.endpoint_id)
+    local cumulative_energy_imported = device:get_field(TOTAL_CUMULATIVE_ENERGY_IMPORTED_MAP) or {}
+    local cumulative_energy_imported_mWh = ib.data.elements.energy.value
+    local cumulative_energy_imported_Wh = utils.round(cumulative_energy_imported_mWh / 1000)
```
hcarter-775, line 1652:
```suggestion
    local cumulative_energy_imported_Wh = utils.round( ib.data.elements.energy.value / 1000) -- convert mWh to Wh
```

hcarter-775: same thing for the periodic energy handler

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1323,18 +1573,126 @@ local function set_rock_mode(driver, device, cmd)
...
+end
+
+local function water_heater_mode_handler(driver, device, ib, response)
+  device.log.info_with({ hub_logs = true },
+    string.format("water_heater_mode_handler mode: %s", ib.data.value))
+
+  local supportWaterHeaterModes = device:get_field(SUPPORTED_WATER_HEATER_MODES)
+  local currentMode = ib.data.value
+  for i, mode in ipairs(supportWaterHeaterModes) do
+    if i - 1 == currentMode then
```
hcarter-775, line 1682: Is this numerical ordering of the modes an assumption we can safely make?

drivers/SmartThings/matter-thermostat/src/test/test_matter_heat_pump.lua
```diff
@@ -0,0 +1,725 @@
...
+  "Assert component to endpoint map",
+  function()
+    local component_to_endpoint_map = mock_device:get_field("__component_to_endpoint_map")
+    assert(component_to_endpoint_map["thermostatOne"] == THERMOSTAT_ONE_EP, string.format("Thermostat One Endpoint must be %d", THERMOSTAT_ONE_EP))
+    assert(component_to_endpoint_map["thermostatTwo"] == THERMOSTAT_TWO_EP, string.format("Thermostat Two Endpoint must be %d", THERMOSTAT_TWO_EP))
+  end
+)
+
+test.register_message_test(
+  "Heating setpoint reports from child thermostat devices should emit correct events to the corret endpoint",
```
hcarter-775, line 206:
```suggestion
  "Heating setpoint reports from component thermostat devices should emit correct events to the corret endpoint",
```

drivers/SmartThings/matter-thermostat/src/test/test_matter_heat_pump.lua
```diff
@@ -0,0 +1,725 @@
...
+    {
+      channel = "capability",
+      direction = "send",
+      message = mock_device:generate_test_message("thermostatTwo", capabilities.thermostatHeatingSetpoint.heatingSetpoint({ value = 23.0, unit = "C" }))
+    }
+  }
+)
+
+test.register_message_test(
+  "Cooling setpoint reports reports from child thermostat devices should emit correct events to the corret endpoint",
```
hcarter-775, line 238:
```suggestion
  "Cooling setpoint reports reports from component thermostat devices should emit correct events to the corret endpoint",
```

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -542,13 +746,36 @@ local function do_configure(driver, device)
...
   local req = im.InteractionRequest(im.InteractionRequest.RequestType.READ, {})
   req:merge(clusters.Thermostat.attributes.ControlSequenceOfOperation:read(device))
   req:merge(clusters.FanControl.attributes.FanModeSequence:read(device))
   req:merge(clusters.FanControl.attributes.WindSupport:read(device))
   req:merge(clusters.FanControl.attributes.RockSupport:read(device))
   device:send(req)
+  local heat_pump_eps = get_endpoints_for_dt(device, HEAT_PUMP_DEVICE_TYPE_ID)
+  local thermostat_eps = get_endpoints_for_dt(device, THERMOSTAT_DEVICE_TYPE_ID)
+  if #heat_pump_eps > 0 then
+    local component_to_endpoint_map = {
```
nickolas-deboom, line 773:
```suggestion
    table.sort(thermostat_eps)
    local component_to_endpoint_map = {
```

drivers/SmartThings/matter-thermostat/profiles/heat-pump.yml
HunsupJung: Okay I will modify it.

drivers/SmartThings/matter-thermostat/profiles/water-heater-power-energy-powerConsumption.yml
HunsupJung: Good point. I will organize it.

drivers/SmartThings/matter-evse/src/init.lua
```diff
@@ -42,13 +43,45 @@ local POWER_CONSUMPTION_REPORT_TIME_INTERVAL = "__pcr_time_interval"
...
+      end
+    end
+  end
+  table.sort(endpoints)
+  return endpoints
+end
+
+local find_default_endpoint = function(device)
+  local evse_eps = get_endpoints_for_dt(device, EVSE_DEVICE_TYPE_ID)
+  local solar_power_eps = get_endpoints_for_dt(device, SOLAR_POWER_DEVICE_TYPE_ID)
```
hcarter-775, line 77:
```suggestion
  local evse_eps = get_endpoints_for_dt(device, EVSE_DEVICE_TYPE_ID)
  local solar_power_eps = get_endpoints_for_dt(device, SOLAR_POWER_DEVICE_TYPE_ID)
  evse_eps.sort()
  solar_power_eps.sort()
```

drivers/SmartThings/matter-evse/src/init.lua
```diff
@@ -115,6 +134,16 @@ local function tbl_contains(array, value)
   return false
 end
 
+local get_total = function(map)
+  if map ~= nil and type(map) == "table" then
```
hcarter-775, line 138: I think we should leave a debug message for the case where `type(map) == "table"`

drivers/SmartThings/matter-evse/src/init.lua
```diff
@@ -154,13 +183,33 @@ local EVSE_FAULT_STATE_ENUM_MAP = {
...
+  local battery_storage_eps = get_endpoints_for_dt(device, BATTERY_STORAGE_DEVICE_TYPE_ID) or {}
+  local eps_to_read = {}
+  utils.merge(eps_to_read, battery_storage_eps)
+  utils.merge(eps_to_read, solar_power_eps)
+  if #eps_to_read > 0 then
+    local read_req = clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyExported:read(device, eps_to_read[1])
+    for i, ep in ipairs(eps_to_read) do
       if i > 1 then
-        read_req:merge(clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyImported:read(device, ep))
+      read_req:merge( clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyExported:read(device, eps_to_read[i]))
```
hcarter-775, line 212:
```suggestion
      read_req:merge( clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyExported:read(device, ep)
```

drivers/SmartThings/matter-evse/src/init.lua
```diff
@@ -479,61 +555,122 @@ local function device_energy_mgmt_mode_attr_handler(driver, device, ib, response
...
 
-    if tbl_contains(cumul_eps, endpoint_id) then
-      -- Since cluster in this endpoint supports both CUME & PERE features, we will prefer
-      -- cumulative_energy_imported_handler to handle the energy report for this endpoint over PeriodicEnergyImported.
-      return
+local function active_power_handler(driver, device, ib, response)
+  -- Consider only Solar Power / Battery Storage devices and sum up in case there are multiple endpoints.
+  local battery_storage_eps = get_endpoints_for_dt(device, SOLAR_POWER_DEVICE_TYPE_ID)
+  local solar_power_eps = get_endpoints_for_dt(device, BATTERY_STORAGE_DEVICE_TYPE_ID)
+  if (tbl_contains(solar_power_eps, ib.endpoint_id) or tbl_contains(battery_storage_eps, ib.endpoint_id)) and ib.data.value then
```
hcarter-775, line 648: This logic doesn't seem necessary.

drivers/SmartThings/matter-evse/src/init.lua
```diff
@@ -479,61 +555,122 @@ local function device_energy_mgmt_mode_attr_handler(driver, device, ib, response
...
+    local endpoint_id = string.format(ib.endpoint_id)
+    local active_power_map = device:get_field(TOTAL_ACTIVE_POWER) or {}
+    local watt_value = ib.data.value / 1000
+
+    active_power_map[endpoint_id] = watt_value
+    local total_active_power = get_total(active_power_map)
+    device:set_field(TOTAL_ACTIVE_POWER, active_power_map)
```
hcarter-775, line 655: what devices do we expect to support multiple active power measurements?

drivers/SmartThings/matter-evse/src/init.lua
nickolas-deboom: I made a similar comment but realized the eps are sorted in `get_endpoints_for_dt`

drivers/SmartThings/matter-evse/src/init.lua
```diff
@@ -173,14 +222,35 @@ local function create_poll_schedule(device)
...
   device:set_field(RECURRING_POLL_TIMER, timer)
 end
 
+local report_energy_to_app = function(device, comp, energy_map, startTime, endTime)
+  local component = device.profile.components[comp]
+  local total_cumulative_energy = get_total(energy_map) or 0
+
+  -- Calculate the energy consumed between the start and the end time
+  local previousTotalConsumptionWh = device:get_latest_state(comp, capabilities.powerConsumptionReport
+  .ID,
```
nickolas-deboom, line 240:
```suggestion
  local previousTotalConsumptionWh = device:get_latest_state(comp, capabilities.powerConsumptionReport.ID,
```

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,132 @@
...
+
+local function thread_version_attribute_handler(driver, device, ib)
+  local version_name = VERSION_TLV_MAP[ib.data.value]
+  if version_name then
+    device:emit_event_for_endpoint(ib.endpoint, threadVersion.threadVersion({ value = version_name }))
+  end
+end
+
+local function thread_interface_enabled_attribute_handler(driver, device, ib)
+    if ib.data.value then
```
nickolas-deboom, line 75: tiny nit: the spacing is off in some places in this file and in the hrap test files

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -78,20 +87,41 @@ local ROCK_MODE_MAP = {
...
 --   2. if the received setpoint command value is in range 41 ~ 104, it is inferred as *F
 local THERMOSTAT_MAX_TEMP_IN_C = 40.0
 local THERMOSTAT_MIN_TEMP_IN_C = 5.0
 
+-- Water Heater:
+--   1. if the received setpoint command value is in range 30 ~ 80, it is inferred as *C
+--   2. if the received setpoint command value is in range 86 ~ 176, it is inferred as *F
+local WATER_HEATER_MAX_TEMP_IN_C = 80.0
+local WATER_HEATER_MIN_TEMP_IN_C = 30.0
+
```
nickolas-deboom, line 124: Should there also be a temperature range defined for Heat Pump? What would the expected range be for that device type? Currently it would use the 5-40 deg C range for thermostats

hcarter-775: No, a heat pump can just use the thermostat range, since it will be attached to regular thermostats.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1323,18 +1573,126 @@ local function set_rock_mode(driver, device, cmd)
...
 end
 
+local function active_power_handler(driver, device, ib, response)
+  if ib.data.value then
+    local watt_value = ib.data.value / 1000
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.powerMeter.power({ value = watt_value, unit = "W" }))
+  end
+end
+
+local function periodic_energy_imported_handler(driver, device, ib, response)
```
nickolas-deboom, line 1604: Could you create test case(s) that exercise this function?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1323,18 +1573,126 @@ local function set_rock_mode(driver, device, cmd)
...
+    table.insert(supportWaterHeaterModes, mode.elements.label.value)
+  end
+  device:set_field(SUPPORTED_WATER_HEATER_MODES, supportWaterHeaterModes, { persist = true })
+  local event = capabilities.mode.supportedModes(supportWaterHeaterModes, { visibility = { displayed = false } })
+  device:emit_event_for_endpoint(ib.endpoint_id, event)
+  event = capabilities.mode.supportedArguments(supportWaterHeaterModes, { visibility = { displayed = false } })
+  device:emit_event_for_endpoint(ib.endpoint_id, event)
+end
+
+local function water_heater_mode_handler(driver, device, ib, response)
```
nickolas-deboom, line 1675: Could we also create a test case for the water heater mode handler?

drivers/SmartThings/matter-evse/profiles/solar-power.yml
```diff
@@ -0,0 +1,18 @@
...
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: SolarPanel
+- id: exportedEnergy
+  capabilities:
+  - id: powerConsumptionReport
```
hcarter-775, line 17: what is the reasoning for displaying the power Consumption report for this?

drivers/SmartThings/matter-thermostat/profiles/heat-pump.yml
```diff
@@ -0,0 +1,40 @@
...
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+- id: thermostatOne
+  label: Thermostat 1
+  capabilities:
+  - id: thermostatMode
+    version: 1
+  - id: relativeHumidityMeasurement
```
hcarter-775, line 20: I don't think this should be a required trait. I think we should separate humidity into a separate profile that uses a `-humidty` for each thermostat in the profile name.

drivers/SmartThings/matter-evse/fingerprints.yml
```diff
@@ -5,3 +5,13 @@ matterGeneric:
       - id: 0x0510
       - id: 0x050C
     deviceProfileName: evse
+  - id: "matter/solar-power"
```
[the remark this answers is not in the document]
HunsupJung: I think so.

drivers/SmartThings/matter-evse/profiles/solar-power.yml
HunsupJung: `powerConsumptionReport` is the capability used for `Energy service`. In the long run, we wait for this information to be reported to the `Energy service`.

drivers/SmartThings/matter-thermostat/profiles/heat-pump.yml
HunsupJung: Okay I will modify it

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: This appears to not be working on lua libs version 10. The energyMeter capability displays as blank.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1323,18 +1573,126 @@ local function set_rock_mode(driver, device, cmd)
...
+
+local function periodic_energy_imported_handler(driver, device, ib, response)
+  local endpoint_id = ib.endpoint_id
+  local cumul_eps = embedded_cluster_utils.get_endpoints(device,
+    clusters.ElectricalEnergyMeasurement.ID,
+    { feature_bitmap = clusters.ElectricalEnergyMeasurement.types.Feature.CUMULATIVE_ENERGY })
+
+  if ib.data then
+    if version.api < 11 then
+      clusters.ElectricalEnergyMeasurement.attributes.PeriodicEnergyImported:augment_type(ib.data)
```
nickolas-deboom, line 1612: This appears to not be working on lua libs version 10. The energyMeter capability displays as blank.

drivers/SmartThings/matter-thermostat/src/init.lua
HunsupJung: I will change it to the original.

drivers/SmartThings/matter-evse/src/init.lua
HunsupJung: It is described as being included in only one endpoint when using EnergyImported in the specification. First of all, it would be nice to assume that there is only one and add it when there is a device that supports multiple EnergyImported.

drivers/SmartThings/matter-evse/src/init.lua
HunsupJung: Solar Power device type

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1362,12 +1604,118 @@ local function set_rock_mode(driver, device, cmd)
   device:send(clusters.FanControl.attributes.RockSetting:write(device, device:component_to_endpoint(cmd.component), rock_mode))
 end
 
+local function set_water_heater_mode(driver, device, cmd)
+  device.log.info_with({ hub_logs = true },
```
ctowns, line 1608: We shouldn't need this in the hub logs, but it would be fine to keep in the device logs if this would be useful beyond just this initial development phase.

ctowns, line 1608:
```suggestion
  device.log.info(,
```
We shouldn't need this in the hub logs, and we could probably remove this log altogether. But it would be fine to keep in just the device logs (remove the `hub_logs = true` part) if this would be useful beyond just this initial development phase.

hcarter-775:
closing this PR, now that [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1968) was merged

PR #1874, opened 2025-01-15
Conditionally override child profiles

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

Devices that utilize the child_device_profile_overrides table may contain other endpoints that should not have their profile overridden when the child device is created. This PR ensures that the profile is only overridden for profiles listed in  child_device_profile_overrides. See the discussion from [PR 1872](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1872/files#r1915492561).

Note that `GENERIC_SWITCH_ID` was removed from `device_type_profile_map`. This is because it is never used in this table and also to make it clear that device types in this table may have their child device profile overridden (buttons do not use parent-child).

# Summary of Completed Tests

Unit test update to verify child device profile override works as intended.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -79,7 +79,6 @@ local device_type_profile_map = {
   [ON_OFF_SWITCH_ID] = "switch-binary",
   [ON_OFF_DIMMER_SWITCH_ID] = "switch-level",
   [ON_OFF_COLOR_DIMMER_SWITCH_ID] = "switch-color-level",
-  [GENERIC_SWITCH_ID] = "button"
 }
 
 local device_type_attribute_map = {
@@ -436,22 +435,6 @@ end
 local function assign_child_profile(device, child_ep)
   local profile
 
-  -- check if device has an overridden child profile that differs from the profile
-  -- that would match the child's device type
-  for _, fingerprint in ipairs(child_device_profile_overrides) do
-    if device.manufacturer_info.vendor_id == fingerprint.vendor_id and
-       device.manufacturer_info.product_id == fingerprint.product_id then
-      if device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID then
-        if child_ep ~= 1 then
-          -- To add Electrical Sensor only to the first EDGE_CHILD(light-power-energy-powerConsumption)
-          -- The profile of the second EDGE_CHILD is determined in the "for" loop below (e.g., light-binary)
-          break
-        end
-      end
-      return fingerprint.child_profile
-    end
-  end
-
   for _, ep in ipairs(device.endpoints) do
     if ep.endpoint_id == child_ep then
       -- Some devices report multiple device types which are a subset of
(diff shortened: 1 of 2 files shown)


hcarter-775:
can we make a unit test or two for this? seems like a worthwhile thing to make sure it doesn't break

nickolas-deboom:
> can we make a unit test or two for this? seems like a worthwhile thing to make sure it doesn't break

Yeah good call - I updated a mock device in `test_multi_switch_parent_child_lights.lua` to use an overridden profile for one of its child devices. Also note that I made a mistake in the initial commit, now it only overrides the profile if the profile is going to be "plug-binary", rather than any profile in `device_type_profile_map`. This is the only situation that we'd want to do this at this point.

hcarter-775 (approved):
good work!

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -452,8 +435,26 @@ local function assign_child_profile(device, child_ep)
...
     end
   end
+
+  -- Check if device has an overridden child profile that differs from the profile that would match
+  -- the child's device type for the following two cases:
+  --   1. The selected profile for the child device is plug-binary, or
+  --   2. To add Electrical Sensor only to the first EDGE_CHILD (light-power-energy-powerConsumption)
+  --      for the Aqara Light Switch H2. The profile of the second EDGE_CHILD for this device is
+  --      determined in the "for" loop above (e.g., light-binary)
+  if profile == "plug-binary" or (device.manufacturer_info.vendor_id == AQARA_MANUFACTURER_ID and child_ep == 1) then
```
ctowns, line 448: Currently this just allows us to override child devices with `plug` device types to a switch profile. Right now the only use case we see in our `child_device_profile_overrides` is forcing a device to use the `switch-binary` device type, but in the future if we need to override the other way around (i.e. force a `switch-binary` to a `plug-binary` type, we'd need to add an additional check for `profile == switch-binary` here.

Ideally this is resolved by the new mounted device types, but just something to call out. I don't have a strong opinion on if the change should be added now because any change to include an override in the future that would require the additional check for `switch-binary` -> `plug-binary` would already require a driver change to  update the override list, so it would go beyond just a fingerprint change anyway. I'm not sure if you know of any devices that might require a switch -> plug override?

nickolas-deboom: This is a good point, and I took a stab at making the logic more generic in a5720821fa5addc2f31a8a57532e2887060e4da4 so that we can do this in the future. I added an 'initial_profile' and a 'target_profile' field to `child_device_profile_overrides` so that we can check any type of profile rather than just "plug-binary". Let me know if that looks alright to you!

ctowns: This looks marvelous, great work Nick!

hcarter-775: yes this is a good idea, glad it got worked on. I had elected it to be "good enough" for now but I like the new system

PR #1904, opened 2025-01-28
Matter Thermostat: Update system_mode_handler

hcarter-775:
# Description of Change
 The ControlSequenceOfOperations attribute handling defines which operations are disallowed for the thermostat. Per the spec: 
- “This attribute SHALL indicate the overall operating environment of the thermostat, and thus the possible system modes that the thermostat can operate in.”

From a spec perspective, the different values in ControlSequenceOfOperationsEnum only specify what is disallowed. Everything else is optional.

The SystemMode attribute, which is handled with system_mode_handler should be limited by this "disallow" list . Per the spec:
- “This attribute SHALL indicate the current operating mode of the thermostat. Its value SHALL be limited by the ControlSequenceOfOperation attribute.”

We currently update the supported modes with whatever is sent, without regard to the ControlSequenceOfOperations. This change will enforce the spec more closely by disallowing certain values, rather than expecting the devices to do this of their own accord.

Lastly, this change makes a small change that puts the AUTO mode on the disallow list as well if that feature bit is not supported.

# Summary of Completed Tests
Unit tests were updated to handle this, ensuring that allowed mode changes could pass the disallow list check, and a new test case was added to ensure the disallow list was correctly disallowing what should not be permitted.
VDA Device and a Physical Thermostat device were tested to ensure systems continue to function as expected.

[MTR-878]: https://smartthings.atlassian.net/browse/MTR-878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -42,16 +42,18 @@ if version.api < 10 then
   clusters.Thermostat.types.ThermostatSystemMode.SLEEP = 0x9
 end
 
+local DISALLOWED_THERMOSTAT_MODES = "__DISALLOWED_CONTROL_OPERATIONS"
+
 local THERMOSTAT_MODE_MAP = {
-  [clusters.Thermostat.types.ThermostatSystemMode.OFF]            = capabilities.thermostatMode.thermostatMode.off,
-  [clusters.Thermostat.types.ThermostatSystemMode.AUTO]           = capabilities.thermostatMode.thermostatMode.auto,
-  [clusters.Thermostat.types.ThermostatSystemMode.COOL]           = capabilities.thermostatMode.thermostatMode.cool,
-  [clusters.Thermostat.types.ThermostatSystemMode.HEAT]           = capabilities.thermostatMode.thermostatMode.heat,
+  [clusters.Thermostat.types.ThermostatSystemMode.OFF]               = capabilities.thermostatMode.thermostatMode.off,
+  [clusters.Thermostat.types.ThermostatSystemMode.AUTO]              = capabilities.thermostatMode.thermostatMode.auto,
+  [clusters.Thermostat.types.ThermostatSystemMode.COOL]              = capabilities.thermostatMode.thermostatMode.cool,
+  [clusters.Thermostat.types.ThermostatSystemMode.HEAT]              = capabilities.thermostatMode.thermostatMode.heat,
   [clusters.Thermostat.types.ThermostatSystemMode.EMERGENCY_HEATING] = capabilities.thermostatMode.thermostatMode.emergency_heat,
-  [clusters.Thermostat.types.ThermostatSystemMode.PRECOOLING]     = capabilities.thermostatMode.thermostatMode.precooling,
-  [clusters.Thermostat.types.ThermostatSystemMode.FAN_ONLY]       = capabilities.thermostatMode.thermostatMode.fanonly,
-  [clusters.Thermostat.types.ThermostatSystemMode.DRY]            = capabilities.thermostatMode.thermostatMode.dryair,
-  [clusters.Thermostat.types.ThermostatSystemMode.SLEEP]          = capabilities.thermostatMode.thermostatMode.asleep
+  [clusters.Thermostat.types.ThermostatSystemMode.PRECOOLING]        = capabilities.thermostatMode.thermostatMode.precooling,
+  [clusters.Thermostat.types.ThermostatSystemMode.FAN_ONLY]          = capabilities.thermostatMode.thermostatMode.fanonly,
+  [clusters.Thermostat.types.ThermostatSystemMode.DRY]               = capabilities.thermostatMode.thermostatMode.dryair,
+  [clusters.Thermostat.types.ThermostatSystemMode.SLEEP]             = capabilities.thermostatMode.thermostatMode.asleep,
 }
 
 local THERMOSTAT_OPERATING_MODE_MAP = {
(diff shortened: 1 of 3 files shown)


ldeora:
Long overdue. It was possible to expand the list of supported modes by sending an unsupported mode. A device with the profile `heating-only` and the supported modes off, auto, heat could become a device that supports cooling.

nickolas-deboom:
LGTM!

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -788,21 +788,13 @@ local function humidity_attr_handler(driver, device, ib, response)
...
-    device:emit_event_for_endpoint(ib.endpoint_id, THERMOSTAT_MODE_MAP[ib.data.value]())
-    local supported_modes = device:get_latest_state(device:endpoint_to_component(ib.endpoint_id), capabilities.thermostatMode.ID, capabilities.thermostatMode.supportedThermostatModes.NAME) or {}
-    -- TODO: remove -- this has been fixed upstream
-    local sm = utils.deep_copy(supported_modes)
-    -- if we get a mode report from the thermostat that isn't in the supported modes, then we need to update the supported modes
-    for _, mode in ipairs(supported_modes) do
-      if mode == THERMOSTAT_MODE_MAP[ib.data.value].NAME then
-        return
-      end
+  local supported_modes = device:get_latest_state(device:endpoint_to_component(ib.endpoint_id), capabilities.thermostatMode.ID, capabilities.thermostatMode.supportedThermostatModes.NAME) or {}
```
ctowns, line 791: This makes sense to me! Do we know why it was done the other way previously? It might be worth knowing if the previous method was meant to address an edge case somehow. It looks like that implementation had been around since the creation of this driver, so I wonder if something changed in the spec to add/enforce the `ControlSequenceOfOperation` attribute.

Overall, I think this new method is a great improvement, I think it'd be good to understand the reasoning for the previous handler before making the changes.

hcarter-775: yes, this is a good point, and has led me into a deeper, more complete re-read of the spec. Therefore, I've re-configured much of the logic to include what the original code was attempting to do (though it was too permissive).

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -788,22 +790,26 @@ local function humidity_attr_handler(driver, device, ib, response)
...
-    table.insert(sm, THERMOSTAT_MODE_MAP[ib.data.value].NAME)
-    local event = capabilities.thermostatMode.supportedThermostatModes(sm, {visibility = {displayed = false}})
-    device:emit_event_for_endpoint(ib.endpoint_id, event)
   end
+  -- if the value is not found in the supported modes list, check if it's disallowed
+  for _, mode in pairs(device:get_field(DISALLOWED_THERMOSTAT_MODES)) do
+    if mode == ib.data.value then
+      return
+    end
+  end
```
nickolas-deboom, line 806:
```suggestion
  local disallowed_thermostat_modes = device:get_field(DISALLOWED_THERMOSTAT_MODES)
  if disallowed_thermostat_modes then
    for _, mode in pairs(disallowed_thermostat_modes) do
      if mode == ib.data.value then
        return
      end
    end
  end
```
If SystemMode is received before ControlSequenceOfOperation then DISALLOWED_THERMOSTAT_MODES wouldn't be populated yet which causes an error here

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -817,25 +823,34 @@ local function running_state_handler(driver, device, ib, response)
...
 
-  local auto = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.AUTOMODE})
-  if #auto > 0 then
-    table.insert(supported_modes, capabilities.thermostatMode.thermostatMode.auto.NAME)
-  end
-
   if ib.data.value <= clusters.Thermostat.attributes.ControlSequenceOfOperation.COOLING_WITH_REHEAT then
     table.insert(supported_modes, capabilities.thermostatMode.thermostatMode.cool.NAME)
-    -- table.insert(supported_modes, capabilities.thermostatMode.thermostatMode.precooling.NAME)
+    table.insert(disallowed_mode_operations, clusters.Thermostat.types.ThermostatSystemMode.HEAT)
```
nickolas-deboom, line 834: Can `ControlSequenceOfOperation` change? It doesn't really seem like it would, but I noticed it's a writable attribute, and while our driver doesn't write to it, another source potentially could. If it does change after this handler runs the first time, `disallowed_mode_operations` would contain the wrong values

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: yes, that's a good point

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: So per spec:

> If an attempt is made to write to this attribute, the server SHALL silently ignore the write
and the value of this attribute SHALL remain unchanged. This behavior is in place for backwards
compatibility with existing thermostats.

However, even if it was written to and we were given an update, the `disallowed_mode_operations` would be reset in the handler since it would always run like so:
```
local disallowed_mode_operations = {}

... inserts to the list based on the enum value ...

device:set_field(DISALLOWED_THERMOSTAT_MODES, disallowed_mode_operations)
```
so `DISALLOWED_THERMOSTAT_MODES` will always have the disallowed values from the most recent read.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -788,22 +790,27 @@ local function humidity_attr_handler(driver, device, ib, response)
...
+      return
     end
-    -- if we get here, then the reported mode was not in our mode map
-    table.insert(sm, THERMOSTAT_MODE_MAP[ib.data.value].NAME)
-    local event = capabilities.thermostatMode.supportedThermostatModes(sm, {visibility = {displayed = false}})
-    device:emit_event_for_endpoint(ib.endpoint_id, event)
   end
+  -- if the value is not found in the supported modes list, check if it's disallowed
+  local disallowed_thermostat_modes = device:get_field(DISALLOWED_THERMOSTAT_MODES)
+  for _, mode in pairs(device:get_field(disallowed_thermostat_modes)) do
```
nickolas-deboom, line 803:
```suggestion
  local disallowed_thermostat_modes = device:get_field(DISALLOWED_THERMOSTAT_MODES) or {}
  for _, mode in pairs(disallowed_thermostat_modes) do
```
I think this is what you meant to do?

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Ah, perfect!

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: yes, thanks. This is what I get for making the change at like 9pm

nickolas-deboom:
The updates you made look good to me :+1:

tpmanley (approved):

ctowns (approved):
Great work @hcarter-775 !

PR #1862, opened 2025-01-10
Add batteryLevel to matter switch

nickolas-deboom:
# Type of Change

- [ ] WWST Certification Request
     - If this is your first time contributing code:
          - [ ] I have reviewed the README.md file
          - [ ] I have reviewed the CODE_OF_CONDUCT.md file
          - [ ] I have signed the CLA
     - [ ] I plan on entering a WWST Certification Request or have entered a request through the WWST Certification console at developer.smartthings.com
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code in hard-to-understand areas
- [x] I have verified my changes by testing with a device or have communicated a plan for testing
- [x] I am adding new behavior, such as adding a sub-driver, and have added and run new unit tests to cover the new behavior

# Description of Change

[CHAD-12161](https://smartthings.atlassian.net/browse/CHAD-12161)

Matter drivers currently assume that `BatPercentRemaining` is available if the `BAT` feature is supported, but this attribute is only optionally required if this feature is present. The changes in this PR improve the profile selection logic in matter-switch by reading the `AttributeList`, checking if `BatPercentRemaining` or `BatChargeLevel` is available, and then profiling the device as needed.

Note that these changes were originally in [PR 1796](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1796) but this PR was split up into 5 separate PRs, one for each affected driver.

# Summary of Completed Tests

Tested with a matter button, which supports `BatPercentRemaining`. Relevant logs:
```
2025-01-30_16:08:45.116 | I | broker     | <Matter Switch (c8775dfd)> <MatterDevice: dd1a6065-2320-4d46-9651-d249e2045b83 [7AE8C08425F969BF-6AB3AA3F0FB2BDDC-11] (Matter Button)> received lifecycle event: added
2025-01-30_16:08:45.161 | I | broker     | <Matter Switch (c8775dfd)> <MatterDevice: dd1a6065-2320-4d46-9651-d249e2045b83 [7AE8C08425F969BF-6AB3AA3F0FB2BDDC-11] (Matter Button)> sending InteractionRequest: <InteractionRequest || type: READ, info_blocks: [<InteractionInfoBlock || cluster: PowerSource, attribute: AttributeList>]>
[...]

PR #1906, opened 2025-01-29
Refactor Matter Switch Initialization

hcarter-775:
# Description of Change
This change is an attempt to condense and clarify some of the main functionality of the switch-button device initialization. This includes clarifying logic blocks by moving them into helper functions and integrating some low hanging optimizations.

# Summary of Completed Tests
- Unit tests continue to pass
- A physical 3-button parent-child switch device correctly onboarded 
- A physical 7-button device correctly onboarded

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -370,10 +370,10 @@ local function mired_to_kelvin(value, minOrMax)
   end
 end
 
---- is_supported_combination_button_switch_device_type helper function used to check
+--- device_type_supports_button_switch_combination helper function used to check
 --- whether the device type for an endpoint is currently supported by a profile for
 --- combination button/switch devices.
-local function is_supported_combination_button_switch_device_type(device, endpoint_id)
+local function device_type_supports_button_switch_combination(device, endpoint_id)
   for _, ep in ipairs(device.endpoints) do
     if ep.endpoint_id == endpoint_id then
       for _, dt in ipairs(ep.device_types) do
@@ -395,23 +395,27 @@ local function get_first_non_zero_endpoint(endpoints)
   return nil
 end
 
---- find_default_endpoint helper function to handle situations where
+--- find_default_endpoint is a helper function to handle situations where
 --- device does not have endpoint ids in sequential order from 1
---- In this case the function returns the lowest endpoint value that isn't 0
---- and supports the OnOff or Switch cluster. This is done to bypass the
---- BRIDGED_NODE_DEVICE_TYPE on bridged devices.
 local function find_default_endpoint(device)
+  local temperature_eps = device:get_endpoints(clusters.TemperatureMeasurement.ID)
+  local humidity_eps = device:get_endpoints(clusters.RelativeHumidityMeasurement.ID)
+  if #temperature_eps > 0 and #humidity_eps > 0 then
+    -- In case of Aqara Climate Sensor W100, in order to sequentially set the button name to button 1, 2, 3
+    return device.MATTER_DEFAULT_ENDPOINT
+  end
+
   local switch_eps = device:get_endpoints(clusters.OnOff.ID)
   local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
   table.sort(switch_eps)
   table.sort(button_eps)
 
-  -- Return the first switch endpoint as the default endpoint if no button endpoints are available
+  -- Return the first switch endpoint as the default endpoint if no button endpoints are present
   if #button_eps == 0 and #switch_eps > 0 then
     return get_first_non_zero_endpoint(switch_eps)
   end
 
-  -- Return the first button endpoint as the default endpoint if no switch endpoints are available
+  -- Return the first button endpoint as the default endpoint if no switch endpoints are present
   if #switch_eps == 0 and #button_eps > 0 then
     return get_first_non_zero_endpoint(button_eps)
   end
@@ -421,7 +425,7 @@ local function find_default_endpoint(device)
   -- default endpoint.
   if #switch_eps > 0 and #button_eps > 0 then
     local main_endpoint = get_first_non_zero_endpoint(switch_eps)
-    if is_supported_combination_button_switch_device_type(device, main_endpoint) then
+    if device_type_supports_button_switch_combination(device, main_endpoint) then
       return main_endpoint
     else
       device.log.warn("The main switch endpoint does not contain a supported device type for a component configuration with buttons")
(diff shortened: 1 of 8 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -706,40 +624,23 @@ local function detect_bridge(device)
...
-      device:set_find_child(find_child)
-    end
-    local main_endpoint = find_default_endpoint(device)
-    for _, ep in ipairs(device.endpoints) do
-      if ep.endpoint_id ~= main_endpoint then
-        local id = 0
-        for _, dt in ipairs(ep.device_types) do
-          id = math.max(id, dt.device_type_id)
-        end
-        for _, attr in pairs(device_type_attribute_map[id] or {}) do
```
nickolas-deboom, line 732: This is needed to ensure that we subscribe to attributes that are specific to child devices. See https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1548

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1210,9 +1111,6 @@ local function device_added(driver, device)
   if device:supports_capability(capabilities.energyMeter) then
     device:emit_event(capabilities.energyMeter.energy({ value = 0.0, unit = "Wh" }))
   end
-
-  -- call device init in case init is not called after added due to device caching
-  device_init(driver, device)
```
nickolas-deboom, line 1215: Are we sure this can be removed? It was added somewhat recently in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1445/files

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: yeah, I figured that out ... I played around with another way of doing it, but the scope grew a little outside of this simple PR, so it's been re-added.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Hmm, maybe it can't? It doesn't exist in any other drivers so I'd assumed it was an outdated addition. @ctowns I see this is your commit; why would this be included here but nowhere else?

hcarter-775: Afaik though, device_init will always be called after init. It's written into the rust-side of things like that.

ctowns: Hey @hcarter-775, the reason `init` is called twice is to make the driver more robust overall by ensuring that the init behavior is called directly from the device added lifecycle event. This essentially means that even if an init event is missing, the driver’s init handler will still be called by the added handler and it will still run the necessary init functions (most importantly, setting up subscriptions). This can result in the device_init function being called twice, once on added and once again if the init event is handled, but this is not an issue and the function is safe to call twice. 

There was an issue where a device init event was sometimes not firing due to a race condition in the scripting engine, but this change was an attempt to make the driver more robust to that issue. The `init` function should be safe to call multiple times, and the handling in `initialize_switch` is gated behind the `SWITCH_INITIALIZED` flag so it is only run once.

hcarter-775: so i guess, in a perfect world this should be included in all matter drivers?

hcarter-775: ah I see, this is mostly directed to the child devices being created in the switch driver case in particular.

ctowns: Exactly. The root cause is due to the `find_child` function being used, so it would only need to be in drivers that use child devices, and in this case that is only the matter-switch driver (now that button has been merged with switch).

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -498,91 +502,90 @@ local function do_configure(driver, device)
...
+        local button_component = "button" .. component_num
+        component_map[button_component] = ep
       end
-      current_component_number = current_component_number + 1
     end
     component_map_used = true
   end
 
+  if component_map_used then
+    device:set_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON, component_map, {persist = true})
```
nickolas-deboom, line 558: You could move this inside the previous if statement and get rid of the `component_map_used` variable

hcarter-775: good idea

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -608,88 +605,62 @@ local function initialize_switch(driver, device)
...
-    -- on each driver init.
     device:set_field(IS_PARENT_CHILD_DEVICE, true, {persist = true})
   end
 
   device:set_field(SWITCH_INITIALIZED, true)
 
-  if component_map_used then
-    device:set_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON, component_map, {persist = true})
-  end
+  -- this is needed in switch_initialize
```
nickolas-deboom, line 616: nit - function name
```suggestion
  -- this is needed in initialize_switch
```

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -723,40 +694,39 @@ local function detect_bridge(device)
...
+  end
+  -- ensure subscription to all endpoint attributes- including those mapped to child devices
+  for _, ep in ipairs(device.endpoints) do
+    if ep.endpoint_id ~= main_endpoint then
+      local id = 0
+      for _, dt in ipairs(ep.device_types) do
+        id = math.max(id, dt.device_type_id)
+      end
+      for _, attr in pairs(device_type_attribute_map[id] or {}) do
+        if id == GENERIC_SWITCH_ID and attr ~= clusters.PowerSource.attributes.BatPercentRemaining then
```
nickolas-deboom, line 721:
```suggestion
        if id == GENERIC_SWITCH_ID and attr ~= clusters.PowerSource.attributes.BatPercentRemaining 
           and attr ~= clusters.PowerSource.attributes.BatChargeLevel then
```
This should have been included in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1862, could you add it here?

nickolas-deboom:
I left some comments but have finished reviewing now and the changes look great! It would also be good to do some regression testing with a few different device types (such as a switch or light, a button, and a switch+button combo). I have some devices you could test out if you'd like

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: yeah no problem

hcarter-775:
> I left some comments but have finished reviewing now and the changes look great! It would also be good to do some regression testing with a few different device types (such as a switch or light, a button, and a switch+button combo). I have some devices you could test out if you'd like

Good idea; see the Completed Tests section of the PR description to see an updated list of tests on 2 physical devices. This should handle the cases that these changes touch.

nickolas-deboom (approved):

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -395,23 +395,27 @@ local function get_first_non_zero_endpoint(endpoints)
...
---- find_default_endpoint helper function to handle situations where
+--- find_default_endpoint is a helper function to handle situations where
 --- device does not have endpoint ids in sequential order from 1
---- In this case the function returns the lowest endpoint value that isn't 0
---- and supports the OnOff or Switch cluster. This is done to bypass the
---- BRIDGED_NODE_DEVICE_TYPE on bridged devices.
 local function find_default_endpoint(device)
+  local temperature_eps = device:get_endpoints(clusters.TemperatureMeasurement.ID)
+  local humidity_eps = device:get_endpoints(clusters.RelativeHumidityMeasurement.ID)
+  if #temperature_eps > 0 and #humidity_eps > 0 then
```
ctowns, line 403: Is this custom code that is only meant to be used for the Aqara Climate sensor? If that is the case, it should be gated on the manufacturer info rather than on the number of temp and humidity endpoints. I believe this would set the return the wrong endpoint for all devices that support humidity and temperature, which could be problematic for bridged devices

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -608,88 +605,62 @@ local function initialize_switch(driver, device)
...
-      if battery_support then
-        local attribute_list_read = im.InteractionRequest(im.InteractionRequest.RequestType.READ, {})
-        attribute_list_read:merge(clusters.PowerSource.attributes.AttributeList:read())
-        device:send(attribute_list_read)
-      else
-        if #temperature_eps > 0 and #humidity_eps > 0 then
-          -- for now, this logic only applies to the Aqara Climate Sensor W100.
-          profile_name = "-temperature-humidity"
-        end
+local function initialize_buttons_and_switches(driver, device, main_endpoint)
```
ctowns, line 640: nit: A handful of the new helper functions used here use the `try_build_<thing>` format. However, there is no return or any indication if the attempt was successful. For clarity, I would just remove the `try` part unless we plan to handle failure cases (which based on previous functionality, I would assume we don't because they are unlikely)

drivers/SmartThings/matter-switch/src/test/test_matter_button.lua
```diff
@@ -319,12 +319,12 @@ test.register_message_test(
     message = {
       mock_device.id,
       clusters.Switch.events.InitialPress:build_test_event_report(
-        mock_device, 1, {new_position = 1}
+        mock_device, 1, {new_position = 1, total_number_of_presses_counted = 1, previous_position = 0}
```
ctowns, line 322: Im not sure I understand why this test condition changed? What is the reason that this behavior changed?

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button.lua
```diff
@@ -106,11 +106,11 @@ local function test_init()
     if i > 1 then subscribe_request:merge(clus:subscribe(mock_device)) end
   end
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
-  test.socket.matter:__expect_send({mock_device.id, subscribe_request})
   test.mock_device.add_test_device(mock_device)
   test.socket.device_lifecycle:__queue_receive({ mock_device.id, "added" })
   local read_attribute_list = clusters.PowerSource.attributes.AttributeList:read()
   test.socket.matter:__expect_send({mock_device.id, read_attribute_list})
+  test.socket.matter:__expect_send({mock_device.id, subscribe_request})
```
ctowns, line 113: Why have these subscribe requests been moved in some of these test files? Is it for aesthetic reasons related to relaxing the channel ordering, or did the behavior of these test cases actually change? Just want to be sure because I don't see a reason why the subscribe behavior should have changed with the refactor.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -608,88 +605,62 @@ local function initialize_switch(driver, device)
...
+  table.sort(button_eps)
 
-        if #button_eps == 1 then
-          profile_name = "button" .. profile_name
-        elseif #button_eps > 1 then
-          profile_name = string.format("%d-button", #button_eps) .. profile_name
-        end
+  -- All button endpoints found will be added as additional components in the profile containing the main_endpoint.
+  -- The resulting endpoint to component map is saved in the COMPONENT_TO_ENDPOINT_MAP_BUTTON field
+  try_build_button_component_map(device, main_endpoint, button_eps)
```
ctowns, line 648: To me personally, it makes it clearer to still have a division between button and switch handilng in this function. For example, we wouldn't need to call this function at all unless `button_eps > 0`, which we already check below on line 654.

Same idea goes for the switch counterpart below.

drivers/SmartThings/matter-switch/src/test/test_matter_button.lua
hcarter-775: This creates the same final result- I had been testing and this was accidentally added, even though it was an unnecessary change. We can remove it.

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button.lua
hcarter-775: All of these tests were edited to more closely resemble the actual logical flow of init. If they were reverted, the tests would still pass

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: agreed :+1:

PR #1577, opened 2024-08-13
CHAD-13781:  Re-profile devices with a bad wind feature map

hcarter-775:
Check all that apply

# Type of Change

- [ ] WWST Certification Request
     - If this is your first time contributing code:
          - [ ] I have reviewed the README.md file
          - [ ] I have reviewed the CODE_OF_CONDUCT.md file
          - [ ] I have signed the CLA
     - [ ] I plan on entering a WWST Certification Request or have entered a request through the WWST Certification console at developer.smartthings.com
- [X] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

- [X] I have performed a self-review of my code
- [X] I have commented my code in hard-to-understand areas
- [ ] I have verified my changes by testing with a device or have communicated a plan for testing
- [X] I am adding new behavior, such as adding a sub-driver, and have added and run new unit tests to cover the new behavior

# Description of Change
[CHAD-13761](https://smartthings.atlassian.net/browse/CHAD-13781) Re-profile devices with a bad wind feature map
See description in this Jira ticket for more information.

# Summary of Completed Tests
Unit test written to catch this case

[CHAD-13761]: https://smartthings.atlassian.net/browse/CHAD-13761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1916, opened 2025-02-07
Include support for a bad wind support mode mapping

hcarter-775:
# Description of Change
There are some devices that claim to support the Wind feature, but whose WindSupport attribute is empty. This update will force a read of that attribute before profiling can occur in the case that the Wind feature is supported to ensure that at least one value is specified. If no value is specified, wind support is not included in the resulting profile.

There are a few paths this logic can follow, so here is the reasoning for each path:
1. `do_configure` runs first -> checks if battery is supported
2. if `battery is supported` -> we read the PowerSource AttributeList.
  a. In `attribute_list_handler` -> `match_profile` runs with the specified battery type
  b. In `match_profile` -> if the wind feature is not supported, a profile is chosen.
  c. Else -> the battery type specified is saved. we read FanControl WindSupport. `match_profile` returns early
    1. In `wind_support_handler` ->  we count the supported modes. a capability event is emitted. `match_profile` runs again.
    2. In `match_profile` -> a profile is chosen.
3. if `battery is not supported` -> `match_profile` runs directly from `do_configure`.
    a. See `b. c. i. and ii.` above for the following logic.

Re-formation of the previous [PR 1577](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1577) 

# Summary of Completed Tests
- Unit tests written to match the failing case, and previous tests were re-written to handle the new profiling system.
- Tested on-device with an empty WindSupport attribute, and it correctly profiled not to include it


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -446,6 +449,19 @@ local function create_thermostat_modes_profile(device)
 end
 
 local function match_profile(driver, device, battery_supported)
+  if battery_supported == nil then
+    return
```
nickolas-deboom, line 453: We might want a log statement here, since it looks like this could only happen if a new call to `match_profile` was added that didn't include `battery_supported`, right?

hcarter-775: Yes, that's true, and for now, that case doesn't actually exist- it did when I was first creating this logic but I buttoned that up. It should definitely include a log here for that case though.

hcarter-775: Actually, in commit [fb42613](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1916/commits/fb426139479942e9eb60e600688a41a1dc6f59f5), I just removed this logic since it's not possible. I also updated the handling a little to make it more bulletproof.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -446,6 +449,16 @@ local function create_thermostat_modes_profile(device)
 end
 
 local function match_profile(driver, device, battery_supported)
+  -- read WindSupport before profiling if the FanControl Wind Feature Flag is supported
+  local wind_eps = device:get_endpoints(clusters.FanControl.ID, { feature_bitmap = clusters.FanControl.types.FanControlFeature.WIND })
+  if #wind_eps > 0 and device:get_field(WIND_MODE_COUNT) == nil then
+    device:set_field(BATTERY_SUPPORT, battery_supported, {persist = true}) -- save value for after WindSupport read
+    local req = im.InteractionRequest(im.InteractionRequest.RequestType.READ, {})
```
ctowns, line 456: This could be simplified
```suggestion
    device:send(clusters.FanControl.attributes.WindSupport:read())
```

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1034,6 +1046,15 @@ local function wind_support_handler(driver, device, ib, response)
   end
   local event = capabilities.windMode.supportedWindModes(supported_wind_modes, {visibility = {displayed = false}})
   device:emit_event_for_endpoint(ib.endpoint_id, event)
+
+  -- save the number of supported wind modes for use in match_profile.
+  device:set_field(WIND_MODE_COUNT, #supported_wind_modes, {persist = true})
```
ctowns, line 1051: This flag is just checking if this value has been used, right? I don't see where the number of wind supported modes is used, it looks like the only check is that there are wind supported modes or not (i.e. modes > 1), so this could be a boolean. However, since this flag is really just tracking if there are supported wind modes or not, I believe we could rely on `device:get_latest_state()` to get the latest capability state for this field. You should also be able to get a count of modes from the capability state too if needed, but I think just checking if the mode has been set or not should be enough for the use case here!

hcarter-775: @ctowns with using `get_latest_state` (which I always forget is an option) do we have to worry about latency issues? As in, `match_profile` will be running milliseconds after the `emit_event_for_endpoint` call in this handler. Can I assume that the capability is set immediately or is there any cloud part of this that may cause it not to be set when it's needed right away?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -446,6 +449,16 @@ local function create_thermostat_modes_profile(device)
 end
 
 local function match_profile(driver, device, battery_supported)
+  -- read WindSupport before profiling if the FanControl Wind Feature Flag is supported
+  local wind_eps = device:get_endpoints(clusters.FanControl.ID, { feature_bitmap = clusters.FanControl.types.FanControlFeature.WIND })
+  if #wind_eps > 0 and device:get_field(WIND_MODE_COUNT) == nil then
+    device:set_field(BATTERY_SUPPORT, battery_supported, {persist = true}) -- save value for after WindSupport read
```
ctowns, line 455: I think it might be easier in the `match_profile` handle to make it so that if `battery_supported` is not provided by the caller, then it is assumed that the current battery support will be maintained. This way, we do not have to start saving battery_support and other capability flags on the device object so that other handlers can call `match_profile` correctly. It will remove that dependency for the caller to know the battery supported value ahead of time. This can be done with something like this:

```
if battery_supported == nil then
   -- continue using existing battery support
    if device:supports_capability(capabilities.battery)
        battery_support = "battery"
   else if device:supports_capability(capabilities.batteryLevel)
       battery_support = "batteryLevel"
   else
       battery_support = None or whatever value is appropriate for None
end
```

This removes the need for the caller to know what the battery support is, which will simplify the driver and prevent us from needing to cache values in the the device object in the event that we need to create other dynamic handlers like the battery and windmode handlers

hcarter-775: but in match_profile, we haven't necessarily profiled the device yet with the correct capabilities, or am I misunderstanding you?

hcarter-775: So I think I see what you're saying and I agree that what we have is not an ideal solution. However, I don't think the solution you've proposed here will work correctly in all cases as things are now. Since at the moment, we pass battery type into match profile through do_configure. Normally this is fine- however, with this introduction, the wind mode stalls the match_profile call since we need to do that read before proceeding. However, at that point, if we don't save this battery value, we lose it. I can envision setting the field to null after it's used to cut down device cached data if that's the issue- I actually think that's a good idea. But I don't think we can cut out the caching altogether as things stand.

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: Great question! The state cache is updated when the event is sent, not when it is received, so there is no dependency on latency from the cloud. You can see this in the source code for the `Device:emit_component_event()` function in our lua libs.

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: I agree, I did not see the early return in the `match_profile` function, so I did not realize the `match_profile` function was being shortcut by the wind mode handling. Could we send the read for windModes in `doConfigure` like we do for battery support? Then, we could update the handlers to both call `match_profile`, and while this may result in two profile switches being made with `try_update_metadata`, that would only occur for devices that claim to support the Wind feature, but whose WindSupport attribute is empty. I don't believe there is a limit on device profile updates so calling try_update_metadata` should not be problematic, especially if the second time is just the same profile.

hcarter-775: The reason I didn't go that route is that in the bad wind support case, calling `try_update_metadata` caused a very strange (looking like things were broken) user onboarding experience. I don't know the details, but I _do_ think there's a reason we should avoid calling `try_update_metadata` twice in such a short period of time.

hcarter-775: Further, the specific implementation you suggested wouldn't work in this case because we couldn't guarantee ordering of which read comes first- windsupport or powerSource attributeList.

nickolas-deboom, line 455: Do we need to persist this field? Looks like we only need it until the wind support handler runs

hcarter-775: No, this doesn't need to be persisted. Also, I may be getting rid of the field altogether per https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1916/files#r1971851770

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: Something I just thought of while going to implement this- I'm not sure that information is actually emitted in all these cases, or is it? Since `emit_component_event` expects the profile to support the capability- but in this case we're dynamically profiling, so there's no guarantee that the capability is in the profile.

hcarter-775:
Will not fix: Closing this ticket

PR #1919, opened 2025-02-10
Persist BooleanState Device Type to Endpoint Fields

hcarter-775:
# Description of Change
Though the Device type to endpoint mapping for BooleanState devices was correctly running in `device_added`, the fields holding this mapping information was not being persisted over a driver restart. To alleviate this, the persist flag was attached to the fields being created, and further, the setup logic was included in `device_init` to re-initialize the fields on the driver reset for all devices already onboarded whose fields are no longer set.  

# Summary of Completed Tests
- Unit tests updated.
- VDA continues to respond as expected for all functionality. The V3 Hub this device was connected to was restart and things continued to respond as expected.

### DIFF
--- a/drivers/SmartThings/matter-sensor/src/init.lua
+++ b/drivers/SmartThings/matter-sensor/src/init.lua
@@ -87,7 +87,7 @@ local function set_boolean_device_type_per_endpoint(driver, device)
       for _, dt in ipairs(ep.device_types) do
           for dt_name, info in pairs(BOOLEAN_DEVICE_TYPE_INFO) do
               if dt.device_type_id == info.id then
-                  device:set_field(dt_name, ep.endpoint_id)
+                  device:set_field(dt_name, ep.endpoint_id, { persist = true })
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -188,13 +188,14 @@ end
 
 local function device_init(driver, device)
   log.info("device init")
+  set_boolean_device_type_per_endpoint(driver, device)
```
ctowns, line 191: You can remove this from the `added` handler if it is being handled here now. Init will always run what a device is added, and also then whenever the driver is restarted. So it will always run in the same power cycle as device added, so we wouldn't need to call this twice (once in added and once again in init)

hcarter-775: good point- it's been removed

nickolas-deboom (approved):

ctowns (approved):

PR #1922, opened 2025-02-11
add support for generic zigbee multi-switch

pInksenberg:
Check all that apply

# Type of Change

          - [x] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
add support for generic zigbee multi-switch

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-switch/fingerprints.yml
+++ b/drivers/SmartThings/zigbee-switch/fingerprints.yml
@@ -2376,3 +2376,11 @@ zigbeeGeneric:
         - 0x0008 #Level
         - 0x0B04 #Electrical Measurment
     deviceProfileName: switch-level-power
+  - id: "genericMeteredSwitch/3"
+    deviceLabel: Zigbee Switch
+    clusters:
+      server:
+        - 0x0006 #OnOff
+        - 0x0702 #Simple Metering
+        - 0x0B04 #Electrical Measurment
+    deviceProfileName: switch-power
--- a/drivers/SmartThings/zigbee-switch/src/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/init.lua
@@ -20,6 +20,7 @@ local configurationMap = require "configurations"
 local SimpleMetering = clusters.SimpleMetering
 local ElectricalMeasurement = clusters.ElectricalMeasurement
 local preferences = require "preferences"
+local device_lib = require "st.device"
 
 local function lazy_load_if_possible(sub_driver_name)
   -- gets the current lua libs api version
@@ -67,7 +68,11 @@ local function endpoint_to_component(device, ep)
   end
 end
 
-local device_init = function(self, device)
+local function find_child(parent, ep_id)
+  return parent:get_child_by_parent_assigned_key(string.format("%02X", ep_id))
+end
+
+local device_init = function(driver, device)
   device:set_component_to_endpoint_fn(component_to_endpoint)
   device:set_endpoint_to_component_fn(endpoint_to_component)
 
@@ -82,6 +87,44 @@ local device_init = function(self, device)
   local ias_zone_config_method = configurationMap.get_ias_zone_config_method(device)
   if ias_zone_config_method ~= nil then
     device:set_ias_zone_config_method(ias_zone_config_method)
+  end
+end
+
+local function is_mcd_device(device)
+  local components = device.profile.components
+  if type(components) == "table" then
+    local component_count = 0
+    for _, component in pairs(components) do
+        component_count = component_count + 1
+    end
+    return component_count >= 2
+  end
+end
+
+local function device_added(driver, device, event)
+  local main_endpoint = device:get_endpoint(clusters.OnOff.ID)
+  if is_mcd_device(device) == false and device.network_type == device_lib.NETWORK_TYPE_ZIGBEE then
+    for _, ep in ipairs(device.zigbee_endpoints) do
+      if ep.id ~= main_endpoint then
+        if device:supports_server_cluster(clusters.OnOff.ID, ep.id) then
+          device:set_find_child(find_child)
+          if find_child(device, ep.id) == nil then
+            local name = string.format("%s %d", device.label, ep.id)
+            local child_profile = "basic-switch"
+            driver:try_create_device(
+              {
+                type = "EDGE_CHILD",
+                label = name,
+                profile = child_profile,
+                parent_device_id = device.id,
+                parent_assigned_child_key = string.format("%02X", ep.id),
+                vendor_provided_label = name
+              }
+            )
+          end
+        end
+      end
+    end
   end
 end
 
@@ -119,15 +162,16 @@ local zigbee_switch_driver_template = {
     lazy_load_if_possible("robb"),
     lazy_load_if_possible("wallhero"),
     lazy_load_if_possible("inovelli-vzm31-sn"),
-	lazy_load_if_possible("laisiao")
+    lazy_load_if_possible("laisiao"),
+    lazy_load_if_possible("tuya-multi")
   },
   lifecycle_handlers = {
     init = device_init,
+    added = device_added,
     infoChanged = info_changed,
     doConfigure = do_configure
   }
 }
-
 defaults.register_for_default_handlers(zigbee_switch_driver_template,
   zigbee_switch_driver_template.supported_capabilities,  {native_capability_cmds_enabled = true})
 local zigbee_switch = ZigbeeDriver("zigbee_switch", zigbee_switch_driver_template)
--- a/drivers/SmartThings/zigbee-switch/src/tuya-multi/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/tuya-multi/init.lua
@@ -1,0 +1,71 @@
+local capabilities = require "st.capabilities"
+local clusters = require "st.zigbee.zcl.clusters"
+local data_types = require "st.zigbee.data_types"
+local zcl_messages = require "st.zigbee.zcl"
+local messages = require "st.zigbee.messages"
+local zb_const = require "st.zigbee.constants"
+local read_attribute = require "st.zigbee.zcl.global_commands.read_attribute"
+
+local TUYA_MFR_HEADER = "_TZ"
+
+local function is_multi_endpoint(device)
+  local main_endpoint = device:get_endpoint(clusters.OnOff.ID)
+  for _, ep in ipairs(device.zigbee_endpoints) do
+    if ep.id ~= main_endpoint then
+      return true
+    end
+  end
+  return false
+end
+
+local function is_tuya_products(opts, driver, device)
+  if string.sub(device:get_manufacturer(),1,3) == TUYA_MFR_HEADER and is_multi_endpoint(device) then  -- if it is a tuya device, then send the magic packet
+      local subdriver = require("tuya-multi")
+      return true, subdriver
+  end
+  return false
+end
+
+local function read_attribute_function(device, cluster_id, attr_id)
+  local read_body = read_attribute.ReadAttribute( attr_id )
+  local zclh = zcl_messages.ZclHeader({
+    cmd = data_types.ZCLCommandId(read_attribute.ReadAttribute.ID)
+  })
+  local addrh = messages.AddressHeader(
+      zb_const.HUB.ADDR,
+      zb_const.HUB.ENDPOINT,
+      device:get_short_address(),
+      device:get_endpoint(clusters.Basic.ID),
+      zb_const.HA_PROFILE_ID,
+      clusters.Basic.ID
+  )
+  local message_body = zcl_messages.ZclMessageBody({
+    zcl_header = zclh,
+    zcl_body = read_body
+  })
+  return messages.ZigbeeMessageTx({
+    address_header = addrh,
+    body = message_body
+  })
+end
+
+local do_configure = function(self, device)
+  device:refresh()
+  device:configure()
+  local magic_spell = {0x0004, 0x0000, 0x0001, 0x0005, 0x0007, 0xfffe}
+  device:send(read_attribute_function(device, clusters.Basic.ID, magic_spell))
+end
+
+
+local tuya_switch_handler = {
+  NAME = "Tuya Switch Handler",
+  lifecycle_handlers = {
+      doConfigure = do_configure
+  },
+  supported_capabilities = {
+      capabilities.switch
+  },
+  can_handle = is_tuya_products
+}
+
+return tuya_switch_handler
--- a/drivers/SmartThings/zigbee-switch/src/test/test_tuya_multi_switch.lua
+++ b/drivers/SmartThings/zigbee-switch/src/test/test_tuya_multi_switch.lua
@@ -1,0 +1,123 @@
+-- Copyright 2025 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- Mock out globals
+local test = require "integration_test"
+local clusters = require "st.zigbee.zcl.clusters"
+local OnOff = clusters.OnOff
+local capabilities = require "st.capabilities"
+local zigbee_test_utils = require "integration_test.zigbee_test_utils"
+local t_utils = require "integration_test.utils"
+
+local mock_simple_device = test.mock_device.build_test_zigbee_device(
+    {
+      profile = t_utils.get_profile_definition("switch-3.yml"),
+      fingerprinted_endpoint_id = 0x01
+    }
+)
+
+zigbee_test_utils.prepare_zigbee_env_info()
+
+local function test_init()
+  test.mock_device.add_test_device(mock_simple_device)
+  zigbee_test_utils.init_noop_health_check_timer()
+end
+
+test.set_test_init_function(test_init)
+
+test.register_message_test(
+    "Reported on off status should be handled: on ep 1",
+    {
+      {
+        channel = "zigbee",
+        direction = "receive",
+        message = { mock_simple_device.id, OnOff.attributes.OnOff:build_test_attr_report(mock_simple_device,
+                                                                                                true):from_endpoint(0x01) }
+      },
+      {
+        channel = "capability",
+        direction = "send",
+        message = mock_simple_device:generate_test_message("switch1",  capabilities.switch.switch.on())
+      }
+    }
+)
+
+test.register_message_test(
+    "Reported on off status should be handled: on ep 2",
+    {
+      {
+        channel = "zigbee",
+        direction = "receive",
+        message = { mock_simple_device.id, OnOff.attributes.OnOff:build_test_attr_report(mock_simple_device,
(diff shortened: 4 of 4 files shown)


drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +100,133 @@ local device_init = function(self, device)
   end
 end
 
+
+local function send_discover_command(device)
+  local discover_attribute_id = 0x0c
```
varzac, line 105: This is not an attribute id

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +100,133 @@ local device_init = function(self, device)
...
+      local high = bytes[i+1]
+      if high then
+          local value = tonumber(high, 16) * 0x100 + tonumber(low, 16)
+          table.insert(result, value)
+      end
+  end
+  return result
+end
+
+local function all_zigbee_message_handler(self, message_channel)
```
varzac, line 214: This is going to change how EVERY message processed by ANY Zigbee Switch on the platform is handled.  I do not think we should make this change.

drivers/SmartThings/zigbee-switch/src/init.lua
tpmanley: I think the intention is to send global command Discovery Attributes which does have a command ID of 0x0C

varzac: Ah, I had forgotten the default for the frame control was a global command.  So that explains the command.  I'm still not sure what the reason for sending it is.

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +100,133 @@ local device_init = function(self, device)
...
+          driver:try_create_device(
+            {
+              type = "EDGE_CHILD",
+              label = name,
+              profile = child_profile,
+              parent_device_id = device.id,
+              parent_assigned_child_key = string.format("%02X", num_switch_server_eps),
+              vendor_provided_label = name
+            }
+          )
```
kdbang, line 183:
```suggestion
          driver:try_create_device(
            {
              type = "EDGE_CHILD",
              label = name,
              profile = "basic-switch",
              parent_device_id = device.id,
              parent_assigned_child_key = string.format("%02X", num_switch_server_eps),
              vendor_provided_label = string.format("%s %d", device.label, num_switch_server_eps)
            }
          )
```
simple and easy to understand

drivers/SmartThings/zigbee-switch/src/tuya-multi/init.lua
```diff
@@ -0,0 +1,123 @@
...
+end
+
+local do_configure = function(self, device)
+    device:refresh()
+    device:configure()
+    send_discover_command(device)
+end
+
+local function discover_response_message_handler(driver, device, zb_rx)
+    if zb_rx.address_header.cluster.value == clusters.Basic.ID and zb_rx.body.zcl_header.cmd.value == DISCOVER_ATTR_RSP_ID then
```
varzac, line 99: If this function is called, both of these have already been validated, so you don't need to check again here.

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +100,133 @@ local device_init = function(self, device)
...
+    zcl_header = zclh,
+    zcl_body = read_body
+  })
+  return messages.ZigbeeMessageTx({
+    address_header = addrh,
+    body = message_body
+})
+end
+
+local function device_added(driver, device, event)
```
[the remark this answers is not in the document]
varzac: This should be moved to the subdriver as this should only be run for the tuya devices.

drivers/SmartThings/zigbee-switch/src/tuya-multi/init.lua
```diff
@@ -0,0 +1,123 @@
...
+      zcl_body = generic_body.GenericBody(discover_command_payload)
+    })
+    local send_message = messages.ZigbeeMessageTx({
+        address_header = addrh,
+        body = message_body
+    })
+    device:send(send_message)
+end
+
+local function read_attribute_function(device, cluster_id, attr_id)
```
varzac, line 47: As mentioned, this code is already defined in the zigbee cluster_base, so you can just require the BasicCluster and then use `BasicCluster.read_attribute(device, cluster_id, attr_id)` instead of duplicating the code here.

drivers/SmartThings/zigbee-switch/src/tuya-multi/init.lua
```diff
@@ -0,0 +1,123 @@
...
+    for i = 1, #rx.body_bytes do
+      out_str = out_str .. string.format(" %02X", string.byte(rx.body_bytes:sub(i,i)))
+    end
+    local bytes = {}
+    for b in out_str:gmatch("%S+") do
+        table.insert(bytes, b)
+    end
+    table.remove(bytes, 1) -- remove first type which means discover attribute command is completed
+    local result = {}
+    for i = 1, #bytes, 3 do   -- remove data type information
```
varzac, line 81: I think this loop can be simplified by using the buf lib ` local buf = buf_lib.Reader(data)` and then you can use

https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/main/lua_libs/st/zigbee/data_types/AttributeId.lua
and
https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/main/lua_libs/st/zigbee/data_types/ZigbeeDataType.lua

using their deserialize functions.

So something like 

```
local attr_list = {}
while buf:remain() > 0 do
     attr_list[#attr_list+1] = AttributeId.deserialize(buf)
     local type_info = ZigbeeDataType.deserialize(buf)
end
```

you could also create a definition for the discover attribute response global command similar to https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/main/lua_libs/st/zigbee/zcl/global_commands/read_attribute_response.lua so that we can support this from any driver instead of custom in this driver.

drivers/SmartThings/zigbee-switch/src/tuya-multi/init.lua
```diff
@@ -0,0 +1,123 @@
...
+local messages = require "st.zigbee.messages"
+local zb_const = require "st.zigbee.constants"
+local read_attribute = require "st.zigbee.zcl.global_commands.read_attribute"
+
+local DISCOVER_ATTR_RSP_ID = 0x0D
+local DISCOVER_ATTR_RST_ID = 0x0C
+local TUYA_MFR_HEADER = "_TZ"
+
+local function is_tuya_products(opts, driver, device)
+    if string.sub(device:get_manufacturer(),1,3) == TUYA_MFR_HEADER then  -- if it is a tuya device, then send the magic packet
```
varzac, line 15: Should we also check that there is more than 1 zigbee endpoint since this handling is only for multi switch devices?

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +89,40 @@ local device_init = function(self, device)
   end
 end
 
+local function device_added(driver, device, event)
+  device:set_find_child(find_child)
```
varzac, line 93: I would move setting this function to only be set if at least 1 child device is created.

drivers/SmartThings/zigbee-switch/src/tuya-multi/init.lua
kdbang: @pInksenberg can you address @varzac 's comment?

pInksenberg: Sure, we will add this logic in next commit tonight.

drivers/SmartThings/zigbee-switch/src/init.lua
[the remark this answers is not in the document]
pInksenberg: we will add S2 device(MCD) fingerprint in added handler to avoid running our logic.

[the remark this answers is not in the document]
inasail: I think this is the last hurdle for this PR. @varzac @tpmanley Do you have any idea to filter this device in added handler? This device doesn't have switch capability in main component, how about checking this condition?

device.profile.components.main.capabilities.switch == nil

Is it too specific?

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +93,50 @@ local device_init = function(self, device)
   end
 end
 
+local function is_mcd_device(device)
+  for _, fingerprint in ipairs(MCD_fingerprint) do
+    if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
```
varzac, line 98: we shouldn't check that the device is specific model, we should just check if the profile on the device object has multiple components

drivers/SmartThings/zigbee-switch/src/init.lua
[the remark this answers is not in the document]
varzac: We should test that the device is a Zigbee device type as well, as EDGE_CHILD devices are added to the driver and will try to run added code as well.

drivers/SmartThings/zigbee-switch/fingerprints.yml
```diff
@@ -2356,13 +2356,13 @@ zigbeeGeneric:
...
-  - id: "genericMeteredSwitch/2"
+  - id: "genericMeteredSwitch"
     deviceLabel: Zigbee Switch
     clusters:
       server:
         - 0x0006 #OnOff
-        - 0x0B04 #Electrical Measurment
-    deviceProfileName: switch-power
+        - 0x0702 #Simple Metering
+    deviceProfileName: switch-power-energy
```
varzac, line 2365: It looks like all the changes in the fingerprints file is just moving things around, which is making the PR more confusing.  I think you should revert all these changes as they don't seem relevant.

qy-fang: Let me explain that, after we moving fingerprints around, we find that a switch which has 0x0b04 & 0x0702 in server cluster will match with switch-power profile. We want these plug to show as a "switch" not a "plug"  in the beginning, then we update its profile to "basic-switch", so from user sides, its  icon will not change.

drivers/SmartThings/zigbee-switch/src/init.lua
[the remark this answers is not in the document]
qy-fang: We have changed it

drivers/SmartThings/zigbee-switch/fingerprints.yml
varzac: Changing the order here probably won't affect anything.  It might make a difference in your testing, because you are removing and republishing the driver for test, this might do something, but for the already published driver, I don't think there is any guarantee that the order here makes any difference.  I think a better solution would be to add a generic fingerprint that includes both clusters, and have it set the profile/icon you want there.

drivers/SmartThings/zigbee-switch/src/tuya-multi/init.lua
```diff
@@ -0,0 +1,71 @@
...
+    address_header = addrh,
+    body = message_body
+  })
+end
+
+local do_configure = function(self, device)
+  device:refresh()
+  device:configure()
+  local magic_spell = {0x0004, 0x0000, 0x0001, 0x0005, 0x0007, 0xfffe}
+  device:send(read_attribute_function(device, clusters.Basic.ID, magic_spell))
```
varzac, line 56: It would be ideal to add a test for this behavior.

drivers/SmartThings/zigbee-switch/src/init.lua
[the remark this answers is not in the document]
varzac: It would be ideal if we could test that this correctly generates the create_device calls, and won't trigger for MCD devices.

drivers/SmartThings/zigbee-switch/fingerprints.yml
qy-fang: We will add a new generic fingerprint.

drivers/SmartThings/zigbee-switch/src/tuya-multi/init.lua
qy-fang: No problem

drivers/SmartThings/zigbee-switch/src/init.lua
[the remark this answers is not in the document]
qy-fang: Since we do not have WWST MCD device in office, we mocked a device which fingerprint and clusters are totally same with a real one. This create_device call will not be triggered.

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +90,55 @@ local device_init = function(self, device)
...
+  if type(components) == "table" then
+    local component_count = 0
+    for _, component in pairs(components) do
+        component_count = component_count + 1
+    end
+    if component_count >= 2 then
+      return true
+    else
+      return false
+    end
```
greens, line 104:
```suggestion
    return component_count >= 2
```

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +90,55 @@ local device_init = function(self, device)
...
+          if find_child(device, num_switch_server_eps) == nil then
+            local name = string.format("%s %d", device.label, num_switch_server_eps)
+            local child_profile = "basic-switch"
+            driver:try_create_device(
+              {
+                type = "EDGE_CHILD",
+                label = name,
+                profile = child_profile,
+                parent_device_id = device.id,
+                parent_assigned_child_key = string.format("%02X", num_switch_server_eps),
```
greens, line 131: So I think you used this number `num_switch_server_eps` in the case that the device has non-contiguous indices for its switch endpoints (i.e. endpoints 1, 3, and 5 are switches, but you want them to appear as 1, 2, and 3), but I think this would break your `find_child` function if it ever happened. I think it's probably safer to just use `ep.id` here. The label can still be whatever you like, though.

greens: Ah, actually it doesn't matter because you increment `num_switch_server_eps` regardless of whether or not the endpoint is a switch endpoint. But if you wanted the numbering to be contiguous you could change it.

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +90,55 @@ local device_init = function(self, device)
...
+    end
+  end
+end
+
+local function device_added(driver, device, event)
+  local num_switch_server_eps = 0
+  local main_endpoint = device:get_endpoint(clusters.OnOff.ID)
+  local updated_flag = false
+  if is_mcd_device(device) == false and device.network_type == device_lib.NETWORK_TYPE_ZIGBEE then
+    for _, ep in ipairs(device.zigbee_endpoints) do
```
greens, line 113: you could probably use this index in the place of `num_switch_server_eps` rather than discard it

drivers/SmartThings/zigbee-switch/src/init.lua
qy-fang: Okay, we will change it

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -85,6 +90,50 @@ local device_init = function(self, device)
...
+      if ep.id ~= main_endpoint then
+        if device:supports_server_cluster(clusters.OnOff.ID, ep.id) and updated_flag == false then
+          device:try_update_metadata({profile="basic-switch"})
+          updated_flag = true
+          device:set_find_child(find_child)
```
greens, line 114: Is there a reason you're switching the main device to a basic switch? I figure it could still use whatever functionality we've fingerprinted it with.

qy-fang: Some switches will match with "switch-power" or "switch-power-energy" profile, but we cannot make power & energy meter value right for every generic switch(some switch use private cluster to meter), so we update their profile to "basic-switch".

greens: If they use a private cluster for metering how would they fingerprint to the power-energy profile?

greens: That's my point. The code you're adding here should be generic and for devices that use standard clusters and standard attributes. Those devices should have no trouble keeping their power and energy capabilities if they've fingerprinted to them using an appropriate generic fingerprint. Treating them as basic switches because it won't work otherwise for Tuya devices is not generic behavior.

Tuya-specific behavior belongs in a Tuya-specific sub-driver.

qy-fang: I get your point, but the code we adding here is not just for Tuya. We cannot get every manufacturer's power & electric coefficient so we want these multi-switch work as basic-switch in generic driver first.
Besides, only "no fingerprint to match" & "multi on/off cluster" & "none MCD device" will be effected with that code.

greens: I understand, but the effort here was to increase functionality for devices we don't have fingerprints for. 

Current functionality for, say, a power strip that supports metering each outlet would be to have one metered switch controllable.

With this change, you gain the ability to control on/off of each endpoint, but you lose the metering.

It seems possible to make these changes such that we could keep metering on the main endpoint while adding on/off control to the remaining endpoints.

I understand that this generic solution would not work for the Tuya devices because they use a custom attribute, but it _should_ work for any device using standard zigbee attributes. We poll the multiplier and divisors as part of the base driver's [configure method](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-switch/src/init.lua#L46).

If the Tuya devices do not work with the base generic driver, then the non-standard behavior should be accommodated in the sub-driver.

qy-fang: I appreciate your comments, it will be grate to support more function.
We try your solutions and find that Aqara Z1 series' switch cannot meter power(power meter value is 0 in plugin), so if we just move this to tuya sub-driver, we will encounter this issue. We wonder your opinions.

greens: Device-specific departures from the zigbee standard are intended to be handled by sub-drivers when the device manufacturer submits the device through the WWST process.

If our base driver uses Zigbee standard behavior, and a device does not function fully with it, that's an issue with the device, (unless you've investigated and found that our driver is acting in a way not up to the standards). 

But if a device fingerprints to our generic fingerprint for a metered switch and uses the simple metering cluster, we would expect to be able to report its power/energy consumption.

A device failing to operate using our base drivers is an argument for adding a sub-driver, not for removing functionality from the base driver.

greens (approved):

drivers/SmartThings/zigbee-switch/src/init.lua
```diff
@@ -119,15 +163,16 @@ local zigbee_switch_driver_template = {
     lazy_load_if_possible("robb"),
     lazy_load_if_possible("wallhero"),
     lazy_load_if_possible("inovelli-vzm31-sn"),
-	lazy_load_if_possible("laisiao")
+	  lazy_load_if_possible("laisiao"),
```
greens, line 166: nit: whitespace

greens (approved):

drivers/SmartThings/zigbee-switch/fingerprints.yml
Mariano-Github: Hi,
Most devices using clusters 0x0702 and 0x0B04 should have the `switch-power-energy` profile.

They use cluster 0x0B04 to send Power with the `Active Power` attribute (0x050B) and cluster 0x0702 to send Total Energy with the `CurrentSummationDelivered` attribute (0x0000).

PR #1924, opened 2025-02-11
Persist fields across multiple Drivers

hcarter-775:
# Description of Change
Matter Switch:
 - CUMULATIVE_REPORTS_NOT_SUPPORTED should be saved. This will not change after driver init.
 - ENERGY_MANAGEMENT_ENDPOINT should be saved. This will not be saved, or re-initialized due to gating after the first driver init.
 - SWITCH_INITIALIZED is used in init, but is set in init. A driver restart would not save this, and so the field is never being used.
 - MOST_RECENT_TEMP should be saved. Low importance, but this doesn't need to be made nil on a driver restart
 - TOTAL_IMPORTED_ENERGY must be persisted. This should never be reset on a driver restart.

Smoke Alarm Sub-Driver:
  - PROFILE_MATCHED is used in init, but is set in init. A driver restart would not save this, and so the field is never being used.
# Summary of Completed Tests
Hub restart attempted in Matter Switch to ensure that init continues to work as expected.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -247,7 +247,7 @@ local function set_poll_report_timer_and_schedule(device, is_cumulative_report)
     clusters.ElectricalEnergyMeasurement.ID,
     {feature_bitmap = clusters.ElectricalEnergyMeasurement.types.Feature.CUMULATIVE_ENERGY })
   if #cumul_eps == 0 then
-    device:set_field(CUMULATIVE_REPORTS_NOT_SUPPORTED, true)
+    device:set_field(CUMULATIVE_REPORTS_NOT_SUPPORTED, true, {persist = true})
   end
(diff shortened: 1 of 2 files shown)


nickolas-deboom (approved):

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -612,7 +612,7 @@ local function try_build_child_switch_profiles(driver, device, switch_eps, main_
     device:set_field(IS_PARENT_CHILD_DEVICE, true, {persist = true})
   end
 
-  device:set_field(SWITCH_INITIALIZED, true)
+  device:set_field(SWITCH_INITIALIZED, true, {persist = true})
```
ctowns, line 615: Could you test a switch device across driver restarts to confirm that persisting this will not cause issues?

hcarter-775: Done :+1: and yes, functionality continues to function as expected.

PR #1946, opened 2025-02-19
Use Persisted State cache for Thermostat SystemMode attribute

hcarter-775:
# Description of Change
As it stands, any optional added SystemMode value will be removed from the SupportedModes table on a driver restart. This PR introduces a persisted table that records all of the mode values added in `system_mode_handler`. Then, on a driver restart, the persisted table (`ALLOWED_OPTIONAL_THERMOSTAT_MODES`) is read in `sequence_of_operation_handler` as the base values to be added to the `supportedThermostatModes` table. 

To note: two similar checks are added, one under the `COOLING_WITH_REHEAT` if-else branch that specifically removes the `emergency heat` and `heat` value if they're found (since this would no longer be allowed) and vice versa with the `HEATING_WITH_REHEAT` branch and the `precooling` and `cool` values. This accounts for the case where the `ControlSequenceOfOperation` attribute is changed, and thus the device no longer supports the same optional values it once did. 

Further, new logic is added to prevent anything from being added to the table twice.

# Summary of Completed Tests
- VDA Room Air Conditioner (which _does_ include optional values) was tested on the previous main driver and the optional values were reset as expected. Then, this updated driver was applied, and its' optional values (`precooling`, `emergency heat`) were persisted over the restart.  
- Unit tests continue to pass.
- VDA Thermostat (which doesn't include any optional values) was tested and runs as expected, including over a restart.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -43,6 +43,7 @@ if version.api < 10 then
 end
 
 local DISALLOWED_THERMOSTAT_MODES = "__DISALLOWED_CONTROL_OPERATIONS"
+local OPTIONAL_THERMOSTAT_MODES_SEEN = "__OPTIONAL_THERMOSTAT_MODES_SEEN"
 
 local THERMOSTAT_MODE_MAP = {
   [clusters.Thermostat.types.ThermostatSystemMode.OFF]               = capabilities.thermostatMode.thermostatMode.off,
@@ -255,12 +256,12 @@ local subscribed_attributes = {
 }
 
 local function tbl_contains(array, value)
-  for _, element in ipairs(array) do
+  for idx, element in ipairs(array) do
     if element == value then
-      return true
+      return true, idx
     end
   end
-  return false
+  return false, nil
 end
 
 local function get_field_for_endpoint(device, field, endpoint)
@@ -824,20 +825,20 @@ end
 local function system_mode_handler(driver, device, ib, response)
   local supported_modes = device:get_latest_state(device:endpoint_to_component(ib.endpoint_id), capabilities.thermostatMode.ID, capabilities.thermostatMode.supportedThermostatModes.NAME) or {}
   -- check that the given mode was in the supported modes list
-  for _, mode in ipairs(supported_modes) do
-    if mode == THERMOSTAT_MODE_MAP[ib.data.value].NAME then
-      device:emit_event_for_endpoint(ib.endpoint_id, THERMOSTAT_MODE_MAP[ib.data.value]())
-      return
-    end
+  if tbl_contains(supported_modes, THERMOSTAT_MODE_MAP[ib.data.value].NAME) then
+    device:emit_event_for_endpoint(ib.endpoint_id, THERMOSTAT_MODE_MAP[ib.data.value]())
+    return
   end
-  -- if the value is not found in the supported modes list, check if it's disallowed
+  -- if the value is not found in the supported modes list, check if it's disallowed and early return if so.
   local disallowed_thermostat_modes = device:get_field(DISALLOWED_THERMOSTAT_MODES) or {}
-  for _, mode in pairs(disallowed_thermostat_modes) do
-    if mode == THERMOSTAT_MODE_MAP[ib.data.value].NAME then
-      return
-    end
+  if tbl_contains(disallowed_thermostat_modes, THERMOSTAT_MODE_MAP[ib.data.value].NAME) then
+    return
   end
   -- if we get here, then the reported mode is allowed and not in our mode map
+  -- add the mode to the OPTIONAL_THERMOSTAT_MODES_SEEN and supportedThermostatModes tables
+  local optional_modes_seen = utils.deep_copy(device:get_field(OPTIONAL_THERMOSTAT_MODES_SEEN)) or {}
+  table.insert(optional_modes_seen, THERMOSTAT_MODE_MAP[ib.data.value].NAME)
+  device:set_field(OPTIONAL_THERMOSTAT_MODES_SEEN, optional_modes_seen, {persist=true})
   local sm_copy = utils.deep_copy(supported_modes)
   table.insert(sm_copy, THERMOSTAT_MODE_MAP[ib.data.value].NAME)
   local supported_modes_event = capabilities.thermostatMode.supportedThermostatModes(sm_copy, {visibility = {displayed = false}})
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -288,6 +289,9 @@ local function component_to_endpoint(device, component_name)
 end
 
 local function device_init(driver, device)
+  if device:get_field(ALLOWED_OPTIONAL_THERMOSTAT_MODES) == nil then
```
ctowns, line 292: Previously, the `system_mode_handler` had this line:
```
local supported_modes = {capabilities.thermostatMode.thermostatMode.off.NAME
```
Where the supported modes type is initialized with the `Off` mode. I think you could just keep that in the system mode handler instead of spreading it out to the init as well. We should receive a system mode report when the initial subscription is set up, so it seems more concise to keep all the handling there.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -829,6 +833,10 @@ local function system_mode_handler(driver, device, ib, response)
     end
   end
   -- if we get here, then the reported mode is allowed and not in our mode map
+  -- add the mode to the ALLOWED_OPTIONAL_THERMOSTAT_MODES and supportedThermostatModes tables
+  local allowed_optional_modes = utils.deep_copy(device:get_field(ALLOWED_OPTIONAL_THERMOSTAT_MODES)) or {}
+  table.insert(allowed_optional_modes, THERMOSTAT_MODE_MAP[ib.data.value].NAME)
+  device:set_field(ALLOWED_OPTIONAL_THERMOSTAT_MODES, allowed_optional_modes, {persist=true})
   local sm_copy = utils.deep_copy(supported_modes)
   table.insert(sm_copy, THERMOSTAT_MODE_MAP[ib.data.value].NAME)
   local supported_modes_event = capabilities.thermostatMode.supportedThermostatModes(sm_copy, {visibility = {displayed = false}})
```
ctowns, line 842: Do we need to use an additional table to store the `ALLOWED_OPTIONAL_THERMOSTAT_MODES`? It looks like this data is the same that will be emitted as a part of the `thermostatMode.supportedThermostatModes` capability event throughout the file. Therefore, you should be able to get the supported modes using `device:get_latest_state()` which is stored in the device's `state_cache`, which should be persisted through a restart as well.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -43,6 +43,7 @@ if version.api < 10 then
 end
 
 local DISALLOWED_THERMOSTAT_MODES = "__DISALLOWED_CONTROL_OPERATIONS"
+local ALLOWED_OPTIONAL_THERMOSTAT_MODES = "__ALLOWED_OPTIONAL_THERMOSTAT_MODES"
```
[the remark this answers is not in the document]
hcarter-775: yes, I think `get_latest_state` will solve the issues you've addressed with this implementation. I did not know that it accessed a `state_cache` that persisted over a restart :+1:

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -851,13 +859,25 @@ local function sequence_of_operation_handler(driver, device, ib, response)
   -- However, we assert here that a Cooling enum value implies that SystemMode supports cooling, and the same for a Heating enum.
   -- We also assert that Off is supported, though per spec this is optional.
   local disallowed_mode_operations = {}
-  local supported_modes = {capabilities.thermostatMode.thermostatMode.off.NAME}
+  local supported_modes = utils.deep_copy(device:get_field(ALLOWED_OPTIONAL_THERMOSTAT_MODES)) or {capabilities.thermostatMode.thermostatMode.off.NAME}
 
   if ib.data.value <= clusters.Thermostat.attributes.ControlSequenceOfOperation.COOLING_WITH_REHEAT then
+    for index, mode in pairs(supported_modes) do
+      if mode == capabilities.thermostatMode.thermostatMode.precooling.NAME then
+        table.remove(supported_modes, index) -- if seen before, remove now
```
nickolas-deboom, line 867: I might be missing something but why do these need to be removed here?

hcarter-775: From the PR description:

To note: two similar checks are added, one under the COOLING_WITH_REHEAT if-else branch that specifically removes the emergency heat value if it's found (since this would no longer be allowed) and vice versa with the HEATING_WITH_REHEAT branch and the precooling value. This accounts for the case where the ControlSequenceOfOperation attribute is changed, and thus the device no longer supports the same optional values it once did.

hcarter-775: though I see these got flipped around in the PR.. so yes, these shouldn't be removed. The comment I sent above should have been the implementation.

nickolas-deboom: Ah! That would make more sense :smile:

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: yeah, there was some reasoning why it was difficult to do as you suggested, but it's been removed in ffd5f71f anyway due to the use of `get_latest_state`. :+1:

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: no, removed in ffd5f71f0e7a02e462fc3e909937b9566a43516b

nickolas-deboom (approved):
Looks good from my perspective!

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -851,30 +860,45 @@ local function sequence_of_operation_handler(driver, device, ib, response)
...
-    table.insert(supported_modes, capabilities.thermostatMode.thermostatMode.auto.NAME)
+    table.insert(modes_for_inclusion, capabilities.thermostatMode.thermostatMode.auto.NAME)
   else
-    table.insert(disallowed_mode_operations, clusters.Thermostat.types.ThermostatSystemMode.AUTO)
+    table.insert(disallowed_mode_operations, capabilities.thermostatMode.thermostatMode.auto.NAME)
   end
 
-  device:set_field(DISALLOWED_THERMOSTAT_MODES, disallowed_mode_operations, {persist = true})
+  -- if a disallowed value was once allowed and added, it should be removed now.
+  for index, mode in pairs(supported_modes) do
```
ctowns, line 890: Shouldn't this be checking `modes_for_inclusion`?

hcarter-775: no, since, per the comment `  -- if a disallowed value was once allowed and added, it should be removed now.`, we're actually removing any modes that were in the original list (`supported_modes`) that should no longer be there due to the updated `ControlSequenceOfOperation` attribute. The `modes_for_inclusion` table does not need to be checked like this, since it only includes a list with the most up-to-date version of that attribute.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -851,30 +860,45 @@ local function sequence_of_operation_handler(driver, device, ib, response)
...
+    )) or {capabilities.thermostatMode.thermostatMode.off.NAME}
 
+  local modes_for_inclusion = {}
   if ib.data.value <= clusters.Thermostat.attributes.ControlSequenceOfOperation.COOLING_WITH_REHEAT then
-    table.insert(supported_modes, capabilities.thermostatMode.thermostatMode.cool.NAME)
-    table.insert(disallowed_mode_operations, clusters.Thermostat.types.ThermostatSystemMode.HEAT)
-    table.insert(disallowed_mode_operations, clusters.Thermostat.types.ThermostatSystemMode.EMERGENCY_HEATING)
+    table.insert(modes_for_inclusion, capabilities.thermostatMode.thermostatMode.cool.NAME)
+    table.insert(disallowed_mode_operations, capabilities.thermostatMode.thermostatMode.heat.NAME)
+    table.insert(disallowed_mode_operations, capabilities.thermostatMode.thermostatMode.emergency_heat.NAME)
```
ctowns, line 871: If a thermostat at one point supported `emergency_heat` (which is optional), but then had a sequence of operation change, wouldn't we then lose that information and be required to receive an `emergency_heat` event again in order to display it to the user?

ctowns: We may need to go back to your `ALLOWED_OPTIONAL_THERMOSTAT_MODES` solution in order to prevent this

hcarter-775: yes :+1: however, a sequence of operation change itself is rare, and is mostly theoretical (the spec says it's not allowed). Further, the use of `emergency heat` and `precooling` seem to be uncommon,  so I elected not to save extra information just for this mix of two uncommon events

nickolas-deboom: The spec does say it's a writable attribute, does it also say that changing it isn't allowed?

nickolas-deboom: I'm a little bit confused about the need for this, since isn't this just the nature of the limitations of this attribute?

hcarter-775: @nickolas-deboom yes it says that changing it isn't allowed, and that the writable aspect is for legacy reasons. Though Tom mentioned that there were companies who'd mentioned that they might write to the state anyway, even if it's against spec.

nickolas-deboom: Gotcha. So in that case I would imagine we don't need to worry about this right?

ctowns (approved):

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: I'd say probably not, but since the work is already done I don't see why not support a weird edge case where there's little loss and not much complexity added.

nickolas-deboom: Oh man, I was looking at the first commit only so I didn't realize it's already implemented :sweat_smile:

PR #1962, opened 2025-02-25
Fix a few places where nil causes crashes in wemo

cjswedes:
This was done based on crashes that have been reported a lot in the field, but should be handled better.

### DIFF
--- a/drivers/SmartThings/wemo/src/protocol.lua
+++ b/drivers/SmartThings/wemo/src/protocol.lua
@@ -107,7 +107,7 @@ function protocol.subscribe(device, listen_ip, listen_port)
     return
   end
 
-  local device_facing_local_ip = find_interface_ip_for_remote(ip)
+  local device_facing_local_ip = find_interface_ip_for_remote(ip) or device.driver.environment_info.hub_ipv4
 
   local response_body = {}
   local resp, code_or_err, headers, status_line = http.request {
--- a/drivers/SmartThings/wemo/src/subscription_server.lua
+++ b/drivers/SmartThings/wemo/src/subscription_server.lua
@@ -93,7 +93,7 @@ function SubscriptionServer.new_server()
         end)
       else
         log.error_with({hub_logs = true},
-          "serve| shutting down tcp server due to unexpected accept error: " .. err)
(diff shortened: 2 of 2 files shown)


drivers/SmartThings/wemo/src/subscription_server.lua
```diff
@@ -124,20 +124,20 @@ function SubscriptionServer.new_server()
...
+            string.format("serve| successfully subscribed to device %s", msg.device.label))
           self.subscriptions[sub_id] = msg.device
           msg.device:set_field("subscription_id", sub_id)
         end
       elseif msg.type == ControlMessageType.Unsubscribe then
-        log.trace("serve| unsubscribing from device " .. msg.device.label)
+        log.trace(string.format("serve| unsubscribing from device %s", msg.device and msg.device.label or nil))
         if msg.device:get_field("subscription_id") then
-          log.warn("serve| no existing subscription for device" .. msg.device.label)
+          log.warn(string.format("serve| no existing subscription for device%s", msg.device and msg.device.label or nil))
```
varzac, line 137: nit: no space between `device` and `%s`

varzac (approved):

drivers/SmartThings/wemo/src/protocol.lua
```diff
@@ -107,7 +107,7 @@ function protocol.subscribe(device, listen_ip, listen_port)
     return
   end
 
-  local device_facing_local_ip = find_interface_ip_for_remote(ip)
```
NoahCornell, line 110: Should this be flipped so that we don't have to do the socket ops if we already have the IP?

NoahCornell (approved):

drivers/SmartThings/wemo/src/protocol.lua
cjswedes: We could, but I wanted to make sure the changes here were as limited in scope as possible to keep existing behavior. In the future we could make a change for all lan drivers to use the environment info primarily, but I dont want to do it here.

cjswedes:
FWIW I did test this with a couple devices on my home setup.

PR #1968, opened 2025-02-28
Matter 1.4 Driver Release

hcarter-775:
Add support for the device types introduced by the Matter 1.4 Spec.
This is a squashed version of the following PR: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1870


ctowns (approved):
I've done a final spot check and this look good to me 👍 
I'll let @nickolas-deboom and @HunsupJung provide the final approval

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -87,6 +89,9 @@ local device_type_profile_map = {
   [ON_OFF_SWITCH_ID] = "switch-binary",
   [ON_OFF_DIMMER_SWITCH_ID] = "switch-level",
   [ON_OFF_COLOR_DIMMER_SWITCH_ID] = "switch-color-level",
+  [MOUNTED_ON_OFF_CONTROL_ID] = "switch-binary",
+  [MOUNTED_DIMMABLE_LOAD_CONTROL_ID] = "switch-level",
+  [GENERIC_SWITCH_ID] = "button"
```
nickolas-deboom, line 94: `GENERIC_SWITCH_ID` should be removed since this table is only used for switch devices rather than buttons.

PR #1991, opened 2025-03-11
Matter Switch: Support fan/light devices

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change adds support for an extended color light/fan combination device. A new profile for this device type was added by [PR 2012](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2012). New logic is included in the matter switch driver to support the component mapping for this profile. This change also refactors some of the initialization logic to help support expanding the driver to cover more multi-type devices in the future.

# Summary of Completed Tests

Tested with an Orein bathroom fan device. New unit tests also included.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -56,6 +56,7 @@ local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 -- containing both button endpoints and switch endpoints will use this field
 -- rather than COMPONENT_TO_ENDPOINT_MAP.
 local COMPONENT_TO_ENDPOINT_MAP_BUTTON = "__component_to_endpoint_map_button"
+local COMPONENT_TO_ENDPOINT_MAP_FAN = "__component_to_endpoint_map_fan"
 local ENERGY_MANAGEMENT_ENDPOINT = "__energy_management_endpoint"
 local IS_PARENT_CHILD_DEVICE = "__is_parent_child_device"
 local COLOR_TEMP_BOUND_RECEIVED_KELVIN = "__colorTemp_bound_received_kelvin"
@@ -87,6 +88,7 @@ local MOUNTED_ON_OFF_CONTROL_ID = 0x010F
 local MOUNTED_DIMMABLE_LOAD_CONTROL_ID = 0x0110
 local GENERIC_SWITCH_ID = 0x000F
 local ELECTRICAL_SENSOR_ID = 0x0510
+local FAN_DEVICE_TYPE_ID = 0x002B
 local device_type_profile_map = {
   [ON_OFF_LIGHT_DEVICE_TYPE_ID] = "light-binary",
   [DIMMABLE_LIGHT_DEVICE_TYPE_ID] = "light-level",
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -403,25 +419,36 @@ local function ignore_initial_color_read(device, attr_bit)
...
         end
       end
     end
   end
-  return false
+end
+
+--- device_type_supports_multicomponent_configuration helper function used to check
+--- whether the device is currently supported by a profile for a multicomponent configuration.
+local function device_type_supports_multicomponent_configuration(device, main_endpoint, secondary_endpoints, secondary_device_type)
```
hcarter-775, line 438: I think pulling both the endpoints and the device type in as separate parameters is a little confusing. It could be better in this case to grab the endpoints based on the device type within this function

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -677,58 +740,61 @@ local function try_build_child_switch_profiles(driver, device, switch_eps, main_
...
-  local num_switch_server_eps = try_build_child_switch_profiles(driver, device, switch_eps, main_endpoint)
+  table.sort(fan_eps)
+  table.sort(switch_eps)
 
   if #button_eps > 0 then
-    build_button_profile(device, main_endpoint, #button_eps)
-    configure_buttons(device)
-    return
+    build_component_map(device, main_endpoint, button_eps, GENERIC_SWITCH_ID)
+    build_profile(device, main_endpoint, button_eps, GENERIC_SWITCH_ID)
```
hcarter-775, line 777: `build_profile` seems a little bit generic?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -56,6 +56,7 @@ local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 -- containing both button endpoints and switch endpoints will use this field
 -- rather than COMPONENT_TO_ENDPOINT_MAP.
 local COMPONENT_TO_ENDPOINT_MAP_BUTTON = "__component_to_endpoint_map_button"
+local COMPONENT_TO_ENDPOINT_MAP_FAN = "__component_to_endpoint_map_fan"
```
hcarter-775, line 59: This seems like a dark road to go down... since it only leads to more and more component to endpoint maps?

hcarter-775: Mainly, I worry about this because this is the kind of system that can't necessarily be removed once we add it.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -194,6 +196,21 @@ local child_device_profile_overrides_per_vendor_id = {
   }
 }
 
+local supported_mcd_configs_per_main_endpoint_device_type = {
```
hcarter-775, line 199: I think modular profiles will make this kind of a moot point, no? Since we can just have all capabilities we could ever need in the main component, and then support any/all of the capabilties we'd ever need in x number of components? Then, we'd only really have to account for that "x" value, and the rest of this would already be handled. 

I understand that this is not the case now, so I it's a little less clear how to proceed.

nickolas-deboom:
Closing in favor of #2032

PR #2010, opened 2025-03-20
Fix button configuration not occuring when a fingerprint exists for a button device

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

When a fingerprint for a button device exists, and matches the profile that the switch driver would select for the device, the `info_changed` event occurs after attempting to update the profile, but since the profile did not change, `configure_buttons` is not called. This leads to the following issues:
- `MultiPressMax` is not read and the driver reports multiple events for a double/multi button press ("Pressed" and "Double pressed" for example)
- `supportedButtonValues` is not emitted, and the button capabilities display all of the possible values for this attribute rather than the ones supported by the device

This change addresses this by always calling `configure_buttons` during button initialization. Then, inside `configure_buttons`, events are only emitted if the profile contains the corresponding component for a given endpoint. This is done to prevent errors if the profile has not been updated yet. If an event is not sent, that is okay because `configure_buttons` is called again in `info_changed` if a profile change occurs.

# Summary of Completed Tests

Updated unit tests and tested with Ikea button. The issue can be reproduced by creating a fingerprint for the device for the same profile that the driver would match the device to. `configure_buttons` would not run previously in this case, so `supportedButtonValues` would never be emitted and this attribute would display all of its possible values in the device card.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -1291,11 +1291,11 @@ local function info_changed(driver, device, event, args)
       device:set_field(COLOR_MODE_ATTRS_BITMAP, 0x0F) -- all bits enabled: Hue (0x01), Saturation (0x02), X (0x04), and Y (0x08)
       device:send(clusters.ColorControl.attributes.ColorMode:read())
     end
-    device:subscribe()
     local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
     if #button_eps > 0 and device.network_type ~= device_lib.NETWORK_TYPE_CHILD then
       configure_buttons(device)
     end
+    device:subscribe()
   end
 end
 
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -295,7 +295,6 @@ local HELD_THRESHOLD = 1
 -- this is the number of buttons for which we have a static profile already made
 local STATIC_BUTTON_PROFILE_SUPPORTED = {1, 2, 3, 4, 5, 6, 7, 8}
 
-local DEFERRED_CONFIGURE = "__DEFERRED_CONFIGURE"
 local BUTTON_DEVICE_PROFILED = "__button_device_profiled"
 
 -- Some switches will send a MultiPressComplete event as part of a long press sequence. Normally the driver will create a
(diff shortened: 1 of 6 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1289,12 +1291,11 @@ local function info_changed(driver, device, event, args)
...
     end
-    device:subscribe()
-    if device:get_field(DEFERRED_CONFIGURE) and device.network_type ~= device_lib.NETWORK_TYPE_CHILD then
-      -- profile has changed, and we deferred setting up our buttons, so do that now
+    local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
+    if #button_eps > 0 and device.network_type ~= device_lib.NETWORK_TYPE_CHILD then
       configure_buttons(device)
-      device:set_field(DEFERRED_CONFIGURE, nil)
     end
+    device:subscribe()
```
hcarter-775, line 1298: really small thing that doesn't need a PR change: why move this line?

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button.lua
```diff
@@ -110,27 +127,14 @@ local function test_init()
...
   test.socket.matter:__expect_send({mock_device.id, read_attribute_list})
+  configure_buttons()
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
   local device_info_copy = utils.deep_copy(mock_device.raw_st_data)
   device_info_copy.profile.id = "5-buttons-battery"
   local device_info_json = dkjson.encode(device_info_copy)
   test.socket.device_lifecycle:__queue_receive({ mock_device.id, "infoChanged", device_info_json })
+  configure_buttons()
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
-
```
hcarter-775, line 119: this occurs across the PR- why get rid of this? It seems like we'd want to ensure these sends are occurring?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -556,26 +573,28 @@ local function configure_buttons(device)
...
-    elseif tbl_contains(msl_eps, ep) then
-      supportedButtonValues_event = capabilities.button.supportedButtonValues({"pushed", "held"}, {visibility = {displayed = false}})
-    elseif tbl_contains(msr_eps, ep) then
-      supportedButtonValues_event = capabilities.button.supportedButtonValues({"pushed", "held"}, {visibility = {displayed = false}})
-      set_field_for_endpoint(device, EMULATE_HELD, ep, true, {persist = true})
-    else -- this switch endpoint only supports momentary switch, no release events
-      supportedButtonValues_event = capabilities.button.supportedButtonValues({"pushed"}, {visibility = {displayed = false}})
-      set_field_for_endpoint(device, INITIAL_PRESS_ONLY, ep, true, {persist = true})
-    end
+    if device.profile.components[endpoint_to_component(device, ep)] then
```
hcarter-775, line 576: It might be a good idea to add some hub logs here for better checking of issues in the future. For example, we could add a line at the top of this if statememt like `device.log.info({hub_logs=true}, "Configuring Supported Values from Generic Switch Endpoint {ep}")` and then we could add an `else` if there isn't a component yet with a similar log like `device.log.info({hub_logs=true}, "Component not Found for Generic Switch Endpoint {ep}. Skipping Supported Value Configuration.")`.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I didn't realize this & will change it back!

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I added some logs as suggested in the latest commit!

hcarter-775 (approved):

PR #2018, opened 2025-03-24
fix: Resolves a conflict between the `main` and `beta` branches

dljsjr:
Conflicts were likely introduced during prior cherry-pick/expedite operations.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/fingerprints.yml
+++ b/drivers/SmartThings/matter-thermostat/fingerprints.yml
@@ -10,25 +10,25 @@ matterManufacturer:
     vendorId: 0x1396
     productId: 0x043A
     deviceProfileName: air-purifier-hepa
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-thermostat/fingerprints.yml
```diff
@@ -8,27 +8,27 @@ matterManufacturer:
   - id: "Leedarson AiDot Air Purifier"
     deviceLabel: AiDot Air Purifier
     vendorId: 0x1396
-    productId: 0x043A
+    productId: 0x0043A
```
nickolas-deboom, line 11: I think this was from a recent PR and was supposed to be changed back to 0x043A

hcarter-775 (approved):

nickolas-deboom (approved):

PR #2027, opened 2025-03-26
Fix wrong colorControl values emitted after device power cycle

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

Following a device power cycle, hub power cycle, or device going offline and then back online, it sends a subscription report with Hue, Saturation, X, and Y values. The ColorMode needs to be checked before emitting the colorControl capability with the hue and saturation values. The issue is that the ColorMode attribute is not guaranteed to be processed by the driver before the other color attributes. Therefore, the ColorMode handler should send a read request for the proper color attributes after checking the ColorMode.

Changes were previously made by [PR 1772](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1772), but that only addressed hub power cycles. This PR handles every case in which the driver might receive a subscription report from the device where this problem could occur.

# Summary of Completed Tests

Tested with a Zemismart color bulb, IKEA color bulb, and by CN team.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -65,13 +65,9 @@ local COLOR_TEMP_MAX = "__color_temp_max"
 local LEVEL_BOUND_RECEIVED = "__level_bound_received"
 local LEVEL_MIN = "__level_min"
 local LEVEL_MAX = "__level_max"
-local COLOR_MODE_ATTRS_BITMAP = "__color_mode_attrs_bitmap"
-local color_mode_attr_bits = {
-  HUE = 0x01, -- CurrentHue
-  SAT = 0x02, -- CurrentSaturation
-  X   = 0x04, -- CurrentX
-  Y   = 0x08  -- CurrentY
-}
+local CURRENT_HUE_SAT = clusters.ColorControl.types.ColorMode.CURRENT_HUE_AND_CURRENT_SATURATION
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -934,15 +910,15 @@ local function level_attr_handler(driver, device, ib, response)
 end
 
 local function hue_attr_handler(driver, device, ib, response)
-  if ignore_initial_color_read(device, color_mode_attr_bits.HUE) or ib.data.value == nil then
+  if device:get_field(COLOR_MODE) == X_Y_COLOR_MODE  or ib.data.value == nil then
```
hcarter-775, line 913: might be better to do:
```suggestion
  if device:get_field(COLOR_MODE) ~= HUE_SAT_COLOR_MODE  or ib.data.value == nil then
```

hcarter-775: and the same with the other handlers

nickolas-deboom: The problem with that is I have seen a Zemismart bulb report that its color mode was `COLOR_TEMPERATURE` in its subscription report following a device power cycle, even though it was previously `CURRENT_HUE_SAT` and the bulb was clearly still emitting a color. It didn't seem to report the proper color mode until a color change command was sent to the device and forced it to update. So with the current check, this problem would be avoided at least.

PR #2028, opened 2025-03-27
Improve test coverage and backwards compatibility for matter switch tests

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Test case update

# Description of Change

Add backwards compatibility for matter switch test cases down to lua libs version v10 and improve overall test coverage. Most of the issues were from test cases expecting native handler registrations, which was added in v11. This PR also fixes a bug in the driver code for lua libs version v10 and lower, which prevented the energyMeter capability from being emitted.

It might be possible to extend support to even lower versions, but there are a few additional api changes that would need to be accounted for.

Also note that to run these tests against lower versions, any capabilities that are not available in that version of the lua libs would need to be manually added to prevent test failures.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -1311,25 +1311,24 @@ end
 
 local temp_attr_handler_factory = function(minOrMax)
   return function(driver, device, ib, response)
-    if ib.data.value == nil then
-      return
-    end
-    local temp = ib.data.value / 100.0
-    local unit = "C"
-    set_field_for_endpoint(device, TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp)
-    local min = get_field_for_endpoint(device, TEMP_BOUND_RECEIVED..TEMP_MIN, ib.endpoint_id)
-    local max = get_field_for_endpoint(device, TEMP_BOUND_RECEIVED..TEMP_MAX, ib.endpoint_id)
-    if min ~= nil and max ~= nil then
-      if min < max then
-        -- Only emit the capability for RPC version >= 5 (unit conversion for
-        -- temperature range capability is only supported for RPC >= 5)
-        if version.rpc >= 5 then
-          device:emit_event_for_endpoint(ib.endpoint_id, capabilities.temperatureMeasurement.temperatureRange({ value = { minimum = min, maximum = max }, unit = unit }))
+    if ib.data.value ~= nil then
+      local temp = ib.data.value / 100.0
+      local unit = "C"
+      set_field_for_endpoint(device, TEMP_BOUND_RECEIVED..minOrMax, ib.endpoint_id, temp)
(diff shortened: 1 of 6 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -363,10 +363,6 @@ local function convert_huesat_st_to_matter(val)
 end
 
 local function mired_to_kelvin(value, minOrMax)
-  if value == 0 then -- shouldn't happen, but has
```
nickolas-deboom, line 366: This code is unreachable because 0 is outside of the range of "sane" values checked at the call site

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -377,10 +373,8 @@ local function mired_to_kelvin(value, minOrMax)
   local rounding_value = 0.5
   if minOrMax == COLOR_TEMP_MIN then
     return utils.round(MIRED_KELVIN_CONVERSION_CONSTANT / (kelvin_step_size * (value + 1)) + rounding_value) * kelvin_step_size
-  elseif minOrMax == COLOR_TEMP_MAX then
+  else -- minOrMax = COLOR_TEMP_MAX
     return utils.round(MIRED_KELVIN_CONVERSION_CONSTANT / (kelvin_step_size * (value - 1)) - rounding_value) * kelvin_step_size
-  else
```
nickolas-deboom, line 382: This else is unreachable because this function is only called with `COLOR_TEMP_MIN` or `COLOR_TEMP_MAX`, so it can be removed

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -392,11 +386,6 @@ local function device_type_supports_button_switch_combination(device, endpoint_i
     if ep.endpoint_id == endpoint_id then
       for _, dt in ipairs(ep.device_types) do
         if dt.device_type_id == DIMMABLE_LIGHT_DEVICE_TYPE_ID then
-          for _, fingerprint in ipairs(child_device_profile_overrides_per_vendor_id[0x115F]) do
```
nickolas-deboom, line 395: This Aqara device uses device type 0x100, while `DIMMABLE_LIGHT_DEVICE_TYPE_ID` is 0x101, so this code can be removed

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -548,9 +536,6 @@ local function do_configure(driver, device)
 end
 
 local function configure_buttons(device)
-  if device.network_type == device_lib.NETWORK_TYPE_CHILD then
```
nickolas-deboom, line 551: This check is already done at both call sites of `configure_buttons` so it can be removed

hcarter-775:
do you think this is worth a rebase? Should we close this? Just checking since it's been open for ~6 months.

nickolas-deboom:
> do you think this is worth a rebase? Should we close this? Just checking since it's been open for ~6 months.

I would say probably not because backwards compatibility was broken by scripting engine changes. Closing

PR #2032, opened 2025-03-28
Matter Switch: Support Fan/Light device

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change adds driver support for a new profile for a fan+color light device added by a previous PR, brings in handlers for fan attributes and capabilities, and adds new logic to select the new profile and create the component map.

# Summary of Completed Tests

Tested with: Orein fan+light and regression tested with Inovelli dimmer switch, Ikea 7 button remote, and VDA water valve device.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -289,7 +289,7 @@ local HELD_THRESHOLD = 1
 -- this is the number of buttons for which we have a static profile already made
 local STATIC_BUTTON_PROFILE_SUPPORTED = {1, 2, 3, 4, 5, 6, 7, 8}
 
-local DEVICE_PROFILED = "__device_profiled"
+local DEVICE_PROFILED = "__button_device_profiled"
 
 -- Some switches will send a MultiPressComplete event as part of a long press sequence. Normally the driver will create a
 -- button capability event on receipt of MultiPressComplete, but in this case that would result in an extra event because
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -50,12 +50,9 @@ local SWITCH_INITIALIZED = "__switch_intialized"
 -- in the device table for devices that joined prior to this transition, and it
 -- will not be set for new devices.
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
--- COMPONENT_TO_ENDPOINT_MAP_BUTTON is for devices with button endpoints, to
--- preserve the MCD functionality for button devices from the matter-button
--- driver after it was merged into the matter-switch driver. Note that devices
--- containing both button endpoints and switch endpoints will use this field
--- rather than COMPONENT_TO_ENDPOINT_MAP.
-local COMPONENT_TO_ENDPOINT_MAP_BUTTON = "__component_to_endpoint_map_button"
+-- COMPONENT_TO_ENDPOINT_MAP_NEW_DEVICES is for new devices that can be supported
+-- by a MCD configuration.
+local COMPONENT_TO_ENDPOINT_MAP_NEW_DEVICES = "__component_to_endpoint_map_button"
 local ENERGY_MANAGEMENT_ENDPOINT = "__energy_management_endpoint"
 local IS_PARENT_CHILD_DEVICE = "__is_parent_child_device"
 local COLOR_TEMP_BOUND_RECEIVED_KELVIN = "__colorTemp_bound_received_kelvin"
@@ -291,7 +288,7 @@ local HELD_THRESHOLD = 1
 -- this is the number of buttons for which we have a static profile already made
 local STATIC_BUTTON_PROFILE_SUPPORTED = {1, 2, 3, 4, 5, 6, 7, 8}
 
-local BUTTON_DEVICE_PROFILED = "__button_device_profiled"
+local DEVICE_PROFILED = "__device_profiled"
 
 -- Some switches will send a MultiPressComplete event as part of a long press sequence. Normally the driver will create a
 -- button capability event on receipt of MultiPressComplete, but in this case that would result in an extra event because
@@ -456,7 +453,7 @@ local function find_default_endpoint(device)
 end
 
 local function component_to_endpoint(device, component)
-  local map = device:get_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON) or device:get_field(COMPONENT_TO_ENDPOINT_MAP) or {}
+  local map = device:get_field(COMPONENT_TO_ENDPOINT_MAP_NEW_DEVICES) or device:get_field(COMPONENT_TO_ENDPOINT_MAP) or {}
   if map[component] then
     return map[component]
   end
@@ -464,7 +461,7 @@ local function component_to_endpoint(device, component)
 end
 
 local function endpoint_to_component(device, ep)
-  local map = device:get_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON) or device:get_field(COMPONENT_TO_ENDPOINT_MAP) or {}
+  local map = device:get_field(COMPONENT_TO_ENDPOINT_MAP_NEW_DEVICES) or device:get_field(COMPONENT_TO_ENDPOINT_MAP) or {}
   for component, endpoint in pairs(map) do
     if endpoint == ep then
       return component
@@ -513,7 +510,7 @@ local function assign_child_profile(device, child_ep)
 end
 
 local function do_configure(driver, device)
-  if device:get_field(BUTTON_DEVICE_PROFILED) then
+  if device:get_field(DEVICE_PROFILED) then
     return
   end
   local level_eps = embedded_cluster_utils.get_endpoints(device, clusters.LevelControl.ID)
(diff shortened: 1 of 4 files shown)


hcarter-775:
Can you do an onboarding test of a light-button device as well?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -585,44 +583,65 @@ local function find_child(parent, ep_id)
...
-local function try_build_button_component_map(device, main_endpoint, button_eps)
-  -- create component mapping on the main profile button endpoints
-  if STATIC_BUTTON_PROFILE_SUPPORTED[#button_eps] then
-    local component_map = {}
-    component_map["main"] = main_endpoint
-    for component_num, ep in ipairs(button_eps) do
-      if ep ~= main_endpoint then
-        local button_component = "button" .. component_num
-        component_map[button_component] = ep
+local function build_component_map(device, main_endpoint, endpoints)
```
hcarter-775, line 586: we don't need to send the endpoints in as a parameter anymore. Let's just set an endpoints variable in the if-else check whether to use button or fan components.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -656,58 +675,63 @@ local function try_build_child_switch_profiles(driver, device, switch_eps, main_
...
-  -- Without support for bindings, only clusters that are implemented as server are counted. This count is handled
-  -- while building switch child profiles
-  local num_switch_server_eps = try_build_child_switch_profiles(driver, device, switch_eps, main_endpoint)
-
+  -- Button/fan endpoints will be added as additional components in the profile containing the main_endpoint.
+  -- The resulting endpoint to component map is saved in the COMPONENT_TO_ENDPOINT_MAP_NEW_DEVICES field
   if #button_eps > 0 then
-    build_button_profile(device, main_endpoint, #button_eps)
+    build_mcd_profile(device, main_endpoint)
+    build_component_map(device, main_endpoint, button_eps)
```
hcarter-775, line 714: We don't need this if #button_eps > 0  and @fan_eps > 0 check here, since the same logic is handled within the `build_mcd_profile` and `build_component_map` function.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -656,58 +675,63 @@ local function try_build_child_switch_profiles(driver, device, switch_eps, main_
...
-    local cluster_id = 0
-    for _, ep in ipairs(device.endpoints) do
-      -- main_endpoint only supports server cluster by definition of get_endpoints()
-      if main_endpoint == ep.endpoint_id then
-        for _, dt in ipairs(ep.device_types) do
-          -- no device type that is not in the switch subset should be considered.
-          if (ON_OFF_SWITCH_ID <= dt.device_type_id and dt.device_type_id <= ON_OFF_COLOR_DIMMER_SWITCH_ID) then
-            cluster_id = math.max(cluster_id, dt.device_type_id)
-          end
+local function handle_light_switch_with_onOff_server_clusters(device, main_endpoint)
```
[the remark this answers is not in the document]
nickolas-deboom: This body of this function was indented two spaces too far so I just un-indented it.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: By that thinking, we also don't need the fan_eps or button_eps reads here.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -585,44 +583,65 @@ local function find_child(parent, ep_id)
...
-        component_map[button_component] = ep
+local function build_component_map(device, main_endpoint, endpoints)
+  local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
+  local fan_eps = device:get_endpoints(clusters.FanControl.ID)
+  local component_name
+
+  if #button_eps > 0 and STATIC_BUTTON_PROFILE_SUPPORTED[#button_eps] then
+    component_name = "button"
+  elseif #fan_eps > 0 then
+    component_name = "fan"
```
ctowns, line 594: Could you remind me why we are making this device type an MCD? If there are no duplicate capabilities in the profile, we could get away with just a single component device, and that seems simpler to implement. I think this solution here is implemented correctly, I just am curious the reasoning behind the choice to go MCD - did the partner ask for it? It would all still be in the same device card regardless

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -585,44 +583,65 @@ local function find_child(parent, ep_id)
...
+    device.log.warn_with({hub_logs = true}, "Device is not supported by a multicomponent configuration")
+    return
+  end
+
+  local component_map = {}
+  component_map["main"] = main_endpoint
+  for component_num, ep in ipairs(endpoints) do
+    if ep ~= main_endpoint then
+      local component = component_name
+      if #endpoints > 1 then
```
ctowns, line 605: nit: this will be run for both fan and button devices, but we don't have support for a multi fan device. Would it make sense to just move this up to the button-specific conditional above and the just directly set the `component_map["fan"] value in the fan conditional? Open to thoughts!

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: I had a similar question, and after asking Nick I believe it came down to the fact that the MCD version looked/felt better on the device card.

nickolas-deboom: Yes that's right. I was also thinking that it's consistent with the light-level-button profile, which only has one button but is still MCD. We can definitely do single component though, is that more preferable?

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: The only issue with that approach would be that we wouldn't know the endpoint number of the fan unless we do a loop like this.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Good call, I made this update in the latest commit

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I updated this in the latest commit. I also realized we similarly didn't need to check for switch endpoints in this function and can do it from `build_child_switch_profiles` instead. All of these functions can safely run for any device without having to worry about which endpoints the device supports.

Also, I decided to name `initialize_buttons_and_switches` back to `initialize_switch` since it deals with fan endpoints too now (I can change that back if you prefer the other name though)

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Yes, even if in this case we decide not to implement this device as MCD for the reasons you've expressed above, I think it might be worthwhile to open a conversation about improving these things, especially with the complex device type support that we can provide with the Matter spec. As it is, lumping an Air Purifier / Air Quality Sensor / Fan together into a single component seems a little inadequate.

nickolas-deboom: Yes, that's what I meant when I said the presentation looked better - it nicely groups the light capabilities and then has a separate section labeled Fan for the fan capabilities. I tried both ways in the past and have to say I like the look of having them separated. But of course, it's not necessary as you outlined, and I'll take another look and see how a single component could simplify things in this PR.

I also agree with Harrison, because as we keep adding support for these compound devices this topic is going to continue to come up.

nickolas-deboom: The latest commit highlights what the changes are for configuring the device as a single component rather than MCD. The difference is less than what I recalled - it's pretty much just the Fan label that is not present in the single component version. Here are screenshots from both configurations for reference:
<details>
<summary>MCD:</summary>


</details>

<details>
<summary>Single component:</summary>


<details>

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -650,64 +665,53 @@ local function try_build_child_switch_profiles(driver, device, switch_eps, main_
...
-local function initialize_buttons_and_switches(driver, device, main_endpoint)
-  local switch_eps = device:get_endpoints(clusters.OnOff.ID)
-  local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
-  table.sort(switch_eps)
-  table.sort(button_eps)
-
-  -- All button endpoints found will be added as additional components in the profile containing the main_endpoint.
-  -- The resulting endpoint to component map is saved in the COMPONENT_TO_ENDPOINT_MAP_BUTTON field
-  try_build_button_component_map(device, main_endpoint, button_eps)
+local function initialize_switch(driver, device, main_endpoint)
```
hcarter-775, line 698: I personally still prefer the initialize_buttons_and_switches name, since button devices are unrelated to switches. At the very least, I think initialize_switch is an outdated name for the current functionality of this configration logic.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -585,46 +586,60 @@ local function find_child(parent, ep_id)
...
+        end
         component_map[button_component] = ep
       end
     end
     device:set_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON, component_map, {persist = true})
   end
 end
 
-local function build_button_profile(device, main_endpoint, num_button_eps)
+local function build_profile(device, main_endpoint)
```
hcarter-775, line 609: I sorta don't see why we're generalizing this functionality anymore, now that the fan device is not a mcd device.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: lol, the irony of that was not lost on me when I was typing this. But in that case, we'd want to specify switches and onOff devices or something like that. I could get down with that.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: True, maybe it would be better to initialize the fan in do_configure

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -585,46 +586,60 @@ local function find_child(parent, ep_id)
   return parent:get_child_by_parent_assigned_key(string.format("%d", ep_id))
 end
 
-local function try_build_button_component_map(device, main_endpoint, button_eps)
```
hcarter-775, line 588: same here, now this function name is a little too generically named

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Aren't onOff devices switches :flushed:

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Yeah either that or do an early return at the top of the initialize_switch function (or whatever we end up calling it) since it is an OnOff (switch) device

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: yes, that's true. I think I'd like to define this function either by the capabilities we map to or by the clusters we map from, not both (even if the word switch pops up for both). But for now, we seem to do most of our variable naming based on the capabilities we map to (switch_eps, button_eps) etc.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Changed back in latest commit

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -650,64 +652,56 @@ local function try_build_child_switch_profiles(driver, device, switch_eps, main_
...
 
-local function initialize_buttons_and_switches(driver, device, main_endpoint)
-  local switch_eps = device:get_endpoints(clusters.OnOff.ID)
+local function initialize_switch(driver, device, main_endpoint)
   local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
-  table.sort(switch_eps)
-  table.sort(button_eps)
-
-  -- All button endpoints found will be added as additional components in the profile containing the main_endpoint.
-  -- The resulting endpoint to component map is saved in the COMPONENT_TO_ENDPOINT_MAP_BUTTON field
```
hcarter-775, line 692: Not sure I agree with removing this comment

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: I still feel relatively strongly that this name should not be changed back to `initialize_switch`

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Restored :+1:

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Sure thing, changed it back :+1:

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I moved it to do_configure, let me know if that looks good to you

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: After some discussion, I changed it to use a single component configuration.

hcarter-775 (approved):

PR #2041, opened 2025-04-01
Matter Switch move button and switch initialization to doConfigure

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [x] Refactor

# Checklist

# Description of Change

This change moves the initialization logic for buttons and switches to `do_configure`. This consolidates all of the profile selection logic to be within doConfigure and allows the removal of logic gates from `device_init` that were there to ensure init code only ran one time.

Also added is a new function that runs at init that can rename or delete persisted fields on the device. The original `__component_to_endpoint_map` field can now be utilized by buttons and other devices because `initialize_buttons_and_switches` is ensured to only run one time, meaning that old MCD switch devices will not be affected. Also, the `__switch_intialized` field can now be deleted from devices.

# Summary of Completed Tests

Tested with various matter buttons and lights to ensure no change in behavior.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
+++ b/drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
@@ -1,3 +1,17 @@
+-- Copyright 2025 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
 local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
 local device_lib = require "st.device"
@@ -19,10 +33,11 @@ local CUBEACTION_TIMER = "__cubeAction_timer"
 local CUBEACTION_TIME = 3
 
 local function is_aqara_cube(opts, driver, device)
-  local name = string.format("%s", device.manufacturer_info.product_name)
-  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
-    string.find(name, "Aqara Cube T1 Pro") then
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    local name = string.format("%s", device.manufacturer_info.product_name)
+    if string.find(name, "Aqara Cube T1 Pro") then
       return true
+    end
   end
   return false
 end
@@ -186,6 +201,12 @@ local function info_changed(driver, device, event, args)
   end
 end
 
+-- override do_configure to prevent it running in the main driver
+local function do_configure(driver, device) end
+
+-- override driver_switched to prevent it running in the main driver
+local function driver_switched(driver, device) end
+
 local function initial_press_event_handler(driver, device, ib, response)
   if get_field_for_endpoint(device, INITIAL_PRESS_ONLY, ib.endpoint_id) then
     local map = device:get_field(COMPONENT_TO_ENDPOINT_MAP_BUTTON) or {}
@@ -214,7 +235,9 @@ local aqara_cube_handler = {
   lifecycle_handlers = {
     init = device_init,
     added = device_added,
-    infoChanged = info_changed
+    infoChanged = info_changed,
+    doConfigure = do_configure,
+    driverSwitched = driver_switched
   },
   matter_handlers = {
     attr = {
--- a/drivers/SmartThings/matter-switch/src/eve-energy/init.lua
+++ b/drivers/SmartThings/matter-switch/src/eve-energy/init.lua
@@ -1,4 +1,4 @@
--- Copyright 2023 SmartThings
+-- Copyright 2025 SmartThings
 --
 -- Licensed under the Apache License, Version 2.0 (the "License");
 -- you may not use this file except in compliance with the License.
@@ -277,6 +277,12 @@ local function device_removed(driver, device)
   delete_poll_schedule(device)
 end
 
+-- override do_configure to prevent it running in the main driver
+local function do_configure(driver, device) end
+
+-- override driver_switched to prevent it running in the main driver
+local function driver_switched(driver, device) end
+
 local function handle_refresh(self, device)
   requestData(device)
 end
@@ -368,6 +374,8 @@ local eve_energy_handler = {
     init = device_init,
     added = device_added,
     removed = device_removed,
+    doConfigure = do_configure,
+    driverSwitched = driver_switched
   },
   matter_handlers = {
     attr = {
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -1,4 +1,4 @@
--- Copyright 2022 SmartThings
+-- Copyright 2025 SmartThings
 --
 -- Licensed under the Apache License, Version 2.0 (the "License");
 -- you may not use this file except in compliance with the License.
@@ -44,19 +44,12 @@ local SWITCH_LEVEL_LIGHTING_MIN = 1
 local CURRENT_HUESAT_ATTR_MIN = 0
 local CURRENT_HUESAT_ATTR_MAX = 254
 
-local SWITCH_INITIALIZED = "__switch_intialized"
--- COMPONENT_TO_ENDPOINT_MAP is here only to preserve the endpoint mapping for
+-- COMPONENT_TO_ENDPOINT_MAP is here to preserve the endpoint mapping for
 -- devices that were joined to this driver as MCD devices before the transition
--- to join all matter-switch devices as parent-child. This value will only exist
--- in the device table for devices that joined prior to this transition, and it
--- will not be set for new devices.
+-- to join switch devices as parent-child. This value will exist in the device
+-- table for devices that joined prior to this transition, and is also used for
+-- button devices that require component mapping.
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
--- COMPONENT_TO_ENDPOINT_MAP_BUTTON is for devices with button endpoints, to
--- preserve the MCD functionality for button devices from the matter-button
--- driver after it was merged into the matter-switch driver. Note that devices
--- containing both button endpoints and switch endpoints will use this field
--- rather than COMPONENT_TO_ENDPOINT_MAP.
-local COMPONENT_TO_ENDPOINT_MAP_BUTTON = "__component_to_endpoint_map_button"
 local ENERGY_MANAGEMENT_ENDPOINT = "__energy_management_endpoint"
 local IS_PARENT_CHILD_DEVICE = "__is_parent_child_device"
 local COLOR_TEMP_BOUND_RECEIVED_KELVIN = "__colorTemp_bound_received_kelvin"
@@ -67,6 +60,12 @@ local LEVEL_BOUND_RECEIVED = "__level_bound_received"
 local LEVEL_MIN = "__level_min"
 local LEVEL_MAX = "__level_max"
 local COLOR_MODE = "__color_mode"
+
+local updated_fields = {
+  { current_field_name = "__component_to_endpoint_map_button", updated_field_name = COMPONENT_TO_ENDPOINT_MAP },
+  { current_field_name = "__switch_intialized", updated_field_name = nil }
+}
+
 local HUE_SAT_COLOR_MODE = clusters.ColorControl.types.ColorMode.CURRENT_HUE_AND_CURRENT_SATURATION
 local X_Y_COLOR_MODE = clusters.ColorControl.types.ColorMode.CURRENTX_AND_CURRENTY
 
@@ -292,8 +291,6 @@ local HELD_THRESHOLD = 1
 -- this is the number of buttons for which we have a static profile already made
 local STATIC_BUTTON_PROFILE_SUPPORTED = {1, 2, 3, 4, 5, 6, 7, 8}
 
-local BUTTON_DEVICE_PROFILED = "__button_device_profiled"
-
 -- Some switches will send a MultiPressComplete event as part of a long press sequence. Normally the driver will create a
 -- button capability event on receipt of MultiPressComplete, but in this case that would result in an extra event because
 -- the "held" capability event is generated when the LongPress event is received. The IGNORE_NEXT_MPC flag is used
@@ -406,6 +403,7 @@ local function device_type_supports_button_switch_combination(device, endpoint_i
 end
 
 local function get_first_non_zero_endpoint(endpoints)
+  table.sort(endpoints)
   for _,ep in ipairs(endpoints) do
     if ep ~= 0 then -- 0 is the matter RootNode endpoint
       return ep
@@ -425,8 +423,6 @@ local function find_default_endpoint(device)
 
   local switch_eps = device:get_endpoints(clusters.OnOff.ID)
   local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
-  table.sort(switch_eps)
-  table.sort(button_eps)
 
   -- Return the first switch endpoint as the default endpoint if no button endpoints are present
   if #button_eps == 0 and #switch_eps > 0 then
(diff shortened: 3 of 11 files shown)


hcarter-775:
I found this line: -- The resulting endpoint to component map is saved in the COMPONENT_TO_ENDPOINT_MAP_BUTTON field
Can you update it to the correct field name?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -472,6 +467,17 @@ local function endpoint_to_component(device, ep)
   return "main"
 end
 
+local function check_field_name_updates(device)
```
hcarter-775, line 470: this is a very critical function. It definitely looks fine to me, but it may be best to make this functionality as explicit as possible. I think being able to reason about this very clearly is important.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -760,6 +722,47 @@ local function device_init(driver, device)
...
 
+local function do_configure(driver, device)
+  if not detect_bridge(device) then
+    local main_endpoint = find_default_endpoint(device)
+    -- initialize the main device card with buttons if applicable, and create child devices as needed for multi-switch devices.
+    initialize_buttons_and_switches(driver, device, main_endpoint)
+    if device:get_field(IS_PARENT_CHILD_DEVICE) then
+      device:set_find_child(find_child)
+    end
+    if device:get_field(BUTTON_DEVICE_PROFILED) then
```
hcarter-775, line 733: since we've moved this logic, I don't think we need the check to include a field anymore, right? We could have the  `initialize_buttons_and_switches` function return whether or not a profile was found mapped to. What do you think?

nickolas-deboom: Yeah good point, and I like that idea

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -760,6 +722,47 @@ local function device_init(driver, device)
   device:subscribe()
 end
 
+local function do_configure(driver, device)
+  if not detect_bridge(device) then
+    local main_endpoint = find_default_endpoint(device)
+    -- initialize the main device card with buttons if applicable, and create child devices as needed for multi-switch devices.
+    initialize_buttons_and_switches(driver, device, main_endpoint)
+    if device:get_field(IS_PARENT_CHILD_DEVICE) then
+      device:set_find_child(find_child)
```
ctowns, line 731: `device:set_find_child(find_child)` will still need to be set on init as it should be called each time the driver restarts.

As a good test case for these changes, we should test various device types and ensure that they continue to work after a driver restart. Onboarding a parent-child device and then confirming that the child devices continue to work as expected after a driver restart would be a good test case!

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -43,19 +44,7 @@ local SWITCH_LEVEL_LIGHTING_MIN = 1
...
--- to join all matter-switch devices as parent-child. This value will only exist
--- in the device table for devices that joined prior to this transition, and it
--- will not be set for new devices.
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
--- COMPONENT_TO_ENDPOINT_MAP_BUTTON is for devices with button endpoints, to
--- preserve the MCD functionality for button devices from the matter-button
--- driver after it was merged into the matter-switch driver. Note that devices
--- containing both button endpoints and switch endpoints will use this field
--- rather than COMPONENT_TO_ENDPOINT_MAP.
-local COMPONENT_TO_ENDPOINT_MAP_BUTTON = "__component_to_endpoint_map_button"
```
ctowns, line 58: What as the original intent behind using a different field specifically for the button MCD devices? Was it because of the check in `init` that essentially bypassed MCD devices? That sounds right to me but I just want to make sure there wasn't another intent

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -760,6 +722,47 @@ local function device_init(driver, device)
   device:subscribe()
 end
 
+local function do_configure(driver, device)
```
ctowns, line 725: Is there a reason to put this in `doConfigure` instead of `added`? I think that we have effectively treated these functions as nearly interchangable "one-time configuration" functions, but it may be worth setting a precedent with this change since this is our biggest and most used driver. Any thoughts on which lifecycle event we should use for this?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -760,6 +722,47 @@ local function device_init(driver, device)
...
+  if not detect_bridge(device) then
+    local main_endpoint = find_default_endpoint(device)
+    -- initialize the main device card with buttons if applicable, and create child devices as needed for multi-switch devices.
+    initialize_buttons_and_switches(driver, device, main_endpoint)
```
ctowns, line 729: Also since `doConfigure` could be called multiple times, we want to make sure functions like this can be run multiple times without issue in the same lifecycle. I don't suspect any issue from what I see but just wanted to call it out.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Thank you, that's a great call out.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: There should be no issue but I do see your concern. However, from my understanding `doConfigure` would only be called multiple times if the device is set to the provisioned state, and it that case it seems that it would be a good idea to re-run these init functions.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: There are some calls in `device_init` that are required to run before `initialize_buttons_and_switches` (`device:set_component_to_endpoint_fn(component_to_endpoint)` for example), so I think `doConfigure` is the best place for this code since it will run after `device_init` (while `added` precedes `device_init`)

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I updated this in the latest commit

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: After trying this out, it felt a bit messy trying to get the profile name from `try_build_button_component_map` back to `doConfigure`. It feels easier to just use the field, but now since we only need it a single time we can set it to nil after using it. Do you agree with this approach?

hcarter-775: Here's what I'm imagining.

In the do_configure handling:
```
    local profile_found = initialize_buttons_and_switches(driver, device, main_endpoint)
    if profile_found then
      return
    end
```

Then, in `initialize_buttons_and_switches`, we could do something like:
```
local function initialize_buttons_and_switches(driver, device, main_endpoint)
  ...
  if #button_eps > 0 then
    build_button_profile(device, main_endpoint, #button_eps)
    configure_buttons(device)
    return true
  end
  ...
  if num_switch_server_eps > 0 and detect_matter_thing(device) then
    handle_light_switch_with_onOff_server_clusters(device, main_endpoint, num_switch_server_eps)
    return true
  end
```
the button path is actually identical per the current logic, and I added another check for the `num_switch_server_eps > 0` case that I think was missed before.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Further, since `doConfigure` is tied to whether or not the provisioning state has been changed to provisioned while `added` is also related to a driver change, `doConfigure` is the more appropriate place for a "true" one time handling like device profiling.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -286,8 +286,6 @@ local HELD_THRESHOLD = 1
 -- this is the number of buttons for which we have a static profile already made
 local STATIC_BUTTON_PROFILE_SUPPORTED = {1, 2, 3, 4, 5, 6, 7, 8}
 
-local BUTTON_DEVICE_PROFILED = "__button_device_profiled"
```
hcarter-775, line 294: we should also set this field to nil since we have that handy function to remove these en masse

nickolas-deboom:
> I found this line: -- The resulting endpoint to component map is saved in the COMPONENT_TO_ENDPOINT_MAP_BUTTON field Can you update it to the correct field name?

Good catch! Fixed in latest commit

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Do you think there is any risk that the command to copy the old field to the new field could fail, and then the next line would set the old field to nil? I kind of doubt that it could happen but I suppose it's possible

hcarter-775: I suppose maybe that could happen? but a driver crash wouldn't cause that, so like lua itself and the lua runtime would have to fail. That's not something we're really capable of worrying about I don't think

nickolas-deboom: Yeah, you're probably right.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -473,6 +469,17 @@ local function endpoint_to_component(device, ep)
   return "main"
 end
 
+local function check_field_name_updates(device)
+  for _, field in ipairs(updated_fields) do
+    if device:get_field(field.field_name) then
```
hcarter-775, line 474: lets call `field_name` `current_field_name` for clarity. also, I don't think we need the `if field.updated_field_name ~= nil`, since this is already handled in the `set_field` definition in lua libs. We also remove the data either way, so it doesn't change the handling if we set data to a nil field that doesn't exist.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Adding to this: another reason to put this handling in `doConfigure` is because it truly is a one time configuration function. `added` for example will be re-run in the standard case of a hub failover, since the device will be put in a new driver.

hcarter-775 (approved):

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I updated the name, but think we should leave the check because I tried it without and get an error if the field is nil:
```
Received unexpected error: ...eboom/Documents/dev/scripting-engine/lua_libs/cosock.lua:252: ...ocuments/dev/scripting-engine/lua_libs/st/dispatcher.lua:270: Error encountered while processing event for <MatterDevice: 00000000-1111-2222-3333-000000000001 [nil] (Matter Switch)>:
    arg1: init
"./init.lua:475: Field key cannot be nil"
stack traceback:
	[C]: in function 'error'
	...om/Documents/dev/scripting-engine/lua_libs/st/thread.lua:88: in function <...om/Documents/dev/scripting-engine/lua_libs/st/thread.lua:55>
stack traceback:
	[C]: in function 'error'
	...eboom/Documents/dev/scripting-engine/lua_libs/cosock.lua:252: in upvalue 'step_thread'
	...eboom/Documents/dev/scripting-engine/lua_libs/cosock.lua:360: in function 'cosock.run'
	...om/Documents/dev/scripting-engine/lua_libs/st/driver.lua:1144: in function 'st.driver.run'
	./init.lua:1654: in main chunk
	[C]: in function 'require'
	[C]: in function 'xpcall'
	.../dev/scripting-engine/lua_libs/integration_test/init.lua:242: in upvalue 'run_configured_test'
	.../dev/scripting-engine/lua_libs/integration_test/init.lua:376: in function 'integration_test.run_registered_tests'
	test/test_multi_switch_parent_child_plugs.lua:357: in main chunk
	[C]: in ?
```

hcarter-775: ah I see, they mark it as an error. That's fine then, thanks for checking :+1:

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -44,19 +44,12 @@ local SWITCH_LEVEL_LIGHTING_MIN = 1
...
+-- to join switch devices as parent-child. This value will exist in the device
+-- table for devices that joined prior to this transition, and is also used for
+-- button devices that require component mapping.
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
--- COMPONENT_TO_ENDPOINT_MAP_BUTTON is for devices with button endpoints, to
--- preserve the MCD functionality for button devices from the matter-button
--- driver after it was merged into the matter-switch driver. Note that devices
--- containing both button endpoints and switch endpoints will use this field
--- rather than COMPONENT_TO_ENDPOINT_MAP.
-local COMPONENT_TO_ENDPOINT_MAP_BUTTON = "__component_to_endpoint_map_button"
```
ctowns, line 59: How do we know that this is safe to remove? If this field is removed, it could leave users who had an older driver stranded with a potentially broken device because we'd no only have the component mapping.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: After discussion offline, we are going to leave the init code in doConfigure. It should also be called when a driverSwitched event is received, so a handler for that has been added.

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -19,10 +33,11 @@ local CUBEACTION_TIMER = "__cubeAction_timer"
 local CUBEACTION_TIME = 3
 
 local function is_aqara_cube(opts, driver, device)
-  local name = string.format("%s", device.manufacturer_info.product_name)
```
nickolas-deboom, line 22: `device.manufacturer_info` does not exist for child devices. Moving this statement inside the if statement prevents an error from occurring here.

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
```diff
@@ -277,6 +277,12 @@ local function device_removed(driver, device)
   delete_poll_schedule(device)
 end
 
+-- override do_configure to prevent it running in the main driver
+local function do_configure(driver, device) end
```
ctowns, line 281: Could the same `init` code that exists here be move to the corresponding `doConfigure` and `driverSwitched` handlers in this sub driver?

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
```diff
@@ -19,10 +33,11 @@ local CUBEACTION_TIMER = "__cubeAction_timer"
...
-  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
-    string.find(name, "Aqara Cube T1 Pro") then
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    local name = string.format("%s", device.manufacturer_info.product_name)
```
ctowns, line 37: I know you didn't write this as part of these changes, but I think we could just reference the `device.manufacturer_info.product_name` directly rather than printing it to a string. I'm not sure if there is a good reason to do that first?

nickolas-deboom: I agree but I'm not sure I'd like to change that in this PR, in case there was a reason it was done this way

drivers/SmartThings/matter-switch/src/eve-energy/init.lua
nickolas-deboom: Perhaps, but a lot of it doesn't apply to the Eve Energy so we'd need to choose the right pieces to make it specific to this device

nickolas-deboom: Perhaps, but a lot of it doesn't apply to the Eve Energy so we'd need to choose the right pieces to make it specific to this device.

nickolas-deboom: I removed the changes to the Eve and Aqara subdrivers from this PR and moved them along with other changes to make them more consistent to https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2116

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
nickolas-deboom: I made this change in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2116 so it can be reviewed separately

drivers/SmartThings/matter-switch/src/aqara-cube/init.lua
nickolas-deboom: I moved this to https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2116

ctowns (approved):

PR #2044, opened 2025-04-02
Early return and Re-read In Case of SystemMode Race Conditions

hcarter-775:
# Description of Change
The current SystemMode / ControlSequenceofOperation handling expects for ControlSequenceofOperation to be handled first in all cases, but this is not necessarily the case. This change ensures a particular ordering by early exiting the SystemMode handler if a table that is initialized in the ControlSequenceofOperation handler has not been initialized yet. Then, in the ControlSequenceofOperation handler, it always does a read of SystemMode. At worst, this will cause a dual handling of a single SystemMode value, which will not do anything detrimental. Further, ControlSequenceofOperation should only be run on a subscription.  

# Summary of Completed Tests
Tested on device by SRPOL.
Uni tests updated.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -1073,6 +1073,11 @@ local function humidity_attr_handler(driver, device, ib, response)
 end
 
 local function system_mode_handler(driver, device, ib, response)
+  if device:get_field(OPTIONAL_THERMOSTAT_MODES_SEEN) == nil then -- this being nil means the sequence_of_operation_handler hasn't run.
+    device.log.info_with({hub_logs = true}, "In the SystemMode handler, ControlSequenceOfOperation has not run yet. Aborting.")
+    return
+  end
+
   local supported_modes = device:get_latest_state(device:endpoint_to_component(ib.endpoint_id), capabilities.thermostatMode.ID, capabilities.thermostatMode.supportedThermostatModes.NAME) or {}
   -- check that the given mode was in the supported modes list
   if tbl_contains(supported_modes, THERMOSTAT_MODE_MAP[ib.data.value].NAME) then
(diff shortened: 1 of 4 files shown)


nickolas-deboom (approved):

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1073,6 +1073,11 @@ local function humidity_attr_handler(driver, device, ib, response)
 end
 
 local function system_mode_handler(driver, device, ib, response)
+  if device:get_field(OPTIONAL_THERMOSTAT_MODES_SEEN) == nil then -- this being nil means the sequence_of_operation_handler hasn't run.
+    device.log.info_with({hub_logs = true}, "In the SystemMode handler, ControlSequenceOfOperation has not run yet. Aborting.")
```
ctowns, line 1077: Why does this need to be info level in our hub logs?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1161,6 +1166,9 @@ local function sequence_of_operation_handler(driver, device, ib, response)
   device:set_field(DISALLOWED_THERMOSTAT_MODES, disallowed_mode_operations)
   local event = capabilities.thermostatMode.supportedThermostatModes(supported_modes, {visibility = {displayed = false}})
   device:emit_event_for_endpoint(ib.endpoint_id, event)
+
+  -- do an extra SystemMode read in case of out of order attribute handling
+  device:send(clusters.Thermostat.attributes.SystemMode:read(device))
```
ctowns, line 1171: Is an extra read neccessary? We have all the information we need at this point, right? So couldn't we just call into a helper function with the relevant information such as the supported modes and the optional modes seen? The same helper function could be used by the attribute handler as well, I'm just not sure why we require an additional device read

PR #2045, opened 2025-04-02
Update smoke co alarm fault support

hcarter-775:
# Description of Change
Support the SmokeCOAlarm attribute BatteryAlert through the hardwareFault capability. Support the batteryLevel capability with the PowerSource attribute BatChargeLevel, and include battery capability support with the PowerSource attribute BatPercentRemaining.

# Summary of Completed Tests
Testd on device. New unit tests written and updated for the updated handling.

### DIFF
--- a/drivers/SmartThings/matter-sensor/src/smoke-co-alarm/init.lua
+++ b/drivers/SmartThings/matter-sensor/src/smoke-co-alarm/init.lua
@@ -159,16 +159,6 @@ local function binary_state_handler_factory(zeroEvent, nonZeroEvent)
   end
 end
 
--- local function bool_handler_factory(trueEvent, falseEvent)
-  -- return function(driver, device, ib, response)
-    -- if ib.data.value and trueEvent ~= nil then
-      -- device:emit_event_for_endpoint(ib.endpoint_id, trueEvent)
-    -- elseif falseEvent ~= nil then
-      -- device:emit_event_for_endpoint(ib.endpoint_id, falseEvent)
-    -- end
-  -- end
--- end
-
 local function test_in_progress_event_handler(driver, device, ib, response)
   if device:supports_capability(capabilities.smokeDetector) then
     if ib.data.value then
@@ -275,7 +265,7 @@ local matter_smoke_co_alarm_handler = {
         [clusters.SmokeCoAlarm.attributes.COState.ID] = binary_state_handler_factory(capabilities.carbonMonoxideDetector.carbonMonoxide.clear(), capabilities.carbonMonoxideDetector.carbonMonoxide.detected()),
         [clusters.SmokeCoAlarm.attributes.BatteryAlert.ID] = battery_alert_attr_handler,
         [clusters.SmokeCoAlarm.attributes.TestInProgress.ID] = test_in_progress_event_handler,
-        [clusters.SmokeCoAlarm.attributes.HardwareFaultAlert.ID] = hardware_fault_alert_handler, -- bool_handler_factory(capabilities.hardwareFault.hardwareFault.detected(), capabilities.hardwareFault.hardwareFault.clear()),
+        [clusters.SmokeCoAlarm.attributes.HardwareFaultAlert.ID] = hardware_fault_alert_handler,
       },
       [clusters.CarbonMonoxideConcentrationMeasurement.ID] = {
         [clusters.CarbonMonoxideConcentrationMeasurement.attributes.MeasuredValue.ID] = carbon_monoxide_attr_handler,
--- a/drivers/SmartThings/matter-sensor/src/smoke-co-alarm/init.lua
+++ b/drivers/SmartThings/matter-sensor/src/smoke-co-alarm/init.lua
@@ -224,8 +224,8 @@ end
 
 local function power_source_attribute_list_handler(driver, device, ib, response)
   for _, attr in ipairs(ib.data.elements) do
-    -- Re-profile the device if BatPercentRemaining (Attribute ID 0x0C) is available
-    if attr.value == 0x0C then
+    -- Re-profile the device if BatPercentRemaining is available
+    if attr.value == clusters.PowerSource.attributes.BatPercentRemaining.ID then
       match_profile(device, battery_support.BATTERY_PERCENTAGE)
       return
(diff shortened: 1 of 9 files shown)


drivers/SmartThings/matter-sensor/src/test/test_matter_smoke_co_alarm.lua
```diff
@@ -81,7 +83,6 @@ local function test_init()
   end
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
   test.mock_device.add_test_device(mock_device)
-  mock_device:expect_metadata_update({ profile = "smoke-co-temp-humidity-comeas" })
 end
```
nickolas-deboom, line 87: Could you add UT(s) that test the hardwareFault capability being set by BatteryAlert?

drivers/SmartThings/matter-sensor/src/smoke-co-alarm/init.lua
```diff
@@ -191,21 +193,70 @@ local function carbon_monoxide_unit_attr_handler(driver, device, ib, response)
...
+    device:emit_event(capabilities.battery.battery(math.floor(ib.data.value / 2.0 + 0.5)))
+  end
+end
+
+local function do_configure(driver, device)
+  local battery_feature_eps = device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY})
+  if #battery_feature_eps > 0 then
+    device:send(clusters.PowerSource.attributes.AttributeList:read())
+  else
+    match_profile(device, battery_support.NO_BATTERY)
```
nickolas-deboom, line 250: Should we create profiles that exclude the battery capabilities? The way it is now, there will be no change in behavior if the battery feature is not supported, but that would mean that the batteryLevel capability would still be included in the profile (and would show up as blank), which isn't a huge deal but we could just create some -nobattery profiles

hcarter-775: Do any smoke alarms not use batteries?

drivers/SmartThings/matter-sensor/src/smoke-co-alarm/init.lua
```diff
@@ -191,21 +193,70 @@ local function carbon_monoxide_unit_attr_handler(driver, device, ib, response)
...
+  hardware_fault_capability_handler(device)
+end
+
+local function power_source_attribute_list_handler(driver, device, ib, response)
+  for _, attr in ipairs(ib.data.elements) do
+    -- Re-profile the device if BatPercentRemaining (Attribute ID 0x0C) is present.
+    if attr.value == 0x0C then
+      match_profile(device, battery_support.BATTERY_PERCENTAGE)
+      return
+    end
```
nickolas-deboom, line 222: If BatPercentRemaining is not available, and so only battery level is supported, would that result in match_profile never being called?

hcarter-775: Yes, but that's expected, no? Since the device will be fingerprinted to a profile including batteryLevel?

nickolas-deboom: The only generic fingerprint is for `smoke-co`, so if the device supports other clusters like temp/humidity/CO measurement, the driver won't switch to the correct profile in this case

nickolas-deboom: We could just add a call to `match_profile` at the end of this function without the battery_support argument to make sure that match_profile is always called

drivers/SmartThings/matter-sensor/src/smoke-co-alarm/init.lua
```diff
@@ -191,21 +194,74 @@ local function carbon_monoxide_unit_attr_handler(driver, device, ib, response)
   device:set_field(CARBON_MONOXIDE_MEASUREMENT_UNIT, unit, { persist = true })
 end
 
+local function hardware_fault_capability_handler(device)
+  local batLevelOk  = device:get_field(BatteryLevel) == clusters.PowerSource.types.BatChargeLevelEnum.OK
+  local batAlertWarnOrCrit  = device:get_field(BatteryAlert) ~= nil and device:get_field(BatteryAlert) ~= clusters.SmokeCoAlarm.types.AlarmStateEnum.NORMAL
+  if (batLevelOk and batAlertWarnOrCrit) or device:get_field(HardwareFaultAlert) == true then
```
nickolas-deboom, line 200: If batAlertWarnOrCrit is true but batLevelOk is false, wouldn't we want to set hardwareFault to detected?

hcarter-775: I don't believe so. If we set it based on that logic, then we'd be seeing the hardwareFault whenever the battery is below normal.

nickolas-deboom: I was just thinking about a case where the battery level is at Warning from the PowerSource cluster but Critical from SmokeCOAlarm for example - I think that the current logic would emit harwareFault as clear.

hcarter-775: I reworked the logic to be more general. At first, I assumed the case you described wouldn't occur and so didn't bother with it, but I think you're right to point it out. I added some logic to handle all cases where BatteryAlert is a higher warning level than BatChargeLevel

hcarter-775: However, this relies pretty heavily on the two attributes being equivalent on the device side except in the case of failure. I think this can be expected, but it is worth pointing out that it allows less room for device-side difference and specific handling.

nickolas-deboom (approved):
LGTM!

drivers/SmartThings/matter-sensor/src/smoke-co-alarm/init.lua
nickolas-deboom: The changes you made make sense from my persepctive :+1

PR #2071, opened 2025-04-14
CHAD-15194 Zigbee locks: don't report code numbers if lock does not s…

greens:
…upport them

The lock/unlock event data field would include the code index reported by the lock, which was usually a default value, even if the lock didn't support lock codes, leading to confusing history messages

### DIFF
--- a/drivers/SmartThings/zigbee-lock/src/test/test_zigbee_samsungsds.lua
+++ b/drivers/SmartThings/zigbee-lock/src/test/test_zigbee_samsungsds.lua
@@ -143,7 +143,7 @@ test.register_message_test(
         channel = "capability",
         direction = "send",
         message = mock_device:generate_test_message("main",
-          capabilities.lock.lock.locked({ data = { codeId = "0", codeName = "Code 0", method = "keypad"} })
+          capabilities.lock.lock.locked({ data = { method = "keypad"} })
         )
       }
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zigbee-lock/src/test/test_zigbee_samsungsds.lua
```diff
@@ -273,7 +273,7 @@ test.register_message_test(
         channel = "capability",
         direction = "send",
         message = mock_device:generate_test_message("main",
-          capabilities.lock.lock.locked({ data = { codeId = "0", codeName = "Code 0", method = "keypad"} })
+          capabilities.lock.lock.locked({ data = { method = "auto"} })
```
wkhenon, line 276: Is this covered by the `init.lua` change made? Or is it this separate clean-up?

greens: this is because we were artificially changing the method to `keypad` if we detected any info about code IDs in the packet from the device

greens: there's a discussion on the jira ticket about it

kdbang (approved):

PR #2077, opened 2025-04-15
feat: Preliminary Support for Viper-backed OAuth in Sonos

dljsjr:
# Type of Change

- [x] New feature

# Checklist

- [x] I have verified my changes by testing with a device or have communicated a plan for testing (*see note*)
- [ ] I am adding new behavior, such as adding a sub-driver, and have added and run new unit tests to cover the new behavior (*see note*)

# Description of Change

This change adds preliminary support for the Sonos OAuth requirements that will take effect in August, allowing us to integrate with the new Hub functionality in the upcoming release that will allow for the Sonos driver to get an OAuth token from the mobile app.

# Note

There are a few small TODO's here based on the shape of errors. We didn't have the ability to readily trigger the failure case for this, and we need to be able to do that to finish those TODO's.

We have a way forward on this now, though, so we'll be able to fill those in shortly.


drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -40,16 +40,26 @@ local favorites_version = ""
...
+      groupId = groupId,
+      playerId = playerId
+    }
+    local maybe_token, err = sonos_conn.driver:get_oauth_token()
+    if err then
+      log.warn(string.format("notice: get_oauth_token -> %s", err))
+    end
+
+    if maybe_token then
+      wss_msg_header.authorization = string.format("Bearer %s", maybe_token.access_token)
```
dljsjr, line 59: For the WSS connection, the `Bearer` token is part of the JSON payload's "header", not the HTTP headers. And the `authorization` key *must* be lowercase, according to the docs.

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -155,11 +168,26 @@ end
...
+  if not token_receive_handle then
+    log.warn(string.format("error creating oauth token receive handle for respawn task: %s", err))
+  end
   cosock.spawn(function()
     local backoff = backoff_builder(60, 1, 0.1)
     while not sonos_conn:is_running() do
-      local start_success = sonos_conn:start()
-      if start_success then return end
+      if sonos_conn.driver.waiting_for_token and token_receive_handle then
+        local token, channel_error = token_receive_handle:receive()
```
dljsjr, line 179: I don't know that we actually want to handle this in a loop; that said, I *believe* that this won't *actually* loop.

`get_oauth_token()` calls through to `security.get_sonos_oauth_token()`, which we have designed to be completely out-of-band async. So we fire-and-forget, with the return value either being nothing or an error due to perms/API compat.

The `token_receive_handle` is half of a cosock channel. We haven't put a timeout on it, and it's on a non-main thread, so it should just yield forever until it gets something. And that "something" should only come through if the send half of the channel gets dropped, or if the token arrives on the environment info update in the augmented store.

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -329,11 +368,14 @@ function SonosConnection.new(driver, device)
...
       end
     else
-      log.warn(string.format("WebSocket Message for %s did not have a data payload: %s", device_name, st_utils.stringify_table(msg)))
+      log.warn(string.format("WebSocket Message for %s did not have a data payload: %s", device_name,
+        st_utils.stringify_table(msg)))
     end
   end
 
   self.on_error = function(uuid, err)
+    -- TODO: Implement a call to `getToken` here on certain failures, once I actually
```
dljsjr, line 377: Once we know what an auth error actually looks like, we'll propagate it out of the Router to the Connection via this callback. This will allow us to handle propagating the oauth flow request to the firmware + marking the speaker offline. I should be able to figure this out tomorrow morning with the new guidance we got from devrel late today.

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -425,7 +468,27 @@ function SonosConnection:start()
     _open_coordinator_socket(self, household_id, player_id)
   end
 
-  self:refresh_subscriptions()
+
+  local reply_tx, reply_rx = cosock.channel.new()
+
+  self:refresh_subscriptions(reply_tx)
+
+  local reply = reply_rx:receive()
```
dljsjr, line 476: From the docs, the WSS connection itself won't actually be refused if we're missing credentials. Instead, the first command we try to send (which is a `subscribe` command in our case) will let us know if the auth fails.

So we've added this reply channel API to be able to make the initial subscription establishment in the `:start()` method wait until we know that the subscription is allowed.

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -178,12 +180,22 @@ local function scan_for_ssdp_updates(driver)
   end)
 end
 
+local startup_state_received = false
```
dljsjr, line 183: We want all device activity to be more or less frozen until we get the start up state/initial contents of the augmented data store, so we have to queue the devices up.

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -178,12 +180,22 @@ local function scan_for_ssdp_updates(driver)
...
 -- We use the same handler for added and init here because at the time of authoring
 -- this driver, there is a bug with LAN Edge Drivers where `init` may not be called
 -- on every device that gets created using `try_create_device`. This makes sure that
 -- a device is fully initialized whether we come from fresh join or restart.
 -- See: https://smartthings.atlassian.net/browse/CHAD-9683
 local function _initialize_device(driver, device)
+  if not startup_state_received then
+    log.warn(string.format("startup state not yet received, delaying init for %s", (device and device.label or "<unknown device>")))
+    if device and device.id then
+      devices_waiting_for_startup_state[device.id] = device
```
dljsjr, line 194: Here's said queue.

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -362,8 +374,81 @@ local function handle_ssdp_discovery(self, ssdp_group_info, callback)
...
+
 --- @type SonosDriver
 local driver = Driver("Sonos", {
+  waiting_for_token = false,
+  oauth_token_tx = oauth_token_tx,
+  oauth = {},
+  get_oauth_token_receive_handle = function(self)
+    oauth_token_tx:subscribe()
+  end,
+  get_oauth_token = function(self)
```
dljsjr, line 389: `get_oauth_token` will only return a token if we actually *have* one. Otherwise, it'll trigger the OAuth activity in the hub firwmare, and immediately return with an error indicating why.

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -362,8 +374,82 @@ local function handle_ssdp_discovery(self, ssdp_group_info, callback)
...
+    else
+      log.warn(st_utils.stringify_table(decoded, "JSON Decode Error while parsing Endpoint App Info", false))
+    end
+
+    if self.hub_augmented_driver_data.sonosOAuthToken then
+      decode_success, decoded = pcall(json.decode, self.hub_augmented_driver_data.sonosOAuthToken)
+      if decode_success then
+        self.oauth.token = decoded
+        self.waiting_for_token = false
+        self.oauth_token_tx:send(decoded)
```
dljsjr, line 438: This is where token receipt actually happens. Notify anyone waiting for one.

drivers/SmartThings/sonos/src/api/sonos_connection.lua
varzac: I'll have to think about this, but I don't know if _no_ timeout is truly what we want.  I could see a very large timeout, but at the same time I understand the potential load caused by thousands of hubs repeatedly checking for a token when the user doesn't link their account.

drivers/SmartThings/sonos/src/types.lua
```diff
@@ -168,13 +168,13 @@ function SonosState.new()
...
   --- @param device SonosDevice|nil
   ret.update_household_info = function(self, id, groups_event, device)
-    log.debug_with({ hub_logs = true },
-      st_utils.stringify_table(
-        {name = (device or { label = "<no device>" }).label, event = groups_event },
-        string.format("Update household info for household %s", id),
-        true
-      )
-    )
+    -- log.trace_with({ hub_logs = false },
```
varzac, line 171: Is this intended?

drivers/SmartThings/sonos/src/old_init.backup
```diff
@@ -0,0 +1,445 @@
+---@module 'result'
```
varzac, line 1: This shouldn't be checked in.

drivers/SmartThings/sonos/src/types.lua
dljsjr: Fixed.

drivers/SmartThings/sonos/src/old_init.backup
dljsjr: 🤦

drivers/SmartThings/sonos/src/lifecycle_handlers.lua
```diff
@@ -0,0 +1,124 @@
...
+  -- threads, as this may involve long-yielding operations.
+  cosock.spawn(function()
+    device.log.debug("top of init loop")
+    local mac_addr = device.device_network_id
+    local player_info_tx, player_info_rx = cosock.channel.new()
+    while true do
+      device.log.debug("sending player info request")
+      driver.ssdp_task:get_player_info(player_info_tx, mac_addr)
+      device.log.debug("selecton on player info receive")
+      local recv_ready, _, select_err = cosock.socket.select({ player_info_rx }, nil, nil)
```
cjswedes, line 59: Im guessing its been this way and there aren't any changes, just a moving of code, but why do we use select for one socket? Wouldn't calling `player_info_rx:receive()` have the same effect?

drivers/SmartThings/sonos/src/ssdp.backup
cjswedes, line 1: nit: needs to be removed prior to merge

cjswedes (commented):
I did my best reviewing this, but it was a difficult review given the refactor and new functionality all in the same change set. I think it would probably be helpful for me to get some high level descriptions of the cosock tasks that are running and their lifecycles.

drivers/SmartThings/sonos/src/ssdp.lua
```diff
@@ -1,194 +1,403 @@
...
-    -- This pattern extracts the Key/Value pairs in to a Lua table via the two capture groups.
-    -- The key capture group is composed entirely of a negating matcher to exclude illegal characters, ending at the `:`.
-    -- The RFC states that after the colon there may be any arbitrary amount of leading space between the colon
-    -- and the value, and that the value shouldn't have any trailing whitespace, so we exclude those as well.
-    -- The original Luncheon implementation of this Lua Pattern used iteration and detected the `;` separator
-    -- that indicates key/value parameters, however, we don't make that distinction here and instead leave parsing
-    -- values with parameters to the consumers of the output of this function.
-    local k, v = string.match(l, '([^%c()<>@,;:\\"/%[%]?={} \t]+):%s*(.-)%s*$')
-    if k == nil or k == "" then
-      return nil, string.format("Couldn't parse header/value pair for line %q", l)
```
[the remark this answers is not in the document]
dljsjr: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2077/files#diff-5156307d746a0a42b10fb3c890e35d267af60ae1a8f3a6b613790ad4ce2aacbeR17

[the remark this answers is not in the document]
cjswedes: FWIW this pattern is slightly different than what luncheon is doing for header kv extraction with a regex:
https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/9cbe65c375a2dda2b1b7d4c1ab1beb0a1d21e061/lua_libs/luncheon/headers.lua#L79-L91

varzac (approved):

drivers/SmartThings/sonos/src/lifecycle_handlers.lua
```diff
@@ -63,22 +67,35 @@ function SonosDriverLifecycleHandlers.initialize_device(driver, device)
           device.log.warn(string.format("error receiving device info: %s", recv_err))
         else
           ---@cast info { ssdp_info: SonosSSDPInfo, discovery_info: SonosDiscoveryInfo, force_refresh: boolean }
-          local success, error, error_code = driver:handle_player_discovery_info(driver:get_api_key(), info,
-            device)
-          if success then
-            return
+          local auth_success, api_key_or_err = driver:check_auth(info)
+          if not auth_success then
```
varzac, line 71: This seems weird to have these 2 checks one after another one a "not" and the other == false

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -59,32 +75,282 @@ function SonosDriver:get_device_by_dni(dni)
...
+    local _, err = self:request_oauth_token()
+    if err then
+      log.error(string.format("Request OAuth token error: %s", err))
+    end
+  end
+end
+
+function SonosDriver:handle_startup_state_received()
+  self:start_ssdp_event_task()
+  self:notify_augmented_data_changed "snapshot"
```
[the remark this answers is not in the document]
dljsjr: The neovim version of luals uses a different formatter than the VS Code one, and it seems to prefer this for some reason. I tried to clean up all the places it does this.

drivers/SmartThings/sonos/src/lifecycle_handlers.lua
dljsjr: The reason it's structured that way is because either the auth check can fail, which is a `false`, or the `check_auth` operation can error, which returns `nil, error`. In both cases, we want to call `device:offline()`, but in the `failed-but-not-error` case, we want to do the additional work in the `== false` branch.

drivers/SmartThings/sonos/src/ssdp.lua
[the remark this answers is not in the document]
dljsjr: The one in Luncheon is "more correct". [I wrote that one](https://github.com/cosock/luncheon/pull/9/files) after I wrote this one. Figured it was about time to cut over.

dljsjr:
With approval from @varzac:

There are QA routes that require the driver to be on a SmartThings owned channel for testing to move forward. Alpha suffices for this, so we're going to merge this now, but I'll continue monitoring for further review and make changes/fixes before we go to Beta.

PR #2081, opened 2025-04-17
CHAD-14927: Driver fixes for lifecycle changes

aleclorimer:
Check all that apply

# Type of Change

- [ ] Refactor

# Checklist

# Description of Change
Because the underlying behavior of drivers is changing and `init` messages are not synthetically generated from within the driver, some tests needed this `init` message to come from the `DeviceLifecycleChannel` in order to have the driver proceed properly.
Secondly, now that drivers have a strict startup process with requiring to wait for an `init` message for each device within the driver, some tests needed to toggle off this functionality with `test.enable_startup_messages(false)`.
Lastly, the internal RPC version for all tests has been set to `9999` to ensure that we are always testing the latest features based on RPC version. These changes fix a few tests where a change in driver behavior from this versioning caused it to fail.


drivers/SmartThings/matter-button/src/test/test_matter_button_parent_child.lua
```diff
@@ -74,6 +74,7 @@ local CLUSTER_SUBSCRIBE_LIST ={
 }
 
 local function test_init()
+  test.enable_startup_messages(false)
```
cjswedes, line 77: Is `true` the default?

cjswedes (commented):
Can you add to the description of the PR when you expect these changes to land, so external developers dont think this is something live. Also can you add a test with `test.enable_startup_messages(true)` so there is an example in the repo of what that looks like?

drivers/SmartThings/matter-button/src/test/test_matter_button_parent_child.lua
aleclorimer: Yes. I would expect these changes to land along with the 0.58 release and the lua-libs update after that.

Should I change it to `test.disable_startup_messages()` and make a `test.enable_startup_messages()` even though that one shouldn't be necessary?

nickolas-deboom:
Can the RPC version still be overridden within test files? Some test cases rely on it to test certain functionality

aleclorimer:
> With the default RPC being set to 9999 can the RPC version still be overridden within test files? Some test cases rely on it to test certain functionality

Yes. `set_rpc_version(#)` still works.

It should now work better than before as previously the RPC version could get set to a different value based on the order of initialization of the files within the script.

greens:
Is this only expected to affect matter drivers? How come?

aleclorimer:
> Is this only expected to affect matter drivers? How come?

It affects all drivers, though matter driver tests were the only tests that needed adjustments to the changes.

ctowns:
What is the reason only these specific matter tests needed the update? Why wouldn't all tests now need to have the `init` lifecycle event provided via the `DeviceLifecycleChannel` channel?

varzac:
> What is the reason only these specific matter tests needed the update? Why wouldn't all tests now need to have the `init` lifecycle event provided via the `DeviceLifecycleChannel` channel?

A number of the matter tests are following a pattern that is not used by other tests.  Specifically setting up test init functions that send `added` lifecycle events for devices before every test.  For Zigbee/Z-Wave typically there is one test for the added flow, and all other tests follow the flow that would happen with a device having already been added to the driver.  So these tests need to be modified.   I would also caution against this pattern when possible, because the test init function runs _before_ the test function starts, there are things that aren't available (such as a `wait_for_events` call) which can lead to a race condition (we have one matter test that we can't currently correct with the new model because of this).

aleclorimer:
Revisiting this set of changes for the 0.58 release. These changes have transitioned over to a new [PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2310). I will close this PR once that new PR gets merged.

PR #2084, opened 2025-04-18
Matter Thermostat: add support for modular profiles

ctowns:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests
Added unit tests for common device configs for Air Purifers, Thermostats, and Room ACs
Onboarded VDA devices for each of these device types
Completed test plan as defined in REQ

### DIFF
--- a/drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
+++ b/drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
@@ -1,0 +1,127 @@
+name: air-purifier-modular
+components:
+  - id: main
+    capabilities:
+      - id: airPurifierFanMode
+        version: 1
+      - id: fanSpeedPercent
+        version: 1
+      - id: fanOscillationMode
+        version: 1
+        optional: true
+      - id: windMode
+        version: 1
+        optional: true
+      - id: thermostatHeatingSetpoint
+        version: 1
+        optional: true
+      - id: thermostatCoolingSetpoint
+        version: 1
+        optional: true
+      - id: thermostatMode
+        version: 1
+        optional: true
+      - id: thermostatOperatingState
+        version: 1
+        config:
+          values:
+            - key: "thermostatOperatingState.value"
+              enabledValues:
+                - idle
+                - cooling
+                - heating
+        optional: true
+      - id: temperatureMeasurement
+        version: 1
+        optional: true
+      - id: relativeHumidityMeasurement
+        version: 1
+        optional: true
+      - id: airQualityHealthConcern
+        version: 1
+        optional: true
+      - id: carbonMonoxideMeasurement
+        version: 1
+        optional: true
+      - id: carbonMonoxideHealthConcern
+        version: 1
+        optional: true
+      - id: carbonDioxideMeasurement
+        version: 1
+        optional: true
+      - id: carbonDioxideHealthConcern
+        version: 1
+        optional: true
+      - id: nitrogenDioxideMeasurement
+        version: 1
+        optional: true
+      - id: nitrogenDioxideHealthConcern
+        version: 1
+        optional: true
+      - id: ozoneMeasurement
+        version: 1
+        optional: true
+      - id: ozoneHealthConcern
+        version: 1
+        optional: true
+      - id: formaldehydeMeasurement
+        version: 1
+        optional: true
+      - id: formaldehydeHealthConcern
+        version: 1
+        optional: true
+      - id: veryFineDustSensor
+        version: 1
+        optional: true
+      - id: veryFineDustHealthConcern
+        version: 1
+        optional: true
+      - id: fineDustHealthConcern
+        version: 1
+        optional: true
+      - id: dustSensor
+        version: 1
+        optional: true
+      - id: dustHealthConcern
+        version: 1
+        optional: true
+      - id: radonMeasurement
+        version: 1
+        optional: true
+      - id: radonHealthConcern
+        version: 1
+        optional: true
+      - id: tvocMeasurement
+        version: 1
+        optional: true
+      - id: tvocHealthConcern
+        version: 1
+        optional: true
+      - id: firmwareUpdate
+        version: 1
+      - id: refresh
+        version: 1
+    categories:
+      - name: AirPurifier
+  - id: activatedCarbonFilter
+    optional: true
+    capabilities:
+      - id: filterState
+        version: 1
+        optional: true
+      - id: filterStatus
+        version: 1
+        optional: true
+    categories:
+      - name: AirPurifier
+  - id: hepaFilter
+    optional: true
+    capabilities:
+      - id: filterState
+        version: 1
+        optional: true
+      - id: filterStatus
+        version: 1
+        optional: true
+    categories:
+      - name: AirPurifier
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -21,6 +21,10 @@ local im = require "st.matter.interaction_model"
 local MatterDriver = require "st.matter.driver"
 local utils = require "st.utils"
 
+local SUPPORTED_COMPONENT_CAPABILITIES = "__supported_component_capabilities"
+-- declare match_profile function for use throughout file
+local match_profile
+
 -- Include driver-side definitions when lua libs api version is < 10
 local version = require "version"
 if version.api < 10 then
@@ -305,6 +309,25 @@ local subscribed_attributes = {
   },
 }
 
+local function supports_capability_by_id_modular(device, capability, component)
+  if not device:get_field(SUPPORTED_COMPONENT_CAPABILITIES) then
+    device.log.warn_with({hub_logs = true}, "Device has overriden supports_capability_by_id, but does not have supported capabilities set.")
+    return false
+  end
+  for _, component_capabilities in ipairs(device:get_field(SUPPORTED_COMPONENT_CAPABILITIES)) do
+    local comp_id = component_capabilities[1]
+    local capability_ids = component_capabilities[2]
+    if (component == nil) or (component == comp_id) then
+        for _, cap in ipairs(capability_ids) do
+          if cap == capability then
+            return true
+          end
+        end
+    end
+  end
+  return false
+end
+
 local function epoch_to_iso8601(time)
   return os.date("!%Y-%m-%dT%H:%M:%SZ", time)
 end
@@ -462,6 +485,11 @@ local function schedule_polls_for_cumulative_energy_imported(device)
 end
 
 local function device_init(driver, device)
+  if device:get_field(SUPPORTED_COMPONENT_CAPABILITIES) then
+    -- assume that device is using a modular profile, override supports_capability_by_id
+    -- library function to utilize optional capabilities
+    device:extend_device("supports_capability_by_id", supports_capability_by_id_modular)
+  end
   device:subscribe()
   device:set_component_to_endpoint_fn(component_to_endpoint)
   device:set_endpoint_to_component_fn(endpoint_to_component)
@@ -478,17 +506,15 @@ local function device_init(driver, device)
 end
 
 local function info_changed(driver, device, event, args)
-  -- Note this is needed because device:subscribe() does not recalculate
-  -- the subscribed attributes each time it is run, that only happens at init.
-  -- This will change in the 0.48.x release of the lua libs.
-  for cap_id, attributes in pairs(subscribed_attributes) do
-    if device:supports_capability_by_id(cap_id) then
-      for _, attr in ipairs(attributes) do
-        device:add_subscribed_attribute(attr)
-      end
-    end
+  if device:get_field(SUPPORTED_COMPONENT_CAPABILITIES) then
+    -- This indicates the device should be using a modular profile, so
+    -- re-up subscription with new capabilities using the modular supports_capability override
+    device:extend_device("supports_capability_by_id", supports_capability_by_id_modular)
   end
-  device:subscribe()
+
+  if device.profile.id ~= args.old_st_store.profile.id then
+    device:subscribe()
+  end
   schedule_polls_for_cumulative_energy_imported(device)
 end
 
@@ -506,7 +532,7 @@ local function get_endpoints_for_dt(device, device_type)
   return endpoints
 end
 
-local function get_device_type(driver, device)
+local function get_device_type(device)
   for _, ep in ipairs(device.endpoints) do
     if ep.device_types ~= nil then
       for _, dt in ipairs(ep.device_types) do
@@ -569,6 +595,27 @@ local function create_level_measurement_profile(device)
     end
   end
   return meas_name, level_name
+end
+
+local function supported_level_measurements(device)
+  local measurement_caps, level_caps = {}, {}
+  for _, details in ipairs(AIR_QUALITY_MAP) do
+    local cap_id  = details[1]
+    local cluster = details[3]
+    -- capability describes either a HealthConcern or Measurement/Sensor
+    if (cap_id:match("HealthConcern$")) then
+      local attr_eps = embedded_cluster_utils.get_endpoints(device, cluster.ID, { feature_bitmap = cluster.types.Feature.LEVEL_INDICATION })
+      if #attr_eps > 0 then
+        table.insert(level_caps, cap_id)
+      end
+    elseif (cap_id:match("Measurement$") or cap_id:match("Sensor$")) then
+      local attr_eps = embedded_cluster_utils.get_endpoints(device, cluster.ID, { feature_bitmap = cluster.types.Feature.NUMERIC_MEASUREMENT })
+      if #attr_eps > 0 then
+        table.insert(measurement_caps, cap_id)
+      end
+    end
+  end
+  return measurement_caps, level_caps
 end
 
 local function create_air_quality_sensor_profile(device)
@@ -652,7 +699,7 @@ local function profiling_data_still_required(device)
   return false
 end
 
-local function match_profile(driver, device)
+local function match_profile_switch(driver, device)
   if profiling_data_still_required(device) then return end
 
   local running_state_supported = device:get_field(profiling_data.THERMOSTAT_RUNNING_STATE_SUPPORT)
@@ -660,7 +707,7 @@ local function match_profile(driver, device)
 
   local thermostat_eps = device:get_endpoints(clusters.Thermostat.ID)
   local humidity_eps = device:get_endpoints(clusters.RelativeHumidityMeasurement.ID)
-  local device_type = get_device_type(driver, device)
+  local device_type = get_device_type(device)
   local profile_name
   if device_type == RAC_DEVICE_TYPE_ID then
     profile_name = "room-air-conditioner"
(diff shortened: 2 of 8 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -794,6 +832,336 @@ local function match_profile(driver, device)
...
+
+  -- clear all profiling data fields after profiling is complete.
+  for _, field in pairs(profiling_data) do
+    device:set_field(field, nil)
+  end
+end
+
+local function match_profile(driver, device)
+  -- must use profile switching on older hubs
+  if version.api < 14 and version.rpc < 7 then
```
varzac, line 1158: I think this should be an or, if either one is too low we use a static profile

ctowns: Good catch - I updated this to an `or` instead, and I also left a note to update the RPC version to 8 once that gets into a build since I bumped the version number for the sake of tracking modular profile support :+1:

drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
```diff
@@ -0,0 +1,54 @@
+name: air-purifier-modular
+components:
+  - id: main
+    capabilities:
+      - id: airPurifierFanMode
```
nickolas-deboom, line 5: Do we want the order of the capabilities to match the order in existing profiles?

ctowns: Good point! I pushed up a commit that puts these in the same order as other profiles to be consitent. I also added in the `version` field for consistency with other profiles, even though it technically isn't needed.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,6 +828,232 @@ local function match_profile(driver, device)
...
+  local thermostat_eps = device:get_endpoints(clusters.Thermostat.ID)
+
+  return version.api >= 14 and version.rpc >= 8 and
+    (device_type == AP_DEVICE_TYPE_ID or
+    (device_type == false and #thermostat_eps > 0))
+end
+
+function match_profile(driver, device)
+  if supports_modular_profile(device) then
+    print("CHT Calling modular profile match")
```
hcarter-775, line 1050: remove?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,6 +828,232 @@ local function match_profile(driver, device)
...
+    return
+  end
+
+  -- clear all profiling data fields after profiling is complete.
+  for _, field in pairs(profiling_data) do
+    device:set_field(field, nil)
+  end
+end
+
+local function supports_modular_profile(device)
```
hcarter-775, line 1039: should we not do this for Room ACs as well? This seems like an equally valid use case?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,6 +828,232 @@ local function match_profile(driver, device)
...
+
+  --re-up subscription with new capabiltiies using the moudlar supports_capability override
+  device:extend_device("supports_capability_by_id", supports_capability_by_id_modular)
+  device:subscribe()
+end
+
+local function match_modular_profile(driver, device)
+  if profiling_data_still_required(device) then return end
+
+  -- TODO: test device refresh
```
hcarter-775, line 1018: remove?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -21,6 +21,9 @@ local im = require "st.matter.interaction_model"
 local MatterDriver = require "st.matter.driver"
 local utils = require "st.utils"
 
+local match_profile
```
hcarter-775, line 24: remove?

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: I originally had room ACs in here and removed it to simplify our initial batch of modular profile devices

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: I defined this at the top so that match_profile could be called from anywhere for simplicity

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -652,15 +691,15 @@ local function profiling_data_still_required(device)
   return false
 end
 
-local function match_profile(driver, device)
-  if profiling_data_still_required(device) then return end
+local function match_profile_switch(driver, device)
+  -- if profiling_data_still_required(device) then return end
```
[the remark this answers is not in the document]
ctowns: You are correct good sir

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,6 +828,232 @@ local function match_profile(driver, device)
...
+
+  local MAIN_COMPONENT_IDX = 1
+  local CAPABILITIES_LIST_IDX = 2
+
+  local humidity_eps = device:get_endpoints(clusters.RelativeHumidityMeasurement.ID)
+  if #humidity_eps > 0 then
+    table.insert(main_component_capabilities, capabilities.relativeHumidityMeasurement.ID)
+  end
+
+  local hepa_filter_eps = embedded_cluster_utils.get_endpoints(device, clusters.HepaFilterMonitoring.ID)
```
nickolas-deboom, line 895: It doesn't really matter but you could use `device:get_endpoints` directly since match_modular_profile will only be called for lua libs v14 and higher which would include these clusters

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,6 +828,232 @@ local function match_profile(driver, device)
...
+  local total_supported_capabilities = optional_supported_component_capabilities
+  table.insert(total_supported_capabilities[MAIN_COMPONENT_IDX][CAPABILITIES_LIST_IDX], capabilities.airPurifierFanMode.ID)
+  table.insert(total_supported_capabilities[MAIN_COMPONENT_IDX][CAPABILITIES_LIST_IDX], capabilities.fanSpeedPercent.ID)
+  table.insert(total_supported_capabilities[MAIN_COMPONENT_IDX][CAPABILITIES_LIST_IDX], capabilities.refresh.ID)
+  table.insert(total_supported_capabilities[MAIN_COMPONENT_IDX][CAPABILITIES_LIST_IDX], capabilities.firmwareUpdate.ID)
+
+  device:set_field(SUPPORTED_COMPONENT_CAPABILITIES, total_supported_capabilities, { persist = true })
+
+  --re-up subscription with new capabiltiies using the moudlar supports_capability override
+  device:extend_device("supports_capability_by_id", supports_capability_by_id_modular)
```
nickolas-deboom: Did we decide whether this should be left here or moved to `info_changed`?

ctowns: Yes, we should move this to `info_changed`! It looks like I pushed the commit up to do that just as you were making this comment - great minds

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: Ah that makes sense, could you add a short comment then to describe what this is doing?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -305,6 +308,21 @@ local subscribed_attributes = {
   },
 }
 
+local function supports_capability_by_id_modular(device, capability, component)
+  for _, component_capabilities in ipairs(device:get_field(SUPPORTED_COMPONENT_CAPABILITIES)) do
```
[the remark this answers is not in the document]
hcarter-775: I'm not sure this is required, since the function would have the same handling without this line.

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: For what its worth, for the near future we will need to support both static and modular profiles simultaneously while the latest FW is rolled out to hubs and other platforms. This means there will be a period during which we still must create and maintain static profiles to support backwards compatibility with hubs and platforms that do not have the modular profiles update. So, even including the room AC modular profile update in this PR would not negate the other PR. However, we could still add room AC now, or the next batch of modular profile updates in the coming weeks. I would opt to hold off for now to keep the scope small, and because I have not done extensive testing on room AC. Also, that would be a good opportunity to reach out to Hunsup for his input since he had worked on that driver.

hcarter-775:
I believe some logic may be needed to introduce this value configuration for thermostatOperatingState that we have in the static profiles, where a device might support all 3, or just idle and cooling / idle and heating
```
    config:
      values:
        - key: "thermostatOperatingState.value"
          enabledValues:
            - idle
            - cooling
            - heating
```

ctowns:
> I believe some logic may be needed to introduce this value configuration for thermostatOperatingState that we have in the static profiles, where a device might support all 3, or just idle and cooling / idle and heating
> 
> ```
>     config:
>       values:
>         - key: "thermostatOperatingState.value"
>           enabledValues:
>             - idle
>             - cooling
>             - heating
> ```

This is something that will be addressed by the new `supportedValues` attributes we added to capabilities right? I don't think we should define config values in modular profiles if we don't have to, because then they are no longer generic.

hcarter-775:
> > I believe some logic may be needed to introduce this value configuration for thermostatOperatingState that we have in the static profiles, where a device might support all 3, or just idle and cooling / idle and heating
> > ```
> >     config:
> >       values:
> >         - key: "thermostatOperatingState.value"
> >           enabledValues:
> >             - idle
> >             - cooling
> >             - heating
> > ```
> 
> This is something that will be addressed by the new `supportedValues` attributes we added to capabilities right? I don't think we should define config values in modular profiles if we don't have to, because then they are no longer generic.

@ctowns we have not added a supported values attribute to this capability yet. Once we do, I agree that this is the way forward, and given this need for it in modular profiles, it should be done soon. In the meantime, should we do the above configure value in the profile anyway? This will at least cut down on the number of values seen, even if it's not completely accurate, and we can remove it when the supported values work is complete.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,10 +840,284 @@ local function match_profile(driver, device)
...
+  if #rock_eps > 0 then
+    table.insert(main_component_capabilities, capabilities.fanOscillationMode.ID)
+  end
+  if #wind_eps > 0 then
+    table.insert(main_component_capabilities, capabilities.windMode.ID)
+  end
+
+  local thermostat_eps = device:get_endpoints(clusters.Thermostat.ID)
+
+  if #thermostat_eps > 0 then
```
hcarter-775, line 922: similar to the change made in [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2131/files), I think we should add some logic for if temperature measurement cluster is supported but not the thermostat cluster

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,10 +840,284 @@ local function match_profile(driver, device)
...
+
+  device:set_field(SUPPORTED_COMPONENT_CAPABILITIES, total_supported_capabilities, { persist = true })
+end
+
+local function match_modular_profile_room_ac(driver, device)
+  local running_state_supported = device:get_field(profiling_data.THERMOSTAT_RUNNING_STATE_SUPPORT)
+  local humidity_eps = device:get_endpoints(clusters.RelativeHumidityMeasurement.ID)
+  local optional_supported_component_capabilities = {}
+  local main_component_capabilities = {}
+  local profile_name = "room-air-conditioner-modular"
```
hcarter-775, line 1024: we still need to add this profile :+1:

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,10 +840,284 @@ local function match_profile(driver, device)
...
+
+  if device_type == AP_DEVICE_TYPE_ID then
+    match_modular_profile_air_purifer(driver, device)
+  elseif device_type == RAC_DEVICE_TYPE_ID then
+    match_modular_profile_room_ac(driver, device)
+  elseif #thermostat_eps > 0 then
+    match_modular_profile_thermostat(driver, device)
+  else
+    device.log.warn_with({hub_logs=true}, "Device type is not supported by modular profile in thermostat driver")
+    return
```
hcarter-775, line 1087: could be nice to call the regular `match_switch_profile` here as a backup

hcarter-775:
@ctowns sorry to keep going on this topic, but it turns out I _did_ create the supportedOperatingStates attribute for the [ThermostatOperatingState capability](https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/main/capabilities/thermostat-operating-state.src/thermostat-operating-state.yml). It may be good to put any new handling we want to add that utilizes this in this PR

drivers/SmartThings/matter-thermostat/src/init.lua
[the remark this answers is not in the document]
nickolas-deboom: That's not really true because if `get_field` returns nil then the `ipairs` call would result in an error

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,10 +830,278 @@ local function match_profile(driver, device)
...
+  if #humidity_eps > 0 then
+    table.insert(main_component_capabilities, capabilities.relativeHumidityMeasurement.ID)
+  end
+
+  local hepa_filter_eps = embedded_cluster_utils.get_endpoints(device, clusters.HepaFilterMonitoring.ID)
+  local ac_filter_eps = embedded_cluster_utils.get_endpoints(device, clusters.ActivatedCarbonFilterMonitoring.ID)
+
+  if #hepa_filter_eps > 0 then
+    -- TODO: only one of these is required by spec
+    table.insert(hepa_filter_component_capabilities, capabilities.filterState.ID)
```
nickolas-deboom, line 890: filterState is set based on the `Condtion` attribute which is only required if the `CON` feature is supported, so you could gate this capability based on that feature. Looks like filterStatus should always be included since `ChangeIndication` is a mandatory attribute

drivers/SmartThings/matter-thermostat/src/init.lua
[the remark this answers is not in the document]
ctowns: Good catch, I added this and added a warning log in case we ever get in this scenario (which should be unlikely)

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: Added a comment for the method declaration

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: Added!

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: Good point, I think I'll just use the same embedded clusters call for consistency since it ends up calling the library function still anyway.

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: Added!

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: Good idea! I added this along with an additional log statement to note that the driver will try the profile switch fallback

drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
```diff
@@ -0,0 +1,120 @@
...
+      - id: thermostatHeatingSetpoint
+        version: 1
+        optional: true
+      - id: thermostatCoolingSetpoint
+        version: 1
+        optional: true
+      - id: thermostatMode
+        version: 1
+        optional: true
+      - id: thermostatOperatingState
```
ctowns, line 24: @hcarter-775 I think we should make the `supportedThermostatOperatingState` changes outside of this PR since this change will go beyond just modular profiles. However, we could set a default config value here. Although, I don't think it would take too much lift to put up a separate PR to get the supported values working, so I imagine that PR could quickly follow up this PR either in this deploy or the next.

ctowns:
Rebased, but no merge conflicts.

drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
hcarter-775: I'm ok with that- I think it would be nice to put the default config value in for the time being, and just remove it immediately if this work ends up being prioritzed quickly.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -789,10 +835,293 @@ local function match_profile(driver, device)
...
+  local total_supported_capabilities = optional_supported_component_capabilities
+  table.insert(main_component_capabilities, capabilities.thermostatMode.ID)
+  table.insert(main_component_capabilities, capabilities.temperatureMeasurement.ID)
+  table.insert(main_component_capabilities, capabilities.refresh.ID)
+  table.insert(main_component_capabilities, capabilities.firmwareUpdate.ID)
+
+  device:set_field(SUPPORTED_COMPONENT_CAPABILITIES, total_supported_capabilities, { persist = true })
+end
+
+local function match_modular_profile_room_ac(driver, device)
```
hcarter-775, line 1021: you know what might be a good idea? To move the configuration logic into a separate folder that we can require, like with a `local profiling = require "profiling"` and then call functions like `profling.match_modular_profile_room_ac()`. Since this PR itself is introducing around 300 new lines of code to the init.lua file.

drivers/SmartThings/matter-thermostat/src/test/test_matter_thermo_featuremap.lua
```diff
@@ -253,7 +253,12 @@ test.register_coroutine_test(
     end
     test.socket.matter:__expect_send({mock_device.id, subscribe_request})
 
-    test.socket.device_lifecycle:__queue_receive(mock_device:generate_info_changed({}))
+    local updates = {
```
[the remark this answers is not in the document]
nickolas-deboom: Nice, I believe this is the same idea as what is done in some other test files, where a copy of `mock_device.raw_st_data` is created and the profile id of the copy is changed to something else before passing it into the info_changed event like this:
```
    local device_info_copy = utils.deep_copy(mock_device.raw_st_data)
    device_info_copy.profile.id = "3-button"
    local device_info_json = dkjson.encode(device_info_copy)
    test.socket.device_lifecycle:__queue_receive({ mock_device.id, "infoChanged", device_info_json })
```
(from [test_matter_multi_button_switch_mcd](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/35c5570bc91821d7ee53ac21f8e48686dad25346/drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua#L204)

drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
```diff
@@ -0,0 +1,120 @@
+name: air-purifier-modular
+components:
+  - id: main
+    capabilities:
```
nickolas-deboom, line 4: I realized that Air Purifiers can optionally support the OnOff cluster - should we add `switch` as an optional capability?

nickolas-deboom (approved):

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: I like where your minds at and I agree this driver needs to be pared down because not all these functions will need to be loaded all the time. I think the real fix to address code size here would be to move these device-specific configuration functions into sub drivers, kind of similar to what is done for the `air-quality-sensor` sub driver. However, I think just moving them to a separate library file would still result in the no memory change because those functions would not be garbage collected, whereas if they were lazy loaded through a sub driver, they would be gc'd if they are not used.

Maybe we could have a follow up PR to "sub-driver-ize" the matter-thermostat driver? Would that accomplish what you're thinking?

drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
ctowns: Hell yeah we can - isn't it nice that we only have to update one profile to do this? :heart_eyes:

drivers/SmartThings/matter-thermostat/src/test/test_matter_thermo_featuremap.lua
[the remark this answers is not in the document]
ctowns: The method you shared above is actually what I did in the `_modular` test files, I just forgot to update this particular test case after I used the shortcut method above. I'll update this one!

drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
ctowns: Oh I see we don't support switch even in our static profiles... Let's add that support as a follow up PR after this goes in. This PR's scope is just to introduce modular profiles with 1:1 static profile support. I can put up another PR to add switch once this one goes in :+1: 

Good catch though! I wonder if On/Off was added with one of the recent releases

drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
ctowns: Good thinking, I added the default config for now and we can plan to quickly get it removed with the dynamic supported values support!

drivers/SmartThings/matter-thermostat/profiles/air-purifier-modular.yml
nickolas-deboom: Yeah that sounds good to me :+1:

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: Honestly, I wasn't necessarily thinking about memory savings, though I think that's a good point and could be a very good idea. In my opinion, I just think It would be easier to read and iterate on if put this code in a separate file. There is no real reason why the init.lua file has to hold everything in my opinion. The sub-driver does seem like a good idea I think, but it would take a little more thought than the simple pull to a separate file I am imagining.

ctowns: Oh I see what you mean, yeah this entire file has become quite large. I think that speaks to the complexity of this driver and the large number of complex device types that it supports. All of the air quality functionality could be moved to a separate file as well and that might be where we would see the most "line" savings. Let's make a larger driver refactor a part of some of the driver improvement discussions we'll be having in the following weeks.

One tip for navigating a file this large is to use the code outline that is available in VS Code. It bookmarks each of the functions, which makes it at least a little easier to get around, but I agree that it can be rather unwieldy at the moment.

drivers/SmartThings/matter-thermostat/src/test/test_matter_thermo_featuremap.lua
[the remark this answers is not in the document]
ctowns: Ended up just leaving this unit test the same and leaving a comment. This is an only unit test, and the rest of the device info is not needed so I left it as is.

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: Yeah, I think it makes sense to move this discussion into the driver improvements planning we've been talking about

hcarter-775 (approved):

hcarter-775:
Looks good to me! I think we have a couple action items coming out of this PR:

1. Use supportedValues for the ThermostatOperatingState capability
2. Re-work the driver to break down some code into separate files / sub-drivers
3. Include the on/off switch in air purifiers.

Anything else to log for next steps?

drivers/SmartThings/matter-thermostat/src/init.lua
ctowns: I was exploring trying to move these to their own file, and one thing I'm seeing is that moving all of these functions to another file will require redefining many of the consts such as `SUPPORTED_COMPONENT_CAPABILITIES` and `profiling_data` which are currently defined in the main driver file, but are also used by the modular functions. So I think we'd also want to maybe have a separate `consts` file or something like that which could be shared between the `init` and `modular_profile_utils` file (name pending). I think that would make it a larger refactor that we can do in a follow up PR after some more discussion on how to best organize this. I think it is a good idea though and we should definitely plan to organize this driver better!

ldeora:
My thermostat radiator valves (former profile `thermostat-heating-only`) look like this now:

Screenshot_20250604_131011_SmartThings

Screenshot_20250604_131431_Samsung Internet

Devices with `supportedThermostatModes` ***off*** and ***heat*** shouldn't have `thermostatCoolingSetpoint.coolingSetpoint`, right?

PR #2093, opened 2025-04-23
Refactoring driver and support for PIR Motion Sensor

pInksenberg:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [x] Refactor

# Checklist

# Description of Change
Refactoring zigbee-sensor driver.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-sensor/fingerprints.yml
+++ b/drivers/SmartThings/zigbee-sensor/fingerprints.yml
@@ -5,4 +5,12 @@ zigbeeGeneric:
       server:
         - 0x0500
     deviceProfileName: generic-sensor
-
+  - id: "generic-battery-sensor"
+    deviceLabel: "Zigbee Generic Sensor"
+    deviceIdentifiers:
+      - 0x0402
+    clusters:
+      server:
+        - 0x0500
+        - 0x0001
+    deviceProfileName: generic-sensor
--- a/drivers/SmartThings/zigbee-sensor/src/contact/init.lua
+++ b/drivers/SmartThings/zigbee-sensor/src/contact/init.lua
@@ -1,0 +1,32 @@
+-- Copyright 2025 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local defaults = require "st.zigbee.defaults"
+local capabilities = require "st.capabilities"
+
+local is_contact_sensor = function(opts, driver, device)
+  if device:supports_capability(capabilities.contactSensor) then
+    return true
+  end
+end
+
+local generic_contact_sensor = {
+  NAME = "Generic Contact Sensor",
+  supported_capabilities = {
+    capabilities.contactSensor
+  },
+  can_handle = is_contact_sensor
+}
+defaults.register_for_default_handlers(generic_contact_sensor, generic_contact_sensor.supported_capabilities)
+return generic_contact_sensor
--- a/drivers/SmartThings/zigbee-sensor/src/init.lua
+++ b/drivers/SmartThings/zigbee-sensor/src/init.lua
@@ -19,6 +19,7 @@ local capabilities = require "st.capabilities"
 local constants = require "st.zigbee.constants"
 local IasZoneType = require "st.zigbee.generated.types.IasZoneType"
 local device_management = require "st.zigbee.device_management"
+local PowerConfiguration = clusters.PowerConfiguration
 
 local CONTACT_SWITCH = IasZoneType.CONTACT_SWITCH
 local MOTION_SENSOR = IasZoneType.MOTION_SENSOR
@@ -28,6 +29,7 @@ local ZIGBEE_GENERIC_SENSOR_PROFILE = "generic-sensor"
 local ZIGBEE_GENERIC_CONTACT_SENSOR_PROFILE = "generic-contact-sensor"
 local ZIGBEE_GENERIC_MOTION_SENSOR_PROFILE = "generic-motion-sensor"
 local ZIGBEE_GENERIC_WATERLEAK_SENSOR_PROFILE = "generic-waterleak-sensor"
+local ZIGBEE_GENERIC_MOTION_ILLUMINANCE_PROFILE = "generic-motion-illuminance"
 
 local ZONETYPE = "ZoneType"
 local IASZone = clusters.IASZone
@@ -42,6 +44,7 @@ local do_configure = function(self, device)
   device:configure()
   device:send(device_management.build_bind_request(device, IASZone.ID, self.environment_info.hub_zigbee_eui))
   device:send(IASZone.attributes.ZoneStatus:configure_reporting(device, 30, 300, 1))
+  device:send(PowerConfiguration.attributes.BatteryPercentageRemaining:read(device))
 end
 
 local function info_changed(driver, device, event, args)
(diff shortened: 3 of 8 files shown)


drivers/SmartThings/zigbee-sensor/fingerprints.yml
```diff
@@ -5,4 +5,20 @@ zigbeeGeneric:
...
+        - 0x0500
+        - 0x0001
+    deviceProfileName: generic-sensor
+  - id: "generic-button-sensor"
+    deviceLabel: "Zigbee Generic Button"
+    clusters:
+      server:
+        - 0x0500
+        - 0xfe00 # EZVIZ private cluster
+        - 0xfe05 # EZVIZ private cluster
```
greens, line 23: we usually capitalize the letters in our hex

drivers/SmartThings/zigbee-sensor/src/contact/init.lua
```diff
@@ -0,0 +1,67 @@
...
+end
+
+
+local ias_zone_status_attr_handler = function(driver, device, zone_status, zb_rx)
+  generate_event_from_zone_status(driver, device, zone_status, zb_rx)
+end
+
+local ias_zone_status_change_handler = function(driver, device, zb_rx)
+  generate_event_from_zone_status(driver, device, zb_rx.body.zcl_body.zone_status, zb_rx)
+end
```
greens, line 47: are we overwriting the defaults somewhere that makes it necessary to redefine them here?

greens (commented):
I'd appreciate if we could separate the zigbee generic sensor driver from the Tuya drivers.

pInksenberg:
> ias_zone_status_change_handler

pInksenberg:
> I'd appreciate if we could separate the zigbee generic sensor driver from the Tuya drivers.
Currently we have divided tuya and zigbee_generic_sensor into different directories, specifically located within "/drivers/generic_zigbee_sensor" and "/Unofficial/tuya" floder. Are you suggesting that we should further separate the Tuya-powered manufacturers, such as Eviz and Meian, into the "/Unofficial/tuya" folder?

drivers/SmartThings/zigbee-sensor/src/contact/init.lua
pInksenberg: you are right, we double check, and the redefine handler is not necessary and will be removed.

greens:
@pInksenberg I meant have them as two separate pull requests

pInksenberg:
> @pInksenberg I meant have them as two separate pull requests

@greens , sure, I separated tuya devices support into another PR, please kindly check~
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2102/files

varzac (approved):

drivers/SmartThings/zigbee-sensor/fingerprints.yml
```diff
@@ -5,4 +5,20 @@ zigbeeGeneric:
...
+        - 0x0001
+    deviceProfileName: generic-sensor
+  - id: "generic-button-sensor"
+    deviceLabel: "Zigbee Generic Button"
+    clusters:
+      server:
+        - 0x0500
+        - 0xFE00 # EZVIZ private cluster
+        - 0xFE05 # EZVIZ private cluster
+    deviceProfileName: generic-remote-control
```
greens, line 24: if this device is a button it should be in the zigbee button driver

pInksenberg: After our team discussions, we have no better solution for this yet. Could you please share your suggestions? How about just put it to Unofficial/ezviz directory?

greens: You could do that or you could put this exact same fingerprint in the `zigbee-button` driver.

drivers/SmartThings/zigbee-sensor/src/waterleak/init.lua
```diff
@@ -0,0 +1,32 @@
...
+end
+
+local generic_water_sensor = {
+  NAME = "Generic Water Sensor",
+  supported_capabilities = {
+    capabilities.waterSensor
+  },
+  can_handle = is_water_sensor
+}
+defaults.register_for_default_handlers(generic_water_sensor, generic_water_sensor.supported_capabilities)
```
greens, line 31: @varzac Does only registering for defaults in a specific sub-driver work like this?

drivers/SmartThings/zigbee-sensor/fingerprints.yml
inasail: @greens 
Actually, this device is non-WWST device, so we can't put exact fingerprint in zigbee-button driver for this device.
As you know, zigbee-sensor have only generic fingerprint for generic purpose and scalability. I think by adding like this, we can support more ezviz button if they use same private cluster in the future(or now) for their new button devices. and If we need to support more other different types of ezviz devices in the future, then let's move this device into Unofficial/ezviz. but now, Let's just keep this device into zigbee-sensor for a while. What do you think?

  - id: "generic-button-sensor"
    deviceLabel: "Zigbee Generic Button"
    clusters:
      server:
        - 0x0500
        - 0xFE00 # EZVIZ private cluster
        - 0xFE05 # EZVIZ private cluster
    deviceProfileName: generic-remote-control

drivers/SmartThings/zigbee-sensor/src/init.lua
```diff
@@ -132,6 +99,14 @@ local zigbee_generic_sensor_template = {
     doConfigure = do_configure,
     infoChanged = info_changed
   },
+  sub_drivers = {
+    require("contact"),
+    require("motion"),
+    require("waterleak"),
+    require("motion-illuminance"),
+    require("meian"),
+    require("ezviz")
```
inasail, line 108: How about changing to 'meian-button' / 'ezviz-button' for the clarity? (including the folder name of each sub drivers)

drivers/SmartThings/zigbee-sensor/fingerprints.yml
greens: put this fingerprint:
```
  - id: "generic-button-sensor"
    deviceLabel: "Zigbee Generic Button"
    clusters:
      server:
        - 0x0500
        - 0xFE00 # EZVIZ private cluster
        - 0xFE05 # EZVIZ private cluster
    deviceProfileName: generic-remote-control
 ```
 
under a new `zigbeeGeneric:` section here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-button/fingerprints.yml

greens: buttons should use the button driver

inasail: @greens  Then, do you agree to add the generic fingerprint only for 'EZVIZ button(non-WWST)' into zigbee-button(public driver)?
I think 'Meian button' can be moved into 'Unofficial/Tuya' driver with exact fingerprint.
Actually, when we made this driver(zigbee-sensor), I misunderstood the meaning of 'sensor' as just 'iot device',
So I thought it would be okay to add the button devices into this driver(zigbee-sensor) under generic fingerprint.

inasail: @greens 
Then, we will proceed as I mentioned, assuming you agree.
Meian button -> move to 'Unofficial/Tuya' with manufacturer specific fingerprint
EZVIZ button -> move to 'zigbee-button' with generic fingerprint
CC: @pInksenberg

drivers/SmartThings/zigbee-sensor/src/waterleak/init.lua
varzac: So I think this will correctly add the zigbee and capability command handlers as it is just adding them on the "driver" table passed in and the subdrivers use the same expected structure as the base driver.  However, there will be some unnecessary stuff added like cluster configurations, which I don't think are handled for subdrivers (https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/main/lua_libs/st/zigbee/defaults/init.lua#L135)

pInksenberg: Are you suggesting that using "register_for_default_handlers" in a sub-driver might cause unnecessary cluster configurations, and therefore we should implement water capability by adding specific command handlers instead of using the default approach? Could you please explain on this with more details?

drivers/SmartThings/zigbee-sensor/fingerprints.yml
greens: what fingerprint were you expecting the Meian button to match with on this PR as-is?

inasail: Meian button have these clusters.
In-cluster: 0x000, 0x0001, 0x0003, 0x0500
out-cluster: 0x0501, 0x0019, 0x000A

I heard this device is matched with fingerprint below.

  - id: "generic-sensor"
    deviceLabel: "Zigbee Generic Sensor"
    clusters:
      server:
        - 0x0500
    deviceProfileName: generic-sensor

pInksenberg:
@inasail revise finished

drivers/SmartThings/zigbee-sensor/src/waterleak/init.lua
greens: I believe what @varzac is saying is that the default cluster configurations will be added to map that represents the sub-driver, but those fields do not do anything when associated with sub-drivers. It's just extra fields taking up memory that will be unused.

I think it's probably okay over re-defining defaults in every sub-driver. Would you agree, Zach?

varzac: Yes, I think it's fine.  Was just pointing out that there is some unused behavior.

greens (approved):

PR #2109, opened 2025-05-01
Matter Switch: Add 13 button profile

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] New profile

# Description of Change

Add new profiles to support 13-button devices.


ldeora:
Is there a reason why the profiles for 9, 10 and 11 buttons are intentionally omitted? A device like the 3rd Reality MK1 Keyboard would work out of the box with these profiles. The device and the platform are Matter certified after all...

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
```diff
@@ -0,0 +1,149 @@
...
+end
+
+local function do_configure(driver, device)
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    match_profile(driver, device)
+  end
+end
+
+local function driver_switched(driver, device)
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER then
```
hcarter-775, line 118: do we need these NETWORK_TYPE_MATTER checks, if we don't expect any child devices to be using this subdriver?

nickolas-deboom: Yeah good point, I don't think they're needed

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
```diff
@@ -0,0 +1,149 @@
...
+
+-------------------------------------------------------------------------------------
+-- Third Reality MK1 specifics
+-------------------------------------------------------------------------------------
+
+local THIRD_REALITY_MANUFACTURER_ID = 0x1407
+local THIRD_REALITY_MK1_PRODUCT_ID = 0x1388
+
+local function is_third_reality_mk1(opts, driver, device)
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
```
hcarter-775, line 31: Adding onto my other comment, this check actually blocks all child devices from joining this subdriver.

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
```diff
@@ -0,0 +1,149 @@
...
+
+local function device_init(driver, device)
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    device:set_endpoint_to_component_fn(endpoint_to_component)
+    subscribe(device)
+  end
+end
+
+local function device_added(driver, device)
+  device_init(driver, device)
```
hcarter-775, line 95: maybe let's just subscribe or even `return` here rather than re-call the whole init? I understand we're mostly doing this call to overwrite the main driver, so I think we could also make a comment of that.

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
nickolas-deboom: Perfect, I removed the other checks

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
nickolas-deboom: The main driver also calls device_init within device_added, which was done to ensure init is called in the case of device caching: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/ed8fe8f57a5d710c9aa11d02622168ef7e2d2f3e

So I think it might be needed here still?

hcarter-775: @ctowns what was the reasoning for adding this init function in added for only matter switch? I see it was added in [this pr](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1445) but I don't see a description.

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
```diff
@@ -0,0 +1,143 @@
...
+-------------------------------------------------------------------------------------
+
+local THIRD_REALITY_MANUFACTURER_ID = 0x1407
+local THIRD_REALITY_MK1_PRODUCT_ID = 0x1388
+
+local function is_third_reality_mk1(opts, driver, device)
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
+     device.manufacturer_info.vendor_id == THIRD_REALITY_MANUFACTURER_ID and
+     device.manufacturer_info.product_id == THIRD_REALITY_MK1_PRODUCT_ID then
+    return true
```
hcarter-775, line 34: can we add a log like `log.info("Using Third Reality MK1 sub driver")`

nickolas-deboom: Added in last commit!

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
```diff
@@ -0,0 +1,144 @@
...
+  if device.profile.id ~= args.old_st_store.profile.id then
+    configure_buttons(device)
+    subscribe(device)
+  end
+end
+
+local function match_profile(driver, device)
+  device:try_update_metadata({profile = "12-button-keyboard"})
+  build_button_component_map(device)
+  configure_buttons(device)
```
hcarter-775, line 107: would we want to use the new buttons.configure call here as well?

nickolas-deboom: Yes but that will need to wait until the other PR 1754 gets merged (or if this one gets merged first, then that other PR will need to do this change)

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
ctowns: This was done to ensure that `subscribe` was called for all devices, because in some cases there is a race condition that causes the `init` event to not fire for some devices. Therefore, calling `init` again in `added` was meant to ensure that the processing in `init` was occuring even if the `init` event was not queued due to the race condition.

Parent-Child devices were primarily affected, so I wouldn't expect it to be an issue here.

There is a detail explanation here: https://smartthings.atlassian.net/browse/CHAD-13449

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
```diff
@@ -0,0 +1,144 @@
...
+end
+
+local function device_added(driver, device)
+  device_init(driver, device)
+end
+
+local function info_changed(driver, device, event, args)
+  if device.profile.id ~= args.old_st_store.profile.id then
+    configure_buttons(device)
+    subscribe(device)
```
ctowns, line 100: You could potential just override the `device:subscribe` function directly by calling:
```
device:extend_device("subscribe", subscribe)
```

which will override the library `subscribe` function directly and point the device to the `subscribe` function you have defined here. I can't think of any cases where this would be needed over the solution you currently have, but if a library function call somewhere calls `device:subscribe`, it might be nice to just have the default `subscribe` itself overridden, since that is what the `extend_device` api is meant to be used for.

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
nickolas-deboom: Sure, updated as suggested.

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
nickolas-deboom: I like this idea and made this change in a recent commit

hcarter-775 (approved):
left a comment about simplifying `added`, otherwise this looks good to go imo since it's been thoroughly tested.

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
```diff
@@ -0,0 +1,137 @@
...
+  device:set_field(COMPONENT_TO_ENDPOINT_MAP, component_map, {persist = true})
+end
+
+local function device_init(driver, device)
+  device:set_endpoint_to_component_fn(endpoint_to_component)
+  device:extend_device("subscribe", subscribe)
+  device:subscribe()
+end
+
+local function device_added(driver, device) end
```
hcarter-775, line 88: can you also add a comment stating why this is an empty handler? like you did for aqara-cube and eve-energy in the other PR?

nickolas-deboom: Done

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
```diff
@@ -0,0 +1,138 @@
...
+  for component, endpoint in pairs(map) do
+    if endpoint == ep then
+      return component
+    end
+  end
+  return "F1"
+end
+
+-- override subscribe function to prevent subscribing to additional events from the main driver
+local function subscribe(device)
```
hcarter-775, line 50: was this required? The main driver will do the same thing, no?

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
```diff
@@ -0,0 +1,138 @@
...
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
+     device.manufacturer_info.vendor_id == THIRD_REALITY_MK1_FINGERPRINT.vendor_id and
+     device.manufacturer_info.product_id == THIRD_REALITY_MK1_FINGERPRINT.product_id then
+    log.info("Using Third Reality MK1 sub driver")
+    return true
+  end
+  return false
+end
+
+local function endpoint_to_component(device, ep)
```
hcarter-775, line 39: is this required? there will always be a map set for this device, no? So the special "F1" override is unneccesary?

nickolas-deboom: Yes it's needed since we are overriding `device_init` and therefore need to call `set_endpoint_to_component_fn` from the subdriver

nickolas-deboom: That's correct

drivers/SmartThings/matter-switch/src/third-reality-mk1/init.lua
hcarter-775: The only reason this would be important would be in the case that the FW for this device updates to handle more robust button presses and will necessitate changes in the future

nickolas-deboom: Yeah that's true.

PR #2111, opened 2025-05-06
Remove unique refresh handler in matter switch

hcarter-775:
# Description of Change
Remove the unique handle_refresh call in matter switch to allow the native capability handler to run instead. This will allow a greater number of distinct attributes to be refreshed in a refresh call, as is expected.

# Summary of Completed Tests
Current unit tests continue to pass.


ctowns (approved):
This looks good to me, I would just do a spot test on a device and it should be good :+1:

nickolas-deboom (approved):

PR #2114, opened 2025-05-07
Add co-comeas-colevel-battery profile

hcarter-775:
# Description of Change
Add new profile to fit a newly found use case.

# Summary of Completed Tests


drivers/SmartThings/matter-sensor/profiles/co-comeas-colevel-battery.yml
```diff
@@ -0,0 +1,20 @@
+name: co-comeas-battery
```
ctowns, line 1: Update name to match profile name

hcarter-775: fixed

nickolas-deboom (approved):

PR #2128, opened 2025-05-15
Fix/unit tests 57 beta

cjswedes:
57.x FW and lua libraries contain support for native attribute registration. This PR updates all the zb/zw/matter tests to include expectations for this new feature. One thing to note about the defaults is that when handling a protocol message that maps to a child device it is often the case that the parent will be registered to be handled natively; rather than was driver cycles determining if the device has a child to avoid the registration, we are relying on the hubs implementation which does not allow for parents or children to be handled natively.

There was also a bug fix in `utils.update` that required some unit test fixes due to the way the integration tests used the function.

I intend to do a 0.57 beta release artifact today and get this merged.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/test/test_electrical_sensor.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_electrical_sensor.lua
@@ -725,6 +725,14 @@ test.register_message_test(
       }
     },
     {
+      channel = "devices",
+      direction = "send",
+      message = {
+        "register_native_capability_attr_handler",
+        { device_uuid = mock_device.id, capability_id = "switchLevel", capability_attr_id = "level" }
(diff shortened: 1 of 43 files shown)


nickolas-deboom:
Just wondering, does this break tests on <= 56?

cjswedes:
> Just wondering, does this break tests on <= 56?

Yes, the 56 lua libs do not have the native attribute registration API, so these test changes will not pass there . You can see in the test results they will throw errors due to functions that dont exist: `test_aqara_wall_switch.lua:71: attempt to call a nil value (method 'expect_native_attr_handler_registration')`. I just finished building the 57 artifact so Ill release that and everything should be passing again.

greens (approved):

PR #2130, opened 2025-05-16
Matter Thermostat: Use larger temperature range

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

For RPC >= 6, the hub converts temperature values from capability commands to Celsius, meaning the driver no longer needs to use separate ranges for Celsius and Fahrenheit in order to determine the unit as Celsius can always be assumed. A range is still used, in order to filter out values received from the device that are likely to be erroneous, but it is now larger and will be able to accommodate the temperature range for more use cases.

A similar change was made to matter-appliance in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1757

# Summary of Completed Tests

Verified with unit tests and tested with VDA Thermostat.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -116,20 +116,21 @@ local SUPPORTED_WATER_HEATER_MODES_WITH_IDX = "__supported_water_heater_modes_wi
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 local MGM3_PPM_CONVERSION_FACTOR = 24.45
 
--- This is a work around to handle when units for temperatureSetpoint is changed for the App.
+-- For RPC version >= 6, we can always assume that the values received from temperatureSetpoint
+-- are in Celsius, but we still limit the setpoint range to somewhat reasonable values.
+-- For RPC <= 5, this is a work around to handle when units for temperatureSetpoint is changed for the App.
 -- When units are switched, we will never know the units of the received command value as the arguments don't contain the unit.
 -- So to handle this we assume the following ranges considering usual thermostat/water-heater temperatures:
 -- Thermostat:
 --   1. if the received setpoint command value is in range 5 ~ 40, it is inferred as *C
 --   2. if the received setpoint command value is in range 41 ~ 104, it is inferred as *F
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -112,20 +112,21 @@ local SUPPORTED_WATER_HEATER_MODES_WITH_IDX = "__supported_water_heater_modes_wi
...
 -- When units are switched, we will never know the units of the received command value as the arguments don't contain the unit.
 -- So to handle this we assume the following ranges considering usual thermostat/water-heater temperatures:
 -- Thermostat:
 --   1. if the received setpoint command value is in range 5 ~ 40, it is inferred as *C
 --   2. if the received setpoint command value is in range 41 ~ 104, it is inferred as *F
-local THERMOSTAT_MAX_TEMP_IN_C = 40.0
-local THERMOSTAT_MIN_TEMP_IN_C = 5.0
-
+local THERMOSTAT_MAX_TEMP_IN_C = version.rpc >= 6 and 100.0 or 40.0
+local THERMOSTAT_MIN_TEMP_IN_C = version.rpc >= 6 and -10.0 or 5.0
```
hcarter-775, line 124: is -10 a reasonable value? Maybe we should let this be 0 if we want a simpler number than 5. otherwise, 5 seems fine to me personally

nickolas-deboom: There is a device that supports down to 1 degC, but yeah -10 might be too far so I might prefer 0 now that I think about it. It's hard to imagine a thermostat that could go below freezing :sweat_smile:. Still, the setpoint min/max values are mandatory attributes so the values listed here are really just a catch-all in case they are for some reason not implemented or just invalid, which we have seen occasionally.

hcarter-775: yeah, I think the freezing and boiling points of water can be a good "base" standard for what a thermostat should be allowed to support

hcarter-775 (approved):

ctowns (approved):

PR #2131, opened 2025-05-19
Include new fingerprint, update comments

hcarter-775:
# Description of Change
Include a new fingerprint variation and update the comments for device type specificity

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-sensor/fingerprints.yml
+++ b/drivers/SmartThings/matter-sensor/fingerprints.yml
@@ -219,30 +219,30 @@ matterGeneric:
   - id: "matter/air-quality-sensor"
     deviceLabel: Matter Air Quality Sensor
     deviceTypes:
-      - id: 0x002C
+      - id: 0x002C # Air Quality Sensor
     deviceProfileName: aqs
   - id: "matter/smoke-co-alarm"
     deviceLabel: Matter Smoke/CO Alarm
     deviceTypes:
-      - id: 0x0076
+      - id: 0x0076 # Smoke CO Alarm
     deviceProfileName: smoke-co
   - id: "matter/rain/sensor"
     deviceLabel: Matter Rain Sensor
     deviceTypes:
-      - id: 0x0044
+      - id: 0x0044 # Rain Sensor
     deviceProfileName: rain-battery
   - id: "matter/freeze/detector"
     deviceLabel: Matter Water Freeze Detector
     deviceTypes:
-      - id: 0x0041
+      - id: 0x0041 # Freeze Detector
     deviceProfileName: freeze-battery
   - id: "matter/leak/detector"
     deviceLabel: Matter Water Leak Detector
     deviceTypes:
-      - id: 0x0043
+      - id: 0x0043 # Leak Sensor
     deviceProfileName: leak-battery
   - id: "matter/flow/sensor"
     deviceLabel: Matter Flow Sensor
     deviceTypes:
-      - id: 0x0306
+      - id: 0x0306 # Flow Sensor
     deviceProfileName: flow-battery
--- a/drivers/SmartThings/matter-thermostat/fingerprints.yml
+++ b/drivers/SmartThings/matter-thermostat/fingerprints.yml
@@ -128,6 +128,14 @@ matterGeneric:
       - id: 0x0301 # Thermostat
       - id: 0x0307 # Humidity Sensor
     deviceProfileName: air-purifier-thermostat-humidity-aqs
+    - id: "matter/ap/aqs/thermostat/humidity"
+    deviceLabel: Matter Air Purifier & Quality Sensor
+    deviceTypes:
+      - id: 0x002D # Air Purifier
+      - id: 0x002C # Air Quality Sensor
+      - id: 0x0302 # Temperature Sensor
+      - id: 0x0307 # Humidity Sensor
+    deviceProfileName: air-purifier-thermostat-humidity-aqs
   - id: "matter/water-heater"
     deviceLabel: Matter Water Heater
     deviceTypes:


drivers/SmartThings/matter-thermostat/fingerprints.yml
```diff
@@ -128,6 +128,14 @@ matterGeneric:
       - id: 0x0301 # Thermostat
       - id: 0x0307 # Humidity Sensor
     deviceProfileName: air-purifier-thermostat-humidity-aqs
+    - id: "matter/ap/aqs/temperature/humidity"
```
nickolas-deboom, line 131: nit - could you de-dent this line?

nickolas-deboom (approved):

drivers/SmartThings/matter-thermostat/fingerprints.yml
hcarter-775: ah, good catch!

drivers/SmartThings/matter-thermostat/fingerprints.yml
```diff
@@ -128,6 +128,14 @@ matterGeneric:
...
       - id: 0x0307 # Humidity Sensor
     deviceProfileName: air-purifier-thermostat-humidity-aqs
+  - id: "matter/ap/aqs/temperature/humidity"
+    deviceLabel: Matter Air Purifier & Quality Sensor
+    deviceTypes:
+      - id: 0x002D # Air Purifier
+      - id: 0x002C # Air Quality Sensor
+      - id: 0x0302 # Temperature Sensor
+      - id: 0x0307 # Humidity Sensor
+    deviceProfileName: air-purifier-thermostat-humidity-aqs
```
ctowns, line 138: Won't this need a new profile? This generic fingerprint does not have a thermostat, but the profile does here.

drivers/SmartThings/matter-sensor/fingerprints.yml
```diff
@@ -219,30 +219,30 @@ matterGeneric:
   - id: "matter/air-quality-sensor"
     deviceLabel: Matter Air Quality Sensor
     deviceTypes:
-      - id: 0x002C
+      - id: 0x002C # Air Quality Sensor
```
ctowns, line 222: Thanks for adding these comments! I'll make sure I include them on future generic fingerprints too :+1:

drivers/SmartThings/matter-thermostat/fingerprints.yml
hcarter-775: The only "thermostat" capability in that profile is `temperatureMeasurement`, which is still the one we'd want here.

hcarter-775: Oh wait, actually there is an issue here- the check `if #thermostat_eps > 0 then
      profile_name = profile_name .. "-thermostat"` won't actually hit since there is no thermostat device type. :thinking: actually we don't support profiling for the "temperature sensor" device type in the matter thermostat driver at the moment, so some changes will need to be made to support it.

hcarter-775: should be fixed in the latest commit. I also updated the "thermostat" profiles that had been created previously to include the generic thermostat capabilities they were missing.

drivers/SmartThings/matter-thermostat/profiles/air-purifier-hepa-ac-thermostat-humidity-fan-nostate-nobattery-aqs-pm25-tvoc-meas.yml
```diff
@@ -9,6 +9,12 @@ components:
     version: 1
   - id: temperatureMeasurement
     version: 1
+  - id: thermostatMode
```
ctowns, line 12: Are there devices that are using this profile? I think we want to be careful adding capabilities to a profile if it is already in use, since those capability may not actually work on device that are already using this profile. We could make a new profile instead that has the superset of the capabilities added here.

drivers/SmartThings/matter-thermostat/profiles/air-purifier-temperature-humidity-aqs.yml
```diff
@@ -0,0 +1,20 @@
+name: air-purifier-temperature-humidity-aqs
```
ctowns, line 1: Is there any profile-switching name-builder logic that would eventually map to this profile? Should their be? It seems that if a device joins with the generic fingerprint that links to this profile, then the profile name would be overwritten during the configure logic to determine the new profile name, right? So wouldn't we need to add a branch in the logic that eventually creates this profile name?

drivers/SmartThings/matter-thermostat/profiles/air-purifier-hepa-ac-thermostat-humidity-fan-nostate-nobattery-aqs-pm25-tvoc-meas.yml
hcarter-775: there shouldn't be. Also, I confirmed with a superset query that no devices are onboarded with this profile.

drivers/SmartThings/matter-thermostat/profiles/air-purifier-temperature-humidity-aqs.yml
hcarter-775: yes, there is logic that can build this profile :+1:

hcarter-775: if you look at the PR, there actually is some added logic for this

drivers/SmartThings/matter-thermostat/profiles/air-purifier-thermostat-humidity-aqs.yml
```diff
@@ -8,10 +8,27 @@ components:
...
+        - key: "thermostatOperatingState.value"
+          enabledValues:
+            - idle
+            - cooling
+            - heating
   - id: relativeHumidityMeasurement
     version: 1
   - id: airQualityHealthConcern
     version: 1
+  - id: battery
```
ctowns, line 30: Is this actually supported? And if it is, should `-battery` be added to the profile name?

ctowns: I only see this below in the AP profile switch section, which seems to only include batteryLevel and no battery. Most APs wouldn't support battery I would think, and most of our AP profiles do not.
```
if battery_supported == battery_support.BATTERY_LEVEL then
      elseif battery_supported == battery_support.NO_BATTERY then
        profile_name = profile_name .. "-nobattery"
      end
```

drivers/SmartThings/matter-thermostat/profiles/air-purifier-temperature-humidity-aqs.yml
ctowns: I saw the temp and humidity stuff below, I was asking about the AQS support that is being added here. However, I see that support already existed via this line which existed prior to this PR.
```
profile_name = profile_name .. create_air_quality_sensor_profile(device)
```

So I understand now that AQS was already supported in the driver configuration logic, we just needed to add a profile for this particular combo of capabilities

drivers/SmartThings/matter-thermostat/profiles/air-purifier-thermostat-humidity-aqs.yml
hcarter-775: ah, I see. Will remove :+1:

drivers/SmartThings/matter-thermostat/profiles/air-purifier-hepa-ac-thermostat-humidity-fan-nostate-nobattery-aqs-pm25-tvoc-meas.yml
hcarter-775: Nope, the creation of this profile in the first place was actually a bit of a mistake honestly, and I'm trying to patch that error here. 

Also, I actually created a third profile just now in this PR that is the same as the one we're discussing but with an extra `-nostate` tag that removes `thermostatOperatingState`, and I made this the new generic fingerprint, since a `-nostate` profile is a more preferable default.

ctowns: No I think what you've said here makes sense to me! If those other profiles were just made accidentally or are no longer needed, we can remove them, but I think you're repurposed them here so that they are more useful. 

I also agree that the `no-state` profile is a more preferable default, and the configuration logic should handle profiles switching in any cases where that default won't work as well.

drivers/SmartThings/matter-thermostat/fingerprints.yml
```diff
@@ -127,7 +127,15 @@ matterGeneric:
...
-    deviceProfileName: air-purifier-thermostat-humidity-aqs
+    deviceProfileName: air-purifier-thermostat-nostate-humidity-aqs
+  - id: "matter/ap/aqs/temperature/humidity"
+    deviceLabel: Matter Air Purifier & Quality Sensor
+    deviceTypes:
+      - id: 0x002D # Air Purifier
+      - id: 0x002C # Air Quality Sensor
+      - id: 0x0302 # Temperature Sensor
+      - id: 0x0307 # Humidity Sensor
+    deviceProfileName: air-purifier-temperature-humidity-aqs
```
ctowns, line 138: Do you need to update this to the `no-state` version?

drivers/SmartThings/matter-thermostat/profiles/air-purifier-hepa-ac-thermostat-humidity-fan-nostate-nobattery-aqs-pm25-tvoc-meas.yml
ctowns: So to sum up each of these profiles that are being added/modified here:
* `air-purifier-hepa-ac-temperature-humidity-fan-aqs-pm25-tvoc-meas.yml` - added for better temp/humidity handling of an existing profile
* `air-purifier-hepa-ac-thermostat-humidity-fan-nostate-nobattery-aqs-pm25-tvoc-meas.yml` - modified to include thermostat capabilities since it is in the name, but they were not previously present.
* `air-purifier-temperature-humidity-aqs` - added as variant of default fingerprint, but with operating state.
* `air-purifier-thermostat-humidity-aqs` - modified to include thermostat capabilities since it is in the name
* `air-purifier-thermostat-nostate-humidity-aqs` added as new default profile for AP/AQS/Thermostat devices.

Is that the gist of these profile changes?

PR #2136, opened 2025-05-20
Add pause, resume, goHome command to matter-rvc driver

HunsupJung:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests


drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -343,6 +452,64 @@ local function rvc_operational_error_attr_handler(driver, device, ib, response)
...
+  )
+  local cap_op_enum = capabilities.robotCleanerOperatingState.operatingState
+  if operating_state ~= cap_op_enum.stopped.NAME and operating_state ~= cap_op_enum.running.NAME and
+     operating_state ~= cap_op_enum.paused.NAME and operating_state ~= cap_op_enum.seekingCharger.NAME and
+     operating_state ~= cap_op_enum.charging.NAME and operating_state ~= cap_op_enum.docked.NAME then
+      operating_state = "Error"
+  end
+
+  -- Set Supported Operating State Commands
+  local cap_op_cmd = capabilities.robotCleanerOperatingState.commands
```
greens, line 496: Are there circumstance where `supportedOperatingStateCommands` does not equal `cap_op_cmd`?

Or where `device:get_field(OPERATING_STATE_SUPPORTED_COMMANDS)` does not equal the `latest_value` of `supportedOperatingStateCommands`?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -77,6 +81,9 @@ end
 local function device_init(driver, device)
   device:subscribe()
   device:set_component_to_endpoint_fn(component_to_endpoint)
+  local req = im.InteractionRequest(im.InteractionRequest.RequestType.READ, {})
+  req:merge(clusters.RvcOperationalState.attributes.AcceptedCommandList:read())
+  device:send(req)
```
hcarter-775:
```suggestion
  device:send(clusters.RvcOperationalState.attributes.AcceptedCommandList:read())
```

hcarter-775: Also, I want to know if we need this read in `init`? Would it not be better to add this attribute to the list in `subscribed_attributes` like:

```
  [capabilities.robotCleanerOperatingState.ID] = {
    clusters.RvcOperationalState.attributes.OperationalState,
    clusters.RvcOperationalState.attributes.OperationalError,
    clusters.RvcOperationalState.attributes.AcceptedCommandList
  },
```

And if we would rather have this be a one-time call, it may be better to put this read in the `doConfigure` handler.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -36,6 +39,7 @@ end
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 local RUN_MODE_SUPPORTED_MODES = "__run_mode_supported_modes"
 local CLEAN_MODE_SUPPORTED_MODES = "__clean_mode_supported_modes"
+local OPERATING_STATE_SUPPORTED_COMMANDS = "__operating_state_supported_commands"
```
hcarter-775, line 42: also, it may be better to utilize the capability attribute `supportedOperatingStateCommands` here instead of adding a new table?

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Thank you for your advice. I will move it to subscribed_attributes

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: When hub set the `supportedOperatingStateCommands`, the Plugin check this value and display. Hub should set this value depend on the status of Operating state and Run mode.
For example, if the device is in the charging dock and is stopped, only the `start(resume)` button must be visible to the user, and the hub should add only the `resume` command to `supportedOperatingStateCommands`

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -46,7 +49,8 @@ local subscribed_attributes = {
   },
   [capabilities.robotCleanerOperatingState.ID] = {
     clusters.RvcOperationalState.attributes.OperationalState,
-    clusters.RvcOperationalState.attributes.OperationalError
+    clusters.RvcOperationalState.attributes.OperationalError,
+    clusters.RvcOperationalState.attributes.AcceptedCommandList
```
hcarter-775: Thank you for adding my suggestion, though I'm still curious, why do we want this to be run multiple times? Should this not be a one-time read event?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -101,14 +105,99 @@ local function info_changed(driver, device, event, args)
 end
 
 -- Helper functions --
+local function is_support_go_home(device)
```
hcarter-775: I think we should handle this like so:

```
local function supports_rvc_operational_state(device, command_name)
  local supported_op_commands = device:get_field(OPERATING_STATE_SUPPORTED_COMMANDS) or {}
  for _, cmd in ipairs(supported_op_commands) do
    if cmd == command_name then return true end
  end
  return false
end
```

Then we can call:
 `supports_rvc_operational_state(device, cap_op_cmds.goHome.NAME)` 
and 
`supports_rvc_operational_state(device, cap_op_cmds.pause.NAME)` 
as needed.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -101,14 +105,99 @@ local function info_changed(driver, device, event, args)
...
+  local supported_op_commands = device:get_field(OPERATING_STATE_SUPPORTED_COMMANDS) or {}
+  for _, cmd in ipairs(supported_op_commands) do
+    if cmd == capabilities.robotCleanerOperatingState.commands.goHome.NAME then
+      return true;
+    end
+  end
+  return false;
+end
+
+local function can_send_go_home_cmd(device, operating_state)
```
hcarter-775: Similar to my suggestion for changes to `supports_rvc_operational_state`, I believe it makes sense to make a master function like:
```
local function can_send_state_command(device, new_operating_state, cur_operating_state, current_tag)
... put a series of if/else statements clarifying the unique logic per `new_operating_state` ... 
end
```
I believe this will help clarify the function under a single main helper function. Further, this will allow one single call to `if operating_state == "Error" then return false end` at the top of this master function instead of spreading this logic across functions.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -101,14 +105,99 @@ local function info_changed(driver, device, event, args)
...
 
-  -- Error state
   if operating_state == "Error" then
+    -- Set Supported Operating State Commands to empty
+    local event = capabilities.robotCleanerOperatingState.supportedOperatingStateCommands(
+      {}, {visibility = {displayed = false}}
+    )
+    device:emit_component_event(device.profile.components["main"], event)
     -- Set runMode to empty
     local event = capabilities.mode.supportedArguments({}, {visibility = {displayed = false}})
```
hcarter-775: remove local here since `event` is already defined now

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -420,6 +585,119 @@ local function robot_cleaner_areas_selection_response_handler(driver, device, ib
...
+    capabilities.robotCleanerOperatingState.operatingState.NAME
+  )
+  local cap_op_enum = capabilities.robotCleanerOperatingState.operatingState
+  if operating_state ~= cap_op_enum.stopped.NAME and operating_state ~= cap_op_enum.running.NAME and
+     operating_state ~= cap_op_enum.paused.NAME and operating_state ~= cap_op_enum.seekingCharger.NAME and
+     operating_state ~= cap_op_enum.charging.NAME and operating_state ~= cap_op_enum.docked.NAME then
+      operating_state = "Error"
+  end
+
+  if can_send_pause_cmd(device, current_tag, operating_state) == true then
```
hcarter-775: is this logic required? If Pause as clicked from the app, it must have been displayed on the device card, which means it is in the supported list at the moment, no?

The other case I can imagine where this handler might be used if not clicked would be through an automation (though I'm not sure if the automation would be deleted if the supported value were removed from the supported values list). Do you know what occurs in this case?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -420,6 +585,119 @@ local function robot_cleaner_areas_selection_response_handler(driver, device, ib
...
+  local operating_state = device:get_latest_state(
+    "main",
+    capabilities.robotCleanerOperatingState.ID,
+    capabilities.robotCleanerOperatingState.operatingState.NAME
+  )
+  local cap_op_enum = capabilities.robotCleanerOperatingState.operatingState
+  if operating_state ~= cap_op_enum.stopped.NAME and operating_state ~= cap_op_enum.running.NAME and
+     operating_state ~= cap_op_enum.paused.NAME and operating_state ~= cap_op_enum.seekingCharger.NAME and
+     operating_state ~= cap_op_enum.charging.NAME and operating_state ~= cap_op_enum.docked.NAME then
+      operating_state = "Error"
```
hcarter-775: Has any work been proposed to add Error to the capability definition enum for RobotCleanerOperatingState? Then we wouldn't need such complex logic to handle the error case

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -420,6 +585,119 @@ local function robot_cleaner_areas_selection_response_handler(driver, device, ib
...
+  end
+
+  if can_send_go_home_cmd(device, current_tag, operating_state) == true then
```
hcarter-775: similar question as [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2136/files#r2136185298). If the capability has been handled, it must be supported, no? There is still the open question about automations though.

hcarter-775: also, if this is required, it should be:
```suggestion
  if can_send_go_home_cmd(device, operating_state) == true then
```
pending some other changes to this logic I suggested (where we have a single master can_send function)

drivers/SmartThings/matter-rvc/src/test/test_matter_rvc.lua
```diff
@@ -15,15 +15,20 @@
 local test = require "integration_test"
 local capabilities = require "st.capabilities"
 local t_utils = require "integration_test.utils"
-
```
hcarter-775, line 18: we need new unit tests to verify this complex functionality works as expected.

drivers/SmartThings/matter-rvc/src/test/test_matter_rvc.lua
```diff
@@ -15,15 +15,20 @@
...
+  clusters.OperationalState = require "OperationalState"
+end
 
-clusters.RvcCleanMode = require "RvcCleanMode"
-clusters.RvcOperationalState = require "RvcOperationalState"
-clusters.RvcRunMode = require "RvcRunMode"
-clusters.OperationalState = require "OperationalState"
-clusters.ServiceArea = require "ServiceArea"
-clusters.Global = require "Global"
+if version.api < 13 then
```
hcarter-775, line 28: is there any documentation that describes what the expected and desired handling for these changes should look like?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -420,6 +585,119 @@ local function robot_cleaner_areas_selection_response_handler(driver, device, ib
...
+    "main",
+    capabilities.robotCleanerOperatingState.ID,
+    capabilities.robotCleanerOperatingState.operatingState.NAME
+  )
+  local cap_op_enum = capabilities.robotCleanerOperatingState.operatingState
+  if operating_state ~= cap_op_enum.stopped.NAME and operating_state ~= cap_op_enum.running.NAME and
+     operating_state ~= cap_op_enum.paused.NAME and operating_state ~= cap_op_enum.seekingCharger.NAME and
+     operating_state ~= cap_op_enum.charging.NAME and operating_state ~= cap_op_enum.docked.NAME then
+      operating_state = "Error"
+  end
```
hcarter-775:
```suggestion
  local operating_state = device:get_latest_state(
    "main",
    capabilities.robotCleanerOperatingState.ID,
    capabilities.robotCleanerOperatingState.operatingState.NAME
  ) or "Error"
```
This occurs another time as well, and should probably look like this? Unless we are expecting the latest_state to be something besides `nil` in these cases, and if so, what are we expecting?

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: You are right. We don't need to call AceeptedCommandList multiple time. I focused on moving the attribute to subscribed_attributes, but if it's important to read one-time, I will move it to doConfiguration.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: It seems to be much more optimized.

drivers/SmartThings/matter-rvc/src/init.lua
hcarter-775: In general, we don't want to send more reads than necessary to the device if we can help it, and making sure this only runs once helps reduce the possibility of edge cases to handle. Thank you!

nickolas-deboom: @hcarter-775 do you think it would be better to leave it as a subscribed attribute rather than introducing a new read? That way we'd get the initial value and then wouldn't need to process it again unless it changed on the device (which seems unlikely but maybe could happen with a FW upgrade?)

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: I will try this

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: If you two decide, I will modify it as it is.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: All enum values except `stopped`, `running`, `paused`, `chargingCharger`, `charging,` and `docked` are error values. The remaining values are expressed as actual UX, so the operating state should not be saved as an error.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: As mentioned above, in Cluster, the `Operational State` value and the `Operational Error` value are divided, but in Capability, this is expressed as `robotCleanerOperatingState` attribute.

https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/main/datatypes/RobotCleanerOperatingState.yml

drivers/SmartThings/matter-rvc/src/init.lua
hcarter-775: @nickolas-deboom, we get a read from all subscribed data every time we re-subscribe, which would happen every time `init` occurs at the moment. Therefore, to mitigate possible edge cases and to avoid re-reading this attribute after the one-time configuration logic, I believe it should be in `doConfigure`.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -105,101 +106,82 @@ local function info_changed(driver, device, event, args)
...
+         current_state == cap_op_enum.charging.NAME then
+        return true
+      end
     end
   end
   return false
 end
 
-local function can_send_change_to_idle_mode_cmd(device, current_tag, operating_state)
+local function can_send_change_to_idle_mode_cmd(device, current_tag, current_state)
```
hcarter-775: not a huge deal but why not add this logic to the `can_send_state_command` function?

drivers/SmartThings/matter-rvc/src/init.lua
hcarter-775: I guess I'm curious, if the `robotCleanerOperatingState` capability is not one of the options that you loop through in this code, what is it? Is it nil, or some data that we can know ahead of time?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -36,6 +38,7 @@ end
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 local RUN_MODE_SUPPORTED_MODES = "__run_mode_supported_modes"
 local CLEAN_MODE_SUPPORTED_MODES = "__clean_mode_supported_modes"
+local OPERATING_STATE_SUPPORTED_COMMANDS = "__operating_state_supported_commands"
```
hcarter-775, line 41: rather than saving this table, I think we should be using the capability attribute: `supportedOperatingStateCommands`. I believe this is what this argument is supposed to be used for, and the presentation for `actions` and `automations` in the capability depends on this attribute.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -343,6 +432,64 @@ local function rvc_operational_error_attr_handler(driver, device, ib, response)
   end
 end
 
+local function handle_rvc_operational_state_accepted_command_list(driver, device, ib, response)
```
nickolas-deboom, line 429: Should this function be reading the accepted commands from `ib.data.elements`? Right now it looks like it's assuming every available command is an accepted command.

drivers/SmartThings/matter-rvc/src/init.lua
nickolas-deboom: I guess I don't see the harm in reading it again considering the logic that is present in `handle_rvc_operational_state_accepted_command_list`, and wouldn't we want to know if it changed so we can set `OPERATING_STATE_SUPPORTED_COMMANDS` properly? Either way, I left a comment below but I think the handler needs to be updated because it appears to be assuming all the commands are accepted commands rather than checking the value of the attribute.

It looks like we subscribe to this attribute in matter-appliance as well.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: The following values can be set.
  - unableToStartOrResume
  - unableToCompleteOperation
  - commandInvalidInState
  - failedToFindChargingDock
  - stuck
  - dustBinMissing
  - dustBinFull
  - waterTankEmpty
  - waterTankMissing
  - waterTankLidOpen
  - mopCleaningPadMissing

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: This function reads the accepted commands from `ib.data.elements`.
I confirmed eufy supports `pause`, `resume` with this code.

https://samsung.slack.com/archives/C038KA8K78S/p1749533882955699?thread_ts=1749477408.863689&cid=C038KA8K78S

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Okay I will add it~!

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: As mentioned above, `supportedOperatingStateCommands` is not used for the purpose you mentioned. 
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2136#discussion_r2131831537

As you said, if we want to support automation, `supportedArguments` must be used like mode capability.
https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/main/capabilities/mode.src/mode.yml#L23-L31

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: I will remove this logic.

HunsupJung: > The other case I can imagine where this handler might be used if not clicked would be through an automation (though I'm not sure if the automation would be deleted if the supported value were removed from the supported values list). Do you know what occurs in this case?

I checked this part. When RVC cannot receive `pause` command, but the Automation send `pause` command to RVC, RVC send `OperationalCommandResponse` with `CommandInvalidInState`
But as you said, this situation doesn't happen. The test was tested by forcibly changing the code so that Automation can send commands.

So, I will remove this logic

drivers/SmartThings/matter-rvc/src/test/test_matter_rvc.lua
HunsupJung: No, I simply synced with `init.lua`. Is there a direction you want?

HunsupJung:
@hcarter-775 , @nickolas-deboom 
We must put this patch before 6/16. Please review and approve before that.

drivers/SmartThings/matter-rvc/src/init.lua
hcarter-775: I see, thanks Hunsup

drivers/SmartThings/matter-rvc/src/init.lua
hcarter-775: I see, I had misread the definition. Sorry about that!

Do you know if there are any plans to support this capability attribute `supportedOperatingStates`? I see it's not currently utilized.

Perhaps, in the future (not this PR) we could use it and set it to the states we have support for. Then we could loop through it instead of using this type of check:
```
  if current_state ~= cap_op_enum.stopped.NAME and current_state ~= cap_op_enum.running.NAME and
     current_state ~= cap_op_enum.paused.NAME and current_state ~= cap_op_enum.seekingCharger.NAME and
     current_state ~= cap_op_enum.charging.NAME and current_state ~= cap_op_enum.docked.NAME then
      current_state = "Error"
  end
```

drivers/SmartThings/matter-rvc/src/init.lua
hcarter-775: Hmm, but in other drivers we just do a single read on the Accepted Attributes List for PowerSource, since we only want that data once. Like you said, there's probably not a lot of harm either way, but I am generally against subscribing to an attribute we only want once. I would prefer this to be a one-time read in `doConfigure`

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -420,6 +562,62 @@ local function robot_cleaner_areas_selection_response_handler(driver, device, ib
...
+      break
+    end
+  end
+  local current_state = device:get_latest_state(
+    "main",
+    capabilities.robotCleanerOperatingState.ID,
+    capabilities.robotCleanerOperatingState.operatingState.NAME
+  )
+  local cap_op_enum = capabilities.robotCleanerOperatingState.operatingState
+  if current_state ~= cap_op_enum.stopped.NAME and current_state ~= cap_op_enum.running.NAME and
```
hcarter-775, line 589: if we set `supportedOperatingStates` in doConfigure to be the values we expect, we can simply loop through this field at this point and the other place this occurs.

drivers/SmartThings/matter-rvc/src/init.lua
hcarter-775: @nickolas-deboom it's doing a check using a table and will add to the list of the table key-value isn't nil.

drivers/SmartThings/matter-rvc/src/test/test_matter_rvc.lua
hcarter-775: No, that's alright. This handling is a little unique so I was wondering if it had been documented from a UI perspective at all.

hcarter-775:
@HunsupJung we can make sure this is merged into the beta channel on 6/16. 

There are 3 threads left on this PR that I haven't resolved (for clarity) that could be addressed before then, but the major one left is that this needs some unit tests verifying the new behavior.

Otherwise, this looks good to me. Thank you!

drivers/SmartThings/matter-rvc/src/test/test_matter_rvc.lua
HunsupJung: I already added new unit tests.
The new unit tests cover all new operations.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: I don't think so. because `supportedOperatingStates` can include error status as like the followings.
If we want to use `supportedOperatingStates` we should add error status that be supported by the device too, so we cannot use supportedOperatingStates at this point.

  - unableToStartOrResume
  - unableToCompleteOperation
  - commandInvalidInState
  - failedToFindChargingDock
  - stuck
  - dustBinMissing
  - dustBinFull
  - waterTankEmpty
  - waterTankMissing
  - waterTankLidOpen
  - mopCleaningPadMissing

drivers/SmartThings/matter-rvc/src/test/test_matter_rvc.lua
hcarter-775: Sorry Hunsup, missed those! Thanks

PR #2152, opened 2025-05-28
Matter Switch: Remove Energy Setters in added

hcarter-775:
# Description of Change
Remove `emit_event` calls in added. In the case of `powerMeter`, this will be populated with a distinct value after `subscribe` occurs either in `device_init` or `info_changed`. It is the same in the case of `energyMeter`. In both cases, setting an initial value of 0 on added does not add any meaningful data to the capability handling, and so this is a wasted call.

# Summary of Completed Tests
One test removed related to ensuring this was running.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -1318,14 +1318,6 @@ local function device_added(driver, device)
     device:send(req)
   end
 
-  -- Reset the values
-  if device:supports_capability(capabilities.powerMeter) then
-    device:emit_event(capabilities.powerMeter.power({ value = 0.0, unit = "W" }))
-  end
-  if device:supports_capability(capabilities.energyMeter) then
-    device:emit_event(capabilities.energyMeter.energy({ value = 0.0, unit = "Wh" }))
-  end
-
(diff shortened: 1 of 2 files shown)


nickolas-deboom (approved):

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1318,14 +1318,6 @@ local function device_added(driver, device)
     device:send(req)
   end
 
-  -- Reset the values
```
ctowns, line 1321: I'm not familiar, but is there a reason this was done originally? I see it is also being done in the Eve subdriver. Is there any kind of delay in receiving this data when testing with a device that prompted the need for this to be zeroed out initially when added?

hcarter-775: I only did this initially because I was less familiar with the handling and it was like this in the Eve subdriver. But in our case, knowing what we know about matter clusters, there is no reason for this to be like so. I can't say for Eve though, which is why I left it.

ctowns: Sounds good to me!

PR #2154, opened 2025-05-29
Add N/A checking for molecular weight of AQS level handling

hcarter-775:
# Description of Change
Add checks for n/a values in unit conversion table, and in that case does not convert the received value.

# Summary of Completed Tests
Unit tests included to handle cases.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -1233,8 +1233,12 @@ local conversion_tables = {
   [units.PPM] = {
     [units.PPM] = function(value) return utils.round(value) end,
     [units.PPB] = function(value) return utils.round(value * (10^3)) end,
-    [units.UGM3] = function(value, molecular_weight) return utils.round((value * molecular_weight * 10^3) / MGM3_PPM_CONVERSION_FACTOR) end,
-    [units.MGM3] = function(value, molecular_weight) return utils.round((value * molecular_weight) / MGM3_PPM_CONVERSION_FACTOR) end,
+    [units.UGM3] = function(value, molecular_weight) 
+      if molecular_weight == "N/A" then return utils.round(value) end
+      return utils.round((value * molecular_weight * 10^3) / MGM3_PPM_CONVERSION_FACTOR) end,
+    [units.MGM3] = function(value, molecular_weight)
+      if molecular_weight == "N/A" then return utils.round(value) end
+      return utils.round((value * molecular_weight) / MGM3_PPM_CONVERSION_FACTOR) end,
   },
   [units.PPB] = {
     [units.PPM] = function(value) return utils.round(value/(10^3)) end,
@@ -1249,7 +1253,9 @@ local conversion_tables = {
   },
   [units.UGM3] = {
     [units.UGM3] = function(value) return utils.round(value) end,
-    [units.PPM] = function(value, molecular_weight) return utils.round((value * MGM3_PPM_CONVERSION_FACTOR) / (molecular_weight * 10^3)) end,
+    [units.PPM] = function(value, molecular_weight)
+      if molecular_weight == "N/A" then return utils.round(value) end
+      return utils.round((value * MGM3_PPM_CONVERSION_FACTOR) / (molecular_weight * 10^3)) end,
   },
   [units.NGM3] = {
     [units.UGM3] = function(value) return utils.round(value/(10^3)) end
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1233,8 +1233,12 @@ local conversion_tables = {
   [units.PPM] = {
     [units.PPM] = function(value) return utils.round(value) end,
     [units.PPB] = function(value) return utils.round(value * (10^3)) end,
-    [units.UGM3] = function(value, molecular_weight) return utils.round((value * molecular_weight * 10^3) / MGM3_PPM_CONVERSION_FACTOR) end,
-    [units.MGM3] = function(value, molecular_weight) return utils.round((value * molecular_weight) / MGM3_PPM_CONVERSION_FACTOR) end,
+    [units.UGM3] = function(value, molecular_weight)
+      if molecular_weight == "N/A" then return utils.round(value) end
```
ctowns, line 1237: Shouldn't this be some kind of error? If we can't do the unit conversion, we should log an error that an invalid unit was received. The value that is returned here would also not be valid, because it wouldn't actually have converted to the correct unit right? I would be more in favor of not showing the user invalid data if we know it is invalid.

hcarter-775: I don't think invalid is necessarily correct, because that assumes that if the type of contaminant doesn't have a molecular value we can use for the conversion, we refuse to show it. This is moreso an issue of forcing dustSensor to be shown in ppm.

Perhaps we could just add more enum values for the units of these capabilities?

But for now, the fact that we can't convert it in a valid way shouldn't stop the value from being displayed, even if it's a little wrong.

hcarter-775: I do agree that we could log something here, but not displaying a value at all seems incorrect to me

ctowns: If that's the case, then we should add the molecular weight for each dust measurement to the molecular weights table and use that for the conversion. That seems like the more complete solution here rather than ignoring the conversion.

hcarter-775: yeah, that would be sweet. The issue there is that from everything I'm seeing on the internet, there is no "molecular weight" for dust. It's just not a thing, since that's not a particular contaminant

ctowns: Yeah I am seeing that too... I think we'll have to expand the capability to include the new unit type.

I don't know how common it is to measure dust in non-ppm levels, but maybe it is more common outside the US? I don't think we've seen that yet which is why the issue with this conversion has not been seen until now.

hcarter-775: yeah, agreed. I'll look into the logs a little more to see what they're sending this in and hopefully get to the bottom of it.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1289,10 +1295,11 @@ local function measurementHandlerFactory(capability_name, attribute, target_unit
     end
 
     if value then
-      device:emit_event_for_endpoint(ib.endpoint_id, attribute({value = value, unit = unit_strings[target_unit]}))
       -- handle case where device profile supports both fineDustLevel and dustLevel
       if capability_name == capabilities.fineDustSensor.NAME and device:supports_capability(capabilities.dustSensor) then
```
nickolas-deboom, line 1299: Maybe I'm reading this wrong, but does it need an additional check like this?
```suggestion
      if (capability_name == capabilities.fineDustSensor.NAME and device:supports_capability(capabilities.dustSensor)) or (capability_name == capabilities.dustSensor.NAME and device:supports_capability(capabilities.fineDustSensor)) then
```

hcarter-775: No, we wouldn't need this check since dustSensor is a superset of fineDustSensor, but not the other way around. So this check handles the edge case that we get a pm25 reading but we don't support fineDustSensor (because we use dustSensor instead)

nickolas-deboom: Ah, thanks for the explanation!

hcarter-775:
gonna close this, since the device that was causing this to be needed was changed to support a different measurement value type that we have support for.

PR #2156, opened 2025-05-29
sonos: Address a few nits from previous PR

dljsjr:
This also has the additional goal of getting another Alpha build that will use the new packaging, which is needed to inject the new API key.


drivers/SmartThings/sonos/src/lifecycle_handlers.lua
```diff
@@ -45,84 +45,63 @@ function SonosDriverLifecycleHandlers.initialize_device(driver, device)
     capabilities.mediaTrackControl.commands.previousTrack.NAME,
   }))
 
-  if not driver:has_received_startup_state() then
+  if api_version >= 14 and not driver:has_received_startup_state() then
```
dljsjr, line 48: @varzac  https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2077#discussion_r2046924200

drivers/SmartThings/sonos/src/ssdp.lua
```diff
@@ -355,7 +355,7 @@ function _ssdp_mt:setwaker(kind, waker)
 
   assert(
     self.waker_ref == nil or waker == nil,
-    "Waker already set, cannot await SSDP serach instance from multiple locations."
+    "Waker already set, cannot await SSDP search instance from multiple locations."
```
dljsjr, line 358: @cjswedes https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2077#discussion_r2070507110

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -316,7 +302,7 @@ function SonosDriver:get_oauth_token()
     if now < expiration then
       -- token is expiring soon, so we pre-emptively refresh
       if math.abs(expiration - now) < ONE_HOUR_IN_SECONDS then
-        local result, err = self:request_oauth_token()
+        local result, err = security.get_sonos_oauth()
```
dljsjr, line 305: This was a potential recursion issue that I caught while testing.

dljsjr:
@cjswedes @varzac One thing I haven't addressed yet is this line: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/8467097af6e3c376347180020ea7d87887b14b8e/drivers/SmartThings/sonos/src/api/sonos_connection.lua#L224

There was discussion on the previous PR about putting a timeout w/ a retry on this. I'm still mostly of the mindset that we don't want to do this. This was something we went over quite a bit during the design stage, and we made a decision that if we don't have a linked Sonos account, then we don't want to "poll"/constantly attempt to create an account link.

I could be convinced otherwise, but we went through a great deal of pain to make sure this specific detail was possible, and I think it could actually cause more problems than it solves.

Again, though, I'm always open to discussion.

dljsjr:
@cjswedes RE: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/8467097af6e3c376347180020ea7d87887b14b8e/drivers/SmartThings/sonos/src/api/sonos_connection.lua#L278-L287 and https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2077#discussion_r2070656137

My intent: Having it immediately after inspecting the error code is a latency optimization, mostly; the reconnect loop *does* make a token request if there isn't one currently in-flight, like you said. This puts one in flight.

I could be swayed to consolidate it, but since the token handling is potentially several seconds of latency + highly async, I thought it would be beneficial to kick off the process ASAP. It might not actually do much in the grand scheme of things, though.

cjswedes (approved):

dljsjr:
> @dljsjr my hub is on 56 for rollback testing and am seeing this when trying to discover my sonos device
> 
> ```
> 2025-05-30_16:20:36.986 | I | broker     | <Sonos (4d111ba2)> Starting Sonos discovery
> 2025-05-30_16:20:38.100 | I | cloud      | Sending srv_notify: 0x0514/1300: c21a6c77-872c-474e-be5b-5f6f11a240ef Sonos: LuaLibv13: runtime error: [string "sonos_driver.lua"]:547: attempt to call a nil value (field 'bus')
> stack traceback:
> 	[string "sonos_driver.lua"]:547: in field 'new_driver_template'
> 	[string "init.lua"]:10: in main chunk
> 2025-05-30_16:20:38.104 | I | errlogger  | add serverNotify: 1300 -> DS03-1300
> 2025-05-30_16:20:38.105 | E | broker     | <Sonos (c21a6c77)> runtime error: [string "sonos_driver.lua"]:547: attempt to call a nil value (field 'bus')
> stack traceback:
> 	[string "sonos_driver.lua"]:547: in field 'new_driver_template'
> 	[string "init.lua"]:10: in main chunk
> ```

:facepalm: `cosock/bus.lua` is vendored, but `cosock/init.lua` isn't, so calling `.bus()` on `cosock` will obviously fail.

That bus is only used for the OAuth flow, so we can probably just avoid instantiating it with an API wrapper check.

drivers/SmartThings/sonos/src/init.lua
```diff
@@ -1,3 +1,9 @@
+local cosock = require "cosock"
```
dljsjr, line 1: @NoahCornell

NoahCornell: Nice that is fixing the crash but now I am seeing
```
2025-05-30T18:33:19.360231487Z WARN Sonos  No SSDP task for driver currently running, cannot perform discovery
```
Looks like that gets spawned as part of `handle_startup_state_received` which won't get called pre-57

dljsjr: @NoahCornell https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2156/files#diff-fd98e5f9cd685da688b3464a6b348e53d4ae6b64630cd49d185ea8d5a05a3ed6R33-R41

NoahCornell: A new crash now, let me know if you need more
```
2025-05-30T21:21:32.405136293Z FATAL Sonos  runtime error: [string "cosock.lua"]:250: [string "sonos_driver.lua"]:300: attempt to index a nil value (field 'hub_augmented_driver_data')
stack traceback:
	[string "sonos_driver.lua"]:300: in method 'get_oauth_token'
	[string "sonos_driver.lua"]:210: in method 'check_auth'
	[string "sonos_driver.lua"]:371: in function <[string "sonos_driver.lua"]:342>
stack traceback:
	[C]: in function 'error'
	[string "cosock.lua"]:250: in upvalue 'step_thread'
	[string "cosock.lua"]:358: in field 'run'
	[string "st/driver.lua"]:1096: in method 'run'
	[string "init.lua"]:43: in main chunk
```

NoahCornell: My device is added and controllable! I'll get back to 57 today and do some more tests

NoahCornell (approved):

PR #2159, opened 2025-05-30
Add reconfiguration for zigbee power meter to reduce reporting

varzac:
This PR adds code to reconfigure devices supporting the powerMeter capability to reduce the minimum reporting interval.  This changes to 5 seconds instead of 1 second.  On device init, if a device is detected as having not performed this migration (based on a field set on the device datastore), and supporting the capability a 5 minute timer is started if one isn't already running.  At the end of the 5 minutes, all devices the driver knows about are checked and the reconfiguration is attempted.  When we hear a success back from the device, the migration is marked as complete.  This will be attempted at most once per device init, but won't be reattempted periodically without another eligible device being initted.  This means for offline devices, this won't be re-attempted until the driver restarts.  I think in a future release I'd like to consider a more general solution in the firmware to handle future configuration changes.

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/configurations.lua
+++ b/drivers/SmartThings/zigbee-switch/src/configurations.lua
@@ -14,9 +14,15 @@
 
 local clusters = require "st.zigbee.zcl.clusters"
 local constants = require "st.zigbee.constants"
+local capabilities = require "st.capabilities"
+local device_def = require "st.device"
 
 local ColorControl = clusters.ColorControl
 local IASZone = clusters.IASZone
+local Status = require "st.zigbee.generated.types.ZclStatus"
+
+local CONFIGURATION_VERSION_KEY = "_configuration_version"
+local CONFIGURATION_ATTEMPTED = "_reconfiguration_attempted"
 
 local devices = {
   IKEA_RGB_BULB = {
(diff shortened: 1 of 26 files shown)


drivers/SmartThings/zigbee-switch/src/configurations.lua
```diff
@@ -60,8 +63,29 @@ local devices = {
   }
 }
 
+
 local configurations = {}
 
+configurations.power_reconfig_wrapper = function(orig_function)
```
cjswedes: Is orig_function supposed to be called from this function?

drivers/SmartThings/zigbee-switch/src/configurations.lua
```diff
@@ -60,8 +63,29 @@ local devices = {
...
+  local new_init = function(driver, device)
+    local config_version = device:get_field(CONFIGURATION_VERSION_KEY)
+    if config_version == nil or config_version < driver.current_config_version then
```
cjswedes, line 93: I think we should also be using `Device:add_configured_attribute` to ensure if configure is called (I think some subdrivers do this from refresh handler) that the new configuration isn't overwritten.

drivers/SmartThings/zigbee-switch/src/configurations.lua
```diff
@@ -60,8 +63,29 @@ local devices = {
...
+          -- Increase minimum reporting interval to 5 seconds
+          device:send(clusters.ElectricalMeasurement.attributes.ActivePower:configure_reporting(device, 5, 600, 5))
+        end
+        if device:supports_server_cluster(clusters.SimpleMetering.ID) then
+          -- Increase minimum reporting interval to 5 seconds
+          device:send(clusters.SimpleMetering.attributes.InstantaneousDemand:configure_reporting(device, 5, 600, 5))
+        end
+      end
+    end
+    device:set_field(CONFIGURATION_VERSION_KEY, driver.current_config_version, {persist = true})
```
tpmanley: If we want to increase our confidence the new config gets applied then instead of setting this field here we could wait for the Configure Reporting Response command. That would help with the case where the device is offilne when when we first try to reconfigure it.

varzac: Discussing with Kianoosh he wanted to get a quick change out with small change, and then review to provide a larger change later.  I think if we want to do a more complete look at the idea of "config versions" and checking for success, I would rather do that in the lua libs for the next release so it can be used in the future with any driver, and give us more time for review/testing.

drivers/SmartThings/zigbee-switch/src/configurations.lua
varzac: lol, yes

drivers/SmartThings/zigbee-switch/src/configurations.lua
```diff
@@ -60,8 +63,29 @@ local devices = {
...
 local configurations = {}
 
+configurations.power_reconfig_wrapper = function(orig_function)
+  local new_init = function(driver, device)
+    local config_version = device:get_field(CONFIGURATION_VERSION_KEY)
+    if config_version == nil or config_version < driver.current_config_version then
+      if device:supports_capability(capabilities.powerMeter) then
+        if device:supports_server_cluster(clusters.ElectricalMeasurement.ID) then
+          -- Increase minimum reporting interval to 5 seconds
+          device:send(clusters.ElectricalMeasurement.attributes.ActivePower:configure_reporting(device, 5, 600, 5))
```
greens, line 97: I think that in the past we've run into race conditions when sending actual zigbee messages inside `init`.

drivers/SmartThings/zigbee-switch/src/test/test_all_capability_zigbee_bulb.lua
```diff
@@ -369,12 +370,31 @@ test.register_coroutine_test(
...
+      assert(mock_device:get_field("_configuration_version") == nil)
+      test.socket.zigbee:__expect_send({mock_device.id, ElectricalMeasurement.attributes.ActivePower:configure_reporting(mock_device, 5, 600, 5)})
+      test.socket.zigbee:__expect_send({mock_device.id, SimpleMetering.attributes.InstantaneousDemand:configure_reporting(mock_device, 5, 600, 5)})
+      test.socket.device_lifecycle:__queue_receive({ mock_device.id, "init" })
+      test.wait_for_events()
+      assert(mock_device:get_field("_configuration_version") == 1)
+    end,
+    {
+      test_init = function()
+        test.disable_startup_messages()
```
varzac: @aleclorimer these got included because I was testing with the lua libs main branch.  I will update these tests to not include these, but after I merge, you will need to add the two new test functions to your PR as well.

drivers/SmartThings/zigbee-switch/src/configurations.lua
varzac: Thinking about this, the only issue I could see is if this is a child device (which is a real issue, and I will need to fix).  Do you have any more context on what the issues were?

drivers/SmartThings/zigbee-switch/src/test/test_all_capability_zigbee_bulb.lua
varzac: Actually, with the way I changed these tests to work with older libs, it works for both old and new, so it doesn't require follow up.

drivers/SmartThings/zigbee-switch/src/configurations.lua
greens: there was a really old issue where the hub EUI occasionally wasn't set yet when some of these sends were called, iirc

it's possible that it has been fixed

varzac: I re-worked this to schedule a timer to then do the actual configuration.

drivers/SmartThings/zigbee-switch/src/configurations.lua
varzac: Done

varzac:
I have addressed the comments and added tests.  I will test with a couple of different physical devices at home, but barring unexpected issues there I think this is ready for review.

kianooshkarami (approved):

varzac:
Testing done:

Switched a single power meter to the driver, verified after 5 minutes new configuration was sent to the device.
Restarted driver, verified configuration does not happen again
Switched 2 more devices in quick succession, waited 5 minutes, verified both newly added devices were configured, but not the first device.
Added a device that doesn't support power meter, wait 5 minutes, verify no configuration was attempted.

drivers/SmartThings/zigbee-switch/src/configurations.lua
cjswedes: > it's possible that it has been fixed

I believe this has been fixed, at least for the driver restart case, however there is still possibilities that the value isn't present after a hub restart due to the zb radio not being up; I have not seen this actually happen though.

cjswedes (approved):

cjswedes:
There is one failed test BTW that seems related to this and should be fixed.

varzac:
> There is one failed test BTW that seems related to this and should be fixed.

Thanks, previously there was a matter test failing because of the capability stuff, and so I hadn't rechecked.  This was because the channel ordering needed to be relaxed since the order we send to the devices was not deterministic.  I've also added a test to cover the case that the device configuration is at 1, which was implicitly covered by the other tests, but I made an explicit one.

drivers/SmartThings/zigbee-switch/src/configurations.lua
varzac: Ok, @tpmanley I decided this was a good idea, so I made the change to only bump the version on success.  This does mean that these devices will re-try this on every driver restart, but that seems acceptable to me, given that it will only happen once per restart, and with the 5 minute timer, drivers in a restart loop won't be sending these.

varzac:
Ok I have re-run all on target tests with the success status check as well as added several tests to cover different outcomes of the configure reporting response.  Technically there are some edge cases (specifically if there is a different configuration that happens while waiting for the response that is a success, but then this request is a failure), but they are unlikely enough, that I think that if any device hits it and stays on the older reporting, I consider that an acceptable outcome for a driver only addressing of this change.

drivers/SmartThings/zigbee-switch/src/configurations.lua
```diff
@@ -60,8 +66,94 @@ local devices = {
...
+    end
+  end
+end
+
+configurations.power_reconfig_wrapper = function(orig_function)
+  local new_init = function(driver, device)
+    local config_version = device:get_field(CONFIGURATION_VERSION_KEY)
+    if config_version == nil or config_version < driver.current_config_version then
+      if driver._reconfig_timer == nil then
+        driver._reconfig_timer = driver:call_with_delay(5*60, configurations.check_and_reconfig_devices, "reconfg_power_devices")
```
tpmanley: nit:

```suggestion
        driver._reconfig_timer = driver:call_with_delay(5*60, configurations.check_and_reconfig_devices, "reconfig_power_devices")
```

drivers/SmartThings/zigbee-switch/src/configurations.lua
```diff
@@ -60,8 +66,94 @@ local devices = {
...
+        if device:supports_server_cluster(clusters.SimpleMetering.ID) then
+          -- Increase minimum reporting interval to 5 seconds
+          device:send(clusters.SimpleMetering.attributes.InstantaneousDemand:configure_reporting(device, 5, 600, 5))
+          device:add_configured_attribute(instantaneous_demand_configuration)
+        end
+      end
+      device:set_field(CONFIGURATION_ATTEMPTED, true, {persist = true})
+    end
+  end
+  driver._reconfig_timer = nil
```
tpmanley, line 109: Just want to confirm I'm understanding correctly, the intention is to attempt the config at most once per device each time the driver starts up vs trying periodically in case the device comes back online?

varzac: Yes, IMO the burden of running a perpetual timer to check all these things is more overhead than it's worth for the edge case of offline devices, and they will be covered the next time the driver restarts.

PR #2171, opened 2025-06-04
fix: memory leak and CPU starvation in SSDP periodic scan

dljsjr:
# Type of Change

- [x] Bug fix

# Checklist

# Description of Change

- We weren't calling `:handled()` on the interval timer, which could put
  the SSDP coroutine in to a hot send loop, and this hot RPC activity
  was starving the process from being able to handle activity on other
  channels
- We were keying in to the wrong table in one location, which was
  causing us to retain a bunch of queue'd messages indefinitely, which
  would lead to unbounded growth if left running long enough.

# Summary of Completed Tests

Tested with real devices


drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
```diff
@@ -96,6 +96,7 @@ local function make_persistent_task_impl(
 
       for _, receiver in ipairs(recv_ready) do
         if receiver == interval_timer then
+          interval_timer:handled()
```
dljsjr, line 99: CPU starvation bug was here

drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
```diff
@@ -298,16 +299,19 @@ function sonos_ssdp.spawn_persistent_ssdp_task()
       event_bus:send(info_to_send)
       local mac_addr = utils.extract_mac_addr(info_to_send.discovery_info.device)
       local waiting_handles = task_handle.waiting_for_unique_key[unique_key] or {}
+
       log.debug(st_utils.stringify_table(waiting_handles, "waiting for unique keys", true))
-      for _, v in pairs(task_handle.waiting_for_unique_key[mac_addr] or {}) do
+      for _, v in pairs(task_handle.waiting_for_mac_addr[mac_addr] or {}) do
```
dljsjr, line 304: Unbounded queue growth was here

varzac (approved):

PR #2173, opened 2025-06-06
fix: Address regressions around group controls and onboarding

dljsjr:
# Type of Change

- [x] Bug fix

# Checklist

# Description of Change

This PR fixes two main issues:

1. There were downstream bugs that were hidden by a previous fix related to an unbounded queue growing; in fixing that, we surfaced another bug that was unintentionally relying on the buggy behavior.
2. There were a few bugs related to speaker groups that were left over from the original refactor, related to devices becoming joined or separated in specific ways, that were surfaced during regression testing of the other fix.

# Summary of Completed Tests

Tested on a real multi-speaker topology

### DIFF
--- a/drivers/SmartThings/sonos/src/utils.lua
+++ b/drivers/SmartThings/sonos/src/utils.lua
@@ -122,6 +122,29 @@ function utils.new_mac_address_keyed_table()
   return setmetatable({}, _mac_addr_key_mt)
 end
 
+---@param tbl table<string,any>
+---@param key string
+local function __case_insensitive_key_index(tbl, key)
+  assert(type(key) == "string", "key for CaseInsensitiveKeyTable must be a string!")
+  local lowercase = key:lower()
+  return rawget(tbl, lowercase)
+end
+
+local function __case_insensitive_key_newindex(tbl, key, value)
+  assert(type(key) == "string", "key for CaseInsensitiveKeyTable must be a string!")
+  rawset(tbl, key:lower(), value)
+end
+
+local _case_insensitive_key_mt = {
+  __index = __case_insensitive_key_index,
+  __newindex = __case_insensitive_key_newindex,
+  __metatable = "CaseInsensitiveKeyTable",
+}
+
+function utils.new_case_insensitive_table()
+  return setmetatable({}, _case_insensitive_key_mt)
+end
+
 ---@param sonos_device_info SonosDeviceInfo
 function utils.extract_mac_addr(sonos_device_info)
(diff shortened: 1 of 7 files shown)


drivers/SmartThings/sonos/src/utils.lua
```diff
@@ -122,6 +122,33 @@ function utils.new_mac_address_keyed_table()
...
+---@param tbl table<string,any>
+---@param key string
+local function __case_insensitive_key_index(tbl, key)
+  assert(type(key) == "string", "key for CaseInsensitiveKeyTable must be a string!")
+  local raw = rawget(tbl, key)
+  if raw ~= nil then
+    return raw
+  end
+  local lowercase = key:lower()
+  return rawget(tbl, lowercase)
```
varzac: Is there a reason to try the raw key first vs just always calling lower?

drivers/SmartThings/sonos/src/utils.lua
```diff
@@ -122,6 +122,33 @@ function utils.new_mac_address_keyed_table()
...
+
+local function __case_insensitive_key_newindex(tbl, key, value)
+  assert(type(key) == "string", "key for CaseInsensitiveKeyTable must be a string!")
+  rawset(tbl, key:lower(), value)
+end
+
+local _case_insensitive_key_mt = {
+  __index = __case_insensitive_key_index,
+  __newindex = __case_insensitive_key_newindex,
+  __metatable = "CaseInsensitiveKeyTable",
```
varzac, line 141: What does setting a metatable to a string do?

varzac (commented):
Couple minor comments but overall looks OK

NoahCornell (approved):

drivers/SmartThings/sonos/src/utils.lua
dljsjr: Setting `__metatable` to anything non-nil creates a "protected" metatable that can't be overridden/replaced by calling `setmetatable`, and it will return that value when `getmetatable` is used instead of the actual metatable itself. It doesn't change the *actual* metatable that Lua uses for resolving metamethods, it just changes how this metatable is exposed to users interacting with the base table.

So it's kinda like setting a `.class` name/value if you use a string.

drivers/SmartThings/sonos/src/utils.lua
dljsjr: Ah, good catch, and probably not. When I originally prototyped this as a solution I wasn't going to use a `__newindex` to normalize inserts until I realized that it could lead to key duping. I forgot to come back to this after I added a `__newindex`.

PR #2174, opened 2025-06-06
philips-hue: Refresh device on online sse update

NoahCornell:
Check all that apply

# Type of Change

- [ ] Bug fix

# Checklist

# Description of Change

Inject refresh command when we receive a zigbee connectivity update from the sse stream. Fixes an issue where device cards are grayed out after discovery when the device was offline and then went online.

# Summary of Completed Tests


drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
```diff
@@ -175,6 +175,13 @@ function AttributeEmitters.connectivity_update(child_device, zigbee_status)
     child_device.log.info_with({hub_logs=true}, "Device zigbee status event, marking device online")
     child_device:online()
     child_device:set_field(Fields.IS_ONLINE, true)
+    if driver then
```
dljsjr: I only recently realized myself that `.driver` is a field on a device record, so you can actually call `child_device.driver` here instead of passing an argument, unless you want to use that argument as a sentinel value; in which case I think it should be a boolean argument.

dljsjr (approved):
One minor comment, not a blocker in any way.

drivers/SmartThings/philips-hue/src/handlers/attribute_emitters.lua
NoahCornell: Ah I'll use that then to stay in line with the other emitters in the file

PR #2181, opened 2025-06-09
Matter Appliance minor refactor

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [x] Refactor

# Checklist

# Description of Change

This change moves some common functionality into common-utils.lua, implements do_configure in the subdrivers rather than the main driver, and also contains some minor fixes involving embedded clusters.

# Summary of Completed Tests

Regression tested with VDA for each supported device type.

### DIFF
--- a/drivers/SmartThings/matter-appliance/src/test/test_laundry_washer.lua
+++ b/drivers/SmartThings/matter-appliance/src/test/test_laundry_washer.lua
@@ -1,0 +1,187 @@
+-- Copyright 2025 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local test = require "integration_test"
+test.set_rpc_version(6)
+local t_utils = require "integration_test.utils"
+local capabilities = require "st.capabilities"
+local clusters = require "st.matter.clusters"
+
+local APPLICATION_ENDPOINT = 1
+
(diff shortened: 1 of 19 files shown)


hcarter-775 (approved):
nice job. seems very necessary :+1:

nickolas-deboom:
I made an update to move some of the temperature attribute and capability handlers into the main driver and remove them from the subdrivers because these have the same behavior in all cases. I left `temperature_setpoint_attr_handler` and `setpoint_limit_handler` defined in the subdrivers because difference min and max values are used in these handlers depending on the device type.

drivers/SmartThings/matter-appliance/src/test/test_laundry_washer.lua
```diff
@@ -0,0 +1,187 @@
+-- Copyright 2024 SmartThings
```
nickolas-deboom: I moved these test cases into their own file in order to properly test the doConfigure logic in `test_init`

hcarter-775:
> I made an update to move some of the temperature attribute and capability handlers into the main driver and remove them from the subdrivers because these have the same behavior in all cases. I left temperature_setpoint_attr_handler and setpoint_limit_handler defined in the subdrivers because difference min and max values are used in these handlers depending on the device type.

do you think it would be too much to put them in the main driver but make the min/max be parameters that can be passed in by the actual implementation in the subdriver? Or is that too much you think

nickolas-deboom:
> > I made an update to move some of the temperature attribute and capability handlers into the main driver and remove them from the subdrivers because these have the same behavior in all cases. I left temperature_setpoint_attr_handler and setpoint_limit_handler defined in the subdrivers because difference min and max values are used in these handlers depending on the device type.
> 
> do you think it would be too much to put them in the main driver but make the min/max be parameters that can be passed in by the actual implementation in the subdriver? Or is that too much you think

In the last commit I attempted this, but ended up leaving the handlers in the subdriver, determine which set of min/max values should be used, and calling a helper function in the common module to do the handling. I don't think there is a way to avoid having the handlers in the subdrivers because the endpoint / component are needed to make the determination on which range to use. Let me know what you think!

hcarter-775:
works for me, I like it :+1:

PR #2189, opened 2025-06-12
Feature/matter britzyhub

KyuminAhn:
Check all that apply

# Type of Change

          - [x] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
Initial Commit of matter-britzyhub edge driver

# Summary of Completed Tests
Tested gas valve
Tested elevator
Tested dimming light

### DIFF
--- a/drivers/SmartThings/sinux-britzyhub/fingerprints.yml
+++ b/drivers/SmartThings/sinux-britzyhub/fingerprints.yml
@@ -12,25 +12,25 @@ matterGeneric:
   - id: "elevator-call-up"
     deviceLabel: 상행 엘리베이터 호출
     deviceTypes:
-      - id: 0x0103 # On Off Switch
+      - id: 0xFF02 # On Off Switch
     deviceProfileName: elevator
   - id: "elevator-call-down"
     deviceLabel: 하행 엘리베이터 호출
     deviceTypes:
-      - id: 0x0103 # On Off Switch
+      - id: 0xFF02 # On Off Switch
     deviceProfileName: elevator
   - id: "elevator-call"
     deviceLabel: 엘리베이터 호출
     deviceTypes:
-      - id: 0x0103 # On Off Switch
+      - id: 0xFF02 # On Off Switch
     deviceProfileName: elevator
   - id: "gas-valve"
     deviceLabel: 가스밸브
     deviceTypes:
-      - id: 0x0303 # Pump
+      - id: 0xFF01 # GasValve
     deviceProfileName: gas-valve
   - id: "ventilator"
     deviceLabel: 전열교환기
     deviceTypes:
-      - id: 0x002D # Air Purifier
+      - id: 0xFF03 # Air Purifier
     deviceProfileName: ventilator
--- a/drivers/SmartThings/sinux-britzyhub/fingerprints.yml
+++ b/drivers/SmartThings/sinux-britzyhub/fingerprints.yml
@@ -1,0 +1,36 @@
+matterGeneric:
+  - id: "dimming-light"
+    deviceLabel: 디밍 조명
+    deviceTypes:
+      - id: 0x0101 # Dimmable Light
+    deviceProfileName: dimming-light
+  #- id: "air-conditioner"
+  #  deviceLabel: 에어컨
+  #  deviceTypes:
+  #    - id: 0x0072 # Room Air Conditioner
+  #  deviceProfileName: air-conditioner
+  - id: "elevator-call-up"
+    deviceLabel: 상행 엘리베이터 호출
+    deviceTypes:
+      - id: 0x0103 # On Off Switch
+    deviceProfileName: elevator
+  - id: "elevator-call-down"
+    deviceLabel: 하행 엘리베이터 호출
+    deviceTypes:
+      - id: 0x0103 # On Off Switch
+    deviceProfileName: elevator
+  - id: "elevator-call"
+    deviceLabel: 엘리베이터 호출
+    deviceTypes:
+      - id: 0x0103 # On Off Switch
+    deviceProfileName: elevator
+  - id: "gas-valve"
+    deviceLabel: 가스밸브
+    deviceTypes:
+      - id: 0x0303 # Pump
+    deviceProfileName: gas-valve
+  - id: "ventilator"
+    deviceLabel: 전열교환기
+    deviceTypes:
+      - id: 0x002D # Air Purifier
+    deviceProfileName: ventilator
--- a/drivers/SmartThings/sinux-britzyhub/profiles/elevator.yml
+++ b/drivers/SmartThings/sinux-britzyhub/profiles/elevator.yml
@@ -1,0 +1,8 @@
+name: elevator
+components:
+  - id: main
+    capabilities:
+      - id: elevatorCall
+        version: 1
+    categories:
+      - name: Elevator
--- a/drivers/SmartThings/sinux-britzyhub/profiles/gas-valve.yml
+++ b/drivers/SmartThings/sinux-britzyhub/profiles/gas-valve.yml
@@ -1,0 +1,8 @@
+name: gas-valve
+components:
+  - id: main
+    capabilities:
+      - id: valve
+        version: 1
+    categories:
+      - name: GasValve
--- a/drivers/SmartThings/sinux-britzyhub/src/air-conditioner/init.lua
+++ b/drivers/SmartThings/sinux-britzyhub/src/air-conditioner/init.lua
@@ -1,0 +1,546 @@
+-- SinuxSoft (c) 2025
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local capabilities = require "st.capabilities"
+local clusters = require "st.matter.clusters"
+local im = require "st.matter.interaction_model"
+local utils = require "st.utils"
+local log = require "log"
+
+local version = require "version"
+
+local SAVED_SYSTEM_MODE_IB = "__saved_system_mode_ib"
+local DISALLOWED_THERMOSTAT_MODES = "__DISALLOWED_CONTROL_OPERATIONS"
+local OPTIONAL_THERMOSTAT_MODES_SEEN = "__OPTIONAL_THERMOSTAT_MODES_SEEN"
+
+local THERMOSTAT_MODE_MAP = {
+  [clusters.Thermostat.types.ThermostatSystemMode.OFF]               = capabilities.thermostatMode.thermostatMode.off,
+  [clusters.Thermostat.types.ThermostatSystemMode.AUTO]              = capabilities.thermostatMode.thermostatMode.auto,
+  [clusters.Thermostat.types.ThermostatSystemMode.COOL]              = capabilities.thermostatMode.thermostatMode.cool,
+  [clusters.Thermostat.types.ThermostatSystemMode.HEAT]              = capabilities.thermostatMode.thermostatMode.heat,
+  [clusters.Thermostat.types.ThermostatSystemMode.EMERGENCY_HEATING] = capabilities.thermostatMode.thermostatMode.emergency_heat,
+  [clusters.Thermostat.types.ThermostatSystemMode.PRECOOLING]        = capabilities.thermostatMode.thermostatMode.precooling,
+  [clusters.Thermostat.types.ThermostatSystemMode.FAN_ONLY]          = capabilities.thermostatMode.thermostatMode.fanonly,
+  [clusters.Thermostat.types.ThermostatSystemMode.DRY]               = capabilities.thermostatMode.thermostatMode.dryair,
+  [clusters.Thermostat.types.ThermostatSystemMode.SLEEP]             = capabilities.thermostatMode.thermostatMode.asleep,
+}
+
+local RAC_DEVICE_TYPE_ID = 0x0072 -- Room Air Conditioner
+local FAN_DEVICE_TYPE_ID = 0x002B
+
+local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
+
+local THERMOSTAT_MAX_TEMP_IN_C = 40.0
+local THERMOSTAT_MIN_TEMP_IN_C = 5.0
+
+local setpoint_limit_device_field = {
+  MIN_SETPOINT_DEADBAND_CHECKED = "MIN_SETPOINT_DEADBAND_CHECKED",
+  MIN_HEAT = "MIN_HEAT",
+  MAX_HEAT = "MAX_HEAT",
+  MIN_COOL = "MIN_COOL",
+  MAX_COOL = "MAX_COOL",
+  MIN_DEADBAND = "MIN_DEADBAND",
+  MIN_TEMP = "MIN_TEMP",
+  MAX_TEMP = "MAX_TEMP"
+}
+
+local profiling_data = {
+  THERMOSTAT_RUNNING_STATE_SUPPORT = "__THERMOSTAT_RUNNING_STATE_SUPPORT"
+}
+
+local subscribed_attributes = {
+  [capabilities.switch.ID] = {
+    clusters.OnOff.attributes.OnOff
+  },
+  [capabilities.temperatureMeasurement.ID] = {
+    clusters.TemperatureMeasurement.attributes.MeasuredValue,
+  },
+  [capabilities.thermostatMode.ID] = {
+    clusters.Thermostat.attributes.SystemMode,
+  },
+  [capabilities.thermostatCoolingSetpoint.ID] = {
+    clusters.Thermostat.attributes.OccupiedCoolingSetpoint,
+  },
+  [capabilities.airConditionerFanMode.ID] = {
+    clusters.FanControl.attributes.FanMode
+  },
+}
+
+local function tbl_contains(array, value)
+  for idx, element in ipairs(array) do
+    if element == value then
+      return true, idx
+    end
+  end
+  return false, nil
+end
+
+local function get_device_type(driver, device)
+  for _, ep in ipairs(device.endpoints) do
+    if ep.device_types ~= nil then
+      for _, dt in ipairs(ep.device_types) do
+        if dt.device_type_id == RAC_DEVICE_TYPE_ID then
+          return RAC_DEVICE_TYPE_ID
+        elseif dt.device_type_id == FAN_DEVICE_TYPE_ID then
+          return FAN_DEVICE_TYPE_ID
+        end
+      end
+    end
+  end
+  return false
+end
+
+local function create_fan_profile(device)
+  local fan_eps = device:get_endpoints(clusters.FanControl.ID)
+  local wind_eps = device:get_endpoints(clusters.FanControl.ID, {feature_bitmap = clusters.FanControl.types.FanControlFeature.WIND})
+  local rock_eps = device:get_endpoints(clusters.FanControl.ID, {feature_bitmap = clusters.FanControl.types.Feature.ROCKING})
+  local profile_name = ""
+  if #fan_eps > 0 then
+    profile_name = profile_name .. "-fan"
+  end
+  if #rock_eps > 0 then
+    profile_name = profile_name .. "-rock"
+  end
+  if #wind_eps > 0 then
+    profile_name = profile_name .. "-wind"
+  end
+  return profile_name
+end
+
+local function create_thermostat_modes_profile(device)
+  local heat_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.HEATING})
+  local cool_eps = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.COOLING})
+
+  local thermostat_modes = ""
+  if #heat_eps == 0 and #cool_eps == 0 then
+    return "No Heating nor Cooling Support"
+  elseif #heat_eps > 0 and #cool_eps == 0 then
+    thermostat_modes = thermostat_modes .. "-heating-only"
+  elseif #cool_eps > 0 and #heat_eps == 0 then
+    thermostat_modes = thermostat_modes .. "-cooling-only"
+  end
+  return thermostat_modes
+end
+
+local function profiling_data_still_required(device)
+    for _, field in pairs(profiling_data) do
+        if device:get_field(field) == nil then
+            return true
+        end
+    end
+    return false
+end
+
+local function match_profile(driver, device)
+  if profiling_data_still_required(device) then return end
+
+  local running_state_supported = device:get_field(profiling_data.THERMOSTAT_RUNNING_STATE_SUPPORT)
+
+  local thermostat_eps = device:get_endpoints(clusters.Thermostat.ID)
+  local humidity_eps = device:get_endpoints(clusters.RelativeHumidityMeasurement.ID)
+  local device_type = get_device_type(driver, device)
+  local profile_name
+  if device_type == RAC_DEVICE_TYPE_ID then
+    profile_name = "room-air-conditioner"
+
+    if #humidity_eps > 0 then
+      profile_name = profile_name .. "-humidity"
+    end
+
+    local fan_name = create_fan_profile(device)
+    fan_name = string.gsub(fan_name, "-rock", "")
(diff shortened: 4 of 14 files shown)


DongHoon-Ryu:
@KyuminAhn 
Hello,
Basically, the EdgeDrivers of individual Bridged Devices under the Bridge do not have to be under the physical Bridge path. This is because the relationship between them is logically established during onboarding.

And, there was no difference between the existing EdgeDriver and the EdgeDriver related to the dimming-light, elevator, gas-valve, etc. you made. If a Bridged Device can operate with the existing Matter Generic EdgeDriver, no separate EdgeDriver code is required. That is, if it can be operated as a Matter Generic EdgeDriver, the EdgeDriver suitable for the Bridged Device is automatically downloaded when the Bridge Device is onboarded without further development.

If you have your own Bridged Device function that cannot be supported by Matter Generic EdgeDriver, you'd better implement the EdgeDriver for the separated directory at this time. However, even in this case, it would be better to implement the function under the directory such as matter-switch or matter-sensor, matter-thermostat, ...  depending on the characteristics of the Bridged Device. This is because duplicate code can be minimized.

@hcarter-775 , @nickolas-deboom 
Could you check this PR?

drivers/SmartThings/sinux-britzyhub/profiles/gas-valve.yml
```diff
@@ -0,0 +1,8 @@
+name: gas-valve
+components:
+  - id: main
+    capabilities:
+      - id: valve
```
DongHoon-Ryu: please use `safetyValve`.

safetyValve capability has only `close` command.

drivers/SmartThings/sinux-britzyhub/src/gas-valve/init.lua
```diff
@@ -0,0 +1,82 @@
+-- SinuxSoft (c) 2025
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
+local clusters = require "st.matter.clusters"
+local log = require "log"
+
+local valve_cap = capabilities.valve
```
DongHoon-Ryu: local valve_cap = capabilities.safetyValve

drivers/SmartThings/sinux-britzyhub/profiles/gas-valve.yml
KyuminAhn: @DongHoon-Ryu 
we will change that, Thanks for your advice,

drivers/SmartThings/sinux-britzyhub/src/gas-valve/init.lua
KyuminAhn: @DongHoon-Ryu 
I will change that, Thanks for your advice,

drivers/SmartThings/sinux-britzyhub/src/air-conditioner/init.lua
```diff
@@ -0,0 +1,530 @@
+-- SinuxSoft (c) 2025
```
nickolas-deboom: I think it would be best to introduce a "is_air_conditioner" function in this subdriver (and similar for the other subdrivers) so that it is only loaded for AC devices. You can reference a similar function in the SmartThings Matter Appliance driver [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/441a709c23c28b135eaff83029f68bd868365a3a/drivers/SmartThings/matter-appliance/src/matter-dishwasher/init.lua#L39) for example. You could just check the device type and then return true if it matches AC.

KyuminAhn: I've applied the source code based on your advice. Thank you.

drivers/SmartThings/sinux-britzyhub/src/init.lua
```diff
@@ -0,0 +1,16 @@
+-- SinuxSoft (c) 2025
+-- Licensed under the Apache License, Version 2.0
+
+local MatterDriver = require "st.matter.driver"
+local log = require "log"
+
+local matter_driver = MatterDriver("britzyhub-matter", {
+  sub_drivers = {
+    require ("elevator"),
```
nickolas-deboom: I assume that 'air-conditioner' and 'dimming-light' are not included here because there would be a conflict with the generic fingerprints for those device types already belonging in other drivers? (matter-thermostat and matter-switch, respectively)

drivers/SmartThings/sinux-britzyhub/src/gas-valve/init.lua
```diff
@@ -0,0 +1,96 @@
...
+  matter_handlers = {
+    attr = {
+      [onoff_cluster.ID] = {
+        [onoff_cluster.attributes.OnOff.ID] = on_off_attr_handler,
+      }
+    }
+  },
+  capability_handlers = {
+    [valve_cap.ID] = {
+      --[valve_cap.commands.open.NAME] = function(driver, device, cmd)
```
nickolas-deboom: This may be commented out accidentally

drivers/SmartThings/sinux-britzyhub/src/air-conditioner/init.lua
```diff
@@ -0,0 +1,530 @@
...
+    elseif ib.data.value == clusters.FanControl.attributes.FanMode.LOW then
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.airConditionerFanMode.fanMode("low"))
+    elseif ib.data.value == clusters.FanControl.attributes.FanMode.MEDIUM then
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.airConditionerFanMode.fanMode("medium"))
+    elseif ib.data.value == clusters.FanControl.attributes.FanMode.HIGH then
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.airConditionerFanMode.fanMode("high"))
+    else
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.airConditionerFanMode.fanMode("auto"))
+    end
+  elseif device:supports_capability_by_id(capabilities.airPurifierFanMode.ID) then
```
nickolas-deboom: The elseif and else clauses can be removed since airConditionerFanMode should be used here.

drivers/SmartThings/sinux-britzyhub/src/air-conditioner/init.lua
```diff
@@ -0,0 +1,530 @@
...
+local function profiling_data_still_required(device)
+    for _, field in pairs(profiling_data) do
+        if device:get_field(field) == nil then
+            return true
+        end
+    end
+    return false
+end
+
+local function match_profile(driver, device)
```
nickolas-deboom: I think that all of the profiling logic should be removed from this sub-driver since the profile used by devices using this driver would be "air-conditioner.yml"

drivers/SmartThings/sinux-britzyhub/src/ventilator/init.lua
```diff
@@ -0,0 +1,213 @@
...
+      [capabilities.airPurifierFanMode.commands.setAirPurifierFanMode.NAME] = set_air_purifier_fan_mode,
+    },
+    [capabilities.fanSpeedPercent.ID] = {
+      [capabilities.fanSpeedPercent.commands.setPercent.NAME] = set_fan_speed_percent,
+    },
+  },
+  supported_capabilities = {
+    capabilities.switch,
+    capabilities.airPurifierFanMode,
+--     capabilities.fanSpeedPercent,
```
nickolas-deboom: I think these may be inadvertently commented out

drivers/SmartThings/sinux-britzyhub/src/ventilator/init.lua
```diff
@@ -0,0 +1,213 @@
...
+  [clusters.FanControl.attributes.FanMode.OFF] = capabilities.airPurifierFanMode.airPurifierFanMode.off(),
+  [clusters.FanControl.attributes.FanMode.LOW] = capabilities.airPurifierFanMode.airPurifierFanMode.low(),
+  [clusters.FanControl.attributes.FanMode.MEDIUM] = capabilities.airPurifierFanMode.airPurifierFanMode.medium(),
+  [clusters.FanControl.attributes.FanMode.HIGH] = capabilities.airPurifierFanMode.airPurifierFanMode.high(),
+  [clusters.FanControl.attributes.FanMode.AUTO] = capabilities.airPurifierFanMode.airPurifierFanMode.auto(),
+}
+
+local function fan_mode_handler(_, device, ib, _)
+  local cap = capabilities.airPurifierFanMode
+  if device:supports_capability_by_id(cap.ID) then
```
nickolas-deboom: This check shouldn't be needed since this capability is present in the ventilator profile

drivers/SmartThings/sinux-britzyhub/src/ventilator/init.lua
```diff
@@ -0,0 +1,213 @@
...
+local function handle_switch_off(_, device, cmd)
+  local ep = component_to_endpoint(device, cmd.component, clusters.OnOff.ID)
+  device:send(clusters.OnOff.server.commands.Off(device, ep))
+end
+
+local function set_air_purifier_fan_mode(_, device, cmd)
+  local args = cmd.args.airPurifierFanMode
+  local mode_map = {
+    low = clusters.FanControl.attributes.FanMode.LOW,
+    sleep = clusters.FanControl.attributes.FanMode.LOW,
```
nickolas-deboom: The enabled values in the profile are low, medium, high, auto, and off, so the other modes here aren't needed to be included

drivers/SmartThings/sinux-britzyhub/src/init.lua
KyuminAhn: I have modified the source code and pushed it to the repository as you advised. Thank you for your help.

drivers/SmartThings/sinux-britzyhub/src/air-conditioner/init.lua
KyuminAhn: I have modified the source code and pushed it to the repository as you advised. Thank you for your help.

drivers/SmartThings/sinux-britzyhub/src/air-conditioner/init.lua
KyuminAhn: I have modified the source code and pushed it to the repository as you advised. Thank you for your help.

drivers/SmartThings/sinux-britzyhub/src/ventilator/init.lua
KyuminAhn: I have modified the source code and pushed it to the repository as you advised. Thank you for your help.

drivers/SmartThings/sinux-britzyhub/src/ventilator/init.lua
KyuminAhn: I have modified the source code and pushed it to the repository as you advised. Thank you for your help.

drivers/SmartThings/sinux-britzyhub/src/ventilator/init.lua
KyuminAhn: I have modified the source code and pushed it to the repository as you advised. Thank you for your help.

drivers/SmartThings/sinux-britzyhub/fingerprints.yml
```diff
@@ -0,0 +1,26 @@
+matterGeneric:
+  - id: "air-conditioner"
+    deviceLabel: Air Conditioner
+    deviceTypes:
+      - id: 0x0072
```
nickolas-deboom: I think that manufacturer specific fingerprints should be used rather than generic because in general we would want AC's and thermostats to use the stock drivers.

drivers/SmartThings/sinux-britzyhub/profiles/vent.yml
```diff
@@ -0,0 +1,17 @@
+name: vent
+components:
+  - id: main
+    capabilities:
+      - id: switch
+        version: 1
+      - id: airPurifierFanMode
```
tpmanley: We have a more generic `fanMode`  capability which I think may be more appropriate than this one which is intended to be specifically for air purifiers. Does the vent actually have a fan on it?

drivers/SmartThings/sinux-britzyhub/fingerprints.yml
KyuminAhn: this AC's driver is for a Bridged Device operating under a Matter Bridge.

As I understand it, I have to use the matterGeneric profile in this situation. However, if there is a way to enable thermostatCooling(Heat)Setpoint - step = 1.0  adjustments for both Thermostat and Air Conditioner devices under the current setup, please let me know. 

I would be happy to update my implementation with that method.

drivers/SmartThings/sinux-britzyhub/profiles/vent.yml
KyuminAhn: I appreciate the advice and will take a look at what you've suggested.

this vent's driver is for a Bridged Device operating under a Matter Bridge.
all of the (Bridged) vent devices  actually have a fan on them

drivers/SmartThings/sinux-britzyhub/profiles/elevator.yml
```diff
@@ -0,0 +1,8 @@
+name: elevator
```
hcarter-775: this is generic to all the introduced profiles, I might suggest including:
```
  - id: firmwareUpdate
    version: 1
  - id: refresh
    version: 1
```
as capabilities. These are handled under the hood and give access to devices what the name suggests (updates and capability refreshes)

hcarter-775:
Hi @KyuminAhn 
Could we move this driver out from under the `*/drivers/SmartThings/` parent directory and into a new one at the same level called `*/drivers/SinuxSoft/`? Then perhaps the driver could be renamed `britzy` if so desired, or could stay the same.

drivers/SmartThings/sinux-britzyhub/fingerprints.yml
hcarter-775: Hi @KyuminAhn
Using a `matterGeneric` fingerprint in this case will cause problems for all thermostat/room air conditioner matter devices that are onboarded to our platform that should be using the stock matter thermostat driver (which also handles the Room Air Conditioner device type). Therefore, we can't have a generic fingerprint be used in a non-stock Matter driver.

Still, since the stock driver handling should be nearly 1-1 with your implementation defined here, this should not cause many issues. To the one difference you note here, that being the step size of 1.0 in your sub-driver versus 0.1 in the stock driver, we strongly believe that all devices of the same device type that are onboarded to our platform should have a consistent user experience, so we recommend letting this device take steps of 0.1.

However, if that's not satisfactory, I might suggest we use the general stock driver option for now (using 0.1 steps), and once you are able to update the britzy bridge firmware to allow for the manufacturer-specific vendor and product ids to be given for connected bridged devices, we can either transition back to using your driver for these devices using a `matterManufacturer`, or we could make changes to the stock driver to handle your particular use case.

drivers/SmartThings/sinux-britzyhub/profiles/elevator.yml
KyuminAhn: From what I understand, this edge driver is used for Bridged Device under Matter Bridge(Aggregator), firmwareUpdate isn't supported

drivers/SmartThings/sinux-britzyhub/fingerprints.yml
KyuminAhn: Hi @hcarter-775 
We've removed the thermostat/room air conditioner item from the Matter Generic Fingerprint based on the information you provided.
thanks

KyuminAhn:
> Hi @KyuminAhn Could we move this driver out from under the `*/drivers/SmartThings/` parent directory and into a new one at the same level called `*/drivers/SinuxSoft/`? Then perhaps the driver could be renamed `britzy` if so desired, or could stay the same.

we have moved all of files in previous dir to '*/drivers/SinuxSoft/britzyhub'
Thanks

drivers/SmartThings/sinux-britzyhub/profiles/elevator.yml
hcarter-775: Ah, that is a good point. Last point here, perhaps you'd want to add the `refresh` capability to the other new profiles here as well? For consistency as well as a possibly better user experience?

KyuminAhn: Sorry to make you confused, I've committed that added refresh capability 
Thanks

drivers/SinuxSoft/britzyhub/profiles/gas-valve.yml
```diff
@@ -0,0 +1,8 @@
+name: gas-valve
```
hcarter-775, line 1: @KyuminAhn sorry I think there may still be a little miscommunication on this last point, I'm suggesting that maybe we add `refresh` to this capability as well?

drivers/SinuxSoft/britzyhub/profiles/vent.yml
```diff
@@ -0,0 +1,17 @@
+name: vent
```
hcarter-775, line 1: and in this one, for consistency.

drivers/SinuxSoft/britzyhub/profiles/vent.yml
```diff
@@ -0,0 +1,17 @@
...
+components:
+  - id: main
+    capabilities:
+      - id: switch
+        version: 1
+      - id: airPurifierFanMode
```
hcarter-775: Also, I know this was mentioned previously, but I do think it may be wise to use `fanMode` in this case rather than `airPurifierFanMode`. It is a more generic option that has identical handling,

drivers/SinuxSoft/britzyhub/profiles/gas-valve.yml
KyuminAhn: Sorry, i missed it. I will do the others, too
Thanks for your help

drivers/SinuxSoft/britzyhub/src/vent/init.lua
```diff
@@ -0,0 +1,152 @@
...
+  device:send(clusters.OnOff.server.commands.On(device, ep))
+end
+
+local function handle_switch_off(driver, device, cmd)
+  local ep = component_to_endpoint(device, cmd.component, clusters.OnOff.ID)
+  device:send(clusters.OnOff.server.commands.Off(device, ep))
+end
+
+local function set_air_purifier_fan_mode(driver, device, cmd)
+  local args = cmd.args.airPurifierFanMode
```
hcarter-775:
```suggestion
  [clusters.FanControl.attributes.FanMode.LOW] = capabilities.fanMode.fanMode.low(),
  [clusters.FanControl.attributes.FanMode.MEDIUM] = capabilities.fanMode.fanMode.medium(),
  [clusters.FanControl.attributes.FanMode.HIGH] = capabilities.fanMode.fanMode.high(),
}

local function fan_mode_handler(driver, device, ib, response)
  local cap = capabilities.fanMode
  if device:supports_capability_by_id(cap.ID) then
    local event = map_enum_value(AC_FAN_MODE_MAP, ib.data.value, cap.fanMode.low())
    device:emit_event_for_endpoint(ib.endpoint_id, event)
  end
end

local function handle_switch_on(driver, device, cmd)
  local ep = component_to_endpoint(device, cmd.component, clusters.OnOff.ID)
  device:send(clusters.OnOff.server.commands.On(device, ep))
end

local function handle_switch_off(driver, device, cmd)
  local ep = component_to_endpoint(device, cmd.component, clusters.OnOff.ID)
  device:send(clusters.OnOff.server.commands.Off(device, ep))
end

local function set_air_purifier_fan_mode(driver, device, cmd)
  local args = cmd.args.fanMode
```

hcarter-775: in the case you choose to change the handling, it would be as simple as doing this ^ the handling is otherwise identical

drivers/SinuxSoft/britzyhub/src/vent/init.lua
```diff
@@ -0,0 +1,152 @@
...
+      },
+    }
+  },
+  capability_handlers = {
+    [capabilities.switch.ID] = {
+      [capabilities.switch.commands.on.NAME] = handle_switch_on,
+      [capabilities.switch.commands.off.NAME] = handle_switch_off,
+    },
+    [capabilities.airPurifierFanMode.ID] = {
+      [capabilities.airPurifierFanMode.commands.setAirPurifierFanMode.NAME] = set_air_purifier_fan_mode,
```
hcarter-775:
```suggestion
    [capabilities.fanMode.ID] = {
      [capabilities.fanMode.commands.setFanMode.NAME] = set_air_purifier_fan_mode,
```
This would be changed like so as well, and maybe the handler name would change. What do you think about this?

drivers/SinuxSoft/britzyhub/src/gas-valve/init.lua
```diff
@@ -0,0 +1,91 @@
...
+
+local function device_init(driver, device)
+  device:set_component_to_endpoint_fn(component_to_endpoint)
+  device:subscribe()
+end
+
+local function info_changed(driver, device, event, args)
+  device:add_subscribed_attribute(onoff_cluster.attributes.OnOff)
+  device:subscribe()
+end
```
hcarter-775, line 49: You should not require this handler. It is already taken care of in your `init` function and will cause more subscribe calls than needed

drivers/SinuxSoft/britzyhub/src/elevator/init.lua
```diff
@@ -0,0 +1,91 @@
...
+
+local function device_init(driver, device)
+  device:set_component_to_endpoint_fn(component_to_endpoint)
+  device:subscribe()
+end
+
+local function info_changed(driver, device, event, args)
+  device:add_subscribed_attribute(onoff_cluster.attributes.OnOff)
+  device:subscribe()
+end
```
hcarter-775, line 49: Repeat opinion: You should not require this handler. It is already taken care of in your init function and will cause more subscribe calls than needed

drivers/SinuxSoft/britzyhub/src/vent/init.lua
```diff
@@ -0,0 +1,152 @@
...
+
+local function info_changed(driver, device, event, args)
+  for _, attr in ipairs({
+    clusters.OnOff.attributes.OnOff,
+    clusters.FanControl.attributes.FanMode,
+  }) do
+    device:add_subscribed_attribute(attr)
+  end
+  device:subscribe()
+end
```
hcarter-775: Repeat opinion: You should not require this handler. It is already taken care of in your init function and will cause more subscribe calls than needed

drivers/SinuxSoft/britzyhub/profiles/vent.yml
KyuminAhn: ​I will make the changes as you've instructed, test the code on the actual device, and then commit it
Thanks

drivers/SinuxSoft/britzyhub/src/vent/init.lua
KyuminAhn: Yes, I will make the changes as you've instructed, test the code on the actual device, and then commit it
Thanks

drivers/SinuxSoft/britzyhub/profiles/vent.yml
hcarter-775: Thank you for the quick responses to my suggestions @KyuminAhn. I appreciate it.

drivers/SinuxSoft/britzyhub/src/vent/init.lua
KyuminAhn: we've changed our code and applied to devices, but does not work, 

and it looks like what we had before was right, there's same code below :
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/matter-pump/src/init.lua#L102-L114

hcarter-775: In the code you mention, there is a comment noting a change in the 0.48x lua libs release that would make the handling there obsolete, and it has now been a couple years since then. This handling should not be required, since the subscribe is taken care of in the init handler :+1:

hcarter-775:
All remaining comments are for minor improvements in capability choice and subscribe handling- these should not affect the basic function of the driver, not should it create any regressions.

tpmanley:
Thanks for making all the updates @KyuminAhn . It looks like the one remaining thing that we suggest gets addressed before merging is changing `airPurifierFanMode` -> `fanMode`. 

Also I just want to note to make sure that you squash all your changes before merging, either with the "Squash and merge" option in Github, or manually squashing and force pushing before doing the Merge.

KyuminAhn:
> Thanks for making all the updates @KyuminAhn . It looks like the one remaining thing that we suggest gets addressed before merging is changing `airPurifierFanMode` -> `fanMode`.
> 
> Also I just want to note to make sure that you squash all your changes before merging, either with the "Squash and merge" option in Github, or manually squashing and force pushing before doing the Merge.

I apologize for the delay in pushing to GitHub due to the late testing with the actual device. I've now switched to fanMode, which was the last issue, and I will finish up by using Squash and merge.

chulspro (approved):

PR #2191, opened 2025-06-13
Matter Sensor: Add motion-illuminance-temperature-humidity-battery profile and fingerprint

hcarter-775:
# Description of Change
Add new profile and corresponding fingerprint.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-sensor/profiles/motion-illuminance-temperature-humidity-battery.yml
+++ b/drivers/SmartThings/matter-sensor/profiles/motion-illuminance-temperature-humidity-battery.yml
@@ -1,0 +1,25 @@
+name: motion-illuminance-temperature-humidity-battery
+components:
+- id: main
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/matter-sensor/profiles/motion-illuminance-temperature-humidity-battery.yml
```diff
@@ -0,0 +1,25 @@
+name: motion-illuminance-temperature-humidity-battery
```
nickolas-deboom, line 1: Do we want to include versions with batteryLevel and no battery?

hcarter-775: No battery already exists, I can add batteryLevel though :+1:

PR #2194, opened 2025-06-16
WWSTCERT-6466 Shelly 1PM Gen4

greens:


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -720,6 +720,11 @@ matterManufacturer:
     vendorId: 0x1490
     productId: 0x1853
     deviceProfileName: plug-binary
+  - id: "5264/4137"
+    deviceLabel: Shelly 1PM Gen4
+    vendorId: 0x1490
+    productId: 0x1029
+    deviceProfileName: plug-power-energy-powerConsumption
```
nickolas-deboom, line 767: I think this is the best fitting profile we have currently, but I wonder if we should make an identical one that has Switch as the category, since after a quick search this device looks like more of a smart switch than a plug.

hcarter-775: How do you propose we separate these during the dynamic profiling step? Just directly profile it based on the vendor/product id in match_profile?

PR #2199, opened 2025-06-16
Matter Switch: Add greater Energy profiling logic for Switches

hcarter-775:
# Description of Change
Update the Matter Switch driver to support the Power Topology cluster. Specifically, include support for the Node and Set features of the cluster. Update unit tests to handle this update. 

# Summary of Completed Tests
Tests updated and continue to pass.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -50,7 +50,6 @@ local CURRENT_HUESAT_ATTR_MAX = 254
 -- table for devices that joined prior to this transition, and is also used for
 -- button devices that require component mapping.
 local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
-local ENERGY_MANAGEMENT_ENDPOINT = "__energy_management_endpoint"
 local IS_PARENT_CHILD_DEVICE = "__is_parent_child_device"
 local COLOR_TEMP_BOUND_RECEIVED_KELVIN = "__colorTemp_bound_received_kelvin"
 local COLOR_TEMP_BOUND_RECEIVED_MIRED = "__colorTemp_bound_received_mired"
@@ -203,6 +202,13 @@ local MINIMUM_ST_ENERGY_REPORT_INTERVAL = (15 * 60) -- 15 minutes, reported in s
 local SUBSCRIPTION_REPORT_OCCURRED = "__subscription_report_occurred"
 local CONVERSION_CONST_MILLIWATT_TO_WATT = 1000 -- A milliwatt is 1/1000th of a watt
 
+local ELECTRICAL_SENSOR_EPS = "__ELECTRICAL_SENSOR_EPS"
+local POWER_CONSUMPTION_REPORT_SUPPORTED = "__POWER_CONSUMPTION_REPORT_SUPPORTED"
+
+local profiling_data = {
+  ELECTRICAL_TOPOLOGY = "__ELECTRICAL_TOPOLOGY",
+}
+
 -- Return an ISO-8061 timestamp in UTC
 local function iso8061Timestamp(time)
   return os.date("!%Y-%m-%dT%H:%M:%SZ", time)
@@ -231,21 +237,13 @@ local function send_import_poll_report(device, latest_total_imported_energy_wh)
   end
 
   -- Report the energy consumed during the time interval. The unit of these values should be 'Wh'
-  if not device:get_field(ENERGY_MANAGEMENT_ENDPOINT) then
-    device:emit_event(capabilities.powerConsumptionReport.powerConsumption({
-      start = iso8061Timestamp(last_time),
-      ["end"] = iso8061Timestamp(current_time - 1),
-      deltaEnergy = energy_delta_wh,
-      energy = latest_total_imported_energy_wh
-    }))
-  else
-    device:emit_event_for_endpoint(device:get_field(ENERGY_MANAGEMENT_ENDPOINT),capabilities.powerConsumptionReport.powerConsumption({
-      start = iso8061Timestamp(last_time),
-      ["end"] = iso8061Timestamp(current_time - 1),
-      deltaEnergy = energy_delta_wh,
-      energy = latest_total_imported_energy_wh
-    }))
-  end
+  local component = device.profile.components["main"]
+  device:emit_component_event(component, capabilities.powerConsumptionReport.powerConsumption({
+    start = iso8061Timestamp(last_time),
+    ["end"] = iso8061Timestamp(current_time - 1),
+    deltaEnergy = energy_delta_wh,
+    energy = latest_total_imported_energy_wh
+  }))
 end
 
 local function create_poll_report_schedule(device)
@@ -275,13 +273,11 @@ local function set_poll_report_timer_and_schedule(device, is_cumulative_report)
     local second_timestamp = os.time()
     local report_interval_secs = second_timestamp - first_timestamp
     device:set_field(IMPORT_REPORT_TIMEOUT, math.max(report_interval_secs, MINIMUM_ST_ENERGY_REPORT_INTERVAL))
-    -- the poll schedule is only needed for devices that support powerConsumption
-    -- and enable powerConsumption when energy management is defined in root endpoint(0).
-    if device:supports_capability(capabilities.powerConsumptionReport) or
-       device:get_field(ENERGY_MANAGEMENT_ENDPOINT) then
+    -- the poll schedule is only needed for devices that support powerConsumptionReport
+    if device:get_field(POWER_CONSUMPTION_REPORT_SUPPORTED) then
       create_poll_report_schedule(device)
     end
-    device:set_field(IMPORT_POLL_TIMER_SETTING_ATTEMPTED, true)
+    device:set_field(IMPORT_POLL_TIMER_SETTING_ATTEMPTED, true, {persist = true})
   end
 end
 
@@ -319,6 +315,22 @@ local function create_multi_press_values_list(size, supportsHeld)
     table.insert(list, string.format("pushed_%dx", i))
   end
   return list
+end
+
+-- check if device type is found on the device. Optionally specify the endpoint to search on
+local function contains_device_type(device, device_type_id, endpoint_id)
+  for _, ep in ipairs(device.endpoints) do
+    for _, dt in ipairs(ep.device_types) do
+      if dt.device_type_id == device_type_id then
+        if endpoint_id and ep.endpoint_id == endpoint_id then
+          return true
+        elseif endpoint_id == nil then
+          return true
+        end
+      end
+    end
+  end
+  return false
 end
 
 local function tbl_contains(array, value)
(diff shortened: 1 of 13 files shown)


drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -274,7 +285,8 @@ local function set_poll_report_timer_and_schedule(device, is_cumulative_report)
     local report_interval_secs = second_timestamp - first_timestamp
     device:set_field(IMPORT_REPORT_TIMEOUT, math.max(report_interval_secs, MINIMUM_ST_ENERGY_REPORT_INTERVAL))
     -- the poll schedule is only needed for devices that support powerConsumptionReport
-    if device:get_field(POWER_CONSUMPTION_REPORT_SUPPORTED) then
```
nickolas-deboom: I'm a bit confused, this field doesn't appear in main, it currently has this:
```
    if device:supports_capability(capabilities.powerConsumptionReport) or
       device:get_field(ENERGY_MANAGEMENT_ENDPOINT) then
      create_poll_report_schedule(device)
    end
```

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -274,7 +285,8 @@ local function set_poll_report_timer_and_schedule(device, is_cumulative_report)
...
+    -- if device:get_field(POWER_CONSUMPTION_REPORT_SUPPORTED) then
```
nickolas-deboom: This comment can be removed

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Ignore my comment about this field, I was looking at a single commit by mistake :sweat_smile:

drivers/SmartThings/matter-switch/src/test/test_aqara_light_switch_h2.lua
```diff
@@ -177,6 +178,8 @@ local function test_init()
     end
   end
   test.socket.matter:__expect_send({aqara_mock_device.id, subscribe_request})
+  aqara_mock_children[aqara_child1_ep]:set_field("__POWER_CONSUMPTION_REPORT_SUPPORTED", true, {persist = true}) -- assume this was set in added
```
nickolas-deboom: This can be removed now, right? (and in other test files?)

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -275,13 +283,11 @@ local function set_poll_report_timer_and_schedule(device, is_cumulative_report)
...
-    -- the poll schedule is only needed for devices that support powerConsumption
-    -- and enable powerConsumption when energy management is defined in root endpoint(0).
-    if device:supports_capability(capabilities.powerConsumptionReport) or
-       device:get_field(ENERGY_MANAGEMENT_ENDPOINT) then
+    -- the poll schedule is only needed for devices that support powerConsumptionReport
+    if device:supports_capability(capabilities.powerConsumptionReport) then
       create_poll_report_schedule(device)
     end
-    device:set_field(IMPORT_POLL_TIMER_SETTING_ATTEMPTED, true)
+    device:set_field(IMPORT_POLL_TIMER_SETTING_ATTEMPTED, true, {persist = true})
```
nickolas-deboom: Is there a reason you made this field persisted? It seems like if the device or hub is offline for a while then it would be better to create a new poll report timer

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -480,43 +496,45 @@ local function check_field_name_updates(device)
...
+
+  if is_child_device then
+    -- Check if child device has an overridden child profile that differs from the child's generic device type profile
+    for _, vendor in pairs(child_device_profile_overrides_per_vendor_id) do
+      for _, fingerprint in ipairs(vendor) do
+        if device.manufacturer_info.product_id == fingerprint.product_id and profile == fingerprint.initial_profile then
+          return fingerprint.target_profile
+        end
       end
     end
```
nickolas-deboom:
```suggestion
  if is_child_device and child_device_profile_overrides_per_vendor_id[device.manufacturer_info.vendor_id] then
    -- Check if child device has an overridden child profile that differs from the child's generic device type profile
    for _, fingerprint in ipairs(child_device_profile_overrides_per_vendor_id[device.manufacturer_info.vendor_id]) do
      if device.manufacturer_info.product_id == fingerprint.product_id and profile == fingerprint.initial_profile then
        return fingerprint.target_profile
      end
    end
```
We could do something like this here to ensure the vendor id matches

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1125,26 +1185,57 @@ local function occupancy_attr_handler(driver, device, ib, response)
...
+      avail_eps_req_complete = false
+    end
+
+    if avail_eps_req_complete then
+      local electrical_tags = ""
+      if info.power then electrical_tags = electrical_tags .. "-power" end
+      if info.energy then electrical_tags = electrical_tags .. "-energy-powerConsumption" end
+      electrical_tags_per_ep[info.availableEndpoints[1].value] = electrical_tags -- set tags on first available endpoint
+    end
+  end
```
nickolas-deboom:
```suggestion
  if electrical_sensor_eps then
    for _, info in ipairs(electrical_sensor_eps) do
      if info.endpoint_id == ib.endpoint_id then
        info.availableEndpoints = ib.data.elements
      elseif info.availableEndpoints == false then -- an endpoint is found that hasn't been updated through this handler.
        avail_eps_req_complete = false
      end

      if avail_eps_req_complete then
        local electrical_tags = ""
        if info.power then electrical_tags = electrical_tags .. "-power" end
        if info.energy then electrical_tags = electrical_tags .. "-energy-powerConsumption" end
        electrical_tags_per_ep[info.availableEndpoints[1].value] = electrical_tags -- set tags on first available endpoint
      end
    end
  end
```
`ipairs` would throw an error if this is nil

drivers/SmartThings/matter-switch/src/test/test_aqara_light_switch_h2.lua
hcarter-775: definitely, good catch

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: nope, you're definitely right

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: good idea

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -1316,8 +1405,22 @@ local function device_added(driver, device)
   if device.network_type == device_lib.NETWORK_TYPE_CHILD then
     local req = clusters.OnOff.attributes.OnOff:read(device)
     device:send(req)
+  elseif device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    collect_and_store_electrical_sensor_info(driver, device)
+    local electrical_sensor_eps = device:get_field(ELECTRICAL_SENSOR_EPS)
```
nickolas-deboom, line 1403: Wouldn't this field still be nil at the time of `added`?

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: FWIW it never will be if this is run, still would be safer

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: gets set in collect_and_store_electrical_sensor_info

nickolas-deboom: Oh I see, I thought it was based on `available_endpoints_handler`, never mind!

nickolas-deboom:
These changes look great! I think it might be a good idea to test on some additional device types since some of the init code is being changed.

nickolas-deboom:
I'm trying to remember but can't quite recall, I think you mentioned that these changes would result in re-profiling for certain devices that would currently not be re-profiled and just rely on fingerprints, is that the case?

hcarter-775:
> I'm trying to remember but can't quite recall, I think you mentioned that these changes would result in re-profiling for certain devices that would currently not be re-profiled and just rely on fingerprints, is that the case?

@nickolas-deboom yes, that's true. See the `assign_switch_profile` call at the bottom of `match_profile`, that will be called for all switch devices that didn't get an early return due to previous specific handling.

nickolas-deboom:
Taking another look at this, "update powerConsumptionReport handling to only send reports to one single capability, even when there are multiple child devices getting energy reports." - wondering where this change is coming from? Was there a request or something?

drivers/SmartThings/matter-switch/src/test/test_aqara_climate_sensor_w100.lua
```diff
@@ -141,6 +141,7 @@ local function test_init()
   end
 
   test.socket.matter:__expect_send({aqara_mock_device.id, subscribe_request})
+  aqara_mock_device:set_field("__ELECTRICAL_TOPOLOGY", {topology = false, tags_on_ep = {}}, {persist = false}) -- since we're assuming this would have happened during device_added in this case.
```
nickolas-deboom, line 144: The Aqara subdriver overrides `device_added` so this field would not be set in the current implementation and so the subdriver may need to be updated to set it. The other subdrivers may also need to account for this

drivers/SmartThings/matter-switch/src/test/test_aqara_light_switch_h2.lua
```diff
@@ -277,7 +280,7 @@ test.register_coroutine_test(
         {
           -- don't use "aqara_mock_children[aqara_child1_ep].id,"
           -- because energy management is at the root endpoint.
```
nickolas-deboom, line 282: This comment can be removed now :sweat_smile:

nickolas-deboom: Actually, is this the correct behavior? If I remember right this device has it's energy management cluster on the root endpoint, so why is it now testing `aqara_child1_ep`?

hcarter-775: It is, this update actually makes the whole aqara child system work better now we use emit_component_event(), side stepping the original issues we faced having the energy management cluster on the root endpoint.

hcarter-775:
> Taking another look at this, "update powerConsumptionReport handling to only send reports to one single capability, even when there are multiple child devices getting energy reports." - wondering where this change is coming from? Was there a request or something?

ST Energy will only take one report per device at the moment, so there's no reason to be sending multiple reports. Before this PR, multiple reports just weren't supported at all, so this isn't really a "change".

drivers/SmartThings/matter-switch/src/test/test_aqara_climate_sensor_w100.lua
hcarter-775: this test doesn't handle a device that's using a subdriver.

nickolas-deboom: Oops, my bad, I saw Aqara and thought this was the Aqara cube test file

PR #2205, opened 2025-06-18
Matter Thermostat: Use fanMode capability

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change replaces the use of `thermostatFanMode` with `fanMode`, to better align with the fan modes and sequences defined by the Fan Control cluster. Note that existing static profiles will continue to utilize `thermostatFanMode`, so this change will only affect newly onboarded devices that use the `thermostat-modular` profile.

# Summary of Completed Tests

Unit tests updated and tested on device

### DIFF
--- a/drivers/SmartThings/matter-thermostat/profiles/thermostat-modular.yml
+++ b/drivers/SmartThings/matter-thermostat/profiles/thermostat-modular.yml
@@ -6,6 +6,9 @@ components:
         version: 1
       - id: thermostatMode
         version: 1
+      - id: fanMode
+        version: 1
+        optional: true
       - id: thermostatFanMode
         version: 1
         optional: true
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -189,9 +189,14 @@ local subscribed_attributes = {
     clusters.Thermostat.attributes.AbsMaxHeatSetpointLimit
   },
   [capabilities.airConditionerFanMode.ID] = {
+    clusters.FanControl.attributes.FanModeSequence,
     clusters.FanControl.attributes.FanMode
   },
   [capabilities.airPurifierFanMode.ID] = {
+    clusters.FanControl.attributes.FanModeSequence,
+    clusters.FanControl.attributes.FanMode
+  },
+  [capabilities.fanMode.ID] = {
     clusters.FanControl.attributes.FanModeSequence,
     clusters.FanControl.attributes.FanMode
   },
@@ -997,7 +1002,7 @@ local function match_modular_profile_thermostat(driver, device)
   local wind_eps = device:get_endpoints(clusters.FanControl.ID, {feature_bitmap = clusters.FanControl.types.FanControlFeature.WIND})
 
   if #fan_eps > 0 then
-    table.insert(main_component_capabilities, capabilities.thermostatFanMode.ID)
+    table.insert(main_component_capabilities, capabilities.fanMode.ID)
   end
   if #rock_eps > 0 then
     table.insert(main_component_capabilities, capabilities.fanOscillationMode.ID)
@@ -1449,7 +1454,7 @@ local function running_state_handler(driver, device, ib, response)
 end
 
 local function sequence_of_operation_handler(driver, device, ib, response)
-  -- The ControlSequenceofOperation attribute only directly specifies what can't be operated by the operating environment, not what can.
+  -- The ControlSequenceOfOperation attribute only directly specifies what can't be operated by the operating environment, not what can.
   -- However, we assert here that a Cooling enum value implies that SystemMode supports cooling, and the same for a Heating enum.
   -- We also assert that Off is supported, though per spec this is optional.
   if device:get_field(OPTIONAL_THERMOSTAT_MODES_SEEN) == nil then
@@ -1480,7 +1485,7 @@ local function sequence_of_operation_handler(driver, device, ib, response)
     table.insert(modes_for_inclusion, capabilities.thermostatMode.thermostatMode.heat.NAME)
   end
 
-  -- check whether the Auto Mode should be supported in SystemMode, though this is unrelated to ControlSequenceofOperation
+  -- check whether the Auto Mode should be supported in SystemMode, though this is unrelated to ControlSequenceOfOperation
   local auto = device:get_endpoints(clusters.Thermostat.ID, {feature_bitmap = clusters.Thermostat.types.ThermostatFeature.AUTOMODE})
   if #auto > 0 then
     table.insert(modes_for_inclusion, capabilities.thermostatMode.thermostatMode.auto.NAME)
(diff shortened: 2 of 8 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1518,150 +1518,71 @@ local function min_deadband_limit_handler(driver, device, ib, response)
...
+    fan_mode_attr = fan_mode_cap.airPurifierFanMode
+    supported_fan_modes = fan_mode_cap.supportedAirPurifierFanModes
+  elseif device:supports_capability_by_id(capabilities.airConditionerFanMode.ID) then
+    fan_mode_cap = capabilities.airConditionerFanMode
+    fan_mode_attr = fan_mode_cap.airConditionerFanMode
+    supported_fan_modes = fan_mode_cap.supportedAcFanModes
+    is_air_conditioner = true
+  end
+  local fan_mode = {
+    ["off"] = is_air_conditioner and "off" or fan_mode_attr.off.NAME,
```
hcarter-775: pretty sure these both map to the same thing. I understand why the `or` was introduced, but `fan_mode_attr.off.NAME` and the others just map to a string, so I'd feel comfortable cutting the `is_air_conditioner` check altogether. You could leave a comment if you wanted to clarify why this is allowed, but for example see how it's done for ConcentrationHealthConcern capability attributes.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1919,46 +1840,36 @@ local cooling_setpoint_limit_handler_factory = function(minOrMax)
   end
 end
 
-local function set_thermostat_fan_mode(driver, device, cmd)
-  local fan_mode_id = nil
-  if cmd.args.mode == capabilities.thermostatFanMode.thermostatFanMode.auto.NAME then
+local function set_fan_mode(driver, device, cmd)
+  local fan_mode_id = clusters.FanControl.attributes.FanMode.OFF
```
hcarter-775: I think the earlier handling is a little more error-safe, since it checks that an actual value was hit before sending a write. I'm not sure how this handler would be hit if none of these were in the cmd parameter, but why remove the safety net

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -2303,23 +2210,21 @@ local matter_driver_template = {
...
     },
     [capabilities.airConditionerFanMode.ID] = {
-      [capabilities.airConditionerFanMode.commands.setFanMode.NAME] = set_fan_mode,
+      [capabilities.airConditionerFanMode.commands.setFanMode.NAME] = set_air_conditioner_fan_mode
     },
     [capabilities.airPurifierFanMode.ID] = {
       [capabilities.airPurifierFanMode.commands.setAirPurifierFanMode.NAME] = set_air_purifier_fan_mode
     },
+    [capabilities.fanMode.ID] = {
+      [capabilities.fanMode.commands.setFanMode.NAME] = set_fan_mode
```
hcarter-775: also, looking at this now, `set_air_conditioner_fan_mode`, `set_fan_mode`, and `set_air_purifier_fan_mode` are identical except for which capability they're sent to. Could be a good time to merge these with a flag parameter to specify which fan mode we're sending to.

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: The prior handling had an else clause so I think this is equivalent

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Agreed!

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: Ah, I see. I still think we may want to introduce this logic that was apparently not done correctly in the previous logic, since I do think having a safety net like that isn't a bad idea.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1518,150 +1518,71 @@ local function min_deadband_limit_handler(driver, device, ib, response)
...
-    elseif ib.data.value == clusters.FanControl.attributes.FanMode.LOW then
-      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.airPurifierFanMode.airPurifierFanMode.low())
-    elseif ib.data.value == clusters.FanControl.attributes.FanMode.MEDIUM then
-      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.airPurifierFanMode.airPurifierFanMode.medium())
-    elseif ib.data.value == clusters.FanControl.attributes.FanMode.HIGH then
-      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.airPurifierFanMode.airPurifierFanMode.high())
-    else
-      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.airPurifierFanMode.airPurifierFanMode.auto())
-    end
+  local fan_mode_cap = capabilities.fanMode
```
hcarter-775, line 1521: shouldn't support be added for `fanMode` as well here?

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Ah nice! I made this change in the latest commit

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Sure thing, I added an else clause to log a warning and return without sending a value

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: I made this change in the latest commit

Huangxiangjie:
Hi @nickolas-deboom. 
We are also looking at fan level in thermostat. We tested this PR with real device and the fan mode capability works fine.

drivers/SmartThings/matter-thermostat/profiles/thermostat-modular.yml
```diff
@@ -6,7 +6,7 @@ components:
         version: 1
       - id: thermostatMode
         version: 1
-      - id: thermostatFanMode
+      - id: fanMode
```
ctowns, line 9: Just calling out that removing `thermostatFanMode` is a destructive change that could result in automation/routines that users may have set up using `thermostatFanMode` to no longer work.

One workaround for this is to just keep `thermostatFanMode` here and then add `fanMode`, and make it so that future devices use `fanMode` when onboarding to this profile, but existing ones can still use `thermostatFanMode`. I could be convinced that it would be better to just remove `thermostatFanMode` though, let's discuss offline.

nickolas-deboom: Good call, I re-added this capability in the latest commit.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1520,150 +1525,81 @@ local function min_deadband_limit_handler(driver, device, ib, response)
...
-    elseif ib.data.value == clusters.FanControl.attributes.FanModeSequence.OFF_ON_AUTO then
-      supportedAcFanModes = {
-        "off",
-        "high",
-        "auto"
-      }
+  local supported_fan_modes_attribute, supportedFanModes
+  if device:supports_capability_by_id(capabilities.thermostatFanMode.ID) then
+    supported_fan_modes_attribute = capabilities.thermostatFanMode.supportedThermostatFanModes
+    -- Our thermostat fan mode control is not granular enough to handle all of the supported modes
```
Huangxiangjie, line 1590: For `thermostatFanMode` capability, the `supportedFanModes` is `{auto, on}` or `{on}`. We are confused as to why it should be handled in this way instead of based on the actual situation reported by the device. One partner is also asking us whether the fan control mode can be added.

nickolas-deboom: We opted to keep `thermostatFanMode` in existing static profiles so as to not break any existing automations/routines, and this capability only supports these fan modes. Note that with the changes in this PR, the `fanMode` capability will be used in place of `thermostatFanMode` in the `thermostat-modular` profile, meaning that going forward the rest of the fan modes reported by the device will be supported.

Cornelius117: @nickolas-deboom Ok, we have understood the purpose and function of this PR. Thank you for your patient explanation.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1520,150 +1525,81 @@ local function min_deadband_limit_handler(driver, device, ib, response)
...
+    [clusters.FanControl.attributes.FanMode.SMART]  = { capabilities.fanMode.fanMode.auto(),
+                                                        capabilities.airConditionerFanMode.fanMode("auto"),
+                                                        capabilities.airPurifierFanMode.airPurifierFanMode.auto(),
+                                                        capabilities.thermostatFanMode.thermostatFanMode.auto() }
+  }
+  local fan_mode_idx = device:supports_capability_by_id(capabilities.fanMode.ID) and 1 or
+    device:supports_capability_by_id(capabilities.airConditionerFanMode.ID) and 2 or
+    device:supports_capability_by_id(capabilities.airPurifierFanMode.ID) and 3 or
+    device:supports_capability_by_id(capabilities.thermostatFanMode.ID) and 4
+  if fan_mode_idx and fan_mode_idx >= 1 and fan_mode_idx <= 4 and fan_mode_event[ib.data.value][fan_mode_idx] then
```
hcarter-775:
```suggestion
  if fan_mode_idx ~= false then
```
I think the rest of this is superfluous based on the handling- since the rest of these will _always_ be true if this ^ is true.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1520,150 +1525,81 @@ local function min_deadband_limit_handler(driver, device, ib, response)
...
-      }
-    end
-    local event = capabilities.airConditionerFanMode.supportedAcFanModes(supportedAcFanModes, {visibility = {displayed = false}})
-    device:emit_event_for_endpoint(ib.endpoint_id, event)
-  elseif device:supports_capability_by_id(capabilities.airPurifierFanMode.ID) then
-    -- Air Purifier
-    local supportedAirPurifierFanModes
+      supportedFanModes = { "on" }
+    end
+  else
```
hcarter-775, line 1597: I see what you're doing by breaking the `thermostatFanMode` logic apart from the other logic, but I think this may overcomplicate the function more than help it. In my opinion, setting the supportedFanModes first in its own logic block, followed by a series of if/else statements per capability reads more easily.

In the handler:
```
if ib.data.value == clusters.FanControl.attributes.FanModeSequence.OFF_LOW_MED_HIGH then
      supportedFanModes = { "off", "low", "medium", "high" }
...
end

if device:supports_capability_by_id(capabilities.thermostatFanMode.ID) then
... special handling that ignores the previously set supportedFanModes ...
elif device:supports_capability_by_id(capabilities.fanMode.ID) then
...
end

local event = supported_fan_modes_attribute(supportedFanModes, {visibility = {displayed = false}})
```

hcarter-775 (approved):

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Ah yeah good call!

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Sure thing, I implemented this pretty much as you described in the latest commit (although I subsequently already squashed commits) - let me know if that looks good!

hcarter-775: looks great, thanks :+1:

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: I realized that the `fan_mode_event[ib.data.value][fan_mode_idx]` check is actually needed since there is a `nil` value in the table. I opened https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2245 to address

PR #2208, opened 2025-06-23
Added support for frient switch devices

marcintyminski:
Check all that apply

# Type of Change

- [ x] WWST Certification Request
          - [x ] I have reviewed the README.md file
          - [x ] I have reviewed the CODE_OF_CONDUCT.md file
          - [x] I have signed the CLA
     - [x ] I plan on entering a WWST Certification Request or have entered a request through the WWST Certification console at developer.smartthings.com
- [ ] Bug fix
- [x ] New feature
- [ ] Refactor

# Checklist

- [x ] I have performed a self-review of my code
- [x ] I have commented my code in hard-to-understand areas
- [x ] I have verified my changes by testing with a device or have communicated a plan for testing
- [x ] I am adding new behavior, such as adding a sub-driver, and have added and run new unit tests to cover the new behavior

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/configurations.lua
+++ b/drivers/SmartThings/zigbee-switch/src/configurations.lua
@@ -19,6 +19,9 @@ local device_def = require "st.device"
 
 local ColorControl = clusters.ColorControl
 local IASZone = clusters.IASZone
+local ElectricalMeasurement = clusters.ElectricalMeasurement
+local SimpleMetering = clusters.SimpleMetering
+local Alarms = clusters.Alarms
 local Status = require "st.zigbee.generated.types.ZclStatus"
 
 local CONFIGURATION_VERSION_KEY = "_configuration_version"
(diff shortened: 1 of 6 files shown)


drivers/SmartThings/zigbee-switch/src/configurations.lua
```diff
@@ -63,7 +66,69 @@ local devices = {
...
+        data_type = SimpleMetering.attributes.CurrentSummationDelivered.base_type,
+        reportable_change = 1
+      },{
+        cluster = Alarms.ID,
+        attribute = Alarms.attributes.AlarmCount.ID,
+        minimum_interval = 1,
+        maximum_interval = 600,
+        data_type = Alarms.attributes.AlarmCount.base_type,
+        reportable_change = 1,
+
```
greens: nit: extra whitespace

drivers/SmartThings/zigbee-switch/src/frient/init.lua
greens, line 1: we generally use 2-space tab widths

drivers/SmartThings/zigbee-switch/src/frient/init.lua
```diff
@@ -0,0 +1,209 @@
...
+    { mfr = "frient A/S", model = "SMRZB-332" },
+    { mfr = "frient A/S", model = "SMRZB-342" },
+}
+
+local POWER_FAILURE_ALARM_CODE = 0x03
+
+
+local function alarm_report_handler(driver, device, zb_rx)
+    local alarm_status = zb_rx.body.zcl_body
+    log.error("Received alarm report:"..util.stringify_table(zb_rx.body.zcl_body, nil, true))
```
greens: This seems like a debug statement, and I expect this message is already logged by the driver framework.

drivers/SmartThings/zigbee-switch/src/frient/init.lua
```diff
@@ -0,0 +1,209 @@
...
+    if ((alarm_status.cluster_identifier.value == SimpleMetering.ID) and (alarm_status.alarm_code.value == POWER_FAILURE_ALARM_CODE)) then
+        device.thread:call_with_delay(2, function(d)
+            device:emit_event_for_endpoint(zb_rx.address_header.src_endpoint.value, capabilities.powerSource.powerSource.unknown())
+        end
+        )
+    end
+end
+
+local function device_added(driver, device)
+end
```
greens, line 59: I don't see anything in the base driver's added method that you would need to override.

drivers/SmartThings/zigbee-switch/src/frient/init.lua
```diff
@@ -0,0 +1,209 @@
...
+
+-- Handler for voltage measurement
+local function voltage_measurement_handler(driver, device, value, zb_rx)
+    local raw_value  = value.value
+    -- By default emit raw value
+    local multiplier = device:get_field(VOLTAGE_MEASUREMENT_MULTIPLIER_KEY) or 1
+    local divisor    = device:get_field(VOLTAGE_MEASUREMENT_DIVISOR_KEY) or 1
+
+    if divisor == 0 then
```
greens: If this is a concern, I would put the check where you write the value to `VOLTAGE_MEASUREMENT_DIVISOR_KEY`.

drivers/SmartThings/zigbee-switch/src/frient/init.lua
```diff
@@ -0,0 +1,209 @@
...
+
+local function voltage_measurement_divisor_handler(driver, device, divisor, zb_rx)
+    log.info("V divisor value = ", divisor.value)
+
+    local raw_value = divisor.value
+    device:set_field(VOLTAGE_MEASUREMENT_DIVISOR_KEY, raw_value, { persist = true })
+end
+
+local function voltage_measurement_multiplier_handler(driver, device, multiplier, zb_rx)
+    log.info("Volt Multiplayer value = ")
```
greens: several log statements here that ought to be removed. This one doesn't actually log any values.

drivers/SmartThings/zigbee-switch/src/configurations.lua
```diff
@@ -63,7 +66,69 @@ local devices = {
...
+        attribute = ElectricalMeasurement.attributes.RMSCurrent.ID,
+        minimum_interval = 5,
+        maximum_interval = 600,
+        data_type = ElectricalMeasurement.attributes.RMSCurrent.base_type,
+        reportable_change = 1
+      },{
+        cluster = ElectricalMeasurement.ID,
+        attribute = ElectricalMeasurement.attributes.ActivePower.ID,
+        minimum_interval = 5,
+        maximum_interval = 600,
```
greens: Our default max here is one hour, rather than 5 minutes. Is there a particular reason you've chosen a shorter max interval? Because of the reportable change, any updates to the value itself should still come in (provided they're not more frequent than the minimum interval), so we're wondering if there's a specific reason you want to see the current value reported every five minutes.

If not, we'd appreciate you sticking with our default configurations.

greens (approved):

drivers/SmartThings/zigbee-switch/src/frient/init.lua
JanJakubiszyn: Hi @greens, this override has been added specifically for SPLZB-132 frient switch model. It has an endpoint which is unused and results in creating a non-functional child device because of the main driver. I think we should bring this function back.

greens: @JanJakubiszyn based on my reading of the code, that should only create extra child devices if the profile has more than one component, though. It doesn't seem like any of the fingerprints here do.

PR #2215, opened 2025-06-24
CHAD-12900 Lazy load for z-wave sensor

greens:


drivers/SmartThings/zwave-sensor/src/aeotec-water-sensor/init.lua
```diff
@@ -26,14 +26,11 @@ local ZWAVE_WATER_TEMP_HUMIDITY_FINGERPRINTS = {
 }
 
 --- Determine whether the passed device is zwave water temperature humidiry sensor
----
---- @param driver Driver driver instance
---- @param device Device device isntance
```
wkhenon, line 31: Do we want to keep these around?

wkhenon (approved):
This looks good! 👍

ctowns (approved):

PR #2216, opened 2025-06-24
Sort EPs before get device type and add unit test

Huangxiangjie:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

- [x I have verified my changes by testing with a device or have communicated a plan for testing

# Description of Change
 We find `device.endpoints` sometimes is disordered, like “ep0, ep5, ep1, ep3, ep4, ep2“. So we plan to sort device endpoints in ASC before judging the device type of equipment,  

# Summary of Completed Tests
 We added Unit Test and tested with real device.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -534,7 +534,14 @@ local function get_endpoints_for_dt(device, device_type)
 end
 
 local function get_device_type(device)
+  local device_types = {}
   for _, ep in ipairs(device.endpoints) do
+    if ep.device_types ~= nil then
+      table.insert(device_types, { endpoint_id = ep.endpoint_id, device_types = ep.device_types })
+    end
+  end
+  table.sort(device_types, function(a, b) return a.endpoint_id < b.endpoint_id end)
+  for _, ep in ipairs(device_types) do
     if ep.device_types ~= nil then
       for _, dt in ipairs(ep.device_types) do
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -534,7 +534,14 @@ local function get_endpoints_for_dt(device, device_type)
 end
 
 local function get_device_type(device)
```
hcarter-775, line 536: I think a slightly more generic version of this function could be:

```
local function get_device_type(device)
  local device_type_found = {}
    if ep.device_types ~= nil then
        device_type_found[dt.device_type_id] = true
      end
    end
  end

  for _, id in ipairs({
    RAC_DEVICE_TYPE_ID,
    AP_DEVICE_TYPE_ID,
    THERMOSTAT_DEVICE_TYPE_ID,
    FAN_DEVICE_TYPE_ID,
    WATER_HEATER_DEVICE_TYPE_ID,
    HEAT_PUMP_DEVICE_TYPE_ID
  }) do if device_type_found[id] then return id end end

  return false
end
```
This ensures that we're not just basing the logic on which device type we see first, but rather the order of "precedence" that we give the device types in the drivers. I tested this and confirmed it works with your test case. Thanks

Huangxiangjie: We also think this is a more generic version and updated PR.
We tested with real device, it works fine, thank you.

PR #2218, opened 2025-06-25
Mitigate deserialization errors due to hub 0.57 FW bug

cjswedes:
0.57 FW has an extra byte as the first byte of zdo message body payloads and is missing the last byte of the body of the payload. This adds a mitigation in the ZdoMessageBody deserialization code which overrides the Lua library function to correct the off by 1 error and inserts a default last byte of 0x01 for every Zdo message payload.

# Testing
- [x] tested with an ikea tradfri 5 button remote
- [x] QA tests with other devices.

### DIFF
--- a/drivers/SmartThings/zigbee-button/src/st/zigbee/zdo/init.lua
+++ b/drivers/SmartThings/zigbee-button/src/st/zigbee/zdo/init.lua
@@ -1,0 +1,155 @@
+-- Copyright 2021 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zigbee-button/src/st/zigbee/zdo/init.lua
```diff
@@ -0,0 +1,155 @@
...
+    local s = {}
+    s.zdo_header = zdo_messages.ZdoHeader.deserialize(buf)
+    -- there is a bug in hub FW that causes the zdo message payload to have an invalid first byte
+    -- and a missing last byte. Default to adding in a 1, which is a good default for the mgmt_bind_response
+    -- binding table entry endpoint_id
+    local version = require "version"
+    if version.rpc == 8 then
+      buf.buf = buf.buf .. "\01"
+      buf:seek(1)
+    end
```
cjswedes, line 106: Here is where the mitigation occurs

cjswedes:
One thing I realized this doesnt address is that any device that was joined while this is broken will remain non functional until it requests the binding table (is reconfigured). I think the approach that I will take is to do a one time reconfiguration of all devices in these subdrivers by setting the provisioning state to TYPED.

varzac:
> One thing I realized this doesnt address is that any device that was joined while this is broken will remain non functional until it requests the binding table (is reconfigured). I think the approach that I will take is to do a one time reconfiguration of all devices in these subdrivers by setting the provisioning state to TYPED.

I would be concerned that the reconfiguration could cause more problems that it solves, and I would just suggest that we ask people to re-onboard their devices if the functionality is affected.

PR #2225, opened 2025-06-30
fix(sonos): tighten up input validation for Sonos State queries.

dljsjr:
Check all that apply

# Type of Change

- [x] Bug fix

# Checklist

# Description of Change

There were a few places missing proper input validation that were leading to nil indexing crashes when using chained lookups.

Out of an abundance of caution, we have tightened up input validation on all paths, even those that aren't part of the identified chained queries causing issues in some situations.

# Summary of Completed Tests

Tested locally by synthetically injecting the error case.

### DIFF
--- a/drivers/SmartThings/sonos/src/sonos_state.lua
+++ b/drivers/SmartThings/sonos/src/sonos_state.lua
@@ -234,6 +234,10 @@ end
 ---@param player_id PlayerId
 ---@return string? device_id
 function SonosState:get_device_id_for_player(household_id, player_id)
+  if type(player_id) ~= "string" then
+    log.error(string.format("invalid player id provided: %s", player_id))
+    return nil
+  end
   local household = _STATE.households[household_id]
   if not household then
     log.error(string.format("No record of Sonos household with id %s", household_id))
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/sonos/src/sonos_state.lua
```diff
@@ -368,7 +388,14 @@ function SonosState:get_coordinator_for_group(household_id, group_id)
         false
       )
   end
-  return household.groups[group_id].coordinatorId
+
+  local group = (household.groups or {})[group_id]
+  if type(group) ~= "table" then
+    log.error(string.format("No known group for id %s in household %s", group_id, household_id))
+    return
```
cjswedes, line 395: Do you want these error logs to go into hub logs? We dont expect this to happen, but if it is happening and we are investigating a hub, it could be a red flag to look into.

cjswedes (approved):

drivers/SmartThings/sonos/src/sonos_state.lua
dljsjr: My thoughts are that this is happening with offline portable devices. So I don't know that we need to clutter hub logs to investigate these.

greens (approved):

PR #2229, opened 2025-06-30
WWSTCERT-6685 Yale Lock with Matter

greens:


drivers/SmartThings/matter-lock/fingerprints.yml
```diff
@@ -94,6 +94,12 @@ matterManufacturer:
     vendorId: 0x147F
     productId: 0x0001
     deviceProfileName: lock-user-pin-battery
+  #Yale
+  - id: "4125/33040"
+    deviceLabel: Yale Lock with Matter
+    vendorId: 0x101D
```
nickolas-deboom, line 110: Does Yale have two different Vendor ID's? I also see it listed in the DCL as 0x144f whereas 0x101d appears to be 'Fortune Brands'

tpmanley (approved):

hcarter-775 (approved):

drivers/SmartThings/matter-lock/fingerprints.yml
tpmanley: Correct, the Yale brand is used by two different companies now, and this PR is about the one used by Fortune Brands

nickolas-deboom (approved):

PR #2231, opened 2025-07-01
Zigbee: opt devices out of scripting-engine health check

greens:


drivers/SmartThings/zigbee-fan/src/test/test_fan_light.lua
```diff
@@ -40,7 +40,7 @@ zigbee_test_utils.prepare_zigbee_env_info()
 
 local function test_init()
     test.mock_device.add_test_device(mock_base_device)
-    zigbee_test_utils.init_noop_health_check_timer()
+    -- zigbee_test_utils.init_noop_health_check_timer()
```
cjswedes: I think we should just remove these test changes rather than comment them out.

drivers/SmartThings/zigbee-sensor/src/contact/init.lua
cjswedes, line 1: The opt out is not per subdriver, it can only be done on the base level driver.

cjswedes (approved):

PR #2235, opened 2025-07-03
Aqara-FP2: Add error handling and modify offline logic

hyundolee27:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

- Added error handling for potential errors that may occur when adding a device.
- Modified the offline detection logic to align with Aqara's logic.
- Reset the device exclusion list on every discovery

# Summary of Completed Tests

Verified operation with 3 FP2 devices connected at the same time to:
- SmartThings Hub V2
- SmartThings Hub V3
- SmartThings Station
- SmartMonitor M5's built-in hub

### DIFF
--- a/drivers/Aqara/aqara-presence-sensor/src/discovery.lua
+++ b/drivers/Aqara/aqara-presence-sensor/src/discovery.lua
@@ -1,10 +1,11 @@
 local log = require "log"
-local discovery = {}
 local fields = require "fields"
 local discovery_mdns = require "discovery_mdns"
 local socket = require "cosock.socket"
 
-local processing_devices = {}
+local discovery = {
+  last_try_time = {}
+}
 
 function discovery.set_device_field(driver, device)
   local device_cache_value = driver.datastore.discovery_cache[device.device_network_id]
@@ -36,6 +37,10 @@ local function try_add_device(driver, device_dni, device_ip)
 
   update_device_discovery_cache(driver, device_dni, device_ip, device_info)
   local create_device_msg = driver.discovery_helper.get_device_create_msg(driver, device_dni, device_ip, device_info)
+  if not create_device_msg then
+    log.error_with({ hub_logs = true }, string.format("Failed to get device info. dni= %s, ip= %s", device_dni, device_ip))
+    return "device info not found"
+  end
 
   local credential = driver.discovery_helper.get_credential(driver, device_dni, device_ip)
 
@@ -52,15 +57,20 @@ local function try_add_device(driver, device_dni, device_ip)
   end
 
   log.info_with({ hub_logs = true }, string.format("try_create_device. dni= %s, ip= %s", device_dni, device_ip))
-  processing_devices[device_dni] = true
-  driver:try_create_device(create_device_msg)
+
+  local success, ret = pcall(driver.try_create_device, driver, create_device_msg)
+  if success then
+    discovery.last_try_time[device_dni] = os.time()
+  else
+    log.error_with({ hub_logs = true }, string.format("Failed to try_create_device. dni= %s, %s", device_dni, ret))
+  end
   return nil
 end
 
 function discovery.device_added(driver, device)
   log.info_with({ hub_logs = true }, string.format("device_added. dni= %s", device.device_network_id))
   discovery.set_device_field(driver, device)
-  processing_devices[device.device_network_id] = nil
+  discovery.last_try_time[device.device_network_id] = nil
   driver.lifecycle_handlers.init(driver, device)
 end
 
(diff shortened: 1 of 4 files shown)


drivers/Aqara/aqara-presence-sensor/src/discovery.lua
```diff
@@ -108,6 +112,7 @@ end
 
 function discovery.do_network_discovery(driver, _, should_continue)
   log.info_with({ hub_logs = true }, string.format("discovery start for Aqara FP2"))
+  processing_devices = {}
```
tpmanley: Can this be removed or was it intended to be used?

hyundolee27: Thank you for your feedback. I have removed this variable. Please review it again.

NoahCornell: It looks like this was being used and was just a file scoped variable which I think might have caused some confusion. It looks like this was guarding against multiple try add device attempts so I am unsure if this is actually okay to remove or not.

drivers/Aqara/aqara-presence-sensor/src/discovery.lua
```diff
@@ -100,7 +100,7 @@ local function discovery_device(driver)
         break
       end
     end
-    if (not processing_devices[dni]) and (not is_already_added) then
```
cjswedes, line 103: It seems like this will cause a lot of extra device added messages from the driver to the cloud if we have already started adding the device to the platform, but the hub hasn't gotten the device record from the platform yet. Why is the `processing_devices` not needed anymore? It seems like its purpose is to prevent sending multiply `try_create_device` requests, which seems like something that we want to do.

hyundolee27:
@cjswedes  @NoahCornell 
Thank you for your review.
As you understand, `processing_devices` is used not to try to register again until `device_added` is called after `try_create_device` is called.

And when the connection was not good in the market, there was an issue that it did not appear to be `device_added` after `try_create_device`. (Unfortunately, we do not know the exact cause with the remaining logs. )
The existing logic initializes `processing_devices` only when `device_added`, so if the device is not added, `processing_device` remains and cannot be added again.

The first patch was simply a modification to initialize `processing_devices` for each discovery.
With this modification, if the device registration fails for some reason, I expected that the problem would be solved by canceling the discovery and calling it again.
In this case, once a problem occurs within a certain discovery, it cannot be solved within the same discovery, so it has been modified not to use `processing_devices`.

However, there is a side effect concern as you mentioned, so I modified `processing_devices` to be used again.
Instead, I added some complementary logic for `processing_devices`.

Please review it again including the new commit.

drivers/Aqara/aqara-presence-sensor/src/discovery.lua
```diff
@@ -100,14 +109,15 @@ local function discovery_device(driver)
...
+    if (not discovery.processing_devices[dni]) and (not is_already_added) then
       try_add_device(driver, dni, ip)
     end
   end
 end
 
 function discovery.do_network_discovery(driver, _, should_continue)
   log.info_with({ hub_logs = true }, string.format("discovery start for Aqara FP2"))
+  discovery.processing_devices = {}
```
cjswedes: I like this approach to only retry a device add request to the platform each time discovery is started/stopped.

drivers/Aqara/aqara-presence-sensor/src/discovery.lua
```diff
@@ -100,14 +109,15 @@ local function discovery_device(driver)
         break
       end
     end
-    if (not processing_devices[dni]) and (not is_already_added) then
+    if (not discovery.processing_devices[dni]) and (not is_already_added) then
```
cjswedes: If you wanted to be more aggressive with retries of the device create request to the platform, you could set the `processing_devices[dni]` value to a timestamp when making the request and check if 5-10 seconds has elapsed since we last tried. Doing this might make it so that a user doesn't need to re-run discovery to get a device onboarded.

I am not requesting any changes, I just thought I would offer a suggestion in case it was something you wanted to do.

hyundolee27: Thank you for your review and good suggestion.
I added the 10-second delay as you suggested. My concerns have also been resolved. Thank you!

Kwang-Hui:
@NoahCornell @cjswedes Could you merge this PR?

cjswedes:
Please squash the commits to 1 or 2 commits, and then we will merge.

hyundolee27:
@cjswedes  I have prepared a squash commit and pushed the changes. Could you please review and merge it?

PR #2236, opened 2025-07-03
Re-add polling for ZLL devices

greens:


drivers/SmartThings/zigbee-switch/src/zll-polling/init.lua
```diff
@@ -0,0 +1,39 @@
...
+local function zll_profile(opts, driver, device, zb_rx, ...)
+  if (device.zigbee_endpoints[device.fingerprinted_endpoint_id].profile == constants.ZLL_PROFILE_ID) then
+    local subdriver = require("zll-polling")
+    return true, subdriver
+  else return false
+  end
+end
+
+local function set_up_zll_polling(driver, device)
+  device.thread:call_on_schedule(5 * 60, function() device:refresh() end, "zll_polling")
```
tpmanley: I believe the older code would only poll the device if we hadn't heard from it in the max poll interval time, is that right? With this change we'd be reading all the clusters every five minutes, even if the device was recently used. I think we may want to be more conservative here and only read the attributes that haven't been refreshed recently to avoid unnecessary messages.

drivers/SmartThings/zigbee-switch/src/zll-polling/init.lua
```diff
@@ -0,0 +1,63 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local device_lib = require "st.device"
+local constants = require "st.zigbee.constants"
+local clusters = require "st.zigbee.zcl.clusters"
+
+-- There are reports of at least one device (SONOFF 01MINIZB) which occasionally
+-- reports this value as an Int8, rather than a Boolean, as per the spec
```
cjswedes: What is meant by this comment? Is it talking about the profile id on the device table?

greens: leftover from copy/paste

PR #2241, opened 2025-07-08
CHAD-15418 workaround

greens:
endpoint ids change from numbers to strings shortly after install. This also fixes some incorrectly-addressed fields for kickstarter sensors.


cbaumler (approved):
I tested these changes using an Innr ZLL bulb and everything was working as expected.

Test Scenario:
- turn ZLL bulb off via mobile app
- Unplug and plug back in (turns on)
- wait 5+ minutes
- Observe device changed from off to on in the mobile app

PR #2249, opened 2025-07-11
zigbee-contact: Remove incorrect device events from init handler

NoahCornell:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

This was a VOC bug that came in where the customer was reporting that there device was incorrectly reporting as open and triggering automations. The logs showed hubcore restarting, so the driver was also starting up, and the driver emitted these events without receiving any zigbee messages. The device record matched this subdriver which is unconditionally sending out `contactSensor:contact`  as `open` and `batteryLevel:battery` as `normal`.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-contact/src/aqara/init.lua
+++ b/drivers/SmartThings/zigbee-contact/src/aqara/init.lua
@@ -55,17 +55,19 @@ local function device_init(driver, device)
     device:add_configured_attribute(attribute)
     device:add_monitored_attribute(attribute)
   end
-
-  device:emit_event(capabilities.batteryLevel.type("CR1632"))
-  device:emit_event(capabilities.batteryLevel.quantity(1))
-  device:emit_event(capabilities.batteryLevel.battery("normal"))
-  device:emit_event(capabilities.contactSensor.contact.open())
 end
 
 local function do_configure(self, device)
   device:configure()
   device:send(cluster_base.write_manufacturer_specific_attribute(device,
     PRIVATE_CLUSTER_ID, PRIVATE_ATTRIBUTE_ID, MFG_CODE, data_types.Uint8, 0x01))
+end
+
+local function added_handler(driver, device)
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zigbee-contact/src/aqara/init.lua
```diff
@@ -58,8 +58,6 @@ local function device_init(driver, device)
 
   device:emit_event(capabilities.batteryLevel.type("CR1632"))
   device:emit_event(capabilities.batteryLevel.quantity(1))
-  device:emit_event(capabilities.batteryLevel.battery("normal"))
-  device:emit_event(capabilities.contactSensor.contact.open())
```
cjswedes, line 62: @seojune79 is this something that should be in an added handler?

greens: None of these should be in `init`. They should all be in `added`.

greens: My guess is they were put in `init` because they were added later and the functionality was desired for devices that had already been installed.

NoahCornell: Is there any concern with adding battery and contact to an added handler since they could still be technically incorrect at that time? There are a few situations where we might send added events after a device has already received one specifically thinking of a failover/replace scenario.

greens: It's the least bad place for them and the generally accepted practice for synthetic events like this.

NoahCornell: Updated the PR so the events are emitted in the added handler now

varzac (approved):

PR #2259, opened 2025-07-15
Add default enum value to supportedAlarmvalues

HunsupJung:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
DoorLockAlarm event is manadatory event in Matter spec and LockJammed is mandatory enum value.
So this patch add unableToLockTheDoor value that is mapped to LockJammed of Matter spec to supportedAlarmValues.

supportedAlarmValues is used in Routine. If supportedAlarmValues is nil, user can select all of lockAlarm Enum value. Even if device does not support some enum values, user can select those values. So we decide to add default value(unableToLockTheDoor) to supportedAlarmVlaues.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -204,6 +204,8 @@ local function info_changed(driver, device, event, args)
     end
   end
   device:subscribe()
+  device:emit_event(capabilities.lockAlarm.alarm.clear({state_change = true}))
+  device:emit_event(capabilities.lockAlarm.supportedAlarmValues({"unableToLockTheDoor"})) -- lockJammed is madatory
 end
 
 -- This function check busy_state and if busy_state is false, set it to true(current time)
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -205,7 +205,7 @@ local function info_changed(driver, device, event, args)
   end
   device:subscribe()
   device:emit_event(capabilities.lockAlarm.alarm.clear({state_change = true}))
-  device:emit_event(capabilities.lockAlarm.supportedAlarmValues({"unableToLockTheDoor"})) -- lockJammed is madatory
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -204,6 +204,8 @@ local function info_changed(driver, device, event, args)
     end
   end
   device:subscribe()
+  device:emit_event(capabilities.lockAlarm.alarm.clear({state_change = true}))
+  device:emit_event(capabilities.lockAlarm.supportedAlarmValues({"unableToLockTheDoor"})) -- lockJammed is madatory
```
Kwang-Hui: How about to add `{visibility = {displayed = false}}` here?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -204,6 +204,8 @@ local function info_changed(driver, device, event, args)
     end
   end
   device:subscribe()
+  device:emit_event(capabilities.lockAlarm.alarm.clear({state_change = true}))
```
nickolas-deboom, line 207: Should this be added in the main driver as well?

Kwang-Hui: I don't think lockAlarm capability is used at main driver.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: Oh yes, I missed that option. I will add it.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: The handler for lockAlarm capability is used at main driver. but main driver does not use profiles that have lockAlarm capability.

nickolas-deboom (approved):

PR #2260, opened 2025-07-15
Wait for Max PIN length before generating credentials

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

There fixes a bug where if the handler for `RequirePINforRemoteOperation` runs before the `maxCodeLength` attribute of the `lockCodes` capability is set, a default pin length of 4 would be used, which may be shorter than the minumim pin length of the device specified by the `MinPINCodeLength` attribute. Therefore, the read was removed from `init` and `added', and was moved to the handler for `MaxPINCodeLength` to guarantee that the max pin length is known before attempting to generate the COTA credential.

# Summary of Completed Tests

Updated UTs


drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -70,8 +70,7 @@ local subscribed_attributes = {
   [capabilities.lockCredentials.ID] = {
     DoorLock.attributes.NumberOfPINUsersSupported,
     DoorLock.attributes.MaxPINCodeLength,
-    DoorLock.attributes.MinPINCodeLength,
-    DoorLock.attributes.RequirePINforRemoteOperation
```
hcarter-775, line 112: I think it may add some unnecessary added complexity to remove this attribute from the subscription (and to add it to the MaxPINCodeLength handler. Obviously in both cases we are ostensibly "subscribing" to the attribute, but if we know we want to do the read, we should let it stay optimized and leave it in the subscribe block.

I think we should go the route we do in matter thermostat with a similar issue around [ControlSequenceOfOperation](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/matter-thermostat/src/init.lua#L1433) (relevant part [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/matter-thermostat/src/init.lua#L1490)) and [SystemMode](http://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/matter-thermostat/src/init.lua#L1394), which allows us to leave the attribute in the subscribe but still gates the order of events.

nickolas-deboom: I addressed this in the latest force push, let me know how that looks!

hcarter-775 (approved):

drivers/SmartThings/matter-lock/src/lock_utils.lua
```diff
@@ -50,7 +50,9 @@ local lock_utils = {
   ENDPOINT_KEY_INDEX = "endpointKeyIndex",
   ENDPOINT_KEY_TYPE = "endpointKeyType",
   DEVICE_KEY_ID = "deviceKeyId",
-  COMMAND_REQUEST_ID = "commandRequestId"
+  COMMAND_REQUEST_ID = "commandRequestId",
+  MAX_PIN_LENGTH = "max_pin_length",
```
hcarter-775:
```suggestion
  MAX_PIN_CODE_LENGTH = "max_pin_code_length",
```
Might be best to just make the variable name exactly like the attribute it's getting its data from

nickolas-deboom: Updated as suggested!

PR #2267, opened 2025-07-17
CHAD-15777: temperatureMeasurement:temperature registering for native…

aleclorimer:
# Checklist

# Description of Change
Registering of native handlers for zigbee and z-wave devices for temperatureMeasurement capabilities using the default handlers.

# Summary of Completed Tests
- [ ] Tested on a physical device using the updated drivers

### DIFF
--- a/drivers/SmartThings/matter-sensor/src/init.lua
+++ b/drivers/SmartThings/matter-sensor/src/init.lua
@@ -233,6 +233,9 @@ local function temperature_attr_handler(driver, device, ib, response)
     local temp = measured_value / 100.0
     local unit = "C"
     device:emit_event_for_endpoint(ib.endpoint_id, capabilities.temperatureMeasurement.temperature({value = temp, unit = unit}))
+    if type(device.register_native_capability_attr_handler) == "function" then
+      device:register_native_capability_attr_handler("temperatureMeasurement", "temperature")
+    end
   end
 end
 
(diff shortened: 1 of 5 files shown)


drivers/SmartThings/zwave-sensor/src/fibaro-flood-sensor/init.lua
```diff
@@ -78,6 +78,9 @@ local function sensor_multilevel_report_handler(self, device, cmd)
     local scale = 'C'
     if (cmd.args.scale == SensorMultilevel.scale.temperature.FAHRENHEIT) then scale = 'F' end
     device:emit_event(capabilities.temperatureMeasurement.temperature({value = cmd.args.sensor_value, unit = scale}))
+    if type(device.register_native_capability_attr_handler) == "function" then
+      device:register_native_capability_attr_handler("temperatureMeasurement","temperature")
+    end
```
cjswedes, line 83: typically I have avoided doing this in a subdriver even if I tested with it. Was this intended to be done or was it a part of testing?

greens:
The `"register_native_capability_cmd_handler"` `"devices"` event will need to be added to the tests, right @cjswedes ?

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -233,6 +233,9 @@ local function temperature_attr_handler(driver, device, ib, response)
     local temp = measured_value / 100.0
     local unit = "C"
     device:emit_event_for_endpoint(ib.endpoint_id, capabilities.temperatureMeasurement.temperature({value = temp, unit = unit}))
+    if type(device.register_native_capability_attr_handler) == "function" then
```
nickolas-deboom, line 236: Will the same be done for matter-thermostat?

drivers/SmartThings/zwave-sensor/src/fibaro-flood-sensor/init.lua
aleclorimer: I threw this in there because I've seen it done a few times and this was the device that I used to develop the NH for Z-wave. This logic follows the default handler logic, but maybe a better route is to remove this method from the subdriver?

PR #2273, opened 2025-07-21
Update configure tests with new zigbee power reporting

varzac:
Update tests to match the new default configuration for power meter reporting

### DIFF
--- a/drivers/SmartThings/zigbee-power-meter/src/test/test_zigbee_power_meter.lua
+++ b/drivers/SmartThings/zigbee-power-meter/src/test/test_zigbee_power_meter.lua
@@ -99,7 +99,7 @@ test.register_coroutine_test(
                                        })
       test.socket.zigbee:__expect_send({
                                          mock_device.id,
-                                         SimpleMetering.attributes.InstantaneousDemand:configure_reporting(mock_device, 1, 3600, 5)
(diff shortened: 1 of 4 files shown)


tpmanley:
You probably noticed this already but some tests are failing:

```
zigbee-switch: 
	test all capability zigbee bulb FAILED on lifecycle configure event should configure device

	test switch power FAILED on Configure should configure all necessary attributes and refresh device


zigbee-power-meter: 
	test zigbee power meter consumption report FAILED on lifecycle configure event should configure device

	test zigbee power meter FAILED on lifecycle configure event should configure device
```

varzac:
> You probably noticed this already but some tests are failing:
> 
> ```
> zigbee-switch: 
> 	test all capability zigbee bulb FAILED on lifecycle configure event should configure device
> 
> 	test switch power FAILED on Configure should configure all necessary attributes and refresh device
> 
> 
> zigbee-power-meter: 
> 	test zigbee power meter consumption report FAILED on lifecycle configure event should configure device
> 
> 	test zigbee power meter FAILED on lifecycle configure event should configure device
> ```

This is because this is still using the last released lua lib artifact which doesn't include the default change yet.  So the tests on this PR will fail until we release a new lua lib artifact.

tpmanley (approved):

varzac:
These tests fixes got pulled in elsewhere

PR #2279, opened 2025-07-22
Support Light with Motion Sensor

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change adds a generic fingerprint and supporting profile for a light with a motion sensor.

# Summary of Completed Tests

New test case added.

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -2847,6 +2847,12 @@ matterGeneric:
       - id: 0x010D # Extended Color Light
       - id: 0x002B # Fan
     deviceProfileName: light-color-level-fan
+  - id: "matter/dimmable/light/motion"
+    deviceLabel: Matter Dimmable Light Occupancy Sensor
+    deviceTypes:
+      - id: 0x0101 # Dimmable Light
+      - id: 0x0107 # Occupancy Sensor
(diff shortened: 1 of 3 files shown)


hcarter-775:
as we do with the fan+light device, I believe we should add this to our `match_profile` logic. Changes coming soon will require this anyway, so I believe it would be best to implement it now

nickolas-deboom:
> as we do with the fan+light device, I believe we should add this to our `match_profile` logic. Changes coming soon will require this anyway, so I believe it would be best to implement it now

Sure, that works too. I guess we can remove the fingerprint then. We also don't need the fingerprint for the fan light I suppose

nickolas-deboom:
Looking at it again, I think we should go with the fingerprint approach until modular profiles are implemented. There are other devices that use profiles with the `motionSensor` capability that rely on fingerprints that would be messed with if we re-profile based on the presense of the OccupancySensing cluster.

hcarter-775:
yeah, I can update the logic in [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2199), which is where I see issues coming up

nickolas-deboom:
I added one more commit to use a range of 1-100 for switchLevel in the profile as we do for other dimmer light profiles.

PR #2280, opened 2025-07-23
Matter Sensor: Use presenceSensor if possible

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change updates the profiling logic to use a profile including the presenceSensor capability over motionSensor if the device supports it based on the feature map.

# Summary of Completed Tests


drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -172,6 +168,21 @@ local function match_profile(driver, device, battery_supported)
...
   profile_name = profile_name .. concatenated_preferences
 
+  if device:supports_capability(capabilities.motionSensor) then
+    -- If the Occupancy Sensing Cluster’s revision is >= 5 (corresponds to Lua Libs version 13+), and any of the AIR / RAD / RFS / VIS
+    -- features are supported by the device, use the presenceSensor capability. Otherwise, use the motionSensor capability.
+    local occupancy_air_eps = device:get_endpoints(clusters.OccupancySensing.ID, {feature_bitmap = clusters.OccupancySensing.types.Feature.ACTIVE_INFRARED}) or {}
+    local occupancy_rad_eps = device:get_endpoints(clusters.OccupancySensing.ID, {feature_bitmap = clusters.OccupancySensing.types.Feature.RADAR}) or {}
+    local occupancy_rfs_eps = device:get_endpoints(clusters.OccupancySensing.ID, {feature_bitmap = clusters.OccupancySensing.types.Feature.RF_SENSING}) or {}
+    local occupancy_vis_eps = device:get_endpoints(clusters.OccupancySensing.ID, {feature_bitmap = clusters.OccupancySensing.types.Feature.VISION}) or {}
+    if profile_name == "-illuminance-temperature-humidity-battery" and version.api >= 13 and
```
hcarter-775: let's put the version check before we search through the eps in the first place

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -172,6 +168,21 @@ local function match_profile(driver, device, battery_supported)
   local concatenated_preferences = supports_sensitivity_preferences(device)
   profile_name = profile_name .. concatenated_preferences
 
+  if device:supports_capability(capabilities.motionSensor) then
+    -- If the Occupancy Sensing Cluster’s revision is >= 5 (corresponds to Lua Libs version 13+), and any of the AIR / RAD / RFS / VIS
+    -- features are supported by the device, use the presenceSensor capability. Otherwise, use the motionSensor capability.
+    local occupancy_air_eps = device:get_endpoints(clusters.OccupancySensing.ID, {feature_bitmap = clusters.OccupancySensing.types.Feature.ACTIVE_INFRARED}) or {}
```
hcarter-775: I don't think we need the `or {}` in these. That's automatically returned by `get_endpoints`

drivers/SmartThings/matter-sensor/src/init.lua
nickolas-deboom: Sure, done in latest commit

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -172,6 +168,23 @@ local function match_profile(driver, device, battery_supported)
...
+    -- features are supported by the device, use the presenceSensor capability. Otherwise, use the motionSensor capability. Currently,
+    -- presenceSensor only used for devices fingerprinting to the motion-illuminance-temperature-humidity-battery profile.
+    if profile_name == "-illuminance-temperature-humidity-battery" and version.api >= 13 then
+      local occupancy_air_eps = device:get_endpoints(clusters.OccupancySensing.ID, {feature_bitmap = clusters.OccupancySensing.types.Feature.ACTIVE_INFRARED})
+      local occupancy_rad_eps = device:get_endpoints(clusters.OccupancySensing.ID, {feature_bitmap = clusters.OccupancySensing.types.Feature.RADAR})
+      local occupancy_rfs_eps = device:get_endpoints(clusters.OccupancySensing.ID, {feature_bitmap = clusters.OccupancySensing.types.Feature.RF_SENSING})
+      local occupancy_vis_eps = device:get_endpoints(clusters.OccupancySensing.ID, {feature_bitmap = clusters.OccupancySensing.types.Feature.VISION})
+      if #occupancy_air_eps > 0 or #occupancy_rad_eps > 0 or #occupancy_rfs_eps > 0 or #occupancy_vis_eps > 0 then
+        occupancy_support = "-presence"
+      end
```
hcarter-775: Perhaps we should put these get_endpoints calls in the if statement rather than calling them each, saving the data, and then doing an if check. This would save some on stored temp data, and would be faster since a valid check would stop the handling more quickly.

hcarter-775: Also, since I cannot recall, can we just check the negation of this? Like, if it supports the other one, we should use motion, or is this purely additive?

nickolas-deboom: Sure thing, updated as suggested in your first comment. For your second question, I might be misinterpreting but I think we could do it that way, the only benefit to the current implementation being that these checks are only needed if the profile is "motion-illuminance-temperature-humidity-battery" and so we can skip it in all other cases.

hcarter-775: oh yeah, that's not what I meant I don't think. I think this looks good as-is either way

hcarter-775 (approved):

PR #2282, opened 2025-07-24
CHAD-15784 Zigbee: WindowShadePreset capability updates

greens:


drivers/SmartThings/zigbee-window-treatment/src/init.lua
```diff
@@ -15,6 +15,23 @@
 local capabilities = require "st.capabilities"
 local ZigbeeDriver = require "st.zigbee"
 local defaults = require "st.zigbee.defaults"
+local window_preset_defaults = require "st.zigbee.defaults.windowShadePreset_defaults"
+
+local function init_handler(self, device)
```
inasail, line 20: Like I mentioned in the [CHAD-15784](https://smartthings.atlassian.net/browse/CHAD-15784?focusedCommentId=1637968), we need this.
device:emit_event(capabilities.windowShadePreset.supportedCommands({"presetPosition", "setPresetPosition"}, { visibility = { displayed = false }})) in init handler.

inasail: And also we need to add these codes in

1. vimar (init overrided)
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-window-treatment/src/vimar/init.lua#L133

2. This is a different driver
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/Unofficial/tuya-zigbee/src/curtain/init.lua#L132

PS: If you hesitate to manage the Unofficial driver, I can request SRCN to do this.

drivers/SmartThings/zigbee-window-treatment/src/init.lua
```diff
@@ -15,9 +15,29 @@
...
+
+    device:emit_event(capabilities.windowShadePreset.position(preset_position, { visibility = {displayed = false}}))
+    device:set_field(window_preset_defaults.PRESET_LEVEL_KEY, preset_position, {persist = true})
+  end
+end
 
 local function added_handler(self, device)
   device:emit_event(capabilities.windowShade.supportedWindowShadeCommands({"open", "close", "pause"}, { visibility = { displayed = false }}))
+  device:emit_event(capabilities.windowShadePreset.supportedCommands({"presetPosition", "setPresetPosition"}, { visibility = { displayed = false }}))
+  device:emit_event(capabilities.windowShadePreset.position(window_preset_defaults.PRESET_LEVEL, { visibility = {displayed = false}}))
```
inasail: I think 'windowShadePreset.position' should only be emitted when there is no latest value, because the added() handler is called in the case of a hub switch-over.
The same goes for Tuya curtain in Unofficial driver.

drivers/SmartThings/zigbee-window-treatment/src/init.lua
```diff
@@ -15,6 +15,24 @@
...
 local defaults = require "st.zigbee.defaults"
+local window_preset_defaults = require "st.zigbee.defaults.windowShadePreset_defaults"
+
+local function init_handler(self, device)
+  if device:supports_capability_by_id(capabilities.windowShadePreset.ID) and
+      device:get_latest_state("main", capabilities.windowShadePreset.ID, capabilities.windowShadePreset.position.NAME) == nil then
+
+    -- These should only ever be nil once (and at the same time) for already-installed devices
+    -- It can be relocated to `added` after migration is complete
+    device:emit_event(capabilities.windowShadePreset.supportedCommands({"presetPosition", "setPresetPosition"}, { visibility = { displayed = false }}))
```
nickolas-deboom, line 26: How is the `setPresetPosition` command used? It doesn't appear in the [presentation](https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/main/capabilities/window-shade-preset.src/presentation.yml) and I couldn't figure out how to change the preset value from the app

nickolas-deboom (approved):

PR #2283, opened 2025-07-25
Add Matter HRAP Support

hcarter-775:
# Description of Change
Implements handling for the TBR and NIM Device Types. Includes a basic thread operational dataset parser.

# Summary of Completed Tests
Tested using CHIP Example App, VDA, and is verified with unit tests.

### DIFF
--- a/drivers/SmartThings/matter-hrap/src/init.lua
+++ b/drivers/SmartThings/matter-hrap/src/init.lua
@@ -1,0 +1,217 @@
+-- Copyright 2025 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+local MatterDriver = require "st.matter.driver"
+local data_types = require "st.matter.data_types"
+local im = require "st.matter.interaction_model"
+local capabilities = require "st.capabilities"
+local clusters = require "st.matter.clusters"
+local lustre_utils = require "lustre.utils"
+local st_utils = require "st.utils"
+local version = require "version"
+local log = require "log"
+
+local CURRENT_ACTIVE_DATASET_TIMESTAMP = "__CURRENT_ACTIVE_DATASET_TIMESTAMP"
+local GET_ACTIVE_DATASET_RETRY_ATTEMPTS = "__GET_ACTIVE_DATASET_RETRY_ATTEMPTS"
+
+-- Include driver-side definitions when lua libs api version is <13
+if version.api < 13 then
+  clusters.ThreadBorderRouterManagement = require "ThreadBorderRouterManagement"
+  clusters.WiFiNetworkManagement = require "WiFiNetworkManagement"
+end
+
+
+--[[ ATTRIBUTE HANDLERS ]]--
+
+local function border_router_name_attribute_handler(driver, device, ib)
+  -- per the spec, the recommended attribute format is <VendorName> <ProductName>._meshcop._udp. This logic removes the MeshCoP suffix IFF it is present
+  local meshCop_name = ib.data.value
+  local terminal_display_char = (string.find(meshCop_name, "._meshcop._udp") or 64) - 1 -- where 64-1=63, the maximum allowed length for BorderRouterName
+  local display_name = string.sub(meshCop_name, 1, terminal_display_char)
+  device:emit_event_for_endpoint(ib.endpoint, capabilities.threadBorderRouter.borderRouterName({ value = display_name }))
+end
+
+local function ssid_attribute_handler(driver, device, ib)
+  if (ib.data.value == string.char(data_types.Null.ID) or ib.data.value == nil) then -- Matter TLV-encoded NULL or Lua-encoded NULL
+    device.log.info("No primary Wi-Fi network is available")
+    return
+  end
+  local valid_utf8, utf8_err = lustre_utils.validate_utf8(ib.data.value)
+  if valid_utf8 then
+    device:emit_event_for_endpoint(ib.endpoint, capabilities.wifiInformation.ssid({ value = ib.data.value }))
+  else
+    device.log.info("UTF-8 validation of SSID failed: Error: '"..utf8_err.."'")
+  end
+end
+
+local function thread_interface_enabled_attribute_handler(driver, device, ib)
+  if ib.data.value then
+    device:emit_event_for_endpoint(ib.endpoint, capabilities.threadBorderRouter.threadInterfaceState("enabled"))
+  else
+    device:emit_event_for_endpoint(ib.endpoint, capabilities.threadBorderRouter.threadInterfaceState("disabled"))
+  end
+end
+
+-- Spec uses TLV encoding of Thread Version, which should be mapped to a more human-readable name
+local VERSION_TLV_MAP = {
+  [1] = "1.0.0",
+  [2] = "1.1.0",
+  [3] = "1.2.0",
+  [4] = "1.3.0",
+  [5] = "1.4.0",
+}
+
+local function thread_version_attribute_handler(driver, device, ib)
+  local version_name = VERSION_TLV_MAP[ib.data.value]
+  if version_name then
+    device:emit_event_for_endpoint(ib.endpoint, capabilities.threadBorderRouter.threadVersion({ value = version_name }))
+  else
+    device.log.warn("The received TLV-encoded Thread version does not have a provided mapping to a human-readable version format")
+  end
+end
+
+local function active_dataset_timestamp_handler(driver, device, ib)
+  if not ib.data.value then
+    device.log.info("No Thread operational dataset configured")
+  elseif ib.data.value ~= device:get_field(CURRENT_ACTIVE_DATASET_TIMESTAMP) then
+    device:send(clusters.ThreadBorderRouterManagement.server.commands.GetActiveDatasetRequest(device, ib.endpoint_id))
+    device:set_field(CURRENT_ACTIVE_DATASET_TIMESTAMP, ib.data.value, { persist = true })
+  end
+end
+
+
+--[[ COMMAND HANDLERS ]]--
+
+local threadNetwork = capabilities.threadNetwork
+local TLV_TYPE_ATTR_MAP = {
+  [0] = threadNetwork.channel,
+  [1] = threadNetwork.panId,
+  [2] = threadNetwork.extendedPanId,
+  [3] = threadNetwork.networkName,
+  -- [4] intentionally omitted, refers to the MeshCoP PSKc
+  [5] = threadNetwork.networkKey,
+}
+
+local function dataset_response_handler(driver, device, ib)
+  if not ib.info_block.data.elements.dataset then
+    device.log.debug_with({ hub_logs = true }, "Received empty Thread operational dataset")
+    return
+  end
+
+  local operational_dataset_length = ib.info_block.data.elements.dataset.byte_length
+  local spec_defined_max_dataset_length = 254
+  if operational_dataset_length > spec_defined_max_dataset_length then
+      device.log.error_with({ hub_logs = true }, "Received Thread operational dataset is too long")
+      return
+  end
+
+  -- parse dataset
+  local operational_dataset = ib.info_block.data.elements.dataset.value
+  local cur_byte = 1
+  while cur_byte + 1 <= operational_dataset_length do
+    local tlv_type = string.byte(operational_dataset, cur_byte)
+    local tlv_length = string.byte(operational_dataset, cur_byte + 1)
+    if (cur_byte + 1 + tlv_length) > operational_dataset_length then
+      device.log.error_with({ hub_logs = true }, "Received Thread operational dataset has a malformed TLV encoding")
+      return
+    end
+    local tlv_mapped_attr = TLV_TYPE_ATTR_MAP[tlv_type]
+    if tlv_mapped_attr then
+      -- extract the value from a TLV-encoded message. Message format: byte tag + byte length + length byte value
+      local tlv_value = operational_dataset:sub(cur_byte + 2, cur_byte + 1 + tlv_length)
+      -- format data as required by threadNetwork attribute properties
+      if tlv_mapped_attr == threadNetwork.channel or tlv_mapped_attr == threadNetwork.panId then
+        tlv_value = st_utils.deserialize_int(tlv_value, tlv_length)
+      elseif tlv_mapped_attr ~= threadNetwork.networkName then
+        tlv_value = st_utils.bytes_to_hex_string(tlv_value)
+      end
+      device:emit_event(tlv_mapped_attr({ value = tlv_value }))
+    end
+    cur_byte = cur_byte + 2 + tlv_length
+  end
+end
+
+local function get_active_dataset_response_handler(driver, device, ib)
+  if ib.status == im.InteractionResponse.Status.FAILURE then
+    -- per spec, on a GetActiveDatasetRequest failure, a failure response is sent over the same command.
+    -- on failure, retry the read up to 3 times before failing out.
+    local retries_attempted = device:get_field(GET_ACTIVE_DATASET_RETRY_ATTEMPTS) or 0
+    if retries_attempted < 3 then
+      device.log.error_with({ hub_logs = true }, "Failed to retrieve Thread operational dataset. Retrying " .. retries_attempted + 1 .. "/3")
+      device:set_field(GET_ACTIVE_DATASET_RETRY_ATTEMPTS, retries_attempted + 1)
+    else
+      -- do not retry again, but reset the count to 0.
+      device:set_field(GET_ACTIVE_DATASET_RETRY_ATTEMPTS, 0)
+    end
+  elseif ib.status == im.InteractionResponse.Status.UNSUPPORTED_ACCESS then
+    device.log.error_with({ hub_logs = true },
+      "Failed to retrieve Thread operational dataset, since the GetActiveDatasetRequest command was not executed over CASE"
+    )
+  end
+end
+
+
+--[[ LIFECYCLE HANDLERS ]]--
+
+local function device_init(driver, device)
+  device:subscribe()
+end
+
+
+--[[ MATTER DRIVER TEMPLATE ]]--
+
+local matter_driver_template = {
+  lifecycle_handlers = {
+    init = device_init,
+  },
+  matter_handlers = {
+    attr = {
+      [clusters.WiFiNetworkManagement.ID] = {
+        [clusters.WiFiNetworkManagement.attributes.Ssid.ID] = ssid_attribute_handler,
+      },
+      [clusters.ThreadBorderRouterManagement.ID] = {
+        [clusters.ThreadBorderRouterManagement.attributes.BorderRouterName.ID] = border_router_name_attribute_handler,
+        [clusters.ThreadBorderRouterManagement.attributes.ThreadVersion.ID] = thread_version_attribute_handler,
+        [clusters.ThreadBorderRouterManagement.attributes.InterfaceEnabled.ID] = thread_interface_enabled_attribute_handler,
+        [clusters.ThreadBorderRouterManagement.attributes.ActiveDatasetTimestamp.ID] = active_dataset_timestamp_handler,
+      }
+    },
+    cmd_response = {
+      [clusters.ThreadBorderRouterManagement.ID] = {
+        [clusters.ThreadBorderRouterManagement.client.commands.DatasetResponse.ID] = dataset_response_handler,
+        [clusters.ThreadBorderRouterManagement.server.commands.GetActiveDatasetRequest.ID] = get_active_dataset_response_handler,
+      }
+    }
+  },
+  subscribed_attributes = {
+    [capabilities.threadBorderRouter.ID] = {
+      clusters.ThreadBorderRouterManagement.attributes.ActiveDatasetTimestamp,
+      clusters.ThreadBorderRouterManagement.attributes.BorderRouterName,
+      clusters.ThreadBorderRouterManagement.attributes.InterfaceEnabled,
+      clusters.ThreadBorderRouterManagement.attributes.ThreadVersion,
+    },
+    [capabilities.wifiInformation.ID] = {
+      clusters.WiFiNetworkManagement.attributes.Ssid,
+    }
+  },
+  supported_capabilities = {
+    capabilities.threadBorderRouter,
+    capabilities.threadNetwork,
+    capabilities.wifiInformation,
+  }
+}
+
+local matter_driver = MatterDriver("matter-hrap", matter_driver_template)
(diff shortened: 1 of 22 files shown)


drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,182 @@
...
+    device:emit_event_for_endpoint(ib.endpoint, capabilities.threadBorderRouter.threadVersion({ value = version_name }))
+  end
+end
+
+
+--[[ COMMAND HANDLERS ]]--
+
+local threadNetwork = capabilities.threadNetwork
+local TLV_TYPE_ATTR_MAP = {
+  [0] = threadNetwork.channel,
```
nickolas-deboom, line 98: Where do these tlv type indices come from? In the thread spec, I see that the channel TLV is type 20 and pan ID is 21 for example. Am I looking in the wrong place?

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,182 @@
...
+  while cur_byte + 1 <= operational_dataset_length do
+    local tlv_type = string.byte(operational_dataset, cur_byte)
+    local tlv_length = string.byte(operational_dataset, cur_byte + 1)
+    if (cur_byte + 1 + tlv_length) > operational_dataset_length then
+      log.error("In dataset_response_handler, received a malformed operational dataaset")
+      return
+    end
+    local tlv_mapped_attr = TLV_TYPE_ATTR_MAP[tlv_type]
+    if tlv_mapped_attr then
+      local tlv_value = operational_dataset:sub(cur_byte + 2, cur_byte + 1 + tlv_length)
```
nickolas-deboom, line 132:
```suggestion
      local tlv_value = operational_dataset:sub(cur_byte + 2, cur_byte + 2 + tlv_length)
```
Should this be '+2', or does the TLV length include the byte containing the TLV length?

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,182 @@
...
+      return
+  end
+
+  -- parse dataset
+  local operational_dataset = ib.info_block.data.elements.dataset.value
+  local cur_byte = 1
+  while cur_byte + 1 <= operational_dataset_length do
+    local tlv_type = string.byte(operational_dataset, cur_byte)
+    local tlv_length = string.byte(operational_dataset, cur_byte + 1)
+    if (cur_byte + 1 + tlv_length) > operational_dataset_length then
```
nickolas-deboom, line 125: Same question here, should this be '+2'?

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: no, because the +1 just accounts for skipping the 1 byte tlv_length

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: Same as above, no, because the +1 just accounts for skipping the 1 byte tlv_length

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: Here's one source of the above handling: https://openthread.io/reference/group/api-operational-dataset

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,182 @@
...
+  subscribed_attributes = {
+    [capabilities.threadBorderRouter.ID] = {
+      clusters.ThreadBorderRouterManagement.attributes.BorderRouterName,
+      clusters.ThreadBorderRouterManagement.attributes.InterfaceEnabled,
+      clusters.ThreadBorderRouterManagement.attributes.ThreadVersion,
+    },
+    [capabilities.wifiInformation.ID] = {
+      clusters.WiFiNetworkManagement.attributes.Ssid,
+    }
+  }
```
nickolas-deboom, line 212: You may want to include a `supported_capabilities` list here, although I don't think it's strictly needed

drivers/SmartThings/matter-hrap/src/init.lua
nickolas-deboom: But what about the byte for the tlv type?

hcarter-775: that's the `cur_byte`

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: added in [cf2b9ff](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2283/commits/cf2b9ffaa5185cd2342d4e72a8076ff68bbb4849)

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: To be clear, this `operational_dataset:sub(cur_byte + 2, cur_byte + 1 + tlv_length)` pulls the substring from the first byte of the tlv value (which is at cur_byte + 2, since this function indexes inclusively) and the last byte inclusively of the tlv value, which is the cur_byte, currently sitting on the type at this point, plus the length byte, plus the whole value length.

So for both of these indices, they are inclusive of the string being taken.

nickolas-deboom: Gotcha, thanks!

drivers/SmartThings/matter-hrap/src/init.lua
nickolas-deboom: Thanks!

nickolas-deboom:
I looked over the cluster-capability mappings again and passed over the driver again and things look good from my perspective. One more question I had is have you tested using the embedded cluster definitions?

hcarter-775:
> I looked over the cluster-capability mappings again and passed over the driver again and things look good from my perspective. One more question I had is have you tested using the embedded cluster definitions?

Good callout! Yes, this has been tested now. There actually was a small typo bug that got fixed in the latest commit

HunsupJung:
@hcarter-775 
I tested it with VDA (1.4.8) and there's no issue.

HunsupJung (approved):

nickolas-deboom (approved):

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,187 @@
...
+if version.api < 13 then
+  clusters.ThreadBorderRouterManagement = require "ThreadBorderRouterManagement"
+  clusters.WiFiNetworkManagement = require "WiFiNetworkManagement"
+end
+
+
+--[[ ATTRIBUTE HANDLERS ]]--
+
+local function border_router_name_attribute_handler(driver, device, ib)
+  -- per the spec, the recommended attribute format is <VendorName> <ProductName>._meshcop._udp. This logic removes the meschop suffix IFF it is present
```
ctowns: nit: 
```suggestion
  -- per the spec, the recommended attribute format is <VendorName> <ProductName>._meshcop._udp. This logic removes the MeshCoP suffix IFF it is present
```

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,187 @@
...
+  if valid_utf8 then
+    device:emit_event_for_endpoint(ib.endpoint, capabilities.wifiInformation.ssid({ value = ib.data.value }))
+  else
+    device.log.info("UTF-8 validation of SSID failed: Error: '"..utf8_err.."'")
+  end
+end
+
+local function thread_interface_enabled_attribute_handler(driver, device, ib)
+  if ib.data.value then
+    device:emit_event_for_endpoint(ib.endpoint, capabilities.threadBorderRouter.threadInterfaceState("enabled"))
```
ctowns, line 60: These nits are more related to potential capability definition improvements in the future and are not blockers to this PR

nit: you could update the `threadBorderRouter` capability to include enum commands so that you can use `capabilities.threadBorderRouter.threadInterfaceState.enabled()` instead of passing a string, which I think provides stronger and clearer typing. You can see an example of these enum commands in the [switch](https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/main/capabilities/switch.src/switch.yml) and [lock](https://github.ecodesamsung.com/iot/SmartThingsCapabilities/blob/main/capabilities/lock.src/lock.yml) capability definitions

You can also utilize the `EnabledState` datatype for fields that just have an enabled/disabled state.

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,187 @@
...
+local function border_router_name_attribute_handler(driver, device, ib)
+  -- per the spec, the recommended attribute format is <VendorName> <ProductName>._meshcop._udp. This logic removes the meschop suffix IFF it is present
+  local meshCop_name = ib.data.value
+  local terminal_display_char = (string.find(meshCop_name, "._meshcop._udp") or 64) - 1 -- where 64-1=63, the maximum allowed length for BorderRouterName
+  local display_name = string.sub(meshCop_name, 1, terminal_display_char)
+  device:emit_event_for_endpoint(ib.endpoint, capabilities.threadBorderRouter.borderRouterName({ value = display_name }))
+end
+
+local function ssid_attribute_handler(driver, device, ib)
+  if ib.data.value == string.char(0x014) or ib.data.value == nil then -- Matter TLV-encoded NULL or Lua-encoded NULL
```
ctowns: nit: could you use the Null.ID value from `st.matter.data_types` instead of hard coding this?

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,187 @@
...
+  [1] = "1.0.0",
+  [2] = "1.1.0",
+  [3] = "1.2.0",
+  [4] = "1.3.0",
+  [5] = "1.4.0",
+}
+
+local function thread_version_attribute_handler(driver, device, ib)
+  local version_name = VERSION_TLV_MAP[ib.data.value]
+  if version_name then
```
ctowns, line 77: It might be worth adding a log here if the version name is not found

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,187 @@
...
+  [0] = threadNetwork.channel,
+  [1] = threadNetwork.panId,
+  [2] = threadNetwork.extendedPanId,
+  [3] = threadNetwork.networkName,
+  [5] = threadNetwork.networkKey,
+}
+
+local function dataset_response_handler(driver, device, ib)
+  if ib.status ~= im.InteractionResponse.Status.SUCCESS then
+    log.error("Failed to retrieve thread operational dataset")
```
ctowns: Should this be elevated to hub logs? Also I would use `device.log` in most cases throughout this function since it just automatically adds device info.

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,187 @@
...
+    return
+  elseif not ib.info_block.data.elements.dataset then
+    log.debug("Received empty thread operational dataset")
```
ctowns: What does it mean if this dataset is empty? Is this an error condition or is it acceptable to just exit and move on without trying another read? It may just be that if it fails, we can't expect to get correct information during this driver lifecycle, but I'm just unfamiliar with this dataset

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,187 @@
...
+
+--[[ COMMAND HANDLERS ]]--
+
+local threadNetwork = capabilities.threadNetwork
+local TLV_TYPE_ATTR_MAP = {
+  [0] = threadNetwork.channel,
+  [1] = threadNetwork.panId,
+  [2] = threadNetwork.extendedPanId,
+  [3] = threadNetwork.networkName,
+  [5] = threadNetwork.networkKey,
```
ctowns, line 103: Just confirming - the 4th index is supposed to be skipped here?

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,187 @@
...
+  while cur_byte + 1 <= operational_dataset_length do
+    local tlv_type = string.byte(operational_dataset, cur_byte)
+    local tlv_length = string.byte(operational_dataset, cur_byte + 1)
+    if (cur_byte + 1 + tlv_length) > operational_dataset_length then
+      log.error("Received thread operational dataset has a malformed TLV encoding")
+      return
+    end
+    local tlv_mapped_attr = TLV_TYPE_ATTR_MAP[tlv_type]
+    if tlv_mapped_attr then
+      local tlv_value = operational_dataset:sub(cur_byte + 2, cur_byte + 1 + tlv_length)
```
ctowns, line 132: I think including a comment related to the explanation that you mentioned [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2283/files#r2243572619) in this code will help make this easier to understand this code in the future, since the byte pointer offsets can be confusing to anyone unfamiliar with the TLV format.

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,187 @@
...
+end
+
+
+--[[ LIFECYCLE HANDLERS ]]--
+
+local function device_init(driver, device)
+  device:subscribe()
+  local tbrm_eps = embedded_cluster_utils.get_endpoints(device, clusters.ThreadBorderRouterManagement.ID)
+  if tbrm_eps and #tbrm_eps > 0 then
+    device:send(clusters.ThreadBorderRouterManagement.server.commands.GetActiveDatasetRequest(device, tbrm_eps[1]))
```
ctowns: Just confirming - is the active dataset expected to change at any point? Why is it read each driver startup, but not cached? And if it is expected to change, why not subscribe to this attribute?

ctowns (commented):
Great work Harrison! I appreciate the helpful comments that are included throughout, and the organization of the handlers if very clean :100:

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: yes, the fourth index is a part of the thread network information that we don't care about parsing.

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: yes, this dataset can change, and we can't subscribe to it because this is a command, not an attribute. Perhaps we could add some better logic to not re-read the data on each and every restart? Can look into this a bit more.

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: ticket made :+1:

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: If this dataset is empty, then either no dataset has been configured, or it is blocked from being shared by permissions on the device end.

In either case, one option would be to add some greater level of logic to re-profile and remove this now unused capability to ensure all fields we expose are being populated.

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: added

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: agree on both accounts, added. Actually, I think any failures in parsing the thread dataset should be elevated to hub logs.

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: agree, added

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: sure, added

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: fixed

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: Also, we may want to introduce some basic retry methods in the case that we get a Matter response failure.

drivers/SmartThings/matter-hrap/src/init.lua
ctowns: For clarity it might be worth just adding a comment here that the 4 index is intentionally skipped

drivers/SmartThings/matter-hrap/src/init.lua
ctowns: I find the comment to be confusing. I think just explaining why you skip the current byte and the following byte (Iength) would make it easier. Most of our TLV parsing is done in the lua libs, so I having it directly done in a driver like this could warrant just a little more explanation and clarity. A simple example in the comment could be helpful.

drivers/SmartThings/matter-hrap/src/init.lua
ctowns: It is strange that this could be expected to change, but the spec doesn't provide an event that we can subscribe to.  I not super familiar with HRAP stuff, so sorry if this is a dumb question - what is a scenario in which this value would change?

Based on a cursory read of the spec, it seems like the `ActiveDatasetTimestamp` attribute contains the most recent timestamp for the active dataset, and shall be updated when a new active dataset is configured. I think then we are supposed to monitor this attribute and then send the command to get the latest dataset when that attribute changes, or when we notice that the timestamp does not match the last timestamp that we cached in the case of a driver restart.

drivers/SmartThings/matter-hrap/src/init.lua
ctowns: We should be able to avoid the null dataset case by only sending the `GetActiveDatasetRequest` command if the `ActiveDatasetTimestamp` is not NULL accoridng to the spec:
> This attribute SHALL be null if the Thread Border Router has no dataset configured, otherwise it SHALL be the timestamp value extracted from the Active Dataset value configured by the Thread Node to which the border router is connected. This attribute SHALL be updated when a new Active dataset is configured on the Thread network to which the border router is connected

See my comment below for why I think the `ActiveDatasetTimestamp` should be an attribute that we subscribe to.

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,192 @@
...
+local function border_router_name_attribute_handler(driver, device, ib)
+  -- per the spec, the recommended attribute format is <VendorName> <ProductName>._meshcop._udp. This logic removes the MeshCoP suffix IFF it is present
+  local meshCop_name = ib.data.value
+  local terminal_display_char = (string.find(meshCop_name, "._meshcop._udp") or 64) - 1 -- where 64-1=63, the maximum allowed length for BorderRouterName
+  local display_name = string.sub(meshCop_name, 1, terminal_display_char)
+  device:emit_event_for_endpoint(ib.endpoint, capabilities.threadBorderRouter.borderRouterName({ value = display_name }))
+end
+
+local function ssid_attribute_handler(driver, device, ib)
+  if (tonumber(st_utils.bytes_to_hex_string(ib.data.value), 16) == data_types.Null.ID or ib.data.value == nil) then -- Matter TLV-encoded NULL or Lua-encoded NULL
```
ctowns: Couldn't this be the same as it was before but just using `data_types.Null.ID`?
```
ib.data.value == string.char(data_types.Null.ID)
``` 

This seems more straightforward to me than a conversion to a hexstring, then to a number.

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: kinda disagree, since there are a bunch of these that are skipped, they just come after?

ctowns: I am really just suggesting this:
```
  [1] = threadNetwork.panId,
  [2] = threadNetwork.extendedPanId,
  [3] = threadNetwork.networkName,
  -- [4] intentionally omitted
  [5] = threadNetwork.networkKey,
```

Because on first pass, it looked to me like this was a sequence of numbers that randomly skipped 1 value, and I didn't know if that was intentional or due to an off by one error during a refactor, or if it was something that was missed during testing. I would say that we should start to err on the side of overdocumenting than underdocumenting, and this is a small comment that just clarifies a little bit more in a case where my pattern seeking mind saw a break in the pattern and wondered why it happened :monkey:

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: to answer your first question, there are a few cases. One would be going from no thread network configured to one being configured, and to move between networks would occur in the case that a user wanted to merge their thread networks across multiples thread border routers, which is a major part of the thread unification initiative.

To the second point, I think that's a good idea and a cleaner solution than arbitrarily retrying on a restart.

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: ah, that's what you think is the issue? I discussed offline with Nick about the comment you quoted and he was having a separate issue with understanding. maybe you could write up an example comment you'd like added?

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: agree, fixed

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: added in latest commit

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: added better comment

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: good point, very simple. added

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: agree, done

drivers/SmartThings/matter-hrap/src/init.lua
nand-nor: Agree with @ctowns, we can subscribe to the active timestamp attribute, which is a reliable indicator of dataset change. This is also how we track changes within hub-core

drivers/SmartThings/matter-hrap/src/init.lua
```diff
@@ -0,0 +1,219 @@
...
+  else
+    device.log.warn("The received TLV-encoded Thread version does not have a provided mapping to a human-readable version format")
+  end
+end
+
+local function active_dataset_timestamp_handler(driver, device, ib)
+  if not ib.data.value then
+    device.log.info("No Thread operational dataset configured")
+  elseif ib.data.value ~= device:get_field(CURRENT_ACTIVE_DATASET_TIMESTAMP) then
+    local tbrm_ep = embedded_cluster_utils.get_endpoints(device, clusters.ThreadBorderRouterManagement.ID)[1]
```
ctowns: I think we could use use the `ib.endpoint_id` instead of getting the endpoint that supports border router management, since that should be the same endpoint that provided this interacton block.

ctowns (approved):
Just one more small comment, otherwise this looks good to me!

drivers/SmartThings/matter-hrap/src/init.lua
hcarter-775: for sure, moved it thoughtlessly out of device_init where this handling was required.

PR #2292, opened 2025-07-29
CHAD-15785 Z-Wave: WindowShadePreset capability updates

greens:


drivers/SmartThings/zwave-window-treatment/src/springs-window-fashion-shade/init.lua
```diff
@@ -37,6 +37,16 @@ local function can_handle_springs_window_fashion_shade(opts, driver, device, ...
   return false
 end
 
+local function init_handler(self, device)
```
inasail, line 40: I think this sub driver doesn't need to be modified because it doesn't have any UI(like preference setting) related with 'presetPosition'.

greens: This device allows the user to set a preset position with the included remote that can then be recalled via a `MultiLevelSwitch:0xFF` command. I did this to be certain that `setPresetPosition` is not supported, as it would be if I did not override the base driver's `init` handler.

inasail (approved):
Looks good to me.

PR #2295, opened 2025-07-29
Matter Window Covering: Update preset handling

nickolas-deboom:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change utilizes the setPresetPosition command that was recently added to the windowShadePreset capability to set the preset rather than using an embedded preference.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-battery.yml
+++ b/drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-battery.yml
@@ -19,7 +19,5 @@ components:
   categories:
   - name: Blind
 preferences:
-  - preferenceId: presetPosition
-    explicit: true
   - preferenceId: reverse
     explicit: true
--- a/drivers/SmartThings/matter-window-covering/src/init.lua
+++ b/drivers/SmartThings/matter-window-covering/src/init.lua
@@ -28,6 +28,8 @@ local battery_support = {
   BATTERY_PERCENTAGE = "BATTERY_PERCENTAGE"
 }
 local REVERSE_POLARITY = "__reverse_polarity"
+local PRESET_LEVEL_KEY = "__preset_level_key"
+local PRESET_LEVEL = 50
 
 local function find_default_endpoint(device, cluster)
   local res = device.MATTER_DEFAULT_ENDPOINT
@@ -76,6 +78,17 @@ end
 
 local function device_init(driver, device)
   device:set_component_to_endpoint_fn(component_to_endpoint)
+  if device:supports_capability_by_id(capabilities.windowShadePreset.ID) and
+    device:get_latest_state("main", capabilities.windowShadePreset.ID, capabilities.windowShadePreset.position.NAME) == nil then
+    -- These should only ever be nil once (and at the same time) for already-installed devices
+    -- It can be removed after migration is complete
+    device:emit_event(capabilities.windowShadePreset.supportedCommands({"presetPosition", "setPresetPosition"}, {visibility = {displayed = false}}))
+    local preset_position = device:get_field(PRESET_LEVEL_KEY) or
+      (device.preferences ~= nil and device.preferences.presetPosition) or
+      PRESET_LEVEL
+    device:emit_event(capabilities.windowShadePreset.position(preset_position, {visibility = {displayed = false}}))
+    device:set_field(PRESET_LEVEL_KEY, preset_position, {persist = true})
+  end
   device:subscribe()
 end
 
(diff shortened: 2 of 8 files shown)


drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -115,9 +128,12 @@ local function info_changed(driver, device, event, args)
...
 local function device_added(driver, device)
-  device:emit_event(
-    capabilities.windowShade.supportedWindowShadeCommands({"open", "close", "pause"}, {visibility = {displayed = false}})
-  )
+  device:emit_event(capabilities.windowShade.supportedWindowShadeCommands({"open", "close", "pause"}, {visibility = {displayed = false}}))
+  device:emit_event(capabilities.windowShadePreset.supportedCommands({"presetPosition", "setPresetPosition"}, {visibility = {displayed = false}}))
+  if device:supports_capability_by_id(capabilities.windowShadePreset.ID) and
+    device:get_latest_state("main", capabilities.windowShadePreset.ID, capabilities.windowShadePreset.position.NAME) == nil then
+    device:emit_event(capabilities.windowShadePreset.position(PRESET_LEVEL, {visibility = {displayed = false}}))
+  end
```
greens: I ultimately decided to omit the additions in the `added` handler for now, since these events will be sent by `init`. After migration is complete we can remove the `init` code and add the `added` events.

nickolas-deboom: Sounds good, updated in the latest commit!

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -129,7 +129,21 @@ local CLUSTER_SUBSCRIBE_LIST_NO_BATTERY = {
...
+    device:generate_test_message(
+      "main", capabilities.windowShadePreset.supportedCommands({"presetPosition", "setPresetPosition"}, {visibility = {displayed = false}})
+    )
+  )
+  test.socket.capability:__expect_send(
+    device:generate_test_message(
+      "main", capabilities.windowShadePreset.position(30, {visibility = {displayed = false}})
+    )
+  )
+end
```
greens, line 112: I'd maybe extend this test just to verify that the updated value is used for the next `presetPosition` command

nickolas-deboom: Good call, I updated the test case called 'Handle preset commands' to verify this behavior in the latest commit

drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-battery.yml
```diff
@@ -19,7 +19,5 @@ components:
   categories:
   - name: Blind
 preferences:
-  - preferenceId: presetPosition
```
hcarter-775, line 22: what was the thinking with getting rid of the preference? this seems like a kinda breaking change for the the user, so I just want to see where this ask is coming from?

nickolas-deboom: The preset can now be set by the `setPresetPosition` command within the `windowShadePreset` capability rather than using this preference.

drivers/SmartThings/matter-window-covering/src/init.lua
hcarter-775: Might want to actually move it into `driverSwitched` rather than `init` since that's a more specific handler.

nickolas-deboom: Existing devices wouldn't get a driverSwitched event unless they switched to a different driver, right?

hcarter-775: A driver update (like us releasing this into prod) will trigger a `driverSwitched` event because this will be a new driver. I've been thinking about this and that handler may be a more natural handler for these "1 time-ish" reconfiguration changes.

Init simply happens anytime a driver starts up, so it would trigger more often technically (like anytime a driver restarts)

nickolas-deboom: I don't think that's true that a driverSwitched event is triggered in this case, otherwise we would see a profile update on every driver release for example https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/4d3ae9aa57c56e48d67baecc1b3323bbaccbf29c/drivers/SmartThings/matter-switch/src/init.lua#L60-L64

hcarter-775: As you know, the `match_profile` function does not always trigger a `try_update_metadata` call, whether it because profiling data is required that gates the functionality or because some devices (lights, switches for example) do not get profile updates. However, I have confirmed in my logs that `driverSwitched` is indeed triggered after a driver release, since a new driver is replacing an old driver.

hcarter-775: Further, a profile update being triggered every driver release is something we very well may see in some cases- we currently don't have any tracking for that kind of thing.

nickolas-deboom: Hmm, I don't see any instances of `driverSwitched` from my hub logs following Monday's driver release. From everything I've seen it's only triggered following an actual driver switch, not just an updated driver. Could you share your logs with me? I'm just wondering why it would be different or why I wouldn't be seeing the same thing

hcarter-775: Really? Yeah I did see it... let me double check. Just to confirm, the hub you checked has drivers running on it that would have been updated from this latest release?

nickolas-deboom: Yes, I'm running the prod ST drivers at home, and after the inventory sync when the new set of drivers were pulled down, I only see `init` events.

hcarter-775: This discussion can be tabled for the moment I believe. I would like to look more closely into this, but it should not block the PR as we would like to get it merged soon.

PR #2302, opened 2025-07-29
fix(sonos): Improved handling of bonded set membership changes.

dljsjr:
# Type of Change

- [x] Bug fix

# Checklist

# Description of Change

We have historically ignored non-primary devices in bonded sets during discovery/onboarding, going back to the pre-Edge days. 

Bonded sets are things like stereo pairs or home theater setups.

These devices cannot be controlled via the WSS LAN API at all, and they aren't treated as being part of a Group in the Sonos group model; the intent is for API consumers to treat the entire bonded set as a single "player". Only one of the devices in the set exposes an API endpoint, which differs from a Group where all the players expose an endpoint. While groups service most commands via calling the API endpoint on the Group Coordinator, certain commands like volume can be sent to non-coordinator players. This is not the case with bonded sets; the entire set always acts atomically, and only the primary device in the set can service commands. Hence, we treat non-primary devices as not controllable in a bonded set, and we don't onboard them.

However, we didn't really have very robust handling of devices that become part of a bonded set at runtime after onboarding. This led to unnecessary CPU and RAM usage by creating a few tight loops due to unexpected failure/edge cases when making certain calls.

Because we would fail to discover this device under normal circumstances, the preferred way to handle this transition is to mark the device as being offline. We discussed emitting a delete for these devices, but it seems that it wouldn't be too uncommon for some users to create and destroy bonded sets ephemerally the way one might do with a Group. For this reason we decided we would avoid deleting the records.

If a non-primary in a bonded set is removed from the bonded set, it will be marked as online again.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/sonos/src/api/event_handlers.lua
+++ b/drivers/SmartThings/sonos/src/api/event_handlers.lua
@@ -1,7 +1,11 @@
 local capabilities = require "st.capabilities"
+local swGenCapability = capabilities["stus.softwareGeneration"]
+
 local log = require "log"
 
 local st_utils = require "st.utils"
+
+local PlayerFields = require "fields".SonosPlayerFields
 
 local CapEventHandlers = {}
 
@@ -12,41 +16,52 @@ CapEventHandlers.PlaybackStatus = {
   Playing = "PLAYBACK_STATE_PLAYING",
 }
 
+local function _do_emit(device, attribute_event)
+  local bonded = device:get_field(PlayerFields.BONDED)
+  if not bonded then
+    device:emit_event(attribute_event)
+  end
+end
+
+function CapEventHandlers.handle_sw_gen(device, sw_gen)
+  _do_emit(device, swGenCapability.generation(string.format("%s", sw_gen)))
+end
+
 function CapEventHandlers.handle_player_volume(device, new_volume, is_muted)
-  device:emit_event(capabilities.audioVolume.volume(new_volume))
+  _do_emit(device, capabilities.audioVolume.volume(new_volume))
   if is_muted then
-    device:emit_event(capabilities.audioMute.mute.muted())
+    _do_emit(device, capabilities.audioMute.mute.muted())
   else
-    device:emit_event(capabilities.audioMute.mute.unmuted())
+    _do_emit(device, capabilities.audioMute.mute.unmuted())
   end
 end
 
 function CapEventHandlers.handle_group_volume(device, new_volume, is_muted)
-  device:emit_event(capabilities.mediaGroup.groupVolume(new_volume))
+  _do_emit(device, capabilities.mediaGroup.groupVolume(new_volume))
   if is_muted then
-    device:emit_event(capabilities.mediaGroup.groupMute.muted())
+    _do_emit(device, capabilities.mediaGroup.groupMute.muted())
   else
-    device:emit_event(capabilities.mediaGroup.groupMute.unmuted())
+    _do_emit(device, capabilities.mediaGroup.groupMute.unmuted())
   end
 end
 
 function CapEventHandlers.handle_group_role_update(device, group_role)
-  device:emit_event(capabilities.mediaGroup.groupRole(group_role))
+  _do_emit(device, capabilities.mediaGroup.groupRole(group_role))
 end
 
 function CapEventHandlers.handle_group_coordinator_update(device, coordinator_id)
-  device:emit_event(capabilities.mediaGroup.groupPrimaryDeviceId(coordinator_id))
+  _do_emit(device, capabilities.mediaGroup.groupPrimaryDeviceId(coordinator_id))
 end
 
 function CapEventHandlers.handle_group_id_update(device, group_id)
-  device:emit_event(capabilities.mediaGroup.groupId(group_id))
+  _do_emit(device, capabilities.mediaGroup.groupId(group_id))
 end
 
 function CapEventHandlers.handle_group_update(device, group_info)
   local groupRole, groupPrimaryDeviceId, groupId = table.unpack(group_info)
-  device:emit_event(capabilities.mediaGroup.groupRole(groupRole))
-  device:emit_event(capabilities.mediaGroup.groupPrimaryDeviceId(groupPrimaryDeviceId))
-  device:emit_event(capabilities.mediaGroup.groupId(groupId))
+  _do_emit(device, capabilities.mediaGroup.groupRole(groupRole))
+  _do_emit(device, capabilities.mediaGroup.groupPrimaryDeviceId(groupPrimaryDeviceId))
+  _do_emit(device, capabilities.mediaGroup.groupId(groupId))
 end
 
 function CapEventHandlers.handle_audio_clip_status(device, clips)
@@ -61,11 +76,11 @@ end
 
 function CapEventHandlers.handle_playback_status(device, playback_state)
   if playback_state == CapEventHandlers.PlaybackStatus.Playing then
-    device:emit_event(capabilities.mediaPlayback.playbackStatus.playing())
+    _do_emit(device, capabilities.mediaPlayback.playbackStatus.playing())
   elseif playback_state == CapEventHandlers.PlaybackStatus.Idle then
-    device:emit_event(capabilities.mediaPlayback.playbackStatus.stopped())
+    _do_emit(device, capabilities.mediaPlayback.playbackStatus.stopped())
   elseif playback_state == CapEventHandlers.PlaybackStatus.Paused then
-    device:emit_event(capabilities.mediaPlayback.playbackStatus.paused())
+    _do_emit(device, capabilities.mediaPlayback.playbackStatus.paused())
   elseif playback_state == CapEventHandlers.PlaybackStatus.Buffering then
     -- TODO the DTH doesn't currently do anything w/ buffering;
     -- might be worth figuring out what to do with this in the future.
@@ -74,7 +89,7 @@ function CapEventHandlers.handle_playback_status(device, playback_state)
 end
 
 function CapEventHandlers.update_favorites(device, new_favorites)
-  device:emit_event(capabilities.mediaPresets.presets(new_favorites))
+  _do_emit(device, capabilities.mediaPresets.presets(new_favorites))
 end
 
 function CapEventHandlers.handle_playback_metadata_update(device, metadata_status_body)
@@ -128,7 +143,7 @@ function CapEventHandlers.handle_playback_metadata_update(device, metadata_statu
   end
 
   if type(audio_track_data.title) == "string" then
-    device:emit_event(capabilities.audioTrackData.audioTrackData(audio_track_data))
+    _do_emit(device, capabilities.audioTrackData.audioTrackData(audio_track_data))
   end
 end
 
--- a/drivers/SmartThings/sonos/src/api/sonos_connection.lua
+++ b/drivers/SmartThings/sonos/src/api/sonos_connection.lua
@@ -166,8 +166,12 @@ local function _open_coordinator_socket(sonos_conn, household_id, self_player_id
       return
     end
 
-    _, err =
-      Router.open_socket_for_player(household_id, coordinator_id, coordinator.websocketUrl, api_key)
+    _, err = Router.open_socket_for_player(
+      household_id,
+      coordinator_id,
+      coordinator.player.websocketUrl,
+      api_key
+    )
     if err ~= nil then
       log.error(
         string.format(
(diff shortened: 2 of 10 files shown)


drivers/SmartThings/sonos/src/api/event_handlers.lua
```diff
@@ -12,41 +14,48 @@ CapEventHandlers.PlaybackStatus = {
   Playing = "PLAYBACK_STATE_PLAYING",
 }
 
+local function _do_emit(device, attribute_event)
```
dljsjr, line 19: The reason we collapsed everything down to this helper function is because certain activity payloads that come in on a single speaker will contain information about all Players in a group/set/zone/etc., and we will dispatch to all appropriate device records that can be associated with one of the payload entries.

Emitting information about that speaker if it's known to be part of a bonded set would lead to the device record being marked as Online in certain situations, which we don't want.

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -42,11 +42,27 @@ local ONE_HOUR_IN_SECONDS = 3600
...
 ---@field private startup_state_received boolean
 ---@field private devices_waiting_for_startup_state SonosDevice[]
+---@field package bonded_devices table<string, boolean> map of Device device_network_id to a boolean indicating if the device is currently known as a bonded device.
 ---
 ---@field public ssdp_task SonosPersistentSsdpTask?
 ---@field private ssdp_event_thread_handle table?
 local SonosDriver = {}
 
+---@param device SonosDevice
+function SonosDriver:update_bonded_device_tracking(device)
```
varzac, line 52: The logic in this function is very confusing to me, can you explain what is intended to be tracked here?

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -483,12 +505,22 @@ function SonosDriver:handle_player_discovery_info(api_key, info, device)
   -- speaker in a bonded set (e.g. a home theater system, a stereo pair, etc).
   -- These aren't the same as speaker groups, and bonded speakers can't be controlled
   -- via websocket at all. So we ignore all bonded non-primary speakers
-  if #info.ssdp_info.group_id == 0 then
-    return nil,
-      string.format(
-        "Player %s is a non-primary bonded Sonos device, ignoring",
-        info.discovery_info.device.name
-      )
+  -- if  then
```
varzac: Commented code

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -358,19 +365,28 @@ function SonosConnection.new(driver, device)
         local household_id, current_coordinator =
           self.driver.sonos:get_coordinator_for_device(self.device)
         local _, player_id = self.driver.sonos:get_player_for_device(self.device)
-        self.driver.sonos:update_household_info(header.householdId, body, self.device)
+        self.driver.sonos:update_household_info(header.householdId, body, self.driver)
```
varzac, line 368: this change seems odd, but looking at the function definition change, it looks like the device was just an extra arg before?

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -358,19 +365,28 @@ function SonosConnection.new(driver, device)
...
         self.driver.sonos:update_device_record_from_state(header.householdId, self.device)
         local _, updated_coordinator = self.driver.sonos:get_coordinator_for_device(self.device)
 
+        local bonded = self.device:get_field(PlayerFields.BONDED)
+        if bonded then
```
tpmanley, line 373: Could this be combined with the `if not bonded` below just to simplify the logic?

drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
```diff
@@ -300,7 +300,7 @@ function sonos_ssdp.spawn_persistent_ssdp_task()
     if info_to_send then
       if not (info_to_send.discovery_info and info_to_send.discovery_info.device) then
         log.error_with(
-          { hub_logs = true },
+          { hub_logs = false },
```
tpmanley, line 303: Is there a reason we don't want to include this in hub logs?

drivers/SmartThings/sonos/src/lifecycle_handlers.lua
```diff
@@ -164,10 +164,12 @@ function SonosDriverLifecycleHandlers.initialize_device(driver, device)
                   return
                 end
                 log.error_with(
-                  { hub_logs = true },
-                  "Error handling Sonos player initialization: %s, error code: %s",
-                  error,
-                  (error_code or "N/A")
+                  { hub_logs = false },
```
tpmanley, line 167: Same question about not including this in hub logs

drivers/SmartThings/sonos/src/sonos_driver.lua
dljsjr: The intent is to track when a device record's bonded status changes.

The device's current bonded status is tracked on the device itself via the `BONDED` field, which we always update as soon as possible so that we can do stuff like short circuiting emits so that we don't mark an offline device as online.

The driver shadows this information in a map that is updated independently and with less urgency, so that it can be used to look up the previous state before updating itself.

This lets us determine if a device's bonded status changes. If a device becomes bonded, we mark it offline. If a device was bonded and is no longer bonded, we re-initialize it.

drivers/SmartThings/sonos/src/api/sonos_connection.lua
dljsjr: Yeah, this looks weird on the surface.

*before* the OAuth refactor, this method would conditionally take a device record as a flag that it needed to perform some updates. Which was a bad heuristic, so when I refactored it, I got rid of that heuristic, but forgot to remove it from the call sites; Lua being Lua, it didn't throw any errors or cause any problems.

With the refactor for the bonded sets, we changed it so that it *always* takes a driver record to be able to do some tracking of bonded/unbonded state transitions, hence the updates to the call sites.

drivers/SmartThings/sonos/src/api/sonos_connection.lua
dljsjr: Not easily; we want to *conditionally* call `:stop()`, but then *always* call `cleanup_unused_sockets`, and we want to `stop` the bonded player entirely before calling that function.

The reason being is that `cleanup_unused_sockets` executes the cleanup logic for when the set of known Coordinators change, so that we get rid of certain namespace subscriptions and connections if a player that was a coordinator is no longer a coordinator.

This cleanup logic is applicable on any Group membership update for the topology regardless of whether or not bonded set membership shifts.

After we do the cleanup, we then want to validate our own coordinator connection (if our coordinator changed) and subscriptions only in the case that we're not part of a bonded set.

drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
dljsjr: It can be chatty, and we've been asked by trim down the amount of logging that the LAN Edge Drivers forward to hub-core. It was set to `true` during initial development of the refactor for debugging purposes but doesn't really need to be that way anymore. With the ability to set the `broker.promoted_driver_logs` flag in LD to promote all logs for a driver, there isn't as much need for this in general unless it's on a path that is *really* pathological, IMO.

drivers/SmartThings/sonos/src/lifecycle_handlers.lua
dljsjr: This can be chatty as well, and based on logs that I've looked at from the fleet, it's almost always only hit by non-primary bonded players, so it's not providing useful information.

cjswedes (approved):

drivers/SmartThings/sonos/src/sonos_state.lua
```diff
@@ -23,6 +22,9 @@ function _household_mt:reset()
   self.groups = utils.new_case_insensitive_table()
   self.players = utils.new_case_insensitive_table()
   self.player_to_group = utils.new_case_insensitive_table()
+  if not self.bonded_players then
```
hcarter-775, line 27: nit: maybe add a comment clarifying the distinction here like `-- previously bonded devices should not be un-bonded after a reset since these should not be treated as distinct devices`

drivers/SmartThings/sonos/src/sonos_state.lua
```diff
@@ -100,8 +102,11 @@ function SonosState:associate_device_record(device, info)
     return
   end
 
-  local group = household.groups[group_id]
+  if not group_id or #group_id == 0 then
+    group_id = household.player_to_group[player_id or ""] or ""
+  end
```
hcarter-775, line 109: why even set `group_id` initially if we can do this?

drivers/SmartThings/sonos/src/sonos_state.lua
```diff
@@ -112,9 +117,11 @@ function SonosState:associate_device_record(device, info)
     return
   end
 
-  local player = household.players[player_id]
+  local player_tbl = household.players[player_id]
+  local player = (player_tbl or {}).player
+  local sonos_device = (player_tbl or {}).device
```
hcarter-775: nit:
```suggestion
  local player = (household.players[player_id] or {}).player
  local sonos_device = (household.players[player_id] or {}).device
```

drivers/SmartThings/sonos/src/sonos_state.lua
```diff
@@ -124,15 +131,24 @@ function SonosState:associate_device_record(device, info)
     return
   end
 
-  household.st_devices[player.id] = device.id
+  household.st_devices[sonos_device.id] = device.id
```
hcarter-775, line 135: just curious, if these both map to the playerId, why even change this? Mostly curious if it's possible for these values to be different, and what that might mean. Should we add a check for that?

drivers/SmartThings/sonos/src/sonos_state.lua
```diff
@@ -221,9 +251,13 @@ function SonosState:update_device_record_group_info(household, group, device)
...
     { persist = true }
   )
-  if field_changed then
+  if not bonded and field_changed then
     CapEventHandlers.handle_group_coordinator_update(device, group.coordinatorId)
   end
+
+  if bonded then
+    device:offline()
+  end
```
hcarter-775, line 261: with this and other places, maybe it would make more sense to move this to the top and early return? could avoid issues in the future with this arbitrary-ish status.

drivers/SmartThings/sonos/src/sonos_state.lua
```diff
@@ -283,25 +319,45 @@ function SonosState:update_household_info(id, groups_event)
...
+
+  for _, player in ipairs(players) do
+    for _, device in ipairs(player.devices) do
+      household.players[device.id] = { player = player, device = device }
+      local previously_bonded = known_bonded_players[device.id] and true or false
+      local currently_bonded
+      local group_id
+      -- non-primary bonded players are excluded from a group's list of PlayerID's so we use the group membership
+      -- of the primary device
+      if type(device.primaryDeviceId) == "string" and #device.primaryDeviceId > 0 then
```
hcarter-775: nit:
```suggestion
      if type(device.primaryDeviceId) == "string" and device.primaryDeviceId ~= "" then
```
might be more clear

NoahCornell:
The changes here look good to me. I can get another speaker and do some testing so that we can merge this.

I'll work on getting a test plan that would cover this and my sonos PR so that thinktank can run through both set of changes in alpha.

hcarter-775 (approved):

drivers/SmartThings/sonos/src/sonos_state.lua
NoahCornell: It looks like we still want to update the device fields in this function if they have changed so I don't think we can return early

drivers/SmartThings/sonos/src/sonos_state.lua
NoahCornell: I believe this can be different. I think the player id would be the same for a stereo pair (bonded devices) but the device id would be unique for each of them

drivers/SmartThings/sonos/src/sonos_state.lua
NoahCornell: I think this should only be case if the device is bonded and in most cases it would be better to rely on what we just got from the device instead of our cached value

NoahCornell:
Now that the driver release is complete, I am planning on merging this at the end of the day so that we can have thinktank run through the regression tests unless there are any objections

PR #2303, opened 2025-07-29
fix: spurious oauth token requests when we already have a valid token.

dljsjr:
# Type of Change

- [x] Bug fix

# Checklist

# Description of Change

There were a handful of situations where the driver was making a request to the cloud for a new copy of the user's Sonos OAuth token, even if the token was cached on the hub and had not yet expired. This attempts to address that.

# Summary of Completed Tests


drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -184,11 +184,15 @@ function SonosDriver:notify_augmented_data_changed(update_kind, update_key, upda
     )
   end
 
-  if
-    self.oauth.endpoint_app_info
+  local maybe_token, _ = self:get_oauth_token()
```
varzac: Doesn't need to be in this PR, but I was initially a bit confused, and I think this funciton would be more descriptive if it was `get_cached_oauth_token` as at first read I assumed this was making to request.

dljsjr: :ack:

dljsjr:
Closing in favor of the solution merged in #2351

PR #2306, opened 2025-07-30
chore: Improve Sonos Memory Usage

dljsjr:
# Type of Change

- [x] Refactor

# Checklist

## Depends On

- [ ] #2302 
- [ ] #2303 

# Description of Change

@NoahCornell did some analysis of driver heap usage, and Sonos was
disproportionately above the mean for the heap-space-per-device line fit
on the dataset.

This PR improves memory usage by limiting the information we keep resident
from API responses to the fields that we actually use in regular operation.

Additional background:

One source of heap usage was unbounded memory growth due to task spawning
on a certain error pathway, which he already fixed in a different PR.

Another source was the fact that we were cacheing the entire API
response for the Player and Group information for the LAN's Sonos topology.
These API payloads are pretty large, and have actually gotten larger over time.

This has been the case for this driver for a very long time; the
decision to store the whole response object was made so that we would
have the information available if we needed it in the future for bug
fixes or enhancements.

It turns out that the information we're utilizing hasn't really changed
much over the last few years, so I'm feeling quite comfortable about
excising the majority of the payload information at this point. We see
pretty signifcant memory savings with these changes, and the savings
should scale appreciably with device count, which is a big win.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
+++ b/drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
@@ -9,7 +9,140 @@ local utils = require "utils"
 local SonosApi = require "api"
 local SSDP_SCAN_INTERVAL_SECONDS = 600
 
+local SONOS_DEFAULT_PORT = 1443
+local SONOS_DEFAULT_WSS_PATH = "websocket/api"
+local SONOS_DEFAULT_REST_PATH = "api/v1"
+
+--- Cached information gathered during discovery scanning, created from a subset of the
+--- found [SonosSSDPInfo](lua://SonosSSDPInfo) and [SonosDiscoveryInfoObject](lua://SonosDiscoveryInfoObject)
+---
+--- @class SpeakerDiscoveryInfo
+--- @field public unique_key UniqueKey
+--- @field public mac_addr string
+--- @field public expires_at integer
+--- @field public ipv4 string
+--- @field public port integer
+--- @field public household_id HouseholdId
+--- @field public player_id PlayerId
+--- @field public name string
+--- @field public model string
+--- @field public model_display_name string
+--- @field public sw_gen integer
+--- @field public wss_url table
+--- @field public rest_url table
+--- @field public is_group_coordinator boolean
+--- @field public group_name string? nil if a speaker is the non-primary in a bonded set
+--- @field public group_id GroupId? nil if a speaker is the non-primary in a bonded set
+--- @field package wss_path string? nil if equivalent to the default value; does not include leading slash!
+--- @field package rest_path string? nil if equivalent to the default value; does not include leading slash!
+local SpeakerDiscoveryInfo = {}
+
+---@param ssdp_info SonosSSDPInfo
+---@param discovery_info SonosDiscoveryInfoObject
+---@return SpeakerDiscoveryInfo info
+function SpeakerDiscoveryInfo.new(ssdp_info, discovery_info)
+  local mac_addr = utils.extract_mac_addr(discovery_info.device)
+  local port, rest_path = string.match(discovery_info.restUrl, "^.*:(%d*)/(.*)$")
+  local _, wss_path = string.match(discovery_info.websocketUrl, "^.*:(%d*)/(.*)$")
+  port = tonumber(port) or SONOS_DEFAULT_PORT
+
+  local ret = {
+    unique_key = utils.sonos_unique_key_from_ssdp(ssdp_info),
+    expires_at = ssdp_info.expires_at,
+    ipv4 = ssdp_info.ip,
+    port = port,
+    mac_addr = mac_addr,
+    household_id = ssdp_info.household_id,
+    player_id = ssdp_info.player_id,
+    name = discovery_info.device.name,
+    model = discovery_info.device.model,
+    model_display_name = discovery_info.device.modelDisplayName,
+    sw_gen = discovery_info.device.swGen,
+    is_group_coordinator = ssdp_info.is_group_coordinator,
+  }
+
+  if type(ssdp_info.group_name) == "string" and #ssdp_info.group_name > 0 then
+    ret.group_name = ssdp_info.group_name
+  end
+
+  if type(ssdp_info.group_id) == "string" and #ssdp_info.group_id > 0 then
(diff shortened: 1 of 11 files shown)


drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
```diff
@@ -9,7 +9,140 @@ local utils = require "utils"
...
+            ret.wss_path or SONOS_DEFAULT_WSS_PATH
+          )
+        )
+      )
+    end
+
+    return rawget(ret, k)
+  end
+
+  local proxy_newindex = function(_, _, _)
```
NoahCornell: Just calling out that this only prevents new keys from being set and will still allow modification of current values. I think this is fine since this is just our driver.

drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
```diff
@@ -9,7 +9,140 @@ local utils = require "utils"
...
+    error("attempt to index a read-only table", 2)
+  end
+
+  for k, v in pairs(SpeakerDiscoveryInfo) do
+    rawset(ret, k, v)
+  end
+
+  return setmetatable(ret, { __index = proxy_index, __newindex = proxy_newindex })
```
NoahCornell, line 122: We can avoid a new metatable and `proxy_index` `proxy_newindex` lua closures from being created each time by defining these outside of the function and by using `self` instead of passing in the `ret` upvalue. Not exactly sure how long these are expected to live but if we have multiple at once then it would save some memory.

drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
```diff
@@ -9,7 +9,140 @@ local utils = require "utils"
...
+    return rawget(ret, k)
+  end
+
+  local proxy_newindex = function(_, _, _)
+    error("attempt to index a read-only table", 2)
+  end
+
+  for k, v in pairs(SpeakerDiscoveryInfo) do
+    rawset(ret, k, v)
+  end
```
NoahCornell, line 120: I think if we define the metatable outside of the function (per comment below this) then `SpeakerDiscoveryInfo` could be the metatable and `__index` could be updated to handle the methods. I don't think it really matters though and this is fine.

drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
```diff
@@ -267,11 +400,12 @@ function sonos_ssdp.spawn_persistent_ssdp_task()
     local maybe_known = task_handle.player_info_by_sonos_ids[unique_key]
     local is_new_information = not (
       maybe_known
-      and maybe_known.ssdp_info.expires_at > os.time()
-      and sonos_ssdp.ssdp_info_eq(maybe_known.ssdp_info, sonos_ssdp_info)
+      and maybe_known.expires_at > os.time()
+      and sonos_ssdp.ssdp_info_eq(maybe_known:as_ssdp_info(), sonos_ssdp_info)
```
NoahCornell: Rather than creating a new table with `as_ssdp_info()` could we just create a new equality checker specifically for `SpeakerDiscoveryInfo` and `SonosSSDPInfo`

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -262,6 +255,11 @@ function SonosDriver:check_auth(info_or_device)
       rest_url = net_url.parse(info_or_device:get_field(PlayerFields.REST_URL))
       household_id = self.sonos:get_sonos_ids_for_device(info_or_device)
       sw_gen = info_or_device:get_field(PlayerFields.SW_GEN)
+    else
+      ---@cast info_or_device SpeakerDiscoveryInfo
+      rest_url = net_url.parse(info_or_device.rest_url)
```
NoahCornell: I don't think this `net_url.parse` is needed since we do that in `__index`

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -527,11 +514,11 @@ function SonosDriver:handle_player_discovery_info(api_key, info, device)
 
   api_key = api_key or self:get_fallback_api_key()
 
-  local rest_url = net_url.parse(info.discovery_info.restUrl)
+  local rest_url = net_url.parse(info.rest_url)
```
NoahCornell: Here as well for the `net_url.parse`

drivers/SmartThings/sonos/src/sonos_state.lua
```diff
@@ -306,18 +305,30 @@ function SonosState:update_device_record_from_state(household_id, device)
...
   local currently_bonded
   local group_id
-
-  -- The primary bonded player will have the same id as the top level player id
+  -- non-primary bonded players are excluded from a group's list of PlayerID's so we use the group membership
+  -- of the primary device
   if sonos_device_id == player.id then
-    currently_bonded = false
-  else
     currently_bonded = true
```
NoahCornell: Swapping these is contrary to my understanding but I might be incorrect. My thought was that if the player ID was equal to the Sonos device ID then that means that this device is the primary device of the bonded set. Other devices included in `devices` or `deviceIds` won't match the player ID and aren't the primary device. Then my thinking was `currently_bonded` is only true if it is not the primary bonded player.

dljsjr: This was a mistake when backporting the fixes from your PR during the rebase, good catch.

NoahCornell: Ah okay I was worried that I had messed something up with the bonded stuff :sweat_smile:

drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
dljsjr: That's deliberate.

drivers/SmartThings/sonos/src/api/sonos_ssdp_discovery.lua
dljsjr: It seemed more straight forward to do it this way and still have the lazily evaluated URL keys.

PR #2310, opened 2025-07-31
58 unit test fixes

cjswedes:
This contains unit test fixes for the following features that are changing in 0.58:
* Integration test lifecycle events are now more explicit. The test framework will queue init (and doConfigure if the mock device overrides the provisioning state) events for mock devices that are added in test init.
* Matter is using modular profiles
* New native handler registrations are available
* Reporting configurations for zigbee power meter are lengthened by default.

### DIFF
--- a/drivers/SmartThings/matter-button/src/test/test_matter_multi_button.lua
+++ b/drivers/SmartThings/matter-button/src/test/test_matter_multi_button.lua
@@ -73,35 +73,47 @@ local CLUSTER_SUBSCRIBE_LIST ={
   clusters.Switch.server.events.MultiPressComplete,
 }
 
+-- All messages queued and expectations set are done before the driver is actually run
 local function test_init()
+  -- we dont want the integration test framework to generate init/doConfigure, we are doing that here
+  -- so we can set the proper expectations on those events.
+  test.disable_startup_messages()
+  test.mock_device.add_test_device(mock_device) -- make sure the cache is populated
+
+  -- added results in a profile update
+  test.socket.device_lifecycle:__queue_receive({ mock_device.id, "added" })
+  mock_device:expect_metadata_update({ profile = "4-button-battery" })
+
+  -- init results in subscription interaction
   local subscribe_request = CLUSTER_SUBSCRIBE_LIST[1]:subscribe(mock_device)
   for i, clus in ipairs(CLUSTER_SUBSCRIBE_LIST) do
     if i > 1 then subscribe_request:merge(clus:subscribe(mock_device)) end
   end
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
-  test.mock_device.add_test_device(mock_device)
-  test.socket.device_lifecycle:__queue_receive({ mock_device.id, "added" })
-  mock_device:expect_metadata_update({ profile = "4-button-battery" })
+  test.socket.device_lifecycle:__queue_receive({ mock_device.id, "init" })
+
+  --doConfigure sets the provisioing state to provisioned
+  test.socket.device_lifecycle:__queue_receive({ mock_device.id, "doConfigure" })
+  mock_device:expect_metadata_update({ provisioning_state = "PROVISIONED" })
+
+  -- simulate the profile change update taking affect and the device info changing
   local device_info_copy = utils.deep_copy(mock_device.raw_st_data)
   device_info_copy.profile.id = "4-buttons-battery"
   local device_info_json = dkjson.encode(device_info_copy)
   test.socket.device_lifecycle:__queue_receive({ mock_device.id, "infoChanged", device_info_json })
-
   test.socket.capability:__expect_send(mock_device:generate_test_message("main", capabilities.button.supportedButtonValues({"pushed"}, {visibility = {displayed = false}})))
   test.socket.capability:__expect_send(mock_device:generate_test_message("main", button_attr.pushed({state_change = false})))
-
   test.socket.capability:__expect_send(mock_device:generate_test_message("button2", capabilities.button.supportedButtonValues({"pushed", "held"}, {visibility = {displayed = false}})))
   test.socket.capability:__expect_send(mock_device:generate_test_message("button2", button_attr.pushed({state_change = false})))
-
   test.socket.capability:__expect_send(mock_device:generate_test_message("button3", capabilities.button.supportedButtonValues({"pushed", "held"}, {visibility = {displayed = false}})))
   test.socket.capability:__expect_send(mock_device:generate_test_message("button3", button_attr.pushed({state_change = false})))
-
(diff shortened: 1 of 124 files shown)


drivers/SmartThings/matter-button/src/test/test_matter_multi_button.lua
```diff
@@ -73,35 +73,47 @@ local CLUSTER_SUBSCRIBE_LIST ={
   clusters.Switch.server.events.MultiPressComplete,
 }
 
+-- All messages queued and expectations set are done before the driver is actually run
 local function test_init()
+  -- we dont want the integration test framework to generate init/doConfigure, we are doing that here
```
cjswedes, line 78: @nickolas-deboom @hcarter-775 @ctowns Most test files will need to be updated with something like this to get the mock devices to go through the added process. This is because often the matter drivers are setting fields and tracking feature support on endpoints during added/init/doConfigure. I would suggest that we make the tests mimick the real added flow where we receive added, then init, and then doConfigure.

drivers/SmartThings/matter-sensor/src/test/test_matter_smoke_co_alarm.lua
```diff
@@ -73,18 +74,20 @@ local cluster_subscribe_list = {
 }
 
 local function test_init()
+  -- The startup messages are enabled, so this device will get an init,
+  -- and doConfigure (because provisioing_state is TYPED on the device).
+  test.mock_device.add_test_device(mock_device)
```
cjswedes, line 79: We dont need to simulate the whole added/init/doConfigure flow since the driver isn't doing anything special (namely setting fields on the device object); the test framework will queue init and doConfigure automatically, which is why they aren't queued here, but we still are setting expectations on the subscribe and update to provisioning state.

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button.lua
```diff
@@ -116,29 +116,42 @@ local function configure_buttons()
...
+  test.disable_startup_messages()
+  test.mock_device.add_test_device(mock_device) -- make sure the cache is populated
+
+  -- added sets a bunch of fields on the device, and calls init
   local subscribe_request = CLUSTER_SUBSCRIBE_LIST[1]:subscribe(mock_device)
   for i, clus in ipairs(CLUSTER_SUBSCRIBE_LIST) do
     if i > 1 then subscribe_request:merge(clus:subscribe(mock_device)) end
   end
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
-  mock_device:set_field("__ELECTRICAL_TOPOLOGY", {topology = false, tags_on_ep = {}}, {persist = false}) -- since we're assuming this would have happened during device_added in this case.
```
cjswedes: Note how this hack is no longer necessary because it is happening in the actual driver added/init/doConfigure handlers.

drivers/SmartThings/matter-sensor/src/test/test_matter_air_quality_sensor.lua
```diff
@@ -249,6 +249,9 @@ local function initialize_mock_device(generic_mock_device, generic_subscribed_at
   test.mock_device.add_test_device(generic_mock_device)
 end
 
+-- TODO add tests for configuration using modular profiles
+test.set_rpc_version(7)
```
cjswedes, line 253: The configure test fail if this is > 7 because then the driver starts using modular profiles. We should add tests for the modular profile usage. cc: @hcarter-775

drivers/SmartThings/matter-thermostat/src/test/test_matter_thermo_battery.lua
```diff
@@ -70,6 +70,17 @@ local cluster_subscribe = {
 }
 
 local function test_init()
+  -- test.socket.device_lifecycle:__queue_receive({ mock_device.id, "added" })
```
nickolas-deboom: Was this intentional?

cjswedes: This was not intentional. There was a bunch of commits that I squashed into the first commit of this PR, and my guess is this was just something that was carried over. I can update this test.

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button.lua
```diff
@@ -116,29 +116,42 @@ local function configure_buttons()
...
+  -- so we can set the proper expectations on those events.
+  test.disable_startup_messages()
+  test.mock_device.add_test_device(mock_device) -- make sure the cache is populated
+
+  -- added sets a bunch of fields on the device, and calls init
   local subscribe_request = CLUSTER_SUBSCRIBE_LIST[1]:subscribe(mock_device)
   for i, clus in ipairs(CLUSTER_SUBSCRIBE_LIST) do
     if i > 1 then subscribe_request:merge(clus:subscribe(mock_device)) end
   end
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
```
aleclorimer, line 131: @ctowns I see that we expect two subscribe requests in this test. The first one because the `added` handler runs the init handler as well. Now that we are guaranteeing that the init message will be sent, can [this init](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/77fb194a61c69e5fa9f274762eaad24aa38da0bf/drivers/SmartThings/matter-switch/src/init.lua#L1418) being kicked off from the added handler be removed, or is this needed still?

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua
```diff
@@ -192,37 +192,64 @@ local function configure_buttons()
...
+-- All messages queued and expectations set are done before the driver is actually run
 local function test_init_mcd_unsupported_switch_device_type()
+  -- we dont want the integration test framework to generate init/doConfigure, we are doing that here
+  -- so we can set the proper expectations on those events.
+  test.disable_startup_messages()
+  test.mock_device.add_test_device(mock_device_mcd_unsupported_switch_device_type) -- make sure the cache is populated
+end
+
+
+local function test_init_mcd_unsupported_switch_device_type_old()
```
nickolas-deboom: Was this meant to be removed?

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button.lua
ctowns: Hey Alec, sorry I missed this - yes you are right we do not need to include that double init in the switch driver now that the init event is guaranteed. I have pushed up a PR to remove that init call for hubs running the latest FW, which we can move forward with when these changes roll out: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2339

Once these test fixes go in, I will make the corresponding fixes to the test cases in the other PR as needed.

hcarter-775 (approved):

PR #2311, opened 2025-07-31
58 Unit test fixes for matter test cases

nickolas-deboom:
This PR covers the matter test case updates associated with #2310.

### DIFF
--- a/drivers/SmartThings/matter-lock/src/test/test_matter_lock_unlatch.lua
+++ b/drivers/SmartThings/matter-lock/src/test/test_matter_lock_unlatch.lua
@@ -54,27 +54,27 @@ local mock_device = test.mock_device.build_test_matter_device({
 })
 
 local function test_init()
+  test.disable_startup_messages()
+  test.mock_device.add_test_device(mock_device)
+  test.socket.device_lifecycle:__queue_receive({ mock_device.id, "added" })
+  test.socket.capability:__expect_send(
+    mock_device:generate_test_message("main", capabilities.lockAlarm.alarm("clear", {state_change = true}))
+  )
+  test.socket.device_lifecycle:__queue_receive({ mock_device.id, "init" })
   local subscribe_request = DoorLock.attributes.LockState:subscribe(mock_device)
   subscribe_request:merge(DoorLock.attributes.OperatingMode:subscribe(mock_device))
   subscribe_request:merge(DoorLock.events.LockOperation:subscribe(mock_device))
   subscribe_request:merge(DoorLock.events.DoorLockAlarm:subscribe(mock_device))
   test.socket["matter"]:__expect_send({mock_device.id, subscribe_request})
-  test.mock_device.add_test_device(mock_device)
+  test.socket.device_lifecycle:__queue_receive({ mock_device.id, "doConfigure" })
+  mock_device:expect_metadata_update({ profile = "lock-unlatch" })
+  mock_device:expect_metadata_update({ provisioning_state = "PROVISIONED" })
+  test.socket.capability:__expect_send(
+    mock_device:generate_test_message("main", capabilities.lock.supportedLockCommands({"lock", "unlock", "unlatch"}, {visibility = {displayed = false}}))
+  )
 end
 
 test.set_test_init_function(test_init)
-
-test.register_coroutine_test(
-  "Assert profile applied over doConfigure",
-  function()
-    test.socket.device_lifecycle:__queue_receive({ mock_device.id, "doConfigure" })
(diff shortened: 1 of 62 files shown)


drivers/SmartThings/matter-lock/src/test/test_matter_lock_unlatch.lua
```diff
@@ -54,28 +54,28 @@ local mock_device = test.mock_device.build_test_matter_device({
...
+  mock_device:expect_metadata_update({ provisioning_state = "PROVISIONED" })
+  test.socket.capability:__expect_send(
+    mock_device:generate_test_message("main", capabilities.lock.supportedLockCommands({"lock", "unlock", "unlatch"}, {visibility = {displayed = false}}))
+  )
 end
 
 test.set_test_init_function(test_init)
 
-test.register_coroutine_test(
-  "Assert profile applied over doConfigure",
```
nickolas-deboom, line 68: This is now tested in `test_init`

aleclorimer:
Tests are looking good. :+1:  Looks like there are 3 failing tests still due to modular profile changes which I think fall outside the scope of these changes (and would be fixed by a different PR).

nickolas-deboom:
> Tests are looking good. 👍 Looks like there are 3 failing tests still due to modular profile changes which I think fall outside the scope of these changes (and would be fixed by a different PR).

Ah yeah, good catch - those will be fixed by #2326!

drivers/SmartThings/matter-lock/src/test/test_new_matter_lock_battery.lua
cjswedes, line 1: Looks like there isnt a doConfigure event in these tests. Was there a reason for that? I think we should include it because it is a part of the path that would get hit in reality, so adding it would add coverage to ensure there isnt something unexpected happening

drivers/SmartThings/matter-media/src/test/test_matter_media_video_player.lua
```diff
@@ -86,22 +86,62 @@ local mock_device_variable_speed = test.mock_device.build_test_matter_device({
...
-  test.socket.matter:__set_channel_ordering("relaxed")
   local subscribe_request = cluster_subscribe_list[1]:subscribe(mock_device)
   for i, cluster in ipairs(cluster_subscribe_list) do
-    print(i)
     if i > 1 then
       subscribe_request:merge(cluster:subscribe(mock_device))
     end
-    print(subscribe_request)
   end
+  test.socket.device_lifecycle:__queue_receive({ mock_device.id, "init" })
```
cjswedes: Is an added event needed, or is this used with tests for where a device was already added, and doConfigure failed? If thats the case I would favor avoiding the disable altogether, and making sure `_provisioning_state = "TYPED"` is set on the mock device: the integration test framework will automatically queue init/doConfigure in that case.

drivers/SmartThings/matter-sensor/src/test/test_matter_freeze_leak_sensor.lua
```diff
@@ -63,29 +63,25 @@ local subscribed_attributes = {
...
 local function test_init_freeze_leak()
+  test.disable_startup_messages()
+  test.mock_device.add_test_device(mock_device_freeze_leak)
   local subscribe_request = subscribed_attributes[1]:subscribe(mock_device_freeze_leak)
   for i, cluster in ipairs(subscribed_attributes) do
     if i > 1 then
       subscribe_request:merge(cluster:subscribe(mock_device_freeze_leak))
     end
   end
+  test.socket.device_lifecycle:__queue_receive({ mock_device_freeze_leak.id, "init" })
```
cjswedes, line 65: added event needed?

drivers/SmartThings/matter-lock/src/test/test_new_matter_lock_battery.lua
nickolas-deboom: I moved doConfigure from the individual test cases to the init functions in the latest commit

drivers/SmartThings/matter-sensor/src/test/test_matter_freeze_leak_sensor.lua
nickolas-deboom: Updated!

drivers/SmartThings/matter-media/src/test/test_matter_media_video_player.lua
nickolas-deboom: Added is needed, fixed in latest commit

nickolas-deboom:
I'm going to leave these commits unsquashed within this PR but use Squash and Merge to merge into the main PR.

PR #2326, opened 2025-08-07
Update supports_modular_profiles to account for Thermostats

nickolas-deboom:
# Description of Change

ce5622b47b6d3db40a2cd1a401f30b5e0c3ee480 updated `get_device_type` to include the Thermostat device type ID, but it wasn't accounted for in `supports_modular_profiles`.

# Summary of Completed Tests

Tested the following cases on device: (1) onboard Thermostat. Device joins modular profile as expected. (2) onboard to production driver, device joins static profile. After updating the driver to include the changes from this PR, device remains on static profile with no unexpected behavior.

Adding the `Thermostat` device type to the mock device within `test_matter_thermostat_modular.lua` verifies that modular profiles are used when the thermostat device type is present and not used otherwise. Before, `get_device_type` returned false for this mock device so it would still use modular profiles.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -539,25 +539,34 @@ local function get_endpoints_for_dt(device, device_type)
 end
 
 local function get_device_type(device)
-  local device_type_found = {}
+  -- For cases where a device has multiple device types, this list indicates which
+  -- device type will be the "main" device type for purposes of selecting a profile
+  -- with an appropriate category. This is done to promote consistency between
+  -- devices with similar device type compositions that may report their device types
+  -- listed in different orders
+  local device_type_priority = {
+    [RAC_DEVICE_TYPE_ID] = 1,
+    [AP_DEVICE_TYPE_ID] = 2,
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1113,18 +1119,18 @@ local function match_modular_profile(driver, device)
 end
 
 local function supports_modular_profile(device)
-  local device_type = get_device_type(device)
-  local thermostat_eps = device:get_endpoints(clusters.Thermostat.ID)
+  local modular_device_type = get_device_type(device)
+  if not tbl_contains({AP_DEVICE_TYPE_ID, RAC_DEVICE_TYPE_ID, THERMOSTAT_DEVICE_TYPE_ID}, modular_device_type) then
```
ctowns: nit: to make this a little more self-documenting, I think you can make this a table called `supported_modular_device_types` or something like that at the top of this function and use that variable here.

nickolas-deboom: Updated as suggested!

PR #2338, opened 2025-08-13
WWSTCERT-7482 Meross Smart Wi-Fi Plug

greens:


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -717,6 +717,12 @@ matterManufacturer:
     vendorId: 0x102E
     productId: 0x2250
     deviceProfileName: button-battery
+#Meross
+  - id: "4933/40962"
+    deviceLabel: Smart Wi-Fi Plug
+    vendorId: 0x1345
+    productId: 0xA002
+    deviceProfileName: switch-binary
```
nickolas-deboom, line 725: I think this plug should use the `plug-power-energy-powerConsumption` profile

PR #2343, opened 2025-08-15
Updating matter rvc ux

HunsupJung:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change
The unnecessary mode capability for Run Mode cluster has been deleted.

# Summary of Completed Tests
This feature is being tested with following TC in SVT.
https://smartthings.atlassian.net/wiki/spaces/STHK/pages/4088660856/REQ-22349+Matter+RVC+UX+Upgrade+TC

### DIFF
--- a/drivers/SmartThings/matter-rvc/src/init.lua
+++ b/drivers/SmartThings/matter-rvc/src/init.lua
@@ -55,8 +55,8 @@ local subscribed_attributes = {
   }
 }
 
-local function find_default_endpoint(device, component_name)
-  local eps = device:get_endpoints(clusters.RvcOperationalState.ID)
+local function find_default_endpoint(device, cluster_id)
+  local eps = device:get_endpoints(cluster_id)
   table.sort(eps)
   for _, v in ipairs(eps) do
     if v ~= 0 then --0 is the matter RootNode endpoint
@@ -85,7 +85,6 @@ end
 
 local function device_init(driver, device)
   device:subscribe()
-  device:set_component_to_endpoint_fn(find_default_endpoint)
 
   -- comp/ep map functionality removed 9/5/25.
   device:set_field("__component_to_endpoint_map", nil)
@@ -501,7 +500,7 @@ end
 -- Capability Handlers --
 local function handle_robot_cleaner_operating_state_start(driver, device, cmd)
   device.log.info("handle_robot_cleaner_operating_state_start")
-  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local endpoint_id = find_default_endpoint(device, clusters.RvcOperationalState.ID)
 
   -- Get current run mode, current tag, current operating state
   local current_run_mode = device:get_field(CURRENT_RUN_MODE)
@@ -530,7 +529,7 @@ local function handle_robot_cleaner_operating_state_start(driver, device, cmd)
     device:send(clusters.RvcOperationalState.commands.Resume(device, endpoint_id))
   elseif can_send_state_command(device, capabilities.mode.commands.setMode.NAME, current_state, current_tag) == true then
     for _, mode in ipairs(supported_run_modes) do
-      endpoint_id = device:component_to_endpoint(cmd.component)
+      endpoint_id = find_default_endpoint(device, clusters.RvcOperationalState.ID)
       if mode.tag == clusters.RvcRunMode.types.ModeTag.CLEANING then
         device:send(clusters.RvcRunMode.commands.ChangeToMode(device, endpoint_id, mode.id))
         return
@@ -541,20 +540,20 @@ end
 
 local function handle_robot_cleaner_operating_state_pause(driver, device, cmd)
   device.log.info("handle_robot_cleaner_operating_state_pause")
-  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local endpoint_id = find_default_endpoint(device, clusters.RvcOperationalState.ID)
   device:send(clusters.RvcOperationalState.commands.Pause(device, endpoint_id))
 end
 
 local function handle_robot_cleaner_operating_state_go_home(driver, device, cmd)
   device.log.info("handle_robot_cleaner_operating_state_go_home")
-  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local endpoint_id = find_default_endpoint(device, clusters.RvcOperationalState.ID)
   device:send(clusters.RvcOperationalState.commands.GoHome(device, endpoint_id))
 end
 
 local function handle_robot_cleaner_mode(driver, device, cmd)
   device.log.info(string.format("handle_robot_cleaner_mode component: %s, mode: %s", cmd.component, cmd.args.mode))
 
-  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local endpoint_id = find_default_endpoint(device, clusters.RvcOperationalState.ID)
   local supported_modes = device:get_field(CLEAN_MODE_SUPPORTED_MODES) or {}
   for _, mode in ipairs(supported_modes) do
     if cmd.args.mode == mode.label then
@@ -573,7 +572,7 @@ local function handle_robot_cleaner_areas_selection(driver, device, cmd)
   for i, areaId in ipairs(cmd.args.areas) do
     table.insert(selectAreas, uint32_dt(areaId))
   end
-  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local endpoint_id = find_default_endpoint(device, clusters.RvcOperationalState.ID)
   if cmd.component == "main" then
     device:send(clusters.ServiceArea.commands.SelectAreas(device, endpoint_id, selectAreas))
   end


drivers/SmartThings/matter-rvc/profiles/rvc-clean-mode-service-area.yml
```diff
@@ -5,6 +5,8 @@ components:
     capabilities:
       - id: robotCleanerOperatingState
```
nickolas-deboom, line 6: Are these profiles currently in use by devices? If so, it might be better to create new profiles rather than removing capabilities to avoid breaking automations / rules for devices in the field.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -57,24 +57,23 @@ local subscribed_attributes = {
...
-  local component_to_endpoint_map = {
-    ["main"] = run_mode_eps[1],
-    ["runMode"] = run_mode_eps[1],
-    ["cleanMode"] = clean_mode_eps[1]
-  }
-  device:set_field(COMPONENT_TO_ENDPOINT_MAP, component_to_endpoint_map, {persist = true})
+local function component_to_endpoint(device, component_name)
+  -- Use the find_default_endpoint function to return the first endpoint that
+  -- supports a given cluster.
+  return find_default_endpoint(device, clusters.RvcOperationalState.ID)
```
nickolas-deboom: What does the endpoint structure typically look like for a RVC? Are the clusters usually all implemented on the same endpoint?

drivers/SmartThings/matter-rvc/profiles/rvc-clean-mode-service-area.yml
hcarter-775: I confirmed these profiles are used by devices, and that automations are being used on the mode operation for at least a handful, though the percentage is small.

Instead of adding new profiles, I believe we should modify the system to use the benefits of modular profiles.

HunsupJung: @hcarter-775 
Can we use the modular profiles by 9/9? If so, could you share the way to use the modular profiles?

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: I not sure, but Roborock implements `Power Source`, `RVC Run Mode`, `RVC Clean Mode`, `RVC Operational State`, and `Service Area` on the endpoint 1.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -93,6 +88,20 @@ local function do_configure(driver, device)
...
   device:try_update_metadata({profile = profile_name})
+end
+
+local function device_init(driver, device)
+  device:subscribe()
+  device:set_component_to_endpoint_fn(component_to_endpoint)
+  if not device:get_field(SERVICE_AREA_PROFILED) and #device:get_endpoints(clusters.ServiceArea.ID) > 0 then
+    match_profile(driver, device)
+    device:set_field(SERVICE_AREA_PROFILED, true)
+  end
```
hcarter-775:
```suggestion
  if not device:get_field(SERVICE_AREA_PROFILED) then
    if #device:get_endpoints(clusters.ServiceArea.ID) > 0 then
      match_profile(driver, device)
    end
    device:set_field(SERVICE_AREA_PROFILED, true, { persist = true })
  end
```
we only want this logic to run once, and we want to set the device as profiled whether or not service area is part of the device

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -93,6 +88,20 @@ local function do_configure(driver, device)
...
+end
+
+local function do_configure(driver, device)
+  match_profile(driver, device)
+  device:set_field(SERVICE_AREA_PROFILED, true)
```
hcarter-775:
```suggestion
  device:set_field(SERVICE_AREA_PROFILED, true, { persist = true })
```
this field should not be reset on a driver restart

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -32,10 +32,11 @@ if version.api < 13 then
   clusters.Global = require "Global"
 end
 
-local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
 local RUN_MODE_SUPPORTED_MODES = "__run_mode_supported_modes"
+local CURRENT_RUN_MODE = "__current_run_mode"
```
hcarter-775, line 36: why do we need to set `CURRENT_RUN_MODE` instead of using the `get_latest_state` api as we did before? And would you mind adding a short comment by this field to explain this?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -32,10 +32,11 @@ if version.api < 13 then
   clusters.Global = require "Global"
 end
 
-local COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
```
hcarter-775, line 35: If we're removing this field, can you please set it as null in the init function? we can leave this local variable deleted and just write something like:

```
-- comp/ep map functionality removed 9/5/25.
device:set_field("__component_to_endpoint_map", nil)
```

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -54,32 +55,26 @@ local subscribed_attributes = {
...
-  if map[component] then
-    return map[component]
-  else
-    return device.MATTER_DEFAULT_ENDPOINT
+local function find_default_endpoint(device, cluster)
+  local res = device.MATTER_DEFAULT_ENDPOINT
+  local eps = device:get_endpoints(cluster)
+  table.sort(eps)
+  for _, v in ipairs(eps) do
+    if v ~= 0 then --0 is the matter RootNode endpoint
```
hcarter-775, line 62: I don't understand why we're ignoring the root endpoint here?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -564,11 +505,7 @@ local function handle_robot_cleaner_operating_state_start(driver, device, cmd)
   local endpoint_id = device:component_to_endpoint(cmd.component)
 
   -- Get current run mode, current tag, current operating state
-  local current_run_mode = device:get_latest_state(
-    "runMode",
-    capabilities.mode.ID,
-    capabilities.mode.mode.NAME
-  )
+  local current_run_mode = device:get_field(CURRENT_RUN_MODE)
```
hcarter-775, line 506: not sure where to comment about this, but line 534 of the current PR is: 
`endpoint_id = device:component_to_endpoint("runMode")`
we need to remove this reference. I wonder also how this was tested where this was missed? Is this logic required at all anymore?

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -54,32 +55,26 @@ local subscribed_attributes = {
...
-local function device_added(driver, device)
-  local run_mode_eps = device:get_endpoints(clusters.RvcRunMode.ID) or {}
-  local clean_mode_eps = device:get_endpoints(clusters.RvcCleanMode.ID) or {}
-  local component_to_endpoint_map = {
-    ["main"] = run_mode_eps[1],
-    ["runMode"] = run_mode_eps[1],
-    ["cleanMode"] = clean_mode_eps[1]
-  }
-  device:set_field(COMPONENT_TO_ENDPOINT_MAP, component_to_endpoint_map, {persist = true})
+local function component_to_endpoint(device, component_name)
```
hcarter-775: I suggest we remove this function in all places in the driver and directly use `find_default_endpoint` with the appropriate cluster that the command handler should be mapping to.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -54,32 +55,26 @@ local subscribed_attributes = {
...
 }
 
-local function component_to_endpoint(device, component)
-  local map = device:get_field(COMPONENT_TO_ENDPOINT_MAP) or {}
-  if map[component] then
-    return map[component]
-  else
-    return device.MATTER_DEFAULT_ENDPOINT
+local function find_default_endpoint(device, cluster)
+  local res = device.MATTER_DEFAULT_ENDPOINT
```
hcarter-775: let's remove the `res` variable setting and just return `default.MATTER_DEFAULT_ENDPOINT` explicitly at the end?

drivers/SmartThings/matter-rvc/profiles/rvc-clean-mode-service-area.yml
```diff
@@ -38,22 +26,15 @@ deviceConfig:
...
     capability: mode
     version: 1
     patch:
     - op: replace
       path: /0/list/command/supportedValues
       value: supportedArguments.value
-  - component: cleanMode
-    capability: mode
+  - component: main
+    capability: serviceArea
```
hcarter-775, line 36: why was the location of this capability moved in this definition?

drivers/SmartThings/matter-rvc/src/init.lua
nickolas-deboom: We should be able to remove this from `init` since it already runs in `do_configure`

hcarter-775: @nickolas-deboom they want to re-profile all previously onboarded devices to use service area if that's supported

nickolas-deboom: Ah gotcha, thanks!

drivers/SmartThings/matter-rvc/profiles/rvc-clean-mode-service-area.yml
HunsupJung: There is no reason. I wanted to emphasize the relation between `mode` and `robotCleanerOperatingState`.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Okay, I got it.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: `mode` capability related to run mode is going to be deleted. So `CURRENT_RUN_MODE` value is needed to store the latest `CurrentMode` value of `RVC Run Mode` Cluster

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Okay, I got it.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: When the manufacturer create a ZAP file through the ZAP tool, number 0 will be used as Root node and the device type that is wanted to add will be 1. Of course, the manufacturer can set the endpoint arbitrarily.
This code is also included in all other Matter drivers, so if we remove it, it would be better to remove it all at once in the future.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Okay, I got it.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Okay, I got it.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Thank you for finding the part. As you mentioned, we need to change `"runMode"` to `cmd.component`.

The reason why it was not a problem during the test is that `component_to_endpoint` finds an endpoint with `RvcOperationalState` cluster regardless of what component is put in.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -55,23 +55,16 @@ local subscribed_attributes = {
   }
 }
 
-local function find_default_endpoint(device, cluster)
-  local res = device.MATTER_DEFAULT_ENDPOINT
-  local eps = device:get_endpoints(cluster)
+local function find_default_endpoint(device, component_name)
```
hcarter-775: sorry, I don't think I was completely clear before. Instead of using the `component_name` as an argument here, I think we should make the cluster be the argument, and call this function for whatever cluster is needed in each handler. I'm thinking something like:

```
local function find_default_endpoint(device, cluster_id)
  local eps = device:get_endpoints(cluster_id)
  ... leave the rest the same ...
end
```
and then change what is passed in at the callsite.

drivers/SmartThings/matter-rvc/src/init.lua
```diff
@@ -92,16 +85,22 @@ end
 
 local function device_init(driver, device)
   device:subscribe()
-  device:set_component_to_endpoint_fn(component_to_endpoint)
-  if not device:get_field(SERVICE_AREA_PROFILED) and #device:get_endpoints(clusters.ServiceArea.ID) > 0 then
-    match_profile(driver, device)
-    device:set_field(SERVICE_AREA_PROFILED, true)
+  device:set_component_to_endpoint_fn(find_default_endpoint)
```
hcarter-775: we no longer require this line if we go the route of my previous suggestion

HunsupJung: Okay, got it.

drivers/SmartThings/matter-rvc/src/init.lua
HunsupJung: Okay, got it.

nickolas-deboom (approved):

PR #2344, opened 2025-08-18
Update new-matter-lock driver to support Aliro feature

HunsupJung:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests
This feature was tested with following TC at SVT.

https://smartthings.atlassian.net/wiki/spaces/STHK/pages/3890216961/REQ-22348+Matter+DoorLock+Aliro+Feature+Support+TC

### DIFF
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -40,8 +40,7 @@ local RESPONSE_STATUS_MAP = {
   [DoorLock.types.DlStatus.OCCUPIED] = "occupied",
   [DoorLock.types.DlStatus.INVALID_FIELD] = "invalidCommand",
   [DoorLock.types.DlStatus.RESOURCE_EXHAUSTED] = "resourceExhausted",
-  [DoorLock.types.DlStatus.NOT_FOUND] = "failure",
-  [DoorLock.types.DlStatus.INVALID_FIELD] = "invalidCommand"
+  [DoorLock.types.DlStatus.NOT_FOUND] = "failure"
 }
 
 local WEEK_DAY_MAP = {
@@ -1954,7 +1953,7 @@ local function set_credential_response_handler(driver, device, ib, response)
     return
   end
   local cmdName = device:get_field(lock_utils.COMMAND_NAME)
-  if cmdName == "addCredential" then
+  if cmdName == "addCredential" or cmdName == "updateCredential" or cmdName == "addCota" then
     set_pin_response_handler(driver, device, ib, response)
   elseif cmdName == "setIssuerKey" then
     set_issuer_key_response_handler(driver, device, ib, response)
@@ -2033,7 +2032,8 @@ end
 -------------------------------
 local function clear_credential_response_handler(driver, device, ib, response)
   local cmdName = device:get_field(lock_utils.COMMAND_NAME)
-  if cmdName ~= "deleteCredential" and cmdName ~= "clearEndpointKey" and cmdName ~= "clearIssuerKey" then
+  if cmdName ~= "deleteCredential" and cmdName ~= "clearEndpointKey" and
+  cmdName ~= "clearIssuerKey" and cmdName ~= "deleteAllCredentials" then
     return
   end
   local status = RESPONSE_STATUS_MAP[ib.status] or "success"
@@ -2041,7 +2041,7 @@ local function clear_credential_response_handler(driver, device, ib, response)
   local userIdx = device:get_field(lock_utils.USER_INDEX)
   local all_user_credentials_removed = false
 
-  if cmdName == "deleteCredential" and status == "success" then
+  if (cmdName == "deleteCredential" or cmdName == "deleteAllCredentials") and status == "success" then
     -- Get result from data saved in relevant, associated fields
     local credIdx = device:get_field(lock_utils.CRED_INDEX)
 
@@ -2087,7 +2087,7 @@ local function clear_credential_response_handler(driver, device, ib, response)
   end
 
   -- Update commandResult
-  if cmdName == "deleteCredential" then
+  if cmdName == "deleteCredential" or cmdName == "deleteAllCredentials" then
     device:emit_event(capabilities.lockCredentials.commandResult(
       command_result_info, {state_change = true, visibility = {displayed = false}}
     ))
--- a/drivers/SmartThings/matter-lock/src/test/test_new_matter_lock.lua
+++ b/drivers/SmartThings/matter-lock/src/test/test_new_matter_lock.lua
@@ -971,7 +971,7 @@ test.register_coroutine_test(
     test.socket.capability:__expect_send(
       mock_device:generate_test_message(
         "main",
-        capabilities.lockUsers.users({{userIndex = 1, userType = "adminMember"}}, {visibility={displayed=false}})
+        capabilities.lockUsers.users({{userIndex = 1, userName="Guest1", userType = "adminMember"}}, {visibility={displayed=false}})
       )
     )
     test.socket.capability:__expect_send(
@@ -1493,7 +1493,6 @@ test.register_coroutine_test(
         ),
       }
     )
-    -- test.wait_for_events()
   end
 )
 
@@ -1748,8 +1747,26 @@ test.register_coroutine_test(
     test.socket.capability:__expect_send(
       mock_device:generate_test_message(
         "main",
+        capabilities.lockUsers.users({}, {visibility={displayed=false}})
+      )
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main",
+        capabilities.lockSchedules.weekDaySchedules({}, {visibility={displayed=false}})
+      )
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main",
+        capabilities.lockSchedules.yearDaySchedules({}, {visibility={displayed=false}})
+      )
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main",
         capabilities.lockCredentials.commandResult(
-          {commandName="deleteAllCredentials", credentialIndex=65534, statusCode="success"},
+          {commandName="deleteAllCredentials", userIndex=65534, credentialIndex=65534, statusCode="success"},
           {state_change=true, visibility={displayed=false}}
         )
       )


drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -2078,6 +2647,385 @@ local function handle_refresh(driver, device, command)
   device:set_field(lock_utils.BUSY_STATE, false, {persist = true})
 end
 
+local function handle_set_reader_config(driver, device, command)
```
tpmanley, line 2514: I thought the plan was to remove this command from the capability and generate the Matter command directly from the driver?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1993,6 +2522,46 @@ local function handle_clear_year_day_schedule(driver, device, command)
   )
 end
 
+------------------------------------
+-- Clear Year Day Schedule Response --
+------------------------------------
+local function clear_year_day_schedule_handler(driver, device, ib, response)
```
tpmanley: Since this isn't related to Aliro can you separate it out into a separate PR?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -509,6 +659,29 @@ local function add_user_to_table(device, userIdx, usrType)
   -- Add new entry to table
   table.insert(user_table, {userIndex = userIdx, userType = usrType})
   device:emit_event(capabilities.lockUsers.users(user_table, {visibility = {displayed = false}}))
+
+  -- If User Type is Guest, add Default Schedule
```
tpmanley: Unless I'm mistaken this doesn't seem related to Aliro either. Since this PR is very large, can you make sure there is one PR with just the Aliro changes and then one or more PRs for the other changes you're making to make it easier to do a good review, and also to test the changes?

0001-Update-new-matter-lock-driver-to-support-Aliro-featu.patch
```diff
@@ -0,0 +1,2061 @@
+From 0b39cef689a2e769589f2fca839a00f32079448c Mon Sep 17 00:00:00 2001
```
hcarter-775: remove this file

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: As you know, Kwang-Hui wants to keep this command.

0001-Update-new-matter-lock-driver-to-support-Aliro-featu.patch
HunsupJung: I removed this file~! Thank you.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I removed this part in this PR.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I removed this part in this PR.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -393,6 +432,117 @@ local function max_year_schedule_of_user_handler(driver, device, ib, response)
...
+  end
+  local octet_string = ""
+  for i = 1, #hex_string, 2 do
+      local hex = hex_string:sub(i, i + 1)
+      octet_string = octet_string .. string.char(tonumber(hex, 16))
+  end
+  return octet_string
+end
+
+local function octet_string_to_hex_string(octet_string)
```
hcarter-775: you can use the utils function `utils.bytes_to_hex_string` here instead I think, you'll just need to do the nil check handling outside of this call (which you do already).

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -393,6 +432,117 @@ local function max_year_schedule_of_user_handler(driver, device, ib, response)
...
+end
+
+-------------------------------
+-- Aliro Group Resolving Key --
+-------------------------------
+local function aliro_protocol_versions_handler(driver, device, ib, response)
+  if ib.data.elements == nil then
+    return
+  end
+  local versions = {}
```
hcarter-775:
```suggestion
  local protocol_versions = {}
```
nit for readability

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -393,6 +432,117 @@ local function max_year_schedule_of_user_handler(driver, device, ib, response)
...
+end
+
+-----------------------------------------------
+-- Aliro Supported BLE UWB Protocol Versions --
+-----------------------------------------------
+local function aliro_supported_ble_uwb_protocol_versions_handler(driver, device, ib, response)
+  if ib.data.elements == nil then
+    return
+  end
+  local versions = {}
```
hcarter-775: same here:
```suggestion
  local protocol_versions = {}
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1507,87 +1746,352 @@ local function set_credential_response_handler(driver, device, ib, response)
...
+    status = "failure"
+  elseif elements.status.value == DoorLock.types.DlStatus.DUPLICATE then
+    status = "duplicate"
+  elseif elements.status.value == DoorLock.types.DlStatus.INVALID_FIELD then
+    status = "failure"
+  elseif elements.status.value == DoorLock.types.DlStatus.RESOURCE_EXHAUSTED then
+    status = "resourceExhausted"
+  elseif elements.status.value == DoorLock.types.DlStatus.NOT_FOUND then
+    status = "failure"
+  end
```
hcarter-775:
```suggestion
  -- Update commandResult
  status = RESPONSE_STATUS_MAP[elements.status.value] or "occupied"
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1507,87 +1746,352 @@ local function set_credential_response_handler(driver, device, ib, response)
...
```
hcarter-775:
```suggestion
  status = RESPONSE_STATUS_MAP[elements.status.value] or "occupied"
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1507,87 +1746,352 @@ local function set_credential_response_handler(driver, device, ib, response)
...
+  -- Get parameters
+  local cmdName = "deleteAllCredentials"
+  local credential = {
+    credential_type = DoorLock.types.CredentialTypeEnum.PIN,
+    credential_index = ALL_INDEX,
+  }
+
+  -- Check busy state
+  local busy = check_busy_state(device)
+  if busy == true then
```
hcarter-775:
```suggestion
  if is_busy_state_set(device) then
```
I see this function `check_busy_state` already existed, and is being used like so in other parts of the code. I think we should change it everywhere to the above, and change the function name.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -2078,6 +2683,389 @@ local function handle_refresh(driver, device, command)
...
+      statusCode = "busy"
+    }
+    local event = capabilities.lockAliro.commandResult(
+      result,
+      {
+        state_change = true,
+        visibility = {displayed = false}
+      }
+    )
+    device:emit_event(event)
```
hcarter-775:
```suggestion
    local command_result_info = {
      commandName = cmdName,
      userIndex = userIdx,
      keyId = keyId,
      requestId = reqId,
      statusCode = "busy"
    }
    device:emit_event(capabilities.lockAliro.commandResult(
      command_result_info, { state_change = true, visibility = {displayed = false} }
    ))
```
This is a little bit of a nit, but all over the place the above organization of cap event sends as `result` -> `event` -> `emit_event`, which is find to be wordy and makes the handling more difficult to read than I think it needs to be. Would you mind changing this handling to something more like my above suggestion? I think by condensing the handling (and changing the variable name from `result` -> `command_result_info` or something similar, this will help readability in the long run.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1507,87 +1746,352 @@ local function set_credential_response_handler(driver, device, ib, response)
...
-    credential_type = DoorLock.types.CredentialTypeEnum.PIN,
-    credential_index = credIdx,
-  }
+-----------------------------------
+-- Set Aliro Credential Response --
+-----------------------------------
+local function set_issuer_key_response_handler(driver, device, ib, response)
+  local cmdName = "setIssuerKey"
+  local userIdx = device:get_field(lock_utils.USER_INDEX)
+  local userType = DoorLock.types.UserTypeEnum.UNRESTRICTED_USER
```
hcarter-775, line 1759: why is this an unrestricted_user?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1507,87 +1746,352 @@ local function set_credential_response_handler(driver, device, ib, response)
...
+  local issuerKeyIndex = device:get_field(lock_utils.ISSUER_KEY_INDEX)
+  local reqId = device:get_field(lock_utils.COMMAND_REQUEST_ID)
 
-  -- Check busy state
-  local busy = check_busy_state(device)
-  if busy == true then
+  local status = "success"
+  local elements = ib.info_block.data.elements
+  if elements.status.value == DoorLock.types.DlStatus.SUCCESS then
```
hcarter-775:
```suggestion
  local elements = ib.info_block.data.elements
  status = RESPONSE_STATUS_MAP[elements.status.value]
  if status == "success" then
    ...
  end
  -- In the case DlStatus returns Occupied, this means the current credential index is in use, so we must try the next one. If there is not a next index (i.e. it is nil), we should mark this as "resourceExhausted" and stop attempting to set the credentials.
  if status == "occupied" and elements.next_credential_index.value == nil then
    status = "resourceExhausted"
  end 
  if status == "occupied" then
    ...
  else
    ...
  end 
end
```
I believe we should lay all the functions that are formatted like this out like this ^
The current handling where there are many different "ways" to talk about the same thing (DlStatus in this case) is difficult to understand and hard to maintain

hcarter-775:
Hi @HunsupJung , I just made some initial comments not changing the handling of your code, but rather trying to maintain a more uniform code standard for readability and maintainability. I think it will be much easier to read and review once the above suggestions for writing have been implemented across the new code.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -27,12 +27,36 @@ end
...
+  ["Sunday"] = 1,
+  ["Monday"] = 2,
+  ["Tuesday"] = 4,
+  ["Wednesday"] = 8,
+  ["Thursday"] = 16,
+  ["Friday"] = 32,
+  ["Saturday"] = 64,
+}
+
+local ENDPOINT_KEY_ENUM = {
```
hcarter-775:
```suggestion
local ALIRO_KEY_TYPE_TO_CRED_ENUM_MAP = {
```
greater clarity in name

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1598,12 +2102,14 @@ local function clear_credential_response_handler(driver, device, ib, response)
     status = "invalidCommand"
   end
 
-  -- Delete User in table
+  -- Delete Pin in table
```
hcarter-775: In this handler, in the event we have a failed status, why would we update the commandResult with a 0 userIndex? Wouldn't it be more accurate not to send the capability any data on this index?

Further, per the capability def:
```
           userIndex:
              type: integer
              minimum: 1
```
so this should not be allowed?

hcarter-775: Also, in the case that the credIdx is ALL_INDEX, aren't we returning a nil index to userIndex?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1627,6 +2133,105 @@ local function clear_credential_response_handler(driver, device, ib, response)
   device:set_field(lock_utils.BUSY_STATE, false, {persist = true})
 end
 
+-------------------------------------
+-- Clear Aliro Credential Response --
+-------------------------------------
+local function clear_issuer_key_response_handler(driver, device, ib, response)
```
hcarter-775: In this and the other sub-functions for `clear_credential_response_handler`, why are we passing the driver and response as a parameter? Also, I don't think we need the ib either, just the ib.status?
```suggestion
local function clear_issuer_key_response_handler(device, ib.status)
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1507,87 +1746,352 @@ local function set_credential_response_handler(driver, device, ib, response)
...
+  local ep = device:component_to_endpoint(command.component)
+  device:send(DoorLock.server.commands.ClearCredential(device, ep, credential))
+end
+
+-----------------------------------
+-- Delete Pin Credential Response --
+-----------------------------------
+local function delete_pin_response_handler(driver, device, ib, response)
   -- Get result
   local cmdName = device:get_field(lock_utils.COMMAND_NAME)
```
hcarter-775, line 2034: no need to make this call here, since it's done in the caller function

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1627,6 +2133,105 @@ local function clear_credential_response_handler(driver, device, ib, response)
...
+local function clear_credential_response_handler(driver, device, ib, response)
+  local cmdName = device:get_field(lock_utils.COMMAND_NAME)
+  if cmdName == "deleteCredential" then
+    delete_pin_response_handler(driver, device, ib, response)
+  elseif cmdName == "clearIssuerKey" then
+    clear_issuer_key_response_handler(driver, device, ib, response)
+  elseif cmdName == "clearEndpointKey" then
+    clear_endpoint_key_response_handler(driver, device, ib, response)
+  end
+end
```
hcarter-775:
```suggestion
local function clear_credential_response_handler(driver, device, ib, response)
  local cmdName = device:get_field(lock_utils.COMMAND_NAME)
  if cmdName ~= "deleteCredential" and cmdName ~= "clearEndpointKey" and cmdName ~= "clearIssuerKey" then
    return
  end
  local status = RESPONSE_STATUS_MAP[ib.status]
  if status ~= "failure" and status ~= "invalidCommand" then
    status = "success" -- treat all non-failure and non-invalid statuses as a success
  end
  local command_result_info = { commandName = cmdName, statusCode = status } -- default command result
  local userIdx = device:get_field(lock_utils.USER_INDEX)
  local credentials_successfully_removed = false

  if cmdName == "deleteCredential" and status == "success" then
    -- Get result from data saved in relevant, associated fields
    local credIdx = device:get_field(lock_utils.CRED_INDEX)

    -- find userIdx associated with credIdx, don't use lock utils field in this case
    userIdx = 0
    userIdx = delete_credential_from_table(device, credIdx)
    if userIdx ~= 0 then
      credentials_successfully_removed = not has_credentials(device, userIdx)
    end

    -- set unique command result fields
    command_result_info.userIndex = userIdx
    command_result_info.credentialIndex = credIdx
  elseif cmdName == "clearIssuerKey" and status == "success" then
    -- Get result from data saved in relevant, associated fields
    local reqId = device:get_field(lock_utils.COMMAND_REQUEST_ID)

    delete_aliro_from_table(device, userIdx, "issuerKey", nil)
    credentials_successfully_removed = not has_credentials(device, userIdx)

    -- set unique command result fields
    command_result_info.userIndex = userIdx
    command_result_info.requestId = reqId
  elseif cmdName == "clearEndpointKey" and status == "success" then
    -- Get result from data saved in relevant, associated fields
    local deviceKeyId = device:get_field(lock_utils.DEVICE_KEY_ID)
    local keyType = device:get_field(lock_utils.ENDPOINT_KEY_TYPE)
    local reqId = device:get_field(lock_utils.COMMAND_REQUEST_ID)

    delete_aliro_from_table(device, userIdx, keyType, deviceKeyId)
    credentials_successfully_removed = not has_credentials(device, userIdx)

    -- set unique command result fields
    command_result_info.userIndex = userIdx
    command_result_info.keyId = deviceKeyId
    command_result_info.requestId = reqId
  end

  -- user data if credentials were removed
  if credentials_successfully_removed then
    delete_user_from_table(device, userIdx)
    delete_week_schedule_from_table_as_user(device, userIdx)
    delete_year_schedule_from_table_as_user(device, userIdx)
  end

  -- Update commandResult
  if cmdName == "deleteCredential" then
    device:emit_event(capabilities.lockCredentials.commandResult(
      command_result_info,
      { state_change = true, visibility = {displayed = false} }
    ))
  else
    device:emit_event(capabilities.lockAliro.commandResult(
      command_result_info,
      { state_change = true, visibility = {displayed = false} }
    ))
  end
  device:set_field(lock_utils.BUSY_STATE, false, {persist = true})
end
```
Based on all the related logic across the sub-functions, I almost think this formatting would make more sense.

hcarter-775: I also wonder about this gate that occurs in all 3 cases: `if has_credentials(device, userIdx) == false then ... end`. I'm wondering what will happen when `has_credentials` evaluates to `true`? It looks like the data would have been removed on the device (considering we are getting a success response) but in this case we haven't removed this data on our end?

Since this is how the credential removal will work moving forward, I wonder what this check is actually for?

hcarter-775: Also, is this `deleteCredential` command how we will handle deletion for all cases moving forward? Since before, we were using `deleteUser`, I wonder how this might change our handling.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: if the other implementation is taken (having one function with multiple branches, this is a moot point

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: again, may be a moot point if the single function approach I suggest below is taken

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: Thank you for letting me know the function.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I agree with you. the suggestion looks good to me.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: The policy is that only unrestricted user can use digital key.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: It is good checking point. Actually, now there is no case for using ALL_INDEX, so I will modify it to returning a nil index to user Index for it.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I didn't know we can use only those parameters.


> if the other implementation is taken (having one function with multiple branches, this is a moot point

you mean we can apply it to other commands, too?

HunsupJung: Oh, I see. I saw your suggestion that is single function right now. I think it's good approach because there is a lot of duplicated code in current code.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I don't know I understand what you said perfectly.
Plugin team will use `deleteCredential` instead of `deleteUser` in R4.

HunsupJung: > I'm wondering what will happen when `has_credentials` evaluates to true? 

We need to keep the user and schedule data, so if `has_credentials` is true, we do nothing.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: Can you also delete this helper, since it's no longer being used?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -215,13 +253,17 @@ local function match_profile_switch(driver, device)
...
 
   local profile_name = "lock"
   if #user_eps > 0 then
     profile_name = profile_name .. "-user"
     if #pin_eps > 0 then
       profile_name = profile_name .. "-pin"
     end
+    if #aliro_eps > 0 then
+      profile_name = profile_name .. "-aliro"
+    end
```
hcarter-775: We should remove this now that it doesn't map to any actual profiles.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: Ok, I believe this makes sense :+1:

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: > I don't know I understand what you said perfectly.
Plugin team will use deleteCredential instead of deleteUser in R4.

OK, that's fine. Thanks for confirming.
 
> We need to keep the user and schedule data, so if has_credentials is true, we do nothing.

Oh ok, I understand now. We're doing this since there are multiple ways for data to be stored for a user. So even a successful credential deletion doesn't mean all credentials are removed.

Given this new understanding, I would change my suggested variable name in the refactor from `credentials_successfully_removed` to `all_user_credentials_removed`

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: Oh yes, I will delete this helper.

hcarter-775:
@HunsupJung Sure you are aware, but the CI needs to be passing before we can merge this PR :+1:

hcarter-775 (approved):

PR #2345, opened 2025-08-18
Bugfix/sonos oauth reconnect

NoahCornell:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

This fixes two issues regarding sonos oauth expiration.

<strike>
**Better Preemptive Token Refreshing:** 

~~This is ensuring that we correctly preemptively refresh the oauth token. We were not setting `refresh` to `true` when calling  [security.get_sonos_oauth](https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/main/lua_libs/st/security.lua#L132-L138). Setting this to `true` ends up setting a query parameter to force refresh the token when making the API calls to get an ISA token. ~~

This change also increases the period where we will attempt to refresh the token to 2 hours. This is done to match up with the SSDP event handler which checks auth hourly so we should never end up in a situation where the token expires. If the token does expire, then our next call on the websocket will fail and we will not handle the next capability command.
</strike>

Removed due to potential issues with force refresh. Will open a new PR to handle.

**Reconnect Task Race Condition:**

If we don't have the proper auth to make websocket requests, then we will receive a `ERROR_NOT_AUTHORIZED` error. The [handling](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/sonos/src/api/sonos_connection.lua#L330-L355) for this error calls [Router.close_socket_for_player](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua#L290-L310).

`Router.close_socket_for_player` will immediately call the `on_close` callbacks for any listeners. For the [sonos_connection](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/sonos/src/api/sonos_connection.lua#L535-L546), this calls [_spawn_reconnect_task](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/sonos/src/api/sonos_connection.lua#L287-L297). 

Both the legacy and oauth flavors of the reconnect task check if the connection is running before attempting to reconnect. This ends up checking in the [websocket router](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua#L213-L218) if we are connected. 

Since there are several tasks at play here, we might do the check if we are connected before [this](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua#L42-L50) runs. The reconnect task will then exit early since it appears as we are already connected. The fix is to call the `on_close` when the socket is actually closed, rather than immediately, so that the reconnect task doesn't exit early.

# Summary of Completed Tests

I have been testing by using the hedge endpoints to manually set the token and it's expiration. I can let it sit and let the expiration naturally hit as well to make sure things are working.

- Test with valid token that is about to expire and modifying SSDP event handler to always check auth. Confirmed that the token is properly refreshed and subsequent capability commands are handled.
- Test with invalid token and modifying SSDP event handler to always check auth. Confirmed that the token is properly refreshed and subsequent capability commands are handled.
- Test with invalid token and sending capability command. Command is dropped but websocket is reconnected succesfully, subsequent commands are handled.
- Test with valid token that is about to expire and sending capability command. Command is handled and token is refreshed.
- Test with unplugging speaker. `on_close` still only called a single time. Reconnects successfully when plugged back in.

### DIFF
--- a/drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua
+++ b/drivers/SmartThings/sonos/src/api/sonos_websocket_router.lua
@@ -44,6 +44,16 @@ cosock.spawn(function()
       if wss ~= nil then
         log.trace(string.format("Closing websocket for player %s", unique_key))
         wss:close(CloseCode.normal(), "Shutdown requested by client")
+        local ws_id = wss.id
+        for _, uuid in ipairs((listener_ids_for_socket[ws_id] or {})) do
+            local listener = listeners[uuid]
+
+            if listener ~= nil then
+              listener.on_close(uuid)
+            end
+            listeners[uuid] = nil
+        end
+        listener_ids_for_socket[ws_id] = nil
       end
       websockets[unique_key] = nil
     end
@@ -292,17 +302,7 @@ function SonosWebSocketRouter.close_socket_for_player(target)
   local ws = websockets[target]
 
   if ws ~= nil then
-    local ws_id = ws.id
     table.insert(pending_close, target)
-    for _, uuid in ipairs((listener_ids_for_socket[ws_id] or {})) do
-      local listener = listeners[uuid]
-
-      if listener ~= nil then
-        listener.on_close(uuid)
-      end
-      listeners[uuid] = nil
-    end
-    listener_ids_for_socket[ws_id] = nil
     return true
   else
     return nil, string.format("No currently open connection for %s", target)


drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -369,8 +373,9 @@ function SonosDriver:get_oauth_token()
     -- token has not expired yet
     if now < expiration then
       -- token is expiring soon, so we pre-emptively refresh
-      if math.abs(expiration - now) < ONE_HOUR_IN_SECONDS then
-        local result, err = security.get_sonos_oauth()
+      if math.abs(expiration - now) < PREEMPTIVE_OAUTH_REFRESH_PERIOD_S then
+        -- Passing in true here forces refresh of the token.
+        local result, err = security.get_sonos_oauth(true)
```
cjswedes: Can you quantify what the API pressure was before when compared with what it is now? We probably should get someone who wrote those apis to signoff on this. IIRC there was discussion of when force_refresh was appropriate to use, and I worry that this may be too aggressive and cause undue stress on our cloud and the sonos auth service.

NoahCornell: Talking with Ingvar it sounds like there are some reasons why we don't want to use force refresh here so I'll remove that commit because the other commit is still valid and a bigger issue. I'll open up a new PR to deal with the token expiration issue and missing capability commands.

cjswedes (approved):

PR #2351, opened 2025-08-21
sonos: Add persistent token refresher task

NoahCornell:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change fixes a potential issue where the driver's oauth token is expired and a capability command comes in before the hourly auth check (via SSDP event handler) happens. The command would not be handled because the driver doesn't have a valid token and there is no retry mechanism.

The crux of this change is adding a persistent token refresher task that tracks when the token expires and refreshes the token when it will expire in under a minute (Same value Viper uses when determining to refresh a token). 

The token refresher task has 3 actions that it can take based off the current state of the driver:
- If oauth is not connected, then the refresher will wait until it is connected. We cannot request tokens if this is not in a connected state.
- If oauth is connected and we have a valid token that is not set to expire within the next minute, then wait for the token to expire.
- If oauth is connected and we don't have a valid token, then request a new token and wait for it to arrive.

Commit by commit this change:
- Removes the previous preemptive token refreshing. This was not valid and would only refresh the token if the time to expire was within 1 minute.
- Add a new oauth endpoint app info bus to alert subscribers of new endpoint app info. This is used to alert the refresher when we go from a disconnected -> connected state.
- Add checks to the oauth token getter such that we only return a token if we are in a connected state. This avoids situation where we might be using an invalid token. Also adds the oauth_is_connected check to be used elsewhere.
- Add the token refresher task.
- Remove token requests from everywhere else in the driver. We don't want to be spamming this elsewhere if we are already on top of it.
- Add a wait for token function for the command handlers to use where it will wait for a token if it it should. It should be very rare where we actually need to wait.

# Summary of Completed Tests

I have tested some basic functionality but will come up with a test plan for thinktank and myself to run through.

### DIFF
--- a/drivers/SmartThings/sonos/src/api/cmd_handlers.lua
+++ b/drivers/SmartThings/sonos/src/api/cmd_handlers.lua
@@ -24,9 +24,9 @@ local function _do_send_to_group(driver, device, payload)
   local household_id, group_id = driver.sonos:get_group_for_device(device)
   payload[1].householdId = household_id
   payload[1].groupId = group_id
-  local maybe_token, err = driver:get_oauth_token()
+  local maybe_token, err = driver:wait_for_oauth_token(30)
   if err then
-    log.warn(string.format("notice: get_oauth_token -> %s", err))
+    log.warn(string.format("notice: wait_for_oauth_token -> %s", err))
   end
 
   if maybe_token then
@@ -40,9 +40,9 @@ local function _do_send_to_self(driver, device, payload)
   local household_id, player_id = driver.sonos:get_player_for_device(device)
   payload[1].householdId = household_id
   payload[1].playerId = player_id
-  local maybe_token, err = driver:get_oauth_token()
+  local maybe_token, err = driver:wait_for_oauth_token(30)
   if err then
-    log.warn(string.format("notice: get_oauth_token -> %s", err))
+    log.warn(string.format("notice: wait_for_oauth_token -> %s", err))
   end
 
   if maybe_token then
--- a/drivers/SmartThings/sonos/src/api/sonos_connection.lua
+++ b/drivers/SmartThings/sonos/src/api/sonos_connection.lua
@@ -246,11 +246,7 @@ end
 ---@param sonos_conn SonosConnection
 local function _oauth_reconnect_task(sonos_conn)
   log.debug("Spawning reconnect task for ", sonos_conn.device.label)
-  local check_auth = sonos_conn.driver:check_auth(sonos_conn.device)
-  local unauthorized = (check_auth == false)
-  if unauthorized and not sonos_conn.driver:is_waiting_for_oauth_token() then
-    sonos_conn.driver:request_oauth_token()
-  end
+  -- Subscribe first so we get all updates on the bus
   local token_receive_handle, err = sonos_conn.driver:oauth_token_event_subscribe()
   if not token_receive_handle then
     log.warn(string.format("error creating oauth token receive handle for respawn task: %s", err))
@@ -258,7 +254,10 @@ local function _oauth_reconnect_task(sonos_conn)
   cosock.spawn(function()
     local backoff = backoff_builder(60, 1, 0.1)
     while not sonos_conn:is_running() do
-      if sonos_conn.driver:is_waiting_for_oauth_token() and token_receive_handle then
+      local check_auth = sonos_conn.driver:check_auth(sonos_conn.device)
+      local unauthorized = (check_auth == false)
+
+      if unauthorized then
         local token, channel_error = token_receive_handle:receive()
         if not token then
           log.warn(string.format("Error requesting token: %s", channel_error))
@@ -275,12 +274,6 @@ local function _oauth_reconnect_task(sonos_conn)
         end
       end
 
-      check_auth = sonos_conn.driver:check_auth(sonos_conn.device)
-      unauthorized = (check_auth == false)
-
-      if unauthorized and not sonos_conn.driver:is_waiting_for_oauth_token() then
-        sonos_conn.driver:request_oauth_token()
-      end
       cosock.socket.sleep(backoff())
     end
     sonos_conn._reconnecting = false
@@ -346,10 +339,6 @@ function SonosConnection.new(driver, device)
             device.log.warn(
               string.format("WebSocket connection no longer authorized, disconnecting")
             )
-            local _, security_err = driver:request_oauth_token()
-            if security_err then
-              log.warn(string.format("Error during request for oauth token: %s", security_err))
-            end
             -- closing the socket directly without calling `:stop()` triggers the reconnect loop,
             -- which is where we wait for the token to come in.
             local unique_key, bad_key_part = utils.sonos_unique_key(household_id, player_id)
--- a/drivers/SmartThings/sonos/src/sonos_driver.lua
+++ b/drivers/SmartThings/sonos/src/sonos_driver.lua
@@ -222,6 +222,10 @@ end
 function SonosDriver:handle_startup_state_received()
   self:start_ssdp_event_task()
   self:notify_augmented_data_changed "snapshot"
+  if api_version >= 14 and security ~= nil then
+    local token_refresher = require "token_refresher"
+    token_refresher.spawn_token_refresher(self)
+  end
   self.startup_state_received = true
   for _, device in pairs(self.devices_waiting_for_startup_state) do
     SonosDriverLifecycleHandlers.initialize_device(self, device)
--- a/drivers/SmartThings/sonos/src/sonos_driver.lua
+++ b/drivers/SmartThings/sonos/src/sonos_driver.lua
@@ -25,8 +25,6 @@ if not security_load_success then
   security = nil
 end
 
-local ONE_HOUR_IN_SECONDS = 3600
-
 ---@class SonosDriver: Driver
 ---
 ---@field public datastore table<string, any> driver persistent store
@@ -384,13 +382,6 @@ function SonosDriver:get_oauth_token()
     local now = os.time()
     -- token has not expired yet
     if now < expiration then
-      -- token is expiring soon, so we pre-emptively refresh
-      if math.abs(expiration - now) < ONE_HOUR_IN_SECONDS then
-        local result, err = security.get_sonos_oauth()
-        if not result then
-          log.warn(string.format("Error requesting OAuth token via Security API: %s", err))
-        end
-      end
       return self.oauth.token
     else
       return nil, "token expired"
--- a/drivers/SmartThings/sonos/src/sonos_driver.lua
+++ b/drivers/SmartThings/sonos/src/sonos_driver.lua
@@ -353,6 +353,34 @@ function SonosDriver:get_oauth_token()
   return nil, "no token"
 end
 
+function SonosDriver:wait_for_oauth_token(timeout)
+  if api_version < 14 or security == nil then
+    return nil, "not supported"
+  end
+
+  if not self:oauth_is_connected() then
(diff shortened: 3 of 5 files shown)


drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -239,61 +239,13 @@ local function _legacy_reconnect_task(sonos_conn)
   end, string.format("%s Reconnect Task", sonos_conn.device.label))
 end
 
----@param sonos_conn SonosConnection
-local function _oauth_reconnect_task(sonos_conn)
```
NoahCornell, line 247: We could keep the logic of trying to reconnect after we receive a token after we determine that we are unauthorized but I was worried about potential race conditions where a token comes in before we have subscribed to the event bus and it didn't seem like the added complexity was worth it.

In most cases, we shouldn't be disconnecting because of auth anymore but there are some potential cases we could hit. I can look at adding this back if there are other opinions about this.

drivers/SmartThings/sonos/src/api/cmd_handlers.lua
```diff
@@ -24,9 +24,9 @@ local function _do_send_to_group(driver, device, payload)
   local household_id, group_id = driver.sonos:get_group_for_device(device)
   payload[1].householdId = household_id
   payload[1].groupId = group_id
-  local maybe_token, err = driver:get_oauth_token()
```
NoahCornell, line 27: 30 seconds for this timeout was just a starting point. It should be very difficult to hit a situation where a command comes in and we don't have a valid token when we should and the request would likely already be in flight.

drivers/SmartThings/sonos/src/token_refresher.lua
```diff
@@ -0,0 +1,128 @@
...
+    end
+  end
+  -- We don't have a valid token or it is about to expire soon.
+  return STATES.REQUEST_TOKEN
+end
+
+--- Waits for a token event with a timeout.
+function Refresher:receive_token(timeout)
+  local token_bus, _ = self.driver:oauth_token_event_subscribe()
+  if token_bus then
```
varzac, line 58: So I think the non-existent `else` block of this if kind of breaks some assumptions I would have about how this function and `wait_for_expire` work.  That is for this function I would expect that after calling this function, we will have a valid token, but it is not clear to me what state we would/should be in if token_bus is falsey.

drivers/SmartThings/sonos/src/token_refresher.lua
```diff
@@ -0,0 +1,128 @@
...
+    token_bus:settimeout(timeout)
+    token_bus:receive()
+  end
+end
+
+--- Waits for either a new token to arrive or the current one to expire.
+function Refresher:wait_for_expire()
```
varzac: The combination of this function name and the doc statement do not make it clear to me what the intended use of this is?  If I call `wait_for_expire` one of the potential return conditions is that we get a fresh token?  That doesn't seem to line up.  I think a more extensive doc comment, and maybe a different name would be better here.

drivers/SmartThings/sonos/src/token_refresher.lua
```diff
@@ -0,0 +1,128 @@
...
+  local now = os.time()
+  local timeout = math.max(expiration - now, 0)
+
+  log.debug(string.format("Token will refresh in %d seconds", timeout))
+  -- Wait while trying to receive the token in case it gets updated for some reason.
+  self:receive_token(timeout)
+end
+
+--- Waits for an oauth info event.
+function Refresher:disconnected()
```
varzac: again the function name here indicates more to me that we would be notifying something of being disconnected, not waiting for any info event.

NoahCornell: Yeah I was just trying to have the function names line up with the states but I see how that is unclear. I will fix them both up

drivers/SmartThings/sonos/src/token_refresher.lua
NoahCornell: Realistically this subscribe should never fail because this bus should never be closed and the api conditions must be true if we are in the token refresher, but it probably should handle it better with some sort of exit.

The name probably doesn't make sense after some changes I made and `wait_for_token_event` would be a more accurate name.

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -384,6 +337,35 @@ function SonosDriver:get_oauth_token()
   return nil, "no token"
 end
 
+function SonosDriver:wait_for_oauth_token(timeout)
```
cjswedes, line 356: Can you add some comments on the logic here for the non initiated with sonos and how the oauth works? The `maybe_token` looks to be the case when auth is already setup and we have a non expired token, but the `subscribe`, `receive` and then re-calling of `get_oauth_token` isnt obvious what that is accomplishing.

drivers/SmartThings/sonos/src/token_refresher.lua
```diff
@@ -0,0 +1,152 @@
...
+--- The likely outcome of this function is to wait the entire expiration timeout. It will
+--- also listen for token events just in case a new token with a new expiration is sent to the driver.
+--- A new token would most likely come from developer testing, but since the new token requests are
+--- not synchronous one could come from an earlier request.
+function Refresher:wait_for_expire_or_token_event()
+  local maybe_token, err = self.driver:get_oauth_token()
+  if not maybe_token then
+    -- Something got funky in the state machine
+    log.warn(string.format("Tried to wait for expiration of non-existent token: %s", err))
+    return
```
cjswedes, line 75: Should we change to `REQUEST_TOKEN`? I think that would help recover if this ever did happen

NoahCornell: We will go to the next correct action in the loop when we call `determine_action`.  It most likely will be `REQUEST_TOKEN` but technically could be any if we were disconnected or if we somehow receive a token on the next check. I'll add a comment here to make that more clear

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -384,6 +337,41 @@ function SonosDriver:get_oauth_token()
...
+  if api_version < 14 or security == nil then
+    return nil, "not supported"
+  end
+
+  if not self:oauth_is_connected() then
```
varzac: If I'm trying to wait for an oauth token wouldn't it be expected that I am not currently connected?

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -384,6 +337,41 @@ function SonosDriver:get_oauth_token()
...
+    token_bus:settimeout(timeout)
+    -- Wait for the new token to come in
+    token_bus:receive()
+    -- Call `SonosDriver:get_oauth_token` again to ensure the token is valid.
+    return self:get_oauth_token()
+  end
+  return nil, err
+end
+
+function SonosDriver:oauth_is_connected()
```
varzac: seeing what this is doing now, I understand a bit.  I wonder if a name like `oauth_app_configured` or `oauth_app_connected` would be a bit clearer.  It's definitely a nit though

NoahCornell: Updated

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -112,6 +112,15 @@ function SonosDriver:oauth_token_event_subscribe()
   return self.oauth_token_bus:subscribe()
 end
 
+---@return (cosock.Bus.Subscription)? receiver the subscription receiver if the bus hasn't been closed, nil if closed
```
hcarter-775, line 126: nit: `receive the subscription receiver if the bus hasn't been closed, nil if closed`
remove the r in the first word

drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -353,23 +346,14 @@ function SonosDriver:request_oauth_token()
 end
 
 ---@return { accessToken: string, expiresAt: number }? the token if a currently valid token is available, nil if not
----@return "token expired"|"no token"|"not supported"|nil reason the reason a token was not provided, nil if there is a valid token available
+---@return "token expired"|"no token"|"not supported"|"not connected"|nil reason the reason a token was not provided, nil if there is a valid token available
```
hcarter-775, line 332: nit: remove the word reason

drivers/SmartThings/sonos/src/api/sonos_connection.lua
hcarter-775: I do not have the context to fully grasp whether removing this reconnect logic is really worth it, so I'll defer to you here. This does seem like a relatively important ability though, no? Mostly checking up to do my due diligence

drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -239,61 +239,13 @@ local function _legacy_reconnect_task(sonos_conn)
...
   if sonos_conn._reconnecting then
     return
   end
   sonos_conn._reconnecting = true
-  if type(api_version) == "number" and api_version >= 14 then
-    _oauth_reconnect_task(sonos_conn)
-  else
-    _legacy_reconnect_task(sonos_conn)
-  end
+  _legacy_reconnect_task(sonos_conn)
```
hcarter-775: don't we want to keep this gated?

hcarter-775 (commented):
makes sense to me from a first pass, what is the current state of testing with this new refresher task framing?

drivers/SmartThings/sonos/src/sonos_driver.lua
NoahCornell: `receiver` here is the name of the return value per the lua language server `@return` annotation https://github.com/LuaLS/lua-language-server/wiki/Annotations#return

This looks a lot better with the language server extension installed with whatever IDE you are using.

drivers/SmartThings/sonos/src/sonos_driver.lua
NoahCornell: Same as above https://github.com/LuaLS/lua-language-server/wiki/Annotations#return

drivers/SmartThings/sonos/src/api/sonos_connection.lua
NoahCornell: Yeah thinking about it a little more, there is some potential to spam the device if oauth isn't connected so I'll look at adding this back

NoahCornell: I also forgot that we can make a websocket connection without being authorized so I think changing this back is the right call. I am less concerned about race conditions because I changed this so that we continue to subscribe to the token bus prior to spawning the task and now we do the auth check at the top of the loop so there should be no way of missing a new token event (which might not have been a problem before either)

NoahCornell:
> makes sense to me from a first pass, what is the current state of testing with this new refresher task framing?

I have had this running on my hub for awhile now with no issues. I have also been testing by injecting invalid tokens into the driver and the refresher task handles everything correctly.

There is a regression test plan that we can have thinktank run through once this and the other PR merges.

NoahCornell:
Same as the other PR, planning on merging this today so we can get thinktank to run through the regression test plan unless there are any objections @hcarter-775 @cjswedes  https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2302#issuecomment-3276336889

PR #2361, opened 2025-08-25
WWSTCERT-7632 LUX TQ1 Smart Thermostat

greens:


drivers/SmartThings/matter-thermostat/fingerprints.yml
```diff
@@ -38,6 +38,12 @@ matterManufacturer:
     vendorId: 0x1527
     productId: 0x0002
     deviceProfileName: thermostat-heating-only-batteryLevel
+  #Lux
+  - id: "4614/1"
+    deviceLabel: LUX TQ1 Smart Thermostat
+    vendorId: 0x1206
+    productId: 0x0001
+    deviceProfileName: thermostat-nostate-nobattery
```
hcarter-775, line 46: the actual profile this would be using is `thermostat-modular` since the 57 release of hubcore. we should probably get this re-fingerprinted.

greens: updated

drivers/SmartThings/matter-thermostat/fingerprints.yml
```diff
@@ -38,6 +38,12 @@ matterManufacturer:
     vendorId: 0x1527
     productId: 0x0002
     deviceProfileName: thermostat-heating-only-batteryLevel
+  #Lux
+  - id: "4614/1"
+    deviceLabel: LUX TQ1 Smart Thermostat
+    vendorId: 0x1206
+    productId: 0x0001
+    deviceProfileName: thermostat-modular
```
nickolas-deboom: For now we decided to not use modular profiles in fingerprints in order to support hubs on lower firmware (the driver will switch devices on supported hubs over to the modular profile)

PR #2366, opened 2025-08-25
Add modular profile support for new matter lock devices

hcarter-775:
# Description of Change
Create the `lock-modular` profile and add handling for new modular profiling. To allow for an embedded device config to be conditionally loaded, there are 2 identical `modular` profiles being introduced. This second one may be removed as the question of moving this logic to the plugin is revisited in the future.

Related to the onboarding flow but not particular to the modular profiling, this PR also changes the PowerSource AttributeList read logic in this driver to match handling seen in other matter drivers, moving the read to the `added` call and handling the result in a `match_profile` function that has been introduced.

# Summary of Completed Tests
- Unit tests updated to ensure regression tests on static profile switching works.
- An array of new unit tests were added to account for modular profiling.
- The VDA runs as expected, as does the Aqara U400. The correct capabilities are added and subscribed to in each case, and all capabilities are controllable.

### DIFF
--- a/drivers/SmartThings/matter-lock/profiles/lock-modular-embedded-unlatch.yml
+++ b/drivers/SmartThings/matter-lock/profiles/lock-modular-embedded-unlatch.yml
@@ -1,0 +1,131 @@
+name: lock-modular-embedded-unlatch
+components:
+- id: main
+  capabilities:
+    - id: lock
+      version: 1
+    - id: lockAlarm
+      version: 1
+    - id: remoteControlStatus
+      version: 1
+    - id: lockUsers
+      version: 1
+      optional: true
+    - id: lockCredentials
+      version: 1
+      optional: true
+    - id: lockSchedules
+      version: 1
+      optional: true
+    - id: battery
+      version: 1
+      optional: true
+    - id: batteryLevel
+      version: 1
+      optional: true
+    - id: firmwareUpdate
+      version: 1
+    - id: refresh
+      version: 1
+  categories:
+    - name: SmartLock
+deviceConfig:
+  dashboard:
+    states:
+      - component: main
+        capability: lock
+        version: 1
+    actions:
+      - component: main
+        capability: lock
+        version: 1
+        visibleCondition: {
+          "capability": "lock",
+          "version": "1",
+          "component": "main",
+          "value": "lock.value",
+          "operator": "DOES_NOT_EQUAL",
+          "operand": "unlatched"
+        }
+  detailView:
+    - component: main
+      capability: lock
+      version: 1
+      values:
+        - key: lock.value
+          alternatives:
+            - key: locked
+              type: inactive
+              value: '{{i18n.attributes.lock.i18n.value.locked.label}}'
+            - key: unlocked
+              value: '{{i18n.attributes.lock.i18n.value.unlocked.label}}'
+            - key: unlatched
+              value: '{{i18n.attributes.lock.i18n.value.unlatched.label}}'
+            - key: not fully locked
+              value: '{{i18n.attributes.lock.i18n.value.not fully locked.label}}'
+      patch:
+        - op: add
+          path: /1
+          value:
+            capability: lock
+            version: 1
+            component: main
+            label: '{{i18n.commands.unlatch.label}}'
+            displayType: pushButton
+            pushButton:
+              command: unlatch
+    - component: main
+      capability: lockAlarm
+      version: 1
+    - component: main
+      capability: remoteControlStatus
+      version: 1
+    - component: main
+      capability: battery
+      version: 1
+    - component: main
+      capability: batteryLevel
+      version: 1
+  automation:
+    conditions:
+      - component: main
+        capability: lock
+        version: 1
+        values:
+          - key: lock.value
+            alternatives:
+              - key: locked
+                type: inactive
+                value: '{{i18n.attributes.lock.i18n.value.locked.label}}'
+              - key: unlocked
+                value: '{{i18n.attributes.lock.i18n.value.unlocked.label}}'
+              - key: unlatched
+                value: '{{i18n.attributes.lock.i18n.value.unlatched.label}}'
+              - key: not fully locked
+                value: '{{i18n.attributes.lock.i18n.value.not fully locked.label}}'
+      - component: main
+        capability: lockAlarm
+        version: 1
+      - component: main
+        capability: remoteControlStatus
+        version: 1
+      - component: main
+        capability: battery
+        version: 1
+      - component: main
+        capability: batteryLevel
+        version: 1
+    actions:
+    - component: main
+      capability: lock
+      version: 1
+      values:
+        - key: '{{enumCommands}}'
+          alternatives:
+            - key: lock
+              type: inactive
+              value: '{{i18n.commands.lock.label}}'
+            - key: unlock
+              value: '{{i18n.commands.unlock.label}}'
+            - key: unlatch
+              value: '{{i18n.commands.unlatch.label}}'
--- a/drivers/SmartThings/matter-lock/profiles/lock-modular.yml
+++ b/drivers/SmartThings/matter-lock/profiles/lock-modular.yml
@@ -1,0 +1,31 @@
+name: lock-modular
+components:
+- id: main
+  capabilities:
+    - id: lock
+      version: 1
+    - id: lockAlarm
+      version: 1
+    - id: remoteControlStatus
+      version: 1
+    - id: lockUsers
+      version: 1
+      optional: true
+    - id: lockCredentials
+      version: 1
+      optional: true
+    - id: lockSchedules
+      version: 1
+      optional: true
+    - id: battery
+      version: 1
+      optional: true
+    - id: batteryLevel
+      version: 1
+      optional: true
+    - id: firmwareUpdate
+      version: 1
+    - id: refresh
+      version: 1
+  categories:
+    - name: SmartLock
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -55,7 +55,15 @@ local NEW_MATTER_LOCK_PRODUCTS = {
   {0x10E1, 0x2002}  -- VDA
 }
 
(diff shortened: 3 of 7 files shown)


drivers/SmartThings/matter-lock/profiles/lock-modular.yml
```diff
@@ -0,0 +1,127 @@
...
+      path: /1
+      value:
+        capability: lock
+        version: 1
+        component: main
+        label: '{{i18n.commands.unlatch.label}}'
+        displayType: pushButton
+        pushButton:
+          command: unlatch
+
```
HunsupJung: Needs to remove empty line.

drivers/SmartThings/matter-lock/profiles/lock-modular.yml
```diff
@@ -0,0 +1,127 @@
...
+  - component: main
+    capability: lockAlarm
+    version: 1
+  - component: main
+    capability: remoteControlStatus
+    version: 1
+  - component: main
+    capability: batteryLevel
+    version: 1
+
```
HunsupJung: Needs to remove empty line

drivers/SmartThings/matter-lock/profiles/lock-modular.yml
hcarter-775: done

drivers/SmartThings/matter-lock/profiles/lock-modular.yml
hcarter-775: done

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -397,27 +482,23 @@ end
...
-      support_battery_level = true
-    end
-  end
-  local profile_name = device:get_field(PROFILE_BASE_NAME)
-  if profile_name ~= nil then
-    if support_battery_percentage then
-      profile_name = profile_name .. "-battery"
-    elseif support_battery_level then
-      profile_name = profile_name .. "-batteryLevel"
+      device:set_field(profiling_data.BATTERY_SUPPORT, battery_support.BATTERY_PERCENTAGE, { persist = true })
```
hcarter-775, line 489: here and elsewhere I have persisted this field- this comes from the thinking that it is a relatively small per-device cost in the driver, and that not doing so would require a re-read of the attribute list anytime the device may be re-profiled for any reason in the future, which we lack the code infrastructure to easily do at the moment (though this could easily be added).

drivers/SmartThings/matter-lock/src/test/test_matter_lock_unlatch.lua
```diff
@@ -55,23 +56,27 @@ local mock_device = test.mock_device.build_test_matter_device({
...
 local function test_init()
   test.disable_startup_messages()
+  -- subscribe_request
+  local subscribe_request = DoorLock.attributes.LockState:subscribe(mock_device)
+  subscribe_request:merge(DoorLock.attributes.OperatingMode:subscribe(mock_device))
+  subscribe_request:merge(DoorLock.events.LockOperation:subscribe(mock_device))
+  subscribe_request:merge(DoorLock.events.DoorLockAlarm:subscribe(mock_device))
+  -- add test device, handle initial subscribe
   test.mock_device.add_test_device(mock_device)
+  test.socket.matter:__expect_send({mock_device.id, subscribe_request})
```
nickolas-deboom, line 66: Why is there an extra subscribe here now vs just the one in `init`?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -148,15 +156,64 @@ local function device_init(driver, device)
...
+
+  local supported_battery_type = device:get_field(profiling_data.BATTERY_SUPPORT)
+  if supported_battery_type == battery_support.BATTERY_LEVEL then
+    table.insert(main_component_capabilities, capabilities.batteryLevel.ID)
+  elseif supported_battery_type == battery_support.BATTERY_PERCENTAGE then
+    table.insert(main_component_capabilities, capabilities.battery.ID)
+  end
+
+  table.insert(enabled_optional_component_capability_pairs, {"main", main_component_capabilities})
+  device:try_update_metadata({profile = modular_profile_name, optional_component_capabilities = enabled_optional_component_capability_pairs})
```
nickolas-deboom, line 208: I believe `SUPPORTED_COMPONENT_CAPABILITIES` needs to be set here so that it can be used in `supports_capability_by_id_modular`

nickolas-deboom: Also to add mandatory capabilities to the list so that they are added to the subscription

drivers/SmartThings/matter-lock/src/test/test_matter_lock_unlatch.lua
hcarter-775: I changed the rpc version for that test to ensure the modular logic was not touched, so therefore another subscribe was needed to handle the initial subscribes from an `add_test_device` since it stopped using some of the new logic.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: since this modular logic is gated behind fw 0.58 api/rpc version, we don't actually need either of the driver workarounds you are describing anymore. As of 0.58, this optional data is now stored in the hub inventories, and only the "correct" capabilities per device are synced to the lua libs/drivers.

nickolas-deboom: Got it, thanks!

drivers/SmartThings/matter-lock/profiles/lock-modular-embedded-unlatch.yml
```diff
@@ -0,0 +1,131 @@
+name: lock-modular-embedded-unlatch
```
hcarter-775, line 1: adding this embedded config data to the plugin directly was discussed, but the implementation details have been deemed complex for the moment, so this workaround has been implemented here. I do believe it is worth discussing further putting this in the app long term.

CC: @HunsupJung would you mind discussing this further with the plugin team? Perhaps this can be included in the r4 app? Let me know what you think about this!

drivers/SmartThings/matter-lock/profiles/lock-modular-embedded-unlatch.yml
```diff
@@ -0,0 +1,131 @@
...
+              - key: not fully locked
+                value: '{{i18n.attributes.lock.i18n.value.not fully locked.label}}'
+      - component: main
+        capability: lockAlarm
+        version: 1
+      - component: main
+        capability: remoteControlStatus
+        version: 1
+      - component: main
+        capability: battery
```
nickolas-deboom, line 113: I forget, do capabilities listed in `detailView` play nicely with disabled optional capabilities? For example if `battery` is not enabled, would it still show up under automation conditions?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -397,27 +482,23 @@ end
...
+    elseif attr.value == 0x0E then
+      device:set_field(profiling_data.BATTERY_SUPPORT, battery_support.BATTERY_LEVEL, { persist = true })
+      match_profile(driver, device)
+      return
     end
-    device.log.info_with({hub_logs=true}, string.format("Updating device profile to %s.", profile_name))
-    device:try_update_metadata({profile = profile_name})
   end
+
+  -- neither BatChargeLevel not BatPercentRemaining were found. Re-profiling without battery.
```
ctowns: nit:
```suggestion
  -- neither BatChargeLevel nor BatPercentRemaining were found. Re-profiling without battery.
```

drivers/SmartThings/matter-lock/profiles/lock-modular-embedded-unlatch.yml
hcarter-775: yes, it does. I checked it out with the VDA and only battery is shown (not batteryLevel, which isn't supported on this particular formatting).

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: fixed in latest (squashed) commit

drivers/SmartThings/matter-lock/profiles/lock-modular-embedded-unlatch.yml
nickolas-deboom: Ok nice!

nickolas-deboom (approved):
LGTM!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -174,15 +231,16 @@ local function do_configure(driver, device)
...
-    device:try_update_metadata({profile = profile_name})
+
+  local supported_battery_type = device:get_field(profiling_data.BATTERY_SUPPORT)
+  if supported_battery_type == battery_support.BATTERY_LEVEL then
+    profile_name = profile_name .. "-batteryLevel"
+  elseif supported_battery_type == battery_support.BATTERY_PERCENTAGE then
+    profile_name = profile_name .. "-battery"
   end
+
+  device.log.info_with({hub_logs=true}, string.format("Updating device profile to %s.", profile_name))
```
ctowns, line 242: nit: you don't need to remove this for now for the sake of older FW, but just confirming - this information will be included by default in logs with hubs using 58+ FW moving forward, right?

hcarter-775: yes, this is now part of the default logs. As you noted, I'm leaving it for older FW for the time being, but eventually we can discuss stripping this out of all Matter drivers.

drivers/SmartThings/matter-lock/profiles/lock-modular-embedded-unlatch.yml
HunsupJung: It's not easy to persuade the Plugin team.
They commented that Plugin could not handle the distinguishing the unlatch or not for UX.

PR #2389, opened 2025-09-10
philips-hue: Split mapping of hue resource IDs by bridge

NoahCornell:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

This change is to help facilitate the migration of an old hue bridge to the pro hue bridge. Currently, the driver holds a mapping of hue resource IDs to smartthings devices records. This becomes a problem when adding the new pro to smartthings because the children of the pro will have the same resource ids as they did when on the old bridge. Child creation is gated by if we already have a record of that hue resource ID. The driver also makes assumptions throughout that the hue resource ID is unique.

By splitting up the mapping by hue bridge, we shouldn't run into any issues with collusions and should be able to re-create the devices on the new pro bridge.

# Summary of Completed Tests

By hue setup is currently offline so I haven't had a chance to test this yet.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/disco/init.lua
+++ b/drivers/SmartThings/philips-hue/src/disco/init.lua
@@ -334,8 +334,13 @@ function HueDiscovery.handle_discovered_child_device(driver, bridge_network_id, 
   end
 
   for _, svc_info in ipairs(primary_services[primary_service_type]) do
+    if driver:get_device_by_dni(v1_dni) then
+      return
+    end
     local v2_resource_id = svc_info.rid or ""
-    if driver:get_device_by_dni(v1_dni) or driver.hue_identifier_to_device_record[v2_resource_id] then return end
+    if (driver.hue_identifier_to_device_record_by_bridge[bridge_network_id] or {})[v2_resource_id] then
+      return
+    end
   end
 
   local api_instance =
--- a/drivers/SmartThings/philips-hue/src/utils/init.lua
+++ b/drivers/SmartThings/philips-hue/src/utils/init.lua
@@ -345,6 +345,7 @@ end
 ---@param driver HueDriver
 ---@param device HueDevice
 ---@param parent_device_id string?
+---@param quiet boolean?
 ---@return HueBridgeDevice? bridge_device
 function utils.get_hue_bridge_for_device(driver, device, parent_device_id, quiet)
   local _ = quiet or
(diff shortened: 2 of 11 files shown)


drivers/SmartThings/philips-hue/src/disco/init.lua
```diff
@@ -334,8 +334,13 @@ function HueDiscovery.handle_discovered_child_device(driver, bridge_network_id,
   end
 
   for _, svc_info in ipairs(primary_services[primary_service_type]) do
+    if driver:get_device_by_dni(v1_dni) then
+      return
+    end
     local v2_resource_id = svc_info.rid or ""
-    if driver:get_device_by_dni(v1_dni) or driver.hue_identifier_to_device_record[v2_resource_id] then return end
```
NoahCornell, line 338: This is what is currently gating us from creating the devices, specifically the last part of the conditional

drivers/SmartThings/philips-hue/src/utils/init.lua
```diff
@@ -371,6 +372,27 @@ function utils.get_hue_bridge_for_device(driver, device, parent_device_id, quiet
...
+  if not bridge then
+    log.warn(string.format("Failed to lookup bridge for device %s", bridge_or_device.label))
+    return nil
+  end
+  local bridge_id = bridge:get_field(Fields.BRIDGE_ID)
+  if not bridge_id then
+    log.warn(string.format("Failed to get bridge id for %s", bridge.label))
+    return nil
+  end
+  driver.hue_identifier_to_device_record_by_bridge[bridge_id] = driver.hue_identifier_to_device_record_by_bridge[bridge_id] or {}
```
nickolas-deboom, line 392: Do we want to log if `driver.hue_identifier_to_device_record_by_bridge[bridge_id]` is nil?

NoahCornell: I don't think so since it would be expected the first time that we try to get a table for a certain bridge

NoahCornell:
Tested on my setup and things are looking good. Batch commands and offline/online tracking via zigbee connectivity status are working correctly.

This just needs testing with the migration path to make sure the new devices are created.

nickolas-deboom:
I tested the migration path and things are looking good on that front as well. The presence of the old hub and devices no longer prevents the creation of bridged devices on the new hub.

PR #2392, opened 2025-09-11
sonos: Fix oauth account linking notifications

NoahCornell:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

This was a miss by me that I thought I had tested in the token refresher PR that was found when during regression testing. When we request a token, if the call returns a 404, then we will make another api call which will trigger a notification to the user guiding them to link their sonos account.

This change adds a function `SonosDriver:alert_unauthorized()` that will go through the token request flow and trigger a notification to the user. This is called when the websocket closes due to being unauthorized, when we determine we are unauthorized in the websocket reconnect task, and in the device init lifecycle handler task when we determine we are unauthorized.

This is done outside of the token refresher task because instead of calling `security.get_sonos_oauth()` with the intent to get a new token, the intent is strictly to notify the user why their device is offline.

`SonosDriver:alert_unauthorized()` will call `security.get_sonos_oauth()` regardless of the oauth connection state because there is a possibility that there is stale data in the augmented driver data table. That flow would look something like this, user onboards sonos speaker with linked sonos account, user unlinks sonos account and device is deleted from smartthings, user re-onboards device via scan nearby, oauth information is still in augmented data so we can connect to the device correctly until the token expires. This may lead to an extra call to `security.get_sonos_oauth()` but doing that outweighs the possibility of the notification not appearing. I can create a ticket to look at clearing out the driver augmented data when a driver has no more devices.

# Summary of Completed Tests

I have tested that the notification now appears when we get disconnected from the device due to auth and on driver startup when we determine we are not authorized.

### DIFF
--- a/drivers/SmartThings/sonos/src/sonos_driver.lua
+++ b/drivers/SmartThings/sonos/src/sonos_driver.lua
@@ -40,6 +40,7 @@ end
 ---@field private oauth { token: {accessToken: string, expiresAt: number}, endpoint_app_info: { state: "connected"|"disconnected" }, force_oauth: boolean? } cached OAuth info
 ---@field private startup_state_received boolean
 ---@field private devices_waiting_for_startup_state SonosDevice[]
+---@field private have_alerted_unauthorized boolean Used to track if we have requested an oauth token, this will trigger the notification used for account linking
 ---@field package bonded_devices table<string, boolean> map of Device device_network_id to a boolean indicating if the device is currently known as a bonded device.
 ---
 ---@field public ssdp_task SonosPersistentSsdpTask?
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -644,6 +664,7 @@ function SonosDriver.new_driver_template()
     oauth_token_bus = oauth_token_bus,
     oauth_info_bus = oauth_info_bus,
     oauth = {},
+    have_alerted_unauthorized = false,
```
cjswedes, line 667: Is the intention that this notification can happen at most once per driver restart?

cjswedes (approved):

drivers/SmartThings/sonos/src/sonos_driver.lua
NoahCornell: Yeah, from my testing it looks like subsequent calls don't go through the trigger preinstall flow due to the error returned from the get_token api changing so just calling it once per driver restart should be sufficient.

PR #2401, opened 2025-09-16
Matter Switch: Refactor Driver File Organization

hcarter-775:
# Description of Change
This PR aims to re-organize the Matter Switch driver to be easier to read and more self-documenting in its nature. 

It does _not_ change any current handling at all, and only re-organizes where the variables and functions are defined, that is, moving them under different files and meta-tables.

The general directory structure is as follows, within `src/`:
- `embedded-clusters`
  - all embedded cluster directories added for backwards compatibility
- `generic-handlers`
  - an event_handlers, attribute_handlers, and capability_handlers file, with some functions renamed
- `sub-drivers`
  - all current sub-drivers, just placed under this directory 
- `utils`
  - move embedded-cluster-utils and color-utils here
  - add a switch_fields file with all global vars, a switch_utils file with nearly all non-handler functions, and a device_configuration file with the remaining non-handler functions that relate to device profiling.

# Summary of Completed Tests
Ensured capability and attribute handlers ran on the following VDA device types: 
- OnOff Switch, 
- Dimmer Switch, 
- Color Temp Light, 
- Extended Color Light, 
- Mounted OnOff Control, 
- Mounted Dimmable Load Control
- Generic Switch

This testing does not handle an exhaustive list of all handlers that could be hit since this driver includes Electrical Sensor and MCD devices that have illuminance, occupancy, motion, fans, etc. To try every combination is out of scope for this refactor. However, this spot testing, plus a series of unit tests hitting every handler ensures that no mapping between handler and attribute was missed, nor that anything was unduly erased during the refactor. Further, I have restarted the driver multiple times with some of these device types (OnOff Switch, Dimmer Switch) and have seen functionality remain constant.

Since this just moves code around, the primary worry here is that something was deleted or that some typo has come up, which breaks some piece of the code. However, since all tested devices do continue to work and all unit tests continue to pass, I am confident that this has not happened. 

Further, I have scanned each handler manually and using inline VScode toolsand ensured each map is expected. I have also manually stepped through all introduced `fields.blank` and `switch_utils.blank` calls and ensured they all map to something meaningful.

Last, I have been able to confirm through unit tests with a backend of older lua libs that the embedded cluster definitions continue to run as expected.


greens (commented):
quick question: Are there any hubcore versions out there in production that don't have these clusters in their lua_libs still?

drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
greens, line 1: missing copyright statement

hcarter-775:
> quick question: Are there any hubcore versions out there in production that don't have these clusters in their lua_libs still?

Yes, there is a non-zero percentage. We are still watching these numbers to see when is appropriate to remove these altogether, but for now have elected to keep them in the driver.

drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
hcarter-775: added

greens (approved):

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -12,704 +12,97 @@
...
-    clusters.Switch.events.ShortRelease,
-    clusters.Switch.events.MultiPressComplete
-  },
-  [ELECTRICAL_SENSOR_ID] = {
-    clusters.ElectricalPowerMeasurement.attributes.ActivePower,
-    clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyImported,
-    clusters.ElectricalEnergyMeasurement.attributes.PeriodicEnergyImported
-  }
-}
+local attribute_handlers = require "generic-handlers.attribute_handlers"
```
cjswedes: nit: IMO we should have our sub folders using snake_case, but its sort of the wild west; the lua libs follow this, but we dont for the driver and subdriver directory names ¯\\_(ツ)_/¯

cjswedes (approved):

drivers/SmartThings/matter-switch/src/utils/embedded_cluster_utils.lua
```diff
@@ -1,12 +1,26 @@
+-- Copyright 2025 SmartThings
```
nickolas-deboom, line 1: Could you rename this file to snake case to match the other utils files?

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: 100% agree, just didn't want to rock the boat too hard

drivers/SmartThings/matter-switch/src/utils/embedded_cluster_utils.lua
hcarter-775: done

ctowns (approved):
This looks good to me - could you provide a little more commentary on the testing you did in the `Summary of Completed Testing` section. Did your spot testing include testing all capabilities, driver restart, etc.? Also, do we have coverage of all device types (have we tested an energy device)? If not, I would call out if any device types are covered by unit tests only. But, the refactor makes sense to me and I assume the unit tests would fail if there was any obvious problem, so overall I think this looks good to me.

Great job with the refactor, this looks very well organized!

drivers/SmartThings/matter-switch/src/generic_handlers/capability_handlers.lua
```diff
@@ -0,0 +1,182 @@
...
+local st_utils = require "st.utils"
+local switch_utils = require "utils.switch_utils"
+local fields = require "utils.switch_fields"
+local version = require "version"
+
+local CapabilityHandlers = {}
+
+-- Include driver-side definitions when lua libs api version is < 11
+if version.api < 11 then
+  clusters.ValveConfigurationAndControl = require "embedded_clusters.ValveConfigurationAndControl"
```
nickolas-deboom, line 26: This require statement and others for embedded clusters fail for some reason, saying `Received unexpected error: ./embedded_clusters/ValveConfigurationAndControl/init.lua:2: module 'ValveConfigurationAndControl.server.attributes' not found`

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -12,704 +12,97 @@
...
+local SwitchLifecycleHandlers = {}
 
-      if supportedButtonValues_event then
-        device:emit_event_for_endpoint(ep, supportedButtonValues_event)
-      end
-      device:emit_event_for_endpoint(ep, capabilities.button.button.pushed({state_change = false}))
-    else
-      device.log.info_with({hub_logs=true}, string.format("Component not found for generic switch endpoint %d. Skipping Supported Value configuration", ep))
-    end
+function SwitchLifecycleHandlers.device_added(driver, device)
```
nickolas-deboom, line 43: Can't these just be local functions?

drivers/SmartThings/matter-switch/src/generic_handlers/capability_handlers.lua
nickolas-deboom: Ah, the embedded cluster definitions just need to be updated to reflect the new path, i.e. 
`local ValveConfigurationAndControlServerAttributes = require "ValveConfigurationAndControl.server.attributes"`
should be 
`local ValveConfigurationAndControlServerAttributes = require "embedded_clusters.ValveConfigurationAndControl.server.attributes"` [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/0bf8794d73c79b4f3519cb02611168d03819139d/drivers/SmartThings/matter-switch/src/embedded_clusters/ValveConfigurationAndControl/init.lua#L2)

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: they can be, but I think it kinda clarifies what they are to give them this prefix. A sorta self-documenting code system imo

drivers/SmartThings/matter-switch/src/generic_handlers/capability_handlers.lua
hcarter-775: fixed

PR #2411, opened 2025-09-18
Remove the adding of monitored attributes in Zigbee drivers

varzac:
This feature is already disabled on these drivers and is moving toward deprecation.  We will remove the unnecessary adding


drivers/SmartThings/zigbee-motion-sensor/src/frient/init.lua
```diff
@@ -111,22 +111,18 @@ local function device_init(driver, device)
   local attribute
   attribute = CONFIGURATIONS[OCCUPANCY_ENDPOINT]
```
greens:
```
Checking drivers/SmartThings/zigbee-motion-sensor/src/frient/init.lua 1 warning

    drivers/SmartThings/zigbee-motion-sensor/src/frient/init.lua:112:3: (W311) value assigned to variable attribute is unused
```

PR #2418, opened 2025-09-22
Matter Energy: Fix non- JSON serializable fields

nickolas-deboom:
Keys in datastore tables need to be strings in order to be JSON serializable. `SUPPORTED_EVSE_MODES_MAP` and
`SUPPORTED_DEVICE_ENERGY_MANAGEMENT_MODES_MAP` need to be updated.

### DIFF
--- a/drivers/SmartThings/matter-energy/src/init.lua
+++ b/drivers/SmartThings/matter-energy/src/init.lua
@@ -512,7 +512,7 @@ local function energy_evse_supported_modes_attr_handler(driver, device, ib, resp
     end
     table.insert(supportedEvseModes, mode.elements.label.value)
   end
-  supportedEvseModesMap[ib.endpoint_id] = supportedEvseModes
+  supportedEvseModesMap[tostring(ib.endpoint_id)] = supportedEvseModes
   device:set_field(SUPPORTED_EVSE_MODES_MAP, supportedEvseModesMap, { persist = true })
   local event = capabilities.mode.supportedModes(supportedEvseModes, { visibility = { displayed = false } })
   device:emit_event_for_endpoint(ib.endpoint_id, event)
@@ -521,10 +521,8 @@ local function energy_evse_supported_modes_attr_handler(driver, device, ib, resp
 end
 
 local function energy_evse_mode_attr_handler(driver, device, ib, response)
-  device.log.info(string.format("energy_evse_modes_attr_handler currentMode: %s", ib.data.value))
-
   local supportedEvseModesMap = device:get_field(SUPPORTED_EVSE_MODES_MAP) or {}
-  local supportedEvseModes = supportedEvseModesMap[ib.endpoint_id] or {}
+  local supportedEvseModes = supportedEvseModesMap[tostring(ib.endpoint_id)] or {}
   local currentMode = ib.data.value
   for i, mode in ipairs(supportedEvseModes) do
     if i - 1 == currentMode then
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-energy/src/init.lua
```diff
@@ -512,7 +512,7 @@ local function energy_evse_supported_modes_attr_handler(driver, device, ib, resp
     end
     table.insert(supportedEvseModes, mode.elements.label.value)
   end
-  supportedEvseModesMap[ib.endpoint_id] = supportedEvseModes
+  supportedEvseModesMap[tostring(ib.endpoint_id)] = supportedEvseModes
```
ctowns: I think these changes make sense and address the immediate issue, but there is something that I think we could improve here: would we be able to use the `set_field_for_endpoint` and `get_field_for_endpoint` commands here instead of directly saving all this endpoint information to a table?

The `set/get_field_for_endpoint` command are already used throughout our other drivers, and in those cases we don't run into the issue of the index being a string because the "stringification" of the key value is handled by the functions themselves. Here are the functions copied from the `matter-switch` driver (although they are used in many drivers):
```
local function get_field_for_endpoint(device, field, endpoint)
  return device:get_field(string.format("%s_%d", field, endpoint))
end

local function set_field_for_endpoint(device, field, endpoint, value, additional_params)
end
```

nickolas-deboom: That would definitely be nice, I just wonder if that would cause issues given that it would be a different format than what is currently used on devices in the field. I supposed that it could be avoided by starting with an empty list in the supported modes handlers rather than adding to the existing lists.

nickolas-deboom: Come to think of it, maybe that's how it should be done, since, although probably unlikely, a device could potentially not support a particular mode after supporting it previously, perhaps following a FW update. So it might make sense to start fresh each time the `SupportedModes` attributes are received.

nickolas-deboom: Never mind, the driver already accounts for that. But still, I like the idea of using get/set_field_for_endpoint and will add an additional commit to this PR with an implementation

nickolas-deboom: Updated in abc9d8647bbc99b520c0c2f52a32f121906ff256

drivers/SmartThings/matter-energy/src/init.lua
```diff
@@ -738,8 +751,7 @@ local function handle_set_mode_command(driver, device, cmd)
     end,
     ["deviceEnergyManagement"] = function( ... )
       local ep = component_to_endpoint(device, cmd.component)
-      local supportedDeviceEnergyMgmtModesMap = device:get_field(SUPPORTED_DEVICE_ENERGY_MANAGEMENT_MODES_MAP)
-      local supportedDeviceEnergyMgmtModes = supportedDeviceEnergyMgmtModesMap[ep] or {}
+      local supportedDeviceEnergyMgmtModes = get_field_for_endpoint(device, SUPPORTED_DEVICE_ENERGY_MANAGEMENT_MODES, ep)
```
hcarter-775:
```suggestion
      local supportedDeviceEnergyMgmtModes = get_field_for_endpoint(device, SUPPORTED_DEVICE_ENERGY_MANAGEMENT_MODES, ep) or {}
```

drivers/SmartThings/matter-energy/src/init.lua
```diff
@@ -726,8 +740,7 @@ local function handle_set_mode_command(driver, device, cmd)
   local set_mode_handlers = {
     ["main"] = function( ... )
       local ep = component_to_endpoint(device, cmd.component)
-      local supportedEvseModesMap = device:get_field(SUPPORTED_EVSE_MODES_MAP)
-      local supportedEvseModes = supportedEvseModesMap[ep] or {}
+      local supportedEvseModes = get_field_for_endpoint(device, SUPPORTED_EVSE_MODES, ep)
```
hcarter-775:
```suggestion
      local supportedEvseModes = get_field_for_endpoint(device, SUPPORTED_EVSE_MODES, ep) or {}
```

hcarter-775 (approved):
spot test with VDA and I think this is good to go

drivers/SmartThings/matter-energy/src/init.lua
nickolas-deboom: Done

nickolas-deboom:
> spot test with VDA and I think this is good to go

Tested with Battery Storage, Solar Power, and EVSE device types. Everything works fine except energyMeter is not populated because of read failures from the VDA for ElectricalEnergyMeasurement. Since this is a VDA issue, the behavior is the same on the production driver and the issue is unrelated to this change.

PR #2422, opened 2025-09-23
Add the processing logic of motion sensor on button devices

Cornelius117:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change
In the past, the button devices with motion sensor did not support their complete functions. Now add relevant processing logic, so that this kind of devices can display the sensor normally in the plugin.

# Summary of Completed Tests
The unit test file is contained in the commit, and all tests can pass.

### DIFF
--- a/drivers/SmartThings/matter-switch/profiles/3-button-motion.yml
+++ b/drivers/SmartThings/matter-switch/profiles/3-button-motion.yml
@@ -26,9 +26,5 @@ components:
     capabilities:
       - id: motionSensor
         version: 1
-      - id: firmwareUpdate
-        version: 1
-      - id: refresh
-        version: 1
     categories:
-    - name: MotionSensor        
+    - name: MotionSensor
--- a/drivers/SmartThings/matter-switch/profiles/6-button-motion.yml
+++ b/drivers/SmartThings/matter-switch/profiles/6-button-motion.yml
@@ -44,9 +44,5 @@ components:
     capabilities:
       - id: motionSensor
         version: 1
-      - id: firmwareUpdate
-        version: 1
-      - id: refresh
-        version: 1   
     categories:
     - name: MotionSensor
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -558,8 +558,8 @@ local function find_child(parent, ep_id)
   return parent:get_child_by_parent_assigned_key(string.format("%d", ep_id))
 end
 
-local function build_button_component_map(device, main_endpoint, button_eps)
-  -- create component mapping on the main profile button endpoints
+local function build_button_component_map(device, main_endpoint, button_eps, motion_eps)
+  -- create component mapping on the main profile button endpoints and motion endpoint
   table.sort(button_eps)
   local component_map = {}
   component_map["main"] = main_endpoint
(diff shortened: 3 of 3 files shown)


drivers/SmartThings/matter-switch/profiles/3-button-motion.yml
```diff
@@ -0,0 +1,34 @@
...
+    capabilities:
+      - id: button
+        version: 1
+    categories:
+    - name: RemoteController
+  - id: motion
+    capabilities:
+      - id: motionSensor
+        version: 1
+      - id: firmwareUpdate
```
nickolas-deboom: `firmwareUpdate` and `refresh` can be removed from the motion component here, since they are already in the main component.

drivers/SmartThings/matter-switch/profiles/6-button-motion.yml
```diff
@@ -0,0 +1,52 @@
...
+    categories:
+    - name: RemoteController
+  - id: motion
+    capabilities:
+      - id: motionSensor
+        version: 1
+      - id: firmwareUpdate
+        version: 1
+      - id: refresh
+        version: 1
```
nickolas-deboom: Same thing here.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -651,15 +662,20 @@ end
 local function initialize_buttons_and_switches(driver, device, main_endpoint)
   local profile_found = false
   local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
+  local motion_eps = device:get_endpoints(clusters.OccupancySensing.ID)
   if tbl_contains(STATIC_BUTTON_PROFILE_SUPPORTED, #button_eps) then
-    build_button_profile(device, main_endpoint, #button_eps)
+    build_button_profile(device, main_endpoint, #button_eps, #motion_eps)
     -- All button endpoints found will be added as additional components in the profile containing the main_endpoint.
     -- The resulting endpoint to component map is saved in the COMPONENT_TO_ENDPOINT_MAP field
     build_button_component_map(device, main_endpoint, button_eps)
```
nickolas-deboom:
```suggestion
    build_button_component_map(device, main_endpoint, button_eps, motion_eps)
```
We could pass in the motion endpoints here just like you did for `build_button_profile`, and then add the motion endpoint to the component-endpoint map within `build_button_component_map` rather than introduce a new `build_motion_component_map` function

nickolas-deboom:
I left a few comments but overall the proposed changes look fine to me, and thank you for providing unit tests. Have you tested this change on the device and confirmed that the plugin renders correctly?

drivers/SmartThings/matter-switch/src/init.lua
Cornelius117: Ok, thank you for your kind advice! We have revised the places you mentioned, please check whether there is still room for improvement.

Cornelius117:
> I left a few comments but overall the proposed changes look fine to me, and thank you for providing unit tests. Have you tested this change on the device and confirmed that the plugin renders correctly?

Yes, we have tested these changes on the device, and the display on the plugin is as expected.

drivers/SmartThings/matter-switch/profiles/3-button-motion.yml
```diff
@@ -0,0 +1,30 @@
...
+    - name: RemoteController
+  - id: button3
+    capabilities:
+      - id: button
+        version: 1
+    categories:
+    - name: RemoteController
+  - id: motion
+    capabilities:
+      - id: motionSensor
```
hcarter-775, line 27: I guess I'm wondering why we're putting motion on its own component in the first place? Seems like this should be in the main component. Same for the other

hcarter-775: To clarify, our current process is to put all capabilities that don't require special handling to be placed on the `main` component. Doing this will make the changes in the component map moot, and the event should be emitted on main by default. Thanks!

nickolas-deboom: Ah yeah, that's a good callout

Cornelius117: We have tried this modification before, but the results are somewhat different from what we expected. We put the details in the comments on this page (https://smartthings.atlassian.net/browse/MTR-939). Please review.

Cornelius117:
Due to code changes, we closed this PR and moved it here (https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2459).

PR #2425, opened 2025-09-24
Use PercentSetting to set fanSpeedPercent

nickolas-deboom:
PercentSetting is subscribed in addition to PercentCurrent to set `fanSpeedPercent`, since it provides an accurate representation of the speed of the fan while helping avoid the following situation:
    (1) The fan speed is changed in the app and PercentSetting is routed to the device
    (2) The fan reports back a value of PercentCurrent that doesn't match PercentSetting because the speed takes a little while to change
    (3) The fanSpeedPercent capability jumps to the value reported by PercentCurrent
    
PercentCurrent is still subscribed to, but its attribute handler is gated on the current fan mode being AUTO, in which case PercentSetting is NULL on the device side and PercentCurrent is used as a fallback for setting the capability.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -196,7 +196,8 @@ local subscribed_attributes = {
     clusters.FanControl.attributes.FanMode
   },
   [capabilities.fanSpeedPercent.ID] = {
-    clusters.FanControl.attributes.PercentCurrent
+    clusters.FanControl.attributes.PercentCurrent,
+    clusters.FanControl.attributes.PercentSetting
   },
   [capabilities.windMode.ID] = {
     clusters.FanControl.attributes.WindSupport,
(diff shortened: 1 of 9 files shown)


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1613,11 +1612,19 @@ local function fan_mode_sequence_handler(driver, device, ib, response)
...
-    speed = utils.clamp_value(ib.data.value, MIN_ALLOWED_PERCENT_VALUE, MAX_ALLOWED_PERCENT_VALUE)
+  local thermostat_mode = device:get_latest_state(
+    device:endpoint_to_component(ib.endpoint_id),
+    capabilities.thermostatMode.ID,
+    capabilities.thermostatMode.thermostatMode.NAME
+  )
+  if thermostat_mode == capabilities.thermostatMode.thermostatMode.auto() then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.fanSpeedPercent.percent(ib.data.value))
   end
-  device:emit_event_for_endpoint(ib.endpoint_id, capabilities.fanSpeedPercent.percent(speed))
```
Huangxiangjie, line 1620: If `fan_speed_percent` value was changed from device, and then report the new value to station. We will ignore the `emit_event`, this will cause status inconsistency issues in this case.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1613,11 +1612,19 @@ local function fan_mode_sequence_handler(driver, device, ib, response)
 end
 
 local function fan_speed_percent_attr_handler(driver, device, ib, response)
-  local speed = 0
-  if ib.data.value ~= nil then
-    speed = utils.clamp_value(ib.data.value, MIN_ALLOWED_PERCENT_VALUE, MAX_ALLOWED_PERCENT_VALUE)
```
hcarter-775: probably shouldn't remove the clamp, right?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1613,11 +1612,19 @@ local function fan_mode_sequence_handler(driver, device, ib, response)
...
+  )
+  if thermostat_mode == capabilities.thermostatMode.thermostatMode.auto() then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.fanSpeedPercent.percent(ib.data.value))
   end
-  device:emit_event_for_endpoint(ib.endpoint_id, capabilities.fanSpeedPercent.percent(speed))
+end
+
+local function fan_speed_setting_attr_handler(driver, device, ib, response)
+  if ib.data.value == nil then return end
+  device:emit_event_for_endpoint(ib.endpoint_id, capabilities.fanSpeedPercent.percent(ib.data.value))
```
hcarter-775: we probably want to clamp this as well

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Restored in latest force push

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Added in latest force push

PR #2426, opened 2025-09-24
Refactor Matter report framework for powerConsumptionReport capability

hcarter-775:
# Description of Change
Remove the 15 minute timer for sending powerConsumptionReport events in favor of sending them as the device sends updated reports through the report handler. Ensure that a minimum of 15 minutes is held between reports.  

# Summary of Completed Tests
Unit tests have been updated to ensure minimal report intervals continue to be followed, and that reports are given as expected.
On-device testing for an Eve Energy device as well as devices supporting the ElectricalEnergyMeasurement cluster that would onboard to the matter switch, matter energy, and matter thermostat drivers have been completed. This testing has confirmed that accurate reports are sent when expected in all cases, and that devices are displayed as offline on the UI when a device is actually offline.


drivers/SmartThings/matter-energy/src/init.lua
```diff
@@ -193,145 +191,30 @@ local BATTERY_CHARGING_STATE_MAP = {
...
+  -- device energy reporting must be handled cumulatively, periodically, or by both simulatanously.
+  -- To ensure a single source of truth, we only handle a device's periodic reporting if cumulative reporting is not supported.
+  local electrical_energy_measurement_eps = embedded_cluster_utils.get_endpoints(device, clusters.ElectricalEnergyMeasurement.ID)
+  if #electrical_energy_measurement_eps > 0 then
+    local cumulative_energy_eps = embedded_cluster_utils.get_endpoints(
+      device,
+      clusters.ElectricalEnergyMeasurement.ID,
+      {feature_bitmap = clusters.ElectricalEnergyMeasurement.types.Feature.CUMULATIVE_ENERGY}
+    )
+    if #cumulative_energy_eps == 0 then device:set_field(CUMULATIVE_REPORTS_NOT_SUPPORTED, true, {persist = false}) end
```
nickolas-deboom, line 216: Did you mean to write persist = true?

drivers/SmartThings/matter-energy/src/init.lua
```diff
@@ -565,93 +447,94 @@ local function device_energy_mgmt_mode_attr_handler(driver, device, ib, response
...
+    last_time = device:get_field(LAST_EXPORTED_REPORT_TIMESTAMP) or 0
+    -- Ensure that the previous report was sent at least 15 minutes ago
+    if MINIMUM_ST_ENERGY_REPORT_INTERVAL >= (current_time - last_time) then return end
+    device:set_field(LAST_EXPORTED_REPORT_TIMESTAMP, current_time, { persist = true })
+  else
+    last_time = device:get_field(LAST_IMPORTED_REPORT_TIMESTAMP) or 0
+    -- Ensure that the previous report was sent at least 15 minutes ago
+    if MINIMUM_ST_ENERGY_REPORT_INTERVAL >= (current_time - last_time) then return end
+    device:set_field(LAST_IMPORTED_REPORT_TIMESTAMP, current_time, { persist = true })
   end
```
nickolas-deboom:
```suggestion
  local last_report_timestamp_field = component.id == "exportedEnergy" and LAST_EXPORTED_REPORT_TIMESTAMP or LAST_IMPORTED_REPORT_TIMESTAMP
  last_time = device:get_field(last_report_timestamp_field) or 0
    -- Ensure that the previous report was sent at least 15 minutes ago
  if MINIMUM_ST_ENERGY_REPORT_INTERVAL >= (current_time - last_time) then return end
  device:set_field(last_report_timestamp_field, current_time, { persist = true })
```
minor suggestion, feel free to ignore :smile:

drivers/SmartThings/matter-energy/src/init.lua
hcarter-775: nope! Since this gets run on device_init, there's no reason to persist it. Mostly I'm just explicitly clarifying it's not persisted, so it's easier to reason about.

drivers/SmartThings/matter-energy/src/init.lua
```diff
@@ -565,93 +447,94 @@ local function device_energy_mgmt_mode_attr_handler(driver, device, ib, response
...
-        -- cumulative_energy_handler to handle the energy report for this endpoint over periodic_energy_handler.
-        return
+    if not is_cumulative_report then
+      if device:get_field(CUMULATIVE_REPORTS_NOT_SUPPORTED) ~= true then
+        return -- if this is a periodic report and cumulative reports ARE supported by the device, ignore the report altogether.
       end
+      energy_Wh = energy_Wh + (total_cumulative_energy[endpoint_id] or 0) -- handle the periodic report
+    end
+    total_cumulative_energy[endpoint_id] = energy_Wh -- in the case that there are multiple electrical sensors, store them in a table.
+    device:set_field(cumulative_import_or_export_field, total_cumulative_energy, { persist = true })
```
nickolas-deboom, line 516: We could potentially also use set_field_for_endpoint rather than storing it all in one table, not big deal either way though

nickolas-deboom:
I'm thinking about a scenario where the device reports energy usage twice within a 15 minute period. The second report won't result in a energy report being sent because 15 minutes has not yet elapsed since the last one (which is a change from before I believe, where previously we would ensure that we are sending a report at least once every 15 minutes and now we are making sure we send a report at most once every 15 minutes - is that correct?). What if the device doesn't report energy usage again for a while? Would that second energy report essentially be skipped?

drivers/SmartThings/matter-energy/src/init.lua
nickolas-deboom: Ah yeah, that's why I was wondering, just wanted to make sure it was on purpose!

drivers/SmartThings/matter-energy/src/init.lua
hcarter-775: we could! But for the purposes of this PR, I believe this should be sufficient.

drivers/SmartThings/matter-energy/src/init.lua
hcarter-775: good idea!

hcarter-775:
> I'm thinking about a scenario where the device reports energy usage twice within a 15 minute period. The second report won't result in a energy report being sent because 15 minutes has not yet elapsed since the last one (which is a change from before I believe, where previously we would ensure that we are sending a report at least once every 15 minutes and now we are making sure we send a report at most once every 15 minutes - is that correct?). What if the device doesn't report energy usage again for a while? Would that second energy report essentially be skipped?

It wouldn't be skipped, since it would be saved in the appropriate field. But yes, it would in your case be a while before another report is sent :+1:, specifically it would be when the device updates us again.

drivers/SmartThings/matter-energy/src/init.lua
hcarter-775: updated

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -193,33 +193,31 @@ local child_device_profile_overrides_per_vendor_id = {
...
 end
 
-local function delete_import_poll_schedule(device)
-  local import_poll_timer = device:get_field(RECURRING_IMPORT_REPORT_POLL_TIMER)
-  if import_poll_timer then
-    device.thread:cancel_timer(import_poll_timer)
-    device:set_field(RECURRING_IMPORT_REPORT_POLL_TIMER, nil)
-    device:set_field(IMPORT_POLL_TIMER_SETTING_ATTEMPTED, nil)
+local function report_power_consumption_to_st_energy(device, latest_total_imported_energy_wh)
+  -- workaround: ignore devices supporting Eve's private energy cluster AND the ElectricalEnergyMeasurement cluster
```
ctowns: If an Eve device supports both, I think we should just use the base driver and not the  sub driver. There shouldn't be a case where both of these are being sent if the sub driver is being used correctly, and we shouldn't require a workaround like this because loading the sub driver should overload any base handling, if the sub driver is needed.

Since this is a special case where the device technically supports two different ways to support this behavior, could we make it so that the can_handle in the Eve sub driver only loads the Eve sub driver if the Eve device is on older FW and only supports the private energy clusters? That should ideally avoid the need for this workaround, and also it will prevent us from loading the sub driver when it is not needed.

hcarter-775: Yeah, I think that would be better to. However, I think I'd like to confirm first that the ElectricalEnergyMeasurement reporting works nicely for Eve devices. To me, the fact that we have historic knowledge that the private cluster approach works for eve devices, I am slightly more inclined to use that instead.

nickolas-deboom:
Has this been tested again after the rebase that incorporated the changes from https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2401/files? Might not be necessary but would be nice to confirm since that was a decently large change.

hcarter-775:
> Has this been tested again after the rebase that incorporated the changes from https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2401/files? Might not be necessary but would be nice to confirm since that was a decently large change.

The base Matter Switch testing was completed on this rebased branch. Cooper's Eve Energy testing was not, though that subdriver was not actually edited by the recent refactor.

Update: re-testing on the latest was done for an Eve Energy device as well.

nickolas-deboom:
> > Has this been tested again after the rebase that incorporated the changes from https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2401/files? Might not be necessary but would be nice to confirm since that was a decently large change.
> 
> The base Matter Switch testing was completed on this rebased branch. Cooper's Eve Energy testing was not, though that subdriver was not actually edited by the recent refactor.

Sweet. I completed another pass and the PR looks good from my perspective, nice work!

drivers/SmartThings/matter-switch/src/init.lua
ctowns: Following up offline discussion - this workaround can be used for now to prevent issues with the sub driver when a device supports both the private clusters and the matter clusters for energy measurements. However, a follow up PR will be created to address the private + matter cluster combo devices in a more direct way in order to keep the scope of this PR smaller, given it is quite large as is.

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/init.lua
```diff
@@ -288,31 +275,18 @@ local function handle_refresh(self, device)
...
-  local startTime = iso8061Timestamp(last_time)
-  local endTime = iso8061Timestamp(current_time - 1)
-
   -- Reset the consumption on the device
   local data = data_types.validate_or_build_type(current_time_2001, data_types.Uint32)
   device:send(cluster_base.write(device, 0x01, PRIVATE_CLUSTER_ID, PRIVATE_ATTR_ID_ACCUMULATED_CONTROL_POINT, nil,
     data))
-
-  -- Report the energy consumed during the time interval. The unit of these values should be 'Wh'
-  device:emit_event(capabilities.powerConsumptionReport.powerConsumption({
```
ctowns, line 308: Don't we still need this for the energy meter reset? Where is the energy being set to zero otherwise?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -112,8 +111,19 @@ function SwitchLifecycleHandlers.device_init(driver, device)
...
+    -- device energy reporting must be handled cumulatively, periodically, or by both simulatanously.
+    -- To ensure a single source of truth, we only handle a device's periodic reporting if cumulative reporting is not supported.
+    local electrical_energy_measurement_eps = embedded_cluster_utils.get_endpoints(device, clusters.ElectricalEnergyMeasurement.ID)
+    if #electrical_energy_measurement_eps > 0 then
+      local cumulative_energy_eps = embedded_cluster_utils.get_endpoints(
+        device,
+        clusters.ElectricalEnergyMeasurement.ID,
+        {feature_bitmap = clusters.ElectricalEnergyMeasurement.types.Feature.CUMULATIVE_ENERGY}
+      )
+      if #cumulative_energy_eps == 0 then device:set_field(fields.CUMULATIVE_REPORTS_NOT_SUPPORTED, true, {persist = false}) end
```
ctowns, line 125: nit: is the `persist = false` needed here?

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/init.lua
hcarter-775: No, we don't. This is because the consumption will be reset for the device with the 
```
local data = data_types.validate_or_build_type(current_time_2001, data_types.Uint32)
device:send(cluster_base.write(device, 0x01, PRIVATE_CLUSTER_ID, PRIVATE_ATTR_ID_ACCUMULATED_CONTROL_POINT, nil, data))
```
call that has been preserved.

Then, with the 1 minute poll timer on this attribute, we will continue to get updates about its state, and since it has been reset on the device, when we get an update, the new `totalConsumptionWh` will be 0 or thereabout (if it has incremented since then), and this check will return 0:
```
  if previous_imported_report and previous_imported_report.energy then
    energy_delta_wh = math.max(latest_total_imported_energy_wh - previous_imported_report.energy, 0.0)
  end
```
which will then be emitted on the regular 15 minute interval.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Sorry to respond to this so quick, it is not. However, I think explicitly stating these optional values in these calls is not a bad idea as it promotes code clarity. This is kinda just something I saw in the sonos/hue drivers and wanted to emulate for matter more.

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/init.lua
ctowns: Sounds good!

drivers/SmartThings/matter-switch/src/init.lua
ctowns: > Sorry to respond to this so quick

Please send your response via mail and wait 3-5 business days before responding next time :)

That sounds good to me!

PR #2428, opened 2025-09-24
Don't include Off for supportedFanModes for thermostats by default

nickolas-deboom:
This change removes Off as a default value from supportedFanModes attribute for the Thermostat device type.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/init.lua
+++ b/drivers/SmartThings/matter-thermostat/src/init.lua
@@ -1565,7 +1565,7 @@ local function fan_mode_sequence_handler(driver, device, ib, response)
     supported_fan_modes_attribute = supported_fan_modes_capability.supportedAcFanModes
   elseif device:supports_capability_by_id(capabilities.thermostatFanMode.ID) then
     supported_fan_modes_capability = capabilities.thermostatFanMode
-    supported_fan_modes_attribute = capabilities.thermostatFanMode.supportedThermostatModes
+    supported_fan_modes_attribute = capabilities.thermostatFanMode.supportedThermostatFanModes
     -- Our thermostat fan mode control is not granular enough to handle all of the supported modes
     if ib.data.value >= clusters.FanControl.attributes.FanModeSequence.OFF_LOW_MED_HIGH_AUTO and
       ib.data.value <= clusters.FanControl.attributes.FanModeSequence.OFF_ON_AUTO then


drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1470,7 +1474,7 @@ local function sequence_of_operation_handler(driver, device, ib, response)
   if device:get_field(OPTIONAL_THERMOSTAT_MODES_SEEN) == nil then
     device:set_field(OPTIONAL_THERMOSTAT_MODES_SEEN, {capabilities.thermostatMode.thermostatMode.off.NAME}, {persist=true})
   end
-  local supported_modes = utils.deep_copy(device:get_field(OPTIONAL_THERMOSTAT_MODES_SEEN))
+  local supported_modes = utils.deep_copy(device:get_field(OPTIONAL_THERMOSTAT_MODES_SEEN) or {})
```
hcarter-775: due to the logic in line 1475, this will literally never be empty, no?

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1578,17 +1582,20 @@ end
...
-    supportedFanModes = { "off", "low", "high", "auto" }
+    supportedFanModes = { "low", "high", "auto" }
   elseif ib.data.value == clusters.FanControl.attributes.FanModeSequence.OFF_HIGH_AUTO then
-    supportedFanModes = { "off", "high", "auto" }
+    supportedFanModes = { "high", "auto" }
   else
-    supportedFanModes = { "off", "high" }
+    supportedFanModes = { "high" }
+  end
+  if get_device_type(device) ~= THERMOSTAT_DEVICE_TYPE_ID then
```
hcarter-775: is `get_device_type()` the best function to use in this case? It might be, but since it has that priority logic built into it maybe it would be better to make a separate function?

hcarter-775: actually, I disagree with this argument. This is actually what we want here I think. marking as resolved

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1578,17 +1582,20 @@ end
 local function fan_mode_sequence_handler(driver, device, ib, response)
   local supportedFanModes, supported_fan_modes_attribute
   if ib.data.value == clusters.FanControl.attributes.FanModeSequence.OFF_LOW_MED_HIGH then
-    supportedFanModes = { "off", "low", "medium", "high" }
+    supportedFanModes = { "low", "medium", "high" }
```
hcarter-775: does this mean we'll be deleting options from currently existing thermostat devices with this capability? What are the numbers on this- specifically routines using off for thermostats? Assume it's very low (maybe 0), but still worth checking. Otherwise this looks good to me

nickolas-deboom: That's a great point, I will try to figure out how many there are if any before proceeding. If there are a lot of existing rules we could opt to only implement this change for new devices

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: Oh yeah, good point 😅

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: In the latest commit I made some changes to include 'off' in the case that 'off' was previously a supported fan mode, in order to avoid any complications with breaking routines

PR #2430, opened 2025-09-29
WWSTCERT-8138: Add support for frient Zigbee Range Extender

marcintyminski:
Check all that apply

# Type of Change

          - [x] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests


drivers/SmartThings/zigbee-range-extender/src/frient/init.lua
```diff
@@ -0,0 +1,80 @@
...
+local function device_added(driver, device)
+  device:emit_event(capabilities.powerSource.powerSource.mains())
+end
+
+local function device_init(driver, device)
+  battery_defaults.build_linear_voltage_init(3.3, 4.1)(driver, device)
+end
+
+local function do_refresh(driver, device)
+  device:send(Basic.attributes.ZCLVersion:read(device))
```
greens: this isn't being handled, so why does it need to be refreshed?

drivers/SmartThings/zigbee-range-extender/src/frient/init.lua
```diff
@@ -0,0 +1,78 @@
...
+  device:emit_event(capabilities.powerSource.powerSource.mains())
+end
+
+local function device_init(driver, device)
+  battery_defaults.build_linear_voltage_init(3.3, 4.1)(driver, device)
+end
+
+local function do_refresh(driver, device)
+  device:send(PowerConfiguration.attributes.BatteryVoltage:read(device))
+  device:send(IASZone.attributes.ZoneStatus:read(device))
```
greens, line 47: Seems like the base driver's `device:refresh` should be enough here. I don't think you need to define a custom handler. A unit test would confirm.

marcintyminski: I tried to delete a custom refresh handler, but then it is only ZCLVersion attribute from Basic cluster that is being refreshed. Unit tests for refreshing fail when there is any attribute other than ZCLVersion.

greens: Ah, I see. This is because for this driver, the IAS Zone attribute has not been configured. So the driver would also not be set up to receive reports from this attribute as-is, either.

In the base driver, you'll need to add [this line](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-contact/src/init.lua#L109) in roughly the same place.

greens: After that is in place, your test results will be different.

PR #2436, opened 2025-09-29
Matter Switch: remove ENERGY_MANAGEMENT_ENDPOINT field

hcarter-775:
# Description of Change
This field is a workaround for the fact that in the Aqara Switch H2 device (and possibly others), the ElectricalEnergyMeasurement cluster is stored on the root endpoint, which is not associated with the main component in out current logic. Therefore, the current handling directly circumvents this issue by writing the event on ep 1 instead of ep 0  by means of this field. 

However, there is a cleaner solution here which circumvents the entire `emit_event_for_endpoint` logic, which I have implemented here, specifically by using the `emit_component_event`, which is what `emit_event_for_endpoint` breaks down into anyway on the backend.

For now, `main` can be directly written here as the component of choice (all devices currently have `powerConsumptionReport` and `energyMeter` on main) but if we expand this logic to become more inclusive of different configurations, this is still a more robust and more clear solution.

# Summary of Completed Tests
All unit tests pass.

Tested with Aqara H2 Switch (the only known device that functions like this) and `energyMeter` and `powerConsumptionReport` events are emitted as expected. Tested with Meross plug and similarly functionality is retained.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
@@ -250,8 +250,6 @@ function AttributeHandlers.active_power_handler(driver, device, ib, response)
   if ib.data.value then
     local watt_value = ib.data.value / fields.CONVERSION_CONST_MILLIWATT_TO_WATT
     device:emit_component_event(component, capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
-  else
-    device:emit_component_event(component, capabilities.powerMeter.power({ value = 0, unit = "W"}))
   end
   if type(device.register_native_capability_attr_handler) == "function" then
     device:register_native_capability_attr_handler("powerMeter","power")
--- a/drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
@@ -16,6 +16,7 @@ local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
 local version = require "version"
 local im = require "st.matter.interaction_model"
+local device_lib = require "st.device"
 
 local st_utils = require "st.utils"
 local fields = require "utils.switch_fields"
@@ -245,17 +246,15 @@ end
 -- [[ ELECTRICAL POWER MEASUREMENT CLUSTER ATTRIBUTES ]] --
 
 function AttributeHandlers.active_power_handler(driver, device, ib, response)
+  local component = device.profile.components["main"]
   if ib.data.value then
     local watt_value = ib.data.value / fields.CONVERSION_CONST_MILLIWATT_TO_WATT
-    if ib.endpoint_id ~= 0 then
-      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
-    else
-      -- when energy management is defined in the root endpoint(0), replace it with the first switch endpoint and process it.
-      device:emit_event_for_endpoint(device:get_field(fields.ENERGY_MANAGEMENT_ENDPOINT), capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
-    end
-    if type(device.register_native_capability_attr_handler) == "function" then
-      device:register_native_capability_attr_handler("powerMeter","power")
-    end
+    device:emit_component_event(component, capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
+  else
+    device:emit_component_event(component, capabilities.powerMeter.power({ value = 0, unit = "W"}))
+  end
+  if type(device.register_native_capability_attr_handler) == "function" then
+    device:register_native_capability_attr_handler("powerMeter","power")
   end
 end
 
@@ -281,25 +280,22 @@ end
 
 function AttributeHandlers.cumul_energy_imported_handler(driver, device, ib, response)
   if ib.data.elements.energy then
+    local energy_component = device.profile.components["main"]
     local watt_hour_value = ib.data.elements.energy.value / fields.CONVERSION_CONST_MILLIWATT_TO_WATT
     device:set_field(fields.TOTAL_IMPORTED_ENERGY, watt_hour_value, {persist = true})
-    if ib.endpoint_id ~= 0 then
-      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
-    else
-      -- when energy management is defined in the root endpoint(0), replace it with the first switch endpoint and process it.
-      device:emit_event_for_endpoint(device:get_field(fields.ENERGY_MANAGEMENT_ENDPOINT), capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
-    end
+    device:emit_component_event(energy_component, capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
     switch_utils.report_power_consumption_to_st_energy(device, device:get_field(fields.TOTAL_IMPORTED_ENERGY))
   end
 end
 
 function AttributeHandlers.per_energy_imported_handler(driver, device, ib, response)
   if ib.data.elements.energy then
+    local energy_component = device.profile.components["main"]
     local watt_hour_value = ib.data.elements.energy.value / fields.CONVERSION_CONST_MILLIWATT_TO_WATT
     local latest_energy_report = device:get_field(fields.TOTAL_IMPORTED_ENERGY) or 0
     local summed_energy_report = latest_energy_report + watt_hour_value
     device:set_field(fields.TOTAL_IMPORTED_ENERGY, summed_energy_report, {persist = true})
-    device:emit_event(capabilities.energyMeter.energy({ value = summed_energy_report, unit = "Wh" }))
+    device:emit_component_event(energy_component, capabilities.energyMeter.energy({ value = summed_energy_report, unit = "Wh" }))
     switch_utils.report_power_consumption_to_st_energy(device, device:get_field(fields.TOTAL_IMPORTED_ENERGY))
   end
 end
@@ -309,7 +305,7 @@ function AttributeHandlers.energy_imported_factory(is_cumulative_report)
     -- workaround: ignore devices supporting Eve's private energy cluster AND the ElectricalEnergyMeasurement cluster
     local EVE_MANUFACTURER_ID, EVE_PRIVATE_CLUSTER_ID = 0x130A, 0x130AFC01
     local eve_private_energy_eps = device:get_endpoints(EVE_PRIVATE_CLUSTER_ID)
-    if device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID and #eve_private_energy_eps > 0 then
+    if device.network_type == device_lib.NETWORK_TYPE_MATTER and device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID and #eve_private_energy_eps > 0 then
       return
     end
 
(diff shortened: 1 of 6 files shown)


drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
```diff
@@ -309,7 +305,7 @@ function AttributeHandlers.energy_imported_factory(is_cumulative_report)
     -- workaround: ignore devices supporting Eve's private energy cluster AND the ElectricalEnergyMeasurement cluster
     local EVE_MANUFACTURER_ID, EVE_PRIVATE_CLUSTER_ID = 0x130A, 0x130AFC01
     local eve_private_energy_eps = device:get_endpoints(EVE_PRIVATE_CLUSTER_ID)
-    if device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID and #eve_private_energy_eps > 0 then
+    if device.network_type == device_lib.NETWORK_TYPE_MATTER and device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID and #eve_private_energy_eps > 0 then
```
hcarter-775, line 306: This check may be required if a child device attempts to call this function for two reasons:
1. a child device may not have `manufacturer_info` filled, which would cause an error.
2. The Eve subdriver only accepts non-child devices, so technically this is a more accurate check since we're trying to _only_ gate devices that would be using the subdriver. 

Both of these are kinda theoretical, but a unit test caught the issue and rather than updating the unit test, I figured the above reasons justified the inclusion anyway.

drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
```diff
@@ -245,17 +246,15 @@ end
...
-    else
-      -- when energy management is defined in the root endpoint(0), replace it with the first switch endpoint and process it.
-      device:emit_event_for_endpoint(device:get_field(fields.ENERGY_MANAGEMENT_ENDPOINT), capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
-    end
-    if type(device.register_native_capability_attr_handler) == "function" then
-      device:register_native_capability_attr_handler("powerMeter","power")
-    end
+    device:emit_component_event(component, capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
+  else
+    device:emit_component_event(component, capabilities.powerMeter.power({ value = 0, unit = "W"}))
```
hcarter-775: If we don't get any data, the UI may emit a NaN value. I have seen this occur after a driver restart when no power is actually being recorded but we call subscribe. This circumvents that.

mocelet: That could run unintended automations with power as trigger when the driver restarts, couldn't it? For instance, user watching the TV has an automation to power off all the devices (TV, amplifier, etc.) with a smart plug to prevent standby when the power usage is below certain watts. The driver restarts, emits 0 W, the TV suddenly powers off and user will not understand what happened.

It's a case similar to the artificial pressed events in buttons, sometimes used during initialization but not actually needed that could trigger unintended automations in other situations.

hcarter-775: That's a good point @mocelet. I struggle a little to think of an example where what you're describing would occur, since a smart plug that's in active use should _not_ report nil and so would not hit this case. However, in principal I agree with this and of the general statement that we should not artificially emit events for UI purposes. 

I will hold off on implementing this in the current PR, but thinking about it now, perhaps a more solid solution to the above issue would be to avoid re-upping the subscription to this attribute in the first place on a restart.

I have removed this extra logic in the latest commit.

mocelet: Thanks @hcarter-775 , I agree it probably won't happen except in rare cases, Matter spec only states that "if the power cannot be measured, a value of null SHALL be returned". Not exactly sure why a device with power monitor could not measure power but since you mentioned seeing null values it's a possibility.

I guess a plug with power monitor sending null will sooner or later report an actual measurement, otherwise I presume it's broken. It's like buttons, even if the driver does not emit the artificial pressed event, as soon as the user checks if the button works it's going to emit them so the UI will be happy in no time.

drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
```diff
@@ -245,17 +246,13 @@ end
 -- [[ ELECTRICAL POWER MEASUREMENT CLUSTER ATTRIBUTES ]] --
 
 function AttributeHandlers.active_power_handler(driver, device, ib, response)
+  local component = device.profile.components["main"]
```
ctowns, line 249: I don't think we should be hard coding in use of the `main` endpoint, because then this will break on a multi-component device. Do we have any multi-component power devices as far as we know?

~Why does `ENERGY_MANAGEMENT_ENDPOINT` need to be saved in the first place when we already know it is on endpoint 0 given the `if ib.endpoint_id ~= 0 then` statement?~ nevermind, I see we need to emit on the first switch endpoint

drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
ctowns: Is this directly related to removing the `ENERGY_MANAGEMENT_ENDPOINT`? If not, I would make this a separate PR. If yes, I'm not sure I understand the connection, could you explain?

drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
ctowns: If we are going to hard code a component, why not just hardcode the endpoint for certain device types? If this is limited to a couple Aqara devices (or only 1?) I might perfer just adding special handling for them.

Is having the energy measurement cluster on the root endpoint typical, or is this the only device?

hcarter-775: We don't have any profiles where energy isn't on the main component. As I noted in the description, even if that comes up, I still think this is a better handling than the current.

hcarter-775: At the moment, this only shows up for Aqara because it is the only device that has energy on a child device. However, this will change, and we know these devices will often have the Electrical Sensor device type on multiple EPs. This would normally be where endpoint_to_component would come in. In this case, I've just circumvented that function and defined the relationship explicitly (since we already know it will be like this in all cases for the foreseeable future since plugs and switches are locked into single component profiles). 

Perhaps endpoint_to_component could be a better option though, but tbh that is just punting this same logic to a helper function.

drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
hcarter-775: It's semi-related. A child device may not have manufacturer info set, which would cause an issue. So in a unit test, that wasn't set. We also don't set the manufacturer info for child devices in our current logic.

drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
ctowns: I think this is all a matter of opinion, but to me we should avoid hard coding components in our function as a rule since we created the `component_to_endpoint` and `endpoint_to_component` apis explicitly for this purpose. I believe the cleanest solution is to handle this mapping in our available APIs, and leave component and endpoint mapping outside of attribute handlers as much as we can. If I am not mistaken, can we not just add the root node to the component to endpoint map and have it map to the `main` component, and then continue to leverage that function as expected?

hcarter-775: At the moment yes, but to me that's basically as hacky as the current handling, since there's no "reason" why the root endpoint should be mapped to main.

I could see an argument for more complex handling of electrical sensor endpoints in the api function, but now that I'm thinking about it, we don't even define that map at the moment for switches and plugs (since everything is assumed to be on main), and so actually it's an even weaker map for those handlers- ~~they just straight call emit_event() which will auto send it to main~~ this is not true upon inspection of the handlers, see the find_child discussion below for why it works in the `on_off_handler` cases.

So to me, this is a question of where we arbitrarily send this to main, unless we start explicitly defining the component map for these switches/plugs (though since they're all single-component, that seems a little unnecessary to me personally).

hcarter-775: also, we do have precedent for this:
```
local function hepa_filter_condition_handler(driver, device, ib, response)
  local component = device.profile.components["hepaFilter"]
  local condition = ib.data.value
  device:emit_component_event(component, capabilities.filterState.filterLifeRemaining(condition))
end
```
(and associated handlers like ac and filterStatus) though it's minimal.

hcarter-775: This actually brings in a deeper issue, in that child devices are assumed to be mapped to a single endpoint:  
```
utils.find_child(parent, ep_id)
end
```
and under the hood, `emit_event_for_endpoint()` is checking this first (before looking at `endpoint_to_component`) when it's deciding where to send the event. If it found this device, everything would work well since it would automatically port the event to the main component (child devices are assumed to be single component).

However, the function I showed above would _not_ give us a real device, cause of the single endpoint to child device assumption. The ep_id that the mapping is expecting is for the endpoint associated with the OnOff cluster, and there is no reason in spec why this has to be the same endpoint as its associated Electrical Sensor endpoint.

So based on this analysis, I think it could be argued that using `endpoint_to_component` in this case would still be a misuse of the apis in place.

nickolas-deboom: I might prefer specific handling for this device, because according to the spec it isn't supported to have clusters with the Application role implemented on the root node, which would include the Electrical Energy+Power Measurement clusters. So it's unlikely we would see this on another device

nickolas-deboom: > If we are going to hard code a component, why not just hardcode the endpoint for certain device types? If this is limited to a couple Aqara devices (or only 1?) I might perfer just adding special handling for them.
> 
> Is having the energy measurement cluster on the root endpoint typical, or is this the only device?

I might prefer specific handling for this device, because according to the spec it isn't supported to have clusters with the Application role implemented on the root node, which would include the Electrical Energy+Power Measurement clusters. So it's unlikely we would see this on another device

nickolas-deboom: But then again, it's already been implemented generically for a while, so I don't necessarily think we need to change it back to the original handling

hcarter-775: The issue is more generic than the fact that it's happening on the root node. It will occur for other devices if 
1. the device is a child device
2. it supports Energy on an endpoint that is different from the endpoint OnOff is on.

From initial testing, I have seen this case show up twice (not counting aqara). It generically can and will occur for generic endpoints, not just root.

ctowns: For that `find_child` problem you mentioned where child device only use one endpoint, we can also update the `find_child` function specifically for our purposes here so that is could allow for multiple endpoints to map to the child device using a new device key that can have multiple endpoints. You could even update it so that we have a specific override of this function that is only set for devices like this where we need multiple endpoints mapped to a child, and other devices can continue to use the generic one.The find child function is device-scoped and not driver-scoped, so you can make a custom one specifically for devices like this and not affect others in the driver

hcarter-775: per this discussion, I have updated the logic to leave a single `emit_event_for_endpoint` call in each of these handlers in this PR: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2444

It cannot be easily broken out of the above PR anymore, since the new logic is inherently tied to the new PowerTopology handling.

hcarter-775:
Given this discussion's conclusion: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2436#discussion_r2399992011

I am closing this PR.

PR #2437, opened 2025-09-29
Matter Switch: Remove color_utils

hcarter-775:
# Description of Change
The comment in `color_utils` states: `--TODO remove the usage of these color utils once 0.48.x has been distributed
-- to all hubs.`

It was discussed that 1 device in the field that is currently online may rely on this code

# Summary of Completed Tests
Ensured correct color functionality on a color extended light VDA device.


nickolas-deboom:
I believe there's something like ~100,000 hubs on FW < 0.48. Is there a way to check if they have any connected devices that would be using `color_utils`?

greens:
Sounds like a question for @tpmanley . I think you'd need to check device type, if possible.

PR #2440, opened 2025-09-30
ZLL: Use Native / Default handlers

wkhenon:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

This pull request defaults handing for `on`, `off`, `set_level`, and `set_color_temperature` in the ZLL sub-driver back to the main `zigbee-switch` driver. 

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/ikea-xy-color-bulb/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/ikea-xy-color-bulb/init.lua
@@ -35,7 +35,13 @@ local IKEA_XY_COLOR_BULB_FINGERPRINTS = {
 }
 
 local function can_handle_ikea_xy_color_bulb(opts, driver, device)
-  return (IKEA_XY_COLOR_BULB_FINGERPRINTS[device:get_manufacturer()] or {})[device:get_model()] or false
+  local can_handle = (IKEA_XY_COLOR_BULB_FINGERPRINTS[device:get_manufacturer()] or {})[device:get_model()]
+  if can_handle then
+    local subdriver = require("ikea-xy-color-bulb")
+    return true, subdriver
+  else
+    return false
+  end
 end
 
 local device_init = function(self, device)
--- a/drivers/SmartThings/zigbee-switch/src/test/test_zll_rgbw_bulb.lua
+++ b/drivers/SmartThings/zigbee-switch/src/test/test_zll_rgbw_bulb.lua
@@ -15,6 +15,7 @@
 local test = require "integration_test"
 local clusters = require "st.zigbee.zcl.clusters"
 local t_utils = require "integration_test.utils"
+local version = require "version"
 local zigbee_test_utils = require "integration_test.zigbee_test_utils"
 
 local OnOff = clusters.OnOff
@@ -140,6 +141,7 @@ test.register_coroutine_test(
     test.socket.zigbee:__set_channel_ordering("relaxed")
     test.timer.__create_and_queue_test_time_advance_timer(2, "oneshot")
     test.socket.capability:__queue_receive({ mock_device.id, { capability = "switch", component = "main", command = "on", args = {} } })
+    if version.api > 15 then mock_device:expect_native_cmd_handler_registration("switch", "on") end
 
     test.socket.zigbee:__expect_send({ mock_device.id, OnOff.commands.On(mock_device) })
 
@@ -160,6 +162,7 @@ test.register_coroutine_test(
     test.socket.zigbee:__set_channel_ordering("relaxed")
     test.timer.__create_and_queue_test_time_advance_timer(2, "oneshot")
     test.socket.capability:__queue_receive({ mock_device.id, { capability = "switch", component = "main", command = "off", args = {} } })
+    if version.api > 15 then mock_device:expect_native_cmd_handler_registration("switch", "off") end
 
     test.socket.zigbee:__expect_send({ mock_device.id, OnOff.commands.Off(mock_device) })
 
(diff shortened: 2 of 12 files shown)


drivers/SmartThings/zigbee-switch/src/test/test_zll_rgbw_bulb.lua
```diff
@@ -200,18 +210,23 @@ test.register_coroutine_test(
...
 
     test.socket.zigbee:__expect_send({ mock_device.id, OnOff.commands.On(mock_device)})
     test.socket.zigbee:__expect_send({ mock_device.id, ColorControl.commands.MoveToColorTemperature(mock_device, 5000, 0x0000)})
 
     test.wait_for_events()
     test.mock_time.advance_time(2)
 
-    test.socket.zigbee:__expect_send({ mock_device.id, OnOff.attributes.OnOff:read(mock_device) })
-    test.socket.zigbee:__expect_send({ mock_device.id, Level.attributes.CurrentLevel:read(mock_device) })
+    if version.api <= 15 then
```
wkhenon: I was having issues running the `setColorTemperature` tests locally with this operator set to `<=` (no issues with `<`). Would this be pointing to a problem with how I'm using the version? Or is there some Lua magic that I'm missing?

greens: Well the way you've written things, wouldn't the initial on/offs _also_ not be sent when the native handlers are used?

@cjswedes probably has more insight here.

cjswedes: Why are there version checks at all for the polling/read messages? I expected there to only be a version check on the native handler registration call. The polling messages after the fact should remain the same whether the lua libs default handler is called or the subdriver is used because.

cjswedes (commented):
The entire `zll-dimmer-bulb` subdriver will no longer be necessary on api version 16 since we will put zll handling into the default handlers, so I think rather than changing the can_handle, it would be preferable to instantiate the subdriver conditionally based on the api version in the base driver

greens:
@cjswedes is there that much difference between having the lazy-loaded `can_handle` check the version vs checking the version before loading at all? Because checking the version before loading looks pretty non-idiomatic.

cjswedes:
> @cjswedes is there that much difference between having the lazy-loaded can_handle check the version vs checking the version before loading at all? Because checking the version before loading looks pretty non-idiomatic.

I think there is a decent sized difference. When we startup and "lazily" load the subdrivers, what were actually doing is the following:
* requiring them all one by one in their entirety
* extracting the can_handle function and any upvalues it needs
* dropping references to everything else
* collecting garbage

So some differences with what I suggested are:
* less latency at startup
* less latency in all event handling (since its one less can_handle)
* small memory savings
* less maintenance of a subdriver that should never be used

I dont think the latency and memory savings are that much, but it would be nice to get rid of a subdriver since it isnt needed anymore. Why cut out every individual handler when you can just cut out the whole subdriver?

drivers/SmartThings/zigbee-switch/src/zll-polling/init.lua
greens, line 1: so I _think_ you just want to change the e13-n11 test case, as I think it was previously incorrect, as it was not accounting for the config values in `configurations.lua`.

wkhenon: @greens Reverted the commit and updated the failed test

drivers/SmartThings/zigbee-switch/src/ikea-xy-color-bulb/init.lua
greens, line 1: It doesn't seem like you did a `git mv` here because it doesn't look like the file at the old location has been `git rm`'d.

wkhenon: @greens Don't we still need the file at the old location for `version < 16` (as a sub-driver of the `zll-dimmer-bulb` sub-driver)? Or are we moving the _entire_ `ikea-xy-color-bulb` sub-driver to the top level regardless of version?

greens: So, because they don't actually share any overlapping handlers:
```
  lifecycle_handlers = {
    init = configurationMap.power_reconfig_wrapper(device_init)
  },
  capability_handlers = {
    [capabilities.colorControl.ID] = {
      [capabilities.colorControl.commands.setColor.NAME] = set_color_handler,
      [capabilities.colorControl.commands.setHue.NAME] = set_hue_handler,
      [capabilities.colorControl.commands.setSaturation.NAME] = set_saturation_handler
    }
  },
  zigbee_handlers = {
    attr = {
      [ColorControl.ID] = {
        [ColorControl.attributes.CurrentX.ID] = current_x_attr_handler,
        [ColorControl.attributes.CurrentY.ID] = current_y_attr_handler
      }
    }
  },
  can_handle = can_handle_ikea_xy_color_bulb
```

vs

```
  lifecycle_handlers = {
    doConfigure = do_configure,
    added = device_added
  },
  capability_handlers = {
    [capabilities.switch.ID] = {
      [capabilities.switch.commands.on.NAME] = handle_switch_on,
      [capabilities.switch.commands.off.NAME] = handle_switch_off
    },
    [capabilities.switchLevel.ID] = {
      [capabilities.switchLevel.commands.setLevel.NAME] = handle_set_level
    },
    [capabilities.colorTemperature.ID] = {
      [capabilities.colorTemperature.commands.setColorTemperature.NAME] = handle_set_color_temperature
    }
  },
  sub_drivers = {
    require("zll-dimmer-bulb/ikea-xy-color-bulb")
  },
  can_handle = can_handle_zll_dimmer_bulb
```

They never needed to be nested in the first place, since the xy handler wasn't overriding anything. So just moving it up is fine and should make things simpler.

wkhenon: @greens Alrighty, it's been fully moved up & out (75517c2e6055a2a43a89dcaaafc9aae581022251)

PR #2444, opened 2025-10-01
Matter Electrical Sensor: Support Power Topology Cluster

hcarter-775:
# Description of Change
Multi-switch plugs that support Electrical Sensor on multiple endpoints are not currently handled correctly in the driver, as the electrical handling is silently dropped when creating a profile for all child devices. This PR aims to fix that by adding handling of the PowerTopology cluster, a required cluster for the Electrical Sensor device type.

This cluster primarily indicates how Electrical Sensors are organized on the device through its feature map, and this PR handles 2 of the permutations, namely SET and NODE. For now, all other permutations (primarily TREE) will be silently dropped if it is supported, though we know of no real devices using this topology.

In the spec, a NODE topology indicates that there is only one source of electrical readings for the entire device, the "node". A SET topology indicates that the electrical sensor supports readings for some subset of the devices, and therefore there are likely >1 of these sensors, each with a SET topology and each supporting readings of different endpoints. These endpoints are specified through the AvailableEndpoints attribute, which we read on an initial join.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
@@ -23,6 +23,9 @@ local fields = require "utils.switch_fields"
 local switch_utils = require "utils.switch_utils"
 local color_utils = require "utils.color_utils"
 
+local cfg = require "utils.device_configuration"
+local device_cfg = cfg.DeviceCfg
+
 local AttributeHandlers = {}
 
 -- [[ ON OFF CLUSTER ATTRIBUTES ]] --
@@ -316,8 +319,33 @@ function AttributeHandlers.energy_imported_factory(is_cumulative_report)
 end
 
 
--- [[ POWER SOURCE CLUSTER ATTRIBUTES ]] --
+-- [[ POWER TOPOLOGY CLUSTER ATTRIBUTES ]] --
 
+function AttributeHandlers.available_endpoints_handler(driver, device, ib, response)
+  local set_topology_eps = device:get_field(fields.SET_TOPOLOGY_EPS)
+  for i, ep in pairs(set_topology_eps or {}) do
+    if ep.endpoint_id == ib.endpoint_id then
+      set_topology_eps[i] = nil -- seen, remove from list
+      local tags = ""
+      if ep[clusters.ElectricalPowerMeasurement.ID] then tags = tags.."-power" end
+      if ep[clusters.ElectricalEnergyMeasurement.ID] then tags = tags.."-energy-powerConsumption" end
+      table.sort(ib.data.elements) -- for consistency, set tags on first listed EP
+      switch_utils.set_field_for_endpoint(device, fields.ELECTRICAL_TAGS_FOR_EP, ib.data.elements[1].value, tags)
+      break
+    end
+  end
+
+  if #set_topology_eps ~= 0 then -- we have not handled all eps
+    device:set_field(fields.SET_TOPOLOGY_EPS, set_topology_eps) -- permanently remove deleted ep
+  else
+    device:set_field(fields.SET_TOPOLOGY_EPS, nil)
+    device:set_field(fields.profiling_data.POWER_TOPOLOGY, clusters.PowerTopology.types.Feature.SET_TOPOLOGY)
+    device_cfg.match_profile(driver, device)
+  end
+end
+
+
+-- [[ POWER SOURCE CLUSTER ATTRIBUTES ]] --
 
 function AttributeHandlers.bat_percent_remaining_handler(driver, device, ib, response)
   if ib.data.value then
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -44,6 +44,8 @@ function SwitchLifecycleHandlers.device_added(driver, device)
   -- was created after the initial subscription report
   if device.network_type == device_lib.NETWORK_TYPE_CHILD then
     device:send(clusters.OnOff.attributes.OnOff:read(device))
+  elseif device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    switch_utils.collect_and_set_electrical_sensor_info(device)
   end
 
   -- call device init in case init is not called after added due to device caching
@@ -90,6 +92,11 @@ function SwitchLifecycleHandlers.device_init(driver, device)
       if ep.endpoint_id ~= main_endpoint then
         local id = 0
         for _, dt in ipairs(ep.device_types) do
+          if dt.device_type_id == fields.ELECTRICAL_SENSOR_ID then
+            for _, attr in pairs(fields.device_type_attribute_map[fields.ELECTRICAL_SENSOR_ID]) do
+              device:add_subscribed_attribute(attr)
+            end
+          end
           id = math.max(id, dt.device_type_id)
         end
         for _, attr in pairs(fields.device_type_attribute_map[id] or {}) do
@@ -171,6 +178,9 @@ local matter_driver_template = {
         [clusters.PowerSource.attributes.AttributeList.ID] = attribute_handlers.power_source_attribute_list_handler,
         [clusters.PowerSource.attributes.BatChargeLevel.ID] = attribute_handlers.bat_charge_level_handler,
         [clusters.PowerSource.attributes.BatPercentRemaining.ID] = attribute_handlers.bat_percent_remaining_handler,
+      },
+      [clusters.PowerTopology.ID] = {
+        [clusters.PowerTopology.attributes.AvailableEndpoints.ID] = attribute_handlers.available_endpoints_handler,
       },
       [clusters.RelativeHumidityMeasurement.ID] = {
         [clusters.RelativeHumidityMeasurement.attributes.MeasuredValue.ID] = attribute_handlers.relative_humidity_measured_value_handler
--- a/drivers/SmartThings/matter-switch/src/test/test_aqara_light_switch_h2.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_aqara_light_switch_h2.lua
@@ -35,7 +35,8 @@ local aqara_mock_device = test.mock_device.build_test_matter_device({
       clusters = {
         {cluster_id = clusters.Basic.ID, cluster_type = "SERVER"},
         {cluster_id = clusters.ElectricalPowerMeasurement.ID, cluster_type = "SERVER", cluster_revision = 1, feature_map = 2 },
-        {cluster_id = clusters.ElectricalEnergyMeasurement.ID, cluster_type = "SERVER", cluster_revision = 1, feature_map = 5 }
+        {cluster_id = clusters.ElectricalEnergyMeasurement.ID, cluster_type = "SERVER", cluster_revision = 1, feature_map = 5 },
+        {cluster_id = clusters.PowerTopology.ID, cluster_type = "SERVER", cluster_revision = 1, feature_map = 1 } -- NODE_TOPOLOGY
       },
       device_types = {
         {device_type_id = 0x0016, device_type_revision = 1}, -- RootNode
--- a/drivers/SmartThings/matter-switch/src/test/test_electrical_sensor.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_electrical_sensor.lua
@@ -16,6 +16,7 @@ local test = require "integration_test"
 local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
 local t_utils = require "integration_test.utils"
+local uint32 = require "st.matter.data_types.Uint32"
 local version = require "version"
 
 if version.api < 11 then
@@ -44,6 +45,7 @@ local mock_device = test.mock_device.build_test_matter_device({
       clusters = {
         { cluster_id = clusters.ElectricalEnergyMeasurement.ID, cluster_type = "SERVER", feature_map = 14, },
         { cluster_id = clusters.ElectricalPowerMeasurement.ID, cluster_type = "SERVER", feature_map = 0, },
+        { cluster_id = clusters.PowerTopology.ID, cluster_type = "SERVER", feature_map = 4, }, -- SET_TOPOLOGY
       },
       device_types = {
         { device_type_id = 0x0510, device_type_revision = 1 }, -- Electrical Sensor
@@ -56,9 +58,29 @@ local mock_device = test.mock_device.build_test_matter_device({
         {cluster_id = clusters.LevelControl.ID, cluster_type = "SERVER", feature_map = 2}
       },
       device_types = {
-        { device_type_id = 0x010A, device_type_revision = 1 } -- OnOff Plug
+        { device_type_id = 0x010B, device_type_revision = 1 }, -- OnOff Dimmable Plug
       }
-    }
+    },
+    {
+      endpoint_id = 3,
+      clusters = {
+        { cluster_id = clusters.ElectricalEnergyMeasurement.ID, cluster_type = "SERVER", feature_map = 14, },
+        { cluster_id = clusters.PowerTopology.ID, cluster_type = "SERVER", feature_map = 4, }, -- SET_TOPOLOGY
+      },
+      device_types = {
+        { device_type_id = 0x0510, device_type_revision = 1 }, -- Electrical Sensor
+      }
+    },
+    {
+      endpoint_id = 4,
+      clusters = {
+        { cluster_id = clusters.OnOff.ID, cluster_type = "SERVER", cluster_revision = 1, feature_map = 0, },
+        { cluster_id = clusters.LevelControl.ID, cluster_type = "SERVER", feature_map = 2},
+      },
+      device_types = {
+        { device_type_id = 0x010B, device_type_revision = 1 }, -- OnOff Dimmable Plug
+      }
+    },
   },
 })
 
@@ -82,18 +104,22 @@ local mock_device_periodic = test.mock_device.build_test_matter_device({
     {
       endpoint_id = 1,
       clusters = {
+        { cluster_id = clusters.OnOff.ID, cluster_type = "SERVER", cluster_revision = 1, feature_map = 0, },
         { cluster_id = clusters.ElectricalEnergyMeasurement.ID, cluster_type = "SERVER", feature_map = 10, },
+        { cluster_id = clusters.PowerTopology.ID, cluster_type = "SERVER", feature_map = 4, } -- SET_TOPOLOGY
       },
       device_types = {
-        { device_type_id = 0x0510, device_type_revision = 1 } -- Electrical Sensor
+        { device_type_id = 0x010A, device_type_revision = 1 }, -- OnOff Plug
+        { device_type_id = 0x0510, device_type_revision = 1 }, -- Electrical Sensor
       }
     },
   },
 })
 
 local subscribed_attributes_periodic = {
-  clusters.ElectricalEnergyMeasurement.attributes.PeriodicEnergyImported,
+  clusters.OnOff.attributes.OnOff,
   clusters.ElectricalEnergyMeasurement.attributes.CumulativeEnergyImported,
+  clusters.ElectricalEnergyMeasurement.attributes.PeriodicEnergyImported,
 }
 local subscribed_attributes = {
   clusters.OnOff.attributes.OnOff,
@@ -138,14 +164,19 @@ local periodic_report_val_23 = {
 }
 
 local function test_init()
+  test.mock_device.add_test_device(mock_device)
   local subscribe_request = subscribed_attributes[1]:subscribe(mock_device)
   for i, cluster in ipairs(subscribed_attributes) do
       if i > 1 then
           subscribe_request:merge(cluster:subscribe(mock_device))
       end
   end
+  test.socket.device_lifecycle:__queue_receive({ mock_device.id, "added" })
+  local read_req = clusters.PowerTopology.attributes.AvailableEndpoints:read(mock_device.id, 1)
+  read_req:merge(clusters.PowerTopology.attributes.AvailableEndpoints:read(mock_device.id, 3))
+  test.socket.matter:__expect_send({ mock_device.id, read_req })
   test.socket.matter:__expect_send({ mock_device.id, subscribe_request })
-  test.mock_device.add_test_device(mock_device)
+  test.socket.matter:__expect_send({ mock_device.id, subscribe_request })
 end
 test.set_test_init_function(test_init)
 
@@ -157,10 +188,12 @@ local function test_init_periodic()
         subscribe_request:merge(cluster:subscribe(mock_device_periodic))
     end
   end
-  test.socket.matter:__expect_send({ mock_device_periodic.id, subscribe_request })
   test.socket.device_lifecycle:__queue_receive({ mock_device_periodic.id, "added" })
+  local read_req = clusters.PowerTopology.attributes.AvailableEndpoints:read(mock_device_periodic.id, 1)
+  test.socket.matter:__expect_send({ mock_device_periodic.id, read_req })
   test.socket.matter:__expect_send({ mock_device_periodic.id, subscribe_request })
   test.socket.device_lifecycle:__queue_receive({ mock_device_periodic.id, "init" })
+  test.socket.matter:__expect_send({ mock_device_periodic.id, subscribe_request })
   test.socket.matter:__expect_send({ mock_device_periodic.id, subscribe_request })
 end
 
@@ -400,9 +433,20 @@ test.register_coroutine_test(
 test.register_coroutine_test(
   "Test profile change on init for Electrical Sensor device type",
   function()
+
     test.socket.device_lifecycle:__queue_receive({ mock_device.id, "doConfigure" })
-    mock_device:expect_metadata_update({ profile = "plug-level-power-energy-powerConsumption" })
     mock_device:expect_metadata_update({ provisioning_state = "PROVISIONED" })
+    test.wait_for_events()
+    test.socket.matter:__queue_receive({ mock_device.id, clusters.PowerTopology.attributes.AvailableEndpoints:build_test_report_data(mock_device, 1, {uint32(2)})})
+    test.socket.matter:__queue_receive({ mock_device.id, clusters.PowerTopology.attributes.AvailableEndpoints:build_test_report_data(mock_device, 3, {uint32(4)})})
+    mock_device:expect_metadata_update({ profile = "plug-level-power-energy-powerConsumption" })
+    mock_device:expect_device_create({
+      type = "EDGE_CHILD",
+      label = "nil 2",
+      profile = "plug-level-energy-powerConsumption",
+      parent_device_id = mock_device.id,
+      parent_assigned_child_key = string.format("%d", 4)
+    })
   end,
   { test_init = test_init }
 )
@@ -411,8 +455,10 @@ test.register_coroutine_test(
   "Test profile change on init for only Periodic Electrical Sensor device type",
   function()
     test.socket.device_lifecycle:__queue_receive({ mock_device_periodic.id, "doConfigure" })
-    mock_device_periodic:expect_metadata_update({ profile = "plug-energy-powerConsumption" })
     mock_device_periodic:expect_metadata_update({ provisioning_state = "PROVISIONED" })
+    test.wait_for_events()
+    test.socket.matter:__queue_receive({ mock_device_periodic.id, clusters.PowerTopology.attributes.AvailableEndpoints:build_test_report_data(mock_device_periodic, 1, {uint32(1)})})
+    mock_device_periodic:expect_metadata_update({ profile = "plug-energy-powerConsumption" })
   end,
   { test_init = test_init_periodic }
 )
--- a/drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
@@ -473,6 +473,8 @@ end
 
 local function test_init_dimmer()
   test.mock_device.add_test_device(mock_device_dimmer)
+  test.socket.device_lifecycle:__queue_receive({ mock_device_dimmer.id, "added" })
+  test.socket.device_lifecycle:__queue_receive({ mock_device_dimmer.id, "init" })
   test.socket.device_lifecycle:__queue_receive({ mock_device_dimmer.id, "doConfigure" })
   mock_device_dimmer:expect_metadata_update({ profile = "switch-level" })
   mock_device_dimmer:expect_metadata_update({ provisioning_state = "PROVISIONED" })
@@ -505,6 +507,7 @@ local function test_init_mounted_on_off_control()
   test.socket.matter:__expect_send({mock_device_mounted_on_off_control.id, subscribe_request})
 
   test.socket.device_lifecycle:__queue_receive({ mock_device_mounted_on_off_control.id, "doConfigure" })
+  mock_device_mounted_on_off_control:expect_metadata_update({ profile = "switch-binary" })
   mock_device_mounted_on_off_control:expect_metadata_update({ provisioning_state = "PROVISIONED" })
 end
 
@@ -526,6 +529,7 @@ local function test_init_mounted_dimmable_load_control()
   test.socket.matter:__expect_send({mock_device_mounted_dimmable_load_control.id, subscribe_request})
 
   test.socket.device_lifecycle:__queue_receive({ mock_device_mounted_dimmable_load_control.id, "doConfigure" })
+  mock_device_mounted_dimmable_load_control:expect_metadata_update({ profile = "switch-level" })
   mock_device_mounted_dimmable_load_control:expect_metadata_update({ provisioning_state = "PROVISIONED" })
 end
 
@@ -566,6 +570,7 @@ local function test_init_parent_child_different_types()
   test.socket.matter:__expect_send({mock_device_parent_child_different_types.id, subscribe_request})
 
   test.socket.device_lifecycle:__queue_receive({ mock_device_parent_child_different_types.id, "doConfigure" })
+  mock_device_parent_child_different_types:expect_metadata_update({ profile = "switch-binary" })
   mock_device_parent_child_different_types:expect_metadata_update({ provisioning_state = "PROVISIONED" })
 
   mock_device_parent_child_different_types:expect_device_create({
@@ -617,6 +622,7 @@ local function test_init_light_level_motion()
   test.socket.matter:__expect_send({mock_device_light_level_motion.id, subscribe_request})
 
   test.socket.device_lifecycle:__queue_receive({ mock_device_light_level_motion.id, "doConfigure" })
+  mock_device_light_level_motion:expect_metadata_update({ profile = "light-level-motion" })
   mock_device_light_level_motion:expect_metadata_update({ provisioning_state = "PROVISIONED" })
 end
 
--- a/drivers/SmartThings/matter-switch/src/test/test_multi_switch_parent_child_lights.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_multi_switch_parent_child_lights.lua
@@ -189,6 +189,7 @@ local function test_init()
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
 
   test.socket.device_lifecycle:__queue_receive({ mock_device.id, "doConfigure" })
+  mock_device:expect_metadata_update({ profile = "light-binary" })
   mock_device:expect_metadata_update({ provisioning_state = "PROVISIONED" })
 
   for _, child in pairs(mock_children) do
@@ -260,6 +261,7 @@ local function test_init_parent_child_endpoints_non_sequential()
   test.socket.matter:__expect_send({mock_device_parent_child_endpoints_non_sequential.id, subscribe_request})
 
   test.socket.device_lifecycle:__queue_receive({ mock_device_parent_child_endpoints_non_sequential.id, "doConfigure" })
+  mock_device_parent_child_endpoints_non_sequential:expect_metadata_update({ profile = "light-binary" })
   mock_device_parent_child_endpoints_non_sequential:expect_metadata_update({ provisioning_state = "PROVISIONED" })
 
   for _, child in pairs(mock_children_non_sequential) do
--- a/drivers/SmartThings/matter-switch/src/test/test_multi_switch_parent_child_plugs.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_multi_switch_parent_child_plugs.lua
@@ -24,6 +24,8 @@ local child_profile_override = t_utils.get_profile_definition("switch-binary.yml
 local parent_ep = 10
 local child1_ep = 20
 local child2_ep = 30
+local child3_ep = 40
+local child4_ep = 50
 
 local mock_device = test.mock_device.build_test_matter_device({
   label = "Matter Switch",
@@ -69,6 +71,24 @@ local mock_device = test.mock_device.build_test_matter_device({
         {device_type_id = 0x010A, device_type_revision = 2} -- On/Off Plug
       }
     },
+    {
+      endpoint_id = child3_ep,
+      clusters = {
+        {cluster_id = clusters.OnOff.ID, cluster_type = "SERVER"},
+      },
+      device_types = {
+        {device_type_id = 0x010A, device_type_revision = 2} -- On/Off Plug
+      }
+    },
+    {
+      endpoint_id = child4_ep,
+      clusters = {
+        {cluster_id = clusters.OnOff.ID, cluster_type = "SERVER"},
+      },
+      device_types = {
+        {device_type_id = 0x010A, device_type_revision = 2} -- On/Off Plug
+      }
+    }
   }
 })
 
@@ -146,6 +166,7 @@ local function test_init()
   test.socket.matter:__expect_send({mock_device.id, subscribe_request})
 
   test.socket.device_lifecycle:__queue_receive({ mock_device.id, "doConfigure" })
+  mock_device:expect_metadata_update({ profile = "plug-binary" })
   mock_device:expect_metadata_update({ provisioning_state = "PROVISIONED" })
 
   for _, child in pairs(mock_children) do
@@ -166,6 +187,22 @@ local function test_init()
     profile = "plug-binary",
     parent_device_id = mock_device.id,
     parent_assigned_child_key = string.format("%d", child2_ep)
+  })
+
+  mock_device:expect_device_create({
+    type = "EDGE_CHILD",
+    label = "Matter Switch 4",
+    profile = "plug-binary",
+    parent_device_id = mock_device.id,
+    parent_assigned_child_key = string.format("%d", child3_ep)
+  })
+
+  mock_device:expect_device_create({
+    type = "EDGE_CHILD",
+    label = "Matter Switch 5",
+    profile = "plug-binary",
+    parent_device_id = mock_device.id,
+    parent_assigned_child_key = string.format("%d", child4_ep)
   })
 end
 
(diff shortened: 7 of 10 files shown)


drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
```diff
@@ -22,6 +22,20 @@ local fields = require "utils.switch_fields"
 local switch_utils = require "utils.switch_utils"
 local color_utils = require "utils.color_utils"
 
+local cfg = require "utils.device_configuration"
+local device_cfg = cfg.DeviceCfg
+
+-- Include driver-side definitions when lua libs api version is < 11
+if version.api < 11 then
+  clusters.ElectricalEnergyMeasurement.ID = 0x0091
+  clusters.ElectricalPowerMeasurement.ID = 0x0090
```
nickolas-deboom: Aren't these ID's already present in the embedded clusters? Or is it just that the IDs are the only thing needed in this module?

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -247,15 +261,10 @@ end
 function AttributeHandlers.active_power_handler(driver, device, ib, response)
   if ib.data.value then
     local watt_value = ib.data.value / fields.CONVERSION_CONST_MILLIWATT_TO_WATT
-    if ib.endpoint_id ~= 0 then
```
nickolas-deboom, line 238: Is this logic not needed anymore? I'm a bit confused

hcarter-775: see this PR: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2436
specifically here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2436#issuecomment-3362828124

drivers/SmartThings/matter-switch/src/generic_handlers/attribute_handlers.lua
hcarter-775: yeah I could have just included the whole thing there, but since only the IDs are being used in this file I just specified the IDs.

hcarter-775: for simplicity, I decided to load the whole cluster here, since it _should_ be loaded anyway in this context

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -267,45 +275,91 @@ end
...
+    if ib.data.elements.energy then
+      local energy_imported_wh = ib.data.elements.energy.value / fields.CONVERSION_CONST_MILLIWATT_TO_WATT
+      if is_periodic_report then
+        -- handle this report only if cumulative reports are not supported
+        if device:get_field(fields.CUMULATIVE_REPORTS_SUPPORTED) then return end
+        energy_imported_wh = energy_imported_wh + energy_meter_latest_state
+      end
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.energyMeter.energy({ value = energy_imported_wh, unit = "Wh" }))
+      local energy_delta_wh = energy_imported_wh - energy_meter_latest_state
+      switch_utils.increment_field(device, fields.TOTAL_IMPORTED_ENERGY, energy_delta_wh, true)
```
nickolas-deboom:
```suggestion
      device:set_field(fields.TOTAL_IMPORTED_ENERGY, energy_delta_wh, {persist = true})
```
Couldn't this just be this?

hcarter-775: no, the energy delta from a single report is not equivalent to the total energy

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -267,45 +275,91 @@ end
...
+
+function AttributeHandlers.available_endpoints_handler(driver, device, ib, response)
+  local set_topology_eps = device:get_field(fields.ELECTRICAL_SENSOR_EPS)
+  for i, ep in pairs(set_topology_eps or {}) do
+    if ep.endpoint_id == ib.endpoint_id then
+      set_topology_eps[i] = nil -- seen, remove from list
+      local tags = ""
+      if ep[clusters.ElectricalPowerMeasurement.ID] then tags = tags.."-power" end
+      if ep[clusters.ElectricalEnergyMeasurement.ID] then tags = tags.."-energy-powerConsumption" end
+      table.sort(ib.data.elements)
```
nickolas-deboom: I think that `augment_type` may be needed before this line if the embedded cluster is being utilized

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -267,45 +275,91 @@ end
...
+      local primary_available_ep = ib.data.elements[1].value -- for consistency, associate data with first listed EP
```
nickolas-deboom: Could the AvailableEndpoints attribute be an empty list?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -205,67 +234,104 @@ function utils.get_endpoints_by_device_type(device, device_type_id)
...
+  local parts_list_req = im.InteractionRequest(im.InteractionRequest.RequestType.READ, {})
+  for _, ep in ipairs(device.endpoints) do
+    if utils.tbl_contains(el_dt_eps, ep.endpoint_id) then
+      local el_ep_info = { endpoint_id = ep.endpoint_id }
+      for _, cluster in ipairs(ep.clusters) do
+        el_ep_info[cluster.cluster_id] = cluster.feature_map -- key the cluster's feature map on each supported cluster id
+      end
+      table.insert(electrical_sensor_eps, el_ep_info)
+      -- these read requests will ONLY be sent if the device supports the TREE_TOPOLOGY or SET_TOPOLOGY features, respectively
+      parts_list_req:merge(clusters.Descriptor.attributes.PartsList:read(device, ep.endpoint_id)) -- TREE read
```
nickolas-deboom: Do we want to use subscriptions rather than reads as we have favored in other places recently?

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -267,45 +275,91 @@ end
...
-    device:set_field(fields.TOTAL_IMPORTED_ENERGY, summed_energy_report, {persist = true})
-    device:emit_event(capabilities.energyMeter.energy({ value = summed_energy_report, unit = "Wh" }))
-    switch_utils.report_power_consumption_to_st_energy(device, device:get_field(fields.TOTAL_IMPORTED_ENERGY))
+
+-- [[ POWER TOPOLOGY CLUSTER ATTRIBUTES ]] --
+
+function AttributeHandlers.available_endpoints_handler(driver, device, ib, response)
+  local set_topology_eps = device:get_field(fields.ELECTRICAL_SENSOR_EPS)
+  for i, ep in pairs(set_topology_eps or {}) do
+    if ep.endpoint_id == ib.endpoint_id then
```
nickolas-deboom: I'm a little confused on what this check is doing. `AvailableEndpoints` is a list of endpoints, so why are we only processing this for `ib.endpoint_id` aka the endpoint containing the `PowerTopology` cluster? I'm thinking about the case where a device implements `PowerTopology` on endpoint 1 and `ElectricalEnergyMeasurement` on endpoint 2 for example.

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
nickolas-deboom: I meant to use `energy_imported_wh`

nickolas-deboom: As in 
```
      device:set_field(fields.TOTAL_IMPORTED_ENERGY, energy_imported_wh, {persist = true})
```

hcarter-775: ah, no, since this field may be written to by multiple endpoints.

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
hcarter-775: It definitely shouldn't, but I suppose it could, so not a bad idea to add checking in case that's true

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
nickolas-deboom: Could you clarify what you mean? It looks like `increment_field` is doing the same thing, just with an extra step

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: I think this should be a separate PR, since the introduced complexity is greater than this PR, and does not directly affect this PR.

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
hcarter-775: `PowerTopology` should never be on a different endpoint than `ElectricalEnergyMeasurement` though, since they are both associated with the Electrical Sensor device type.

nickolas-deboom: Ah thanks I see, I didn't realize its a mandatory cluster for ElectricalSensor

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -267,45 +275,91 @@ end
...
+
+function AttributeHandlers.parts_list_handler(driver, device, ib, response)
+  local tree_topology_eps = device:get_field(fields.ELECTRICAL_SENSOR_EPS)
+  for i, ep in pairs(tree_topology_eps or {}) do
+    if ep.endpoint_id == ib.endpoint_id then
+      tree_topology_eps[i] = nil -- seen, remove from list
+      local tags = ""
+      if ep[clusters.ElectricalPowerMeasurement.ID] then tags = tags.."-power" end
+      if ep[clusters.ElectricalEnergyMeasurement.ID] then tags = tags.."-energy-powerConsumption" end
+      table.sort(ib.data.elements)
```
nickolas-deboom: augment_type also needed here

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
hcarter-775: increment field is not using the `energy_imported_wh` field, it is taking its own field and incrementing the delta. The delta can come from multiple different endpoints, each of which may have a different total value. These are being summed in this field.

nickolas-deboom: If there were multiple endpoints reporting imported energy, would they all overwrite the energyMeter capability since we only have one capability? I'm just wondering what `local energy_delta_wh = energy_imported_wh - energy_meter_latest_state` really means, since `energy_meter_latest_state` is being pulled from energyMeter on the main component rather than something specific to an endpoint

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -176,8 +193,9 @@ function utils.endpoint_to_component(device, ep)
   return "main"
 end
 
-function utils.find_child(parent, ep_id)
-  return parent:get_child_by_parent_assigned_key(string.format("%d", ep_id))
+function utils.find_child(parent_device, ep_id)
+  local primary_ep_key = utils.get_field_for_endpoint(parent_device, fields.PRIMARY_ASSOCIATED_EP, ep_id) or ep_id
```
nickolas-deboom: Could you clarify why this change is needed? Does this mean that all events for child devices will be routed through the primary child device? If there were other, non-electrical sensor child devices, would they be affected?

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
hcarter-775: actually, not sure it would. This is not a complex element list (it's just a bunch of integers, so there's no type to attempt to conform to

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
hcarter-775: as stated [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2444#discussion_r2535504129), I think this is the same

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -205,67 +234,104 @@ function utils.get_endpoints_by_device_type(device, device_type_id)
...
-  for _, capability in ipairs(fields.supported_capabilities) do
-    if device:supports_capability(capability) then
-      return false
-    end
+  -- every profile except for matter-thing supports at least 2 capabilities (refresh, firmwareUpdate)
+  for i, _ in pairs(device.profile.components.main.capabilities) do
+    if i > 1 then return false end
   end
-  return device:supports_capability(capabilities.refresh)
+  return true
```
nickolas-deboom:
```suggestion
  return #device.profile.components.main.capabilities < 2
```
Could simplify further maybe?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: if the "primary associated ep" field is populated for an endpoint, it will be used in place of that endpoint. if not, functionality will remain the same as before. This is to allow association of multiple endpoints to one child device, since a child device is only accessible through a single endpoint id, not multiple.

Therefore, we'd need something like: 

(child device) <- keyed to with endpoint 2 (for example) <- keyed to through primary associated ep field through endpoint 4 (for example).

This is how endpoint 4 could send an event to the child device

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: unfortunately we cannot do that, since the metatable len functionality is not defined for the capabilities on older firmwares. What this means in practice is #capabilities is always 0 for all profiles

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
nickolas-deboom: Ah nice

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
hcarter-775: added https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2444/commits/2b99576efaeccccaab74b72ff118822533da4261

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
nickolas-deboom: Ah, interesting.

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
hcarter-775: Multiple endpoints would not be handled by the same `energyMeter` capability, that is not feasible based on the current profiling logic. Each supported energy cluster would map only to one `energyMeter` capability, mapped across different child devices. So though it's a little confusing where the source of truth is, even using the latest state, we are still sourcing everything directly from specific endpoint event emission.

nickolas-deboom: Thanks for the explanation! that makes sense now

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -95,26 +91,18 @@ SwitchFields.COLOR_MODE = "__color_mode"
...
+  [0x1321] = { -- SONOFF_MANUFACTURER_ID
     [0x000C] = { target_profile = "switch-binary", initial_profile = "plug-binary" },
     [0x000D] = { target_profile = "switch-binary", initial_profile = "plug-binary" },
   },
   [0x115F] = { -- AQARA_MANUFACTURER_ID
-    [0x1003] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 2 Buttons(Generic Switch), 1 Channel(On/Off Light)
-    [0x1004] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 2 Buttons(Generic Switch), 2 Channels(On/Off Light)
-    [0x1005] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 4 Buttons(Generic Switch), 3 Channels(On/Off Light)
-    [0x1008] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 2 Buttons(Generic Switch), 1 Channel(On/Off Light)
-    [0x1009] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 4 Buttons(Generic Switch), 2 Channels(On/Off Light)
```
nickolas-deboom, line 116: I feel like this may have come up in a previous pr but I can't recall. Why are these being removed?

hcarter-775: this is now the default logic, so no override is required

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -281,67 +309,104 @@ function utils.get_endpoints_by_device_type(device, device_type_id)
...
+  end
+
+  local electrical_ep = electrical_sensor_eps[1] or {}
+  device:set_field(fields.POWER_CONSUMPTION_REPORT_EP, electrical_ep.endpoint_id, { persist = true })
+
+  local electrical_ep_has_feature = function(feature_name)
+      local feature = clusters.PowerTopology.types.Feature[feature_name]
+      if feature then
+        return clusters.PowerTopology.are_features_supported(feature, electrical_ep[clusters.PowerTopology.ID] or 0)
+      end
```
nickolas-deboom:
```suggestion
    return clusters.PowerTopology.are_features_supported(feature, electrical_ep[clusters.PowerTopology.ID] or 0)
```
and then pass in the full feature name at the call sites i.e. `clusters.PowerTopology.types.Feature.SET_TOPOLOGY`

nickolas-deboom: If you do this, maybe just remove the helper entirely since it would just be one line at that point

hcarter-775: I agree about the first suggestion, though the one liner is still a little unwieldy imo. I'd like to keep the mini helper function I think

hcarter-775: changed

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -95,26 +97,19 @@ SwitchFields.COLOR_MODE = "__color_mode"
...
 SwitchFields.updated_fields = {
   { current_field_name = "__component_to_endpoint_map_button", updated_field_name = SwitchFields.COMPONENT_TO_ENDPOINT_MAP },
-  { current_field_name = "__switch_intialized", updated_field_name = nil }
+  { current_field_name = "__switch_intialized", updated_field_name = nil },
+  { current_field_name = "__energy_management_endpoint", updated_field_name = nil },
+  { current_field_name = "__total_imported_energy", updated_field_name = nil },
 }
 
-SwitchFields.HUE_SAT_COLOR_MODE = clusters.ColorControl.types.ColorMode.CURRENT_HUE_AND_CURRENT_SATURATION
-SwitchFields.X_Y_COLOR_MODE = clusters.ColorControl.types.ColorMode.CURRENTX_AND_CURRENTY
```
hcarter-775, line 103: these were defined twice, this removes one of their definitions.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -45,6 +45,14 @@ function utils.set_field_for_endpoint(device, field, endpoint, value, additional
   device:set_field(string.format("%s_%d", field, endpoint), value, additional_params)
 end
 
+function utils.remove_field_value(device, field_name, value)
```
nickolas-deboom: nit- call it `index` rather than `value`?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -340,62 +348,57 @@ function utils.report_power_consumption_to_st_energy(device, total_imported_ener
...
+    table.insert(electrical_sensor_eps_info, utils.get_endpoint_info(device, ep_id))
+    parts_list_req:merge(clusters.Descriptor.attributes.PartsList:read(device, ep_id))
+    available_eps_req:merge(clusters.PowerTopology.attributes.AvailableEndpoints:read(device, ep_id))
   end
 
-  if electrical_ep_has_feature(clusters.PowerTopology.types.Feature.SET_TOPOLOGY) then
-    device:set_field(fields.ELECTRICAL_SENSOR_EPS, electrical_sensor_eps) -- assume any other stored EPs also have a SET topology
+  -- check the feature map for the first (or only) Electrical Sensor EP
+  local power_topology_cluster_info = utils.find_cluster_for_ep(electrical_sensor_eps_info[1], clusters.PowerTopology.ID)
+  local power_topology_feature_map = power_topology_cluster_info.feature_map or 0
```
nickolas-deboom: Looks like `find_cluster_for_ep` could return nil so might need 
```
local power_topology_feature_map = power_topology_cluster_info and power_topology_cluster_info.feature_map or 0
```

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -267,45 +270,74 @@ end
...
-    device:set_field(fields.TOTAL_IMPORTED_ENERGY, watt_hour_value, {persist = true})
-    if ib.endpoint_id ~= 0 then
-      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.energyMeter.energy({ value = watt_hour_value, unit = "Wh" }))
+function AttributeHandlers.energy_imported_factory(is_periodic_report)
+  return function(driver, device, ib, response)
+    if version.api < 11 then
+      clusters.ElectricalEnergyMeasurement.types.EnergyMeasurementStruct:augment_type(ib.data)
+    end
+    if ib.data.elements.energy then
+      local energy_imported_wh = ib.data.elements.energy.value / 1000 -- convert received milliwatt to watt
```
ctowns: nit
```suggestion
      local energy_imported_wh = ib.data.elements.energy.value / 1000 -- convert received milliwatt-hour to watt-hour
```

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -297,50 +318,86 @@ function utils.detect_bridge(device)
...
+    device:set_field(fields.ELECTRICAL_SENSOR_EPS, electrical_sensor_eps_info) -- assume any other stored EPs also have a SET topology
+    device:send(available_eps_req)
+    return
+  elseif clusters.PowerTopology.are_features_supported(clusters.PowerTopology.types.Feature.TREE_TOPOLOGY, power_topology_feature_map) then
+    device:set_field(fields.ELECTRICAL_SENSOR_EPS, electrical_sensor_eps_info) -- assume any other stored EPs also have a TREE topology
+    device:send(parts_list_req)
+    return
+  elseif clusters.PowerTopology.are_features_supported(clusters.PowerTopology.types.Feature.NODE_TOPOLOGY, power_topology_feature_map) then
+    -- EP has a NODE topology, so there is only ONE Electrical Sensor EP
+    device:set_field(fields.profiling_data.POWER_TOPOLOGY, clusters.PowerTopology.types.Feature.NODE_TOPOLOGY, {persist=true})
```
ctowns, line 405: Does this field also need to be set for the `SET` and `TREE` topologies also?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -297,50 +318,86 @@ function utils.detect_bridge(device)
...
-      deltaEnergy = energy_delta_wh,
-      energy = latest_total_imported_energy_wh
-    }))
+function utils.set_fields_for_electrical_sensor_endpoint(device, electrical_ep_info, associated_endpoint_ids)
+  local tags = ""
+  if utils.find_cluster_for_ep(electrical_ep_info, clusters.ElectricalPowerMeasurement.ID) then tags = tags.."-power" end
+  if utils.find_cluster_for_ep(electrical_ep_info, clusters.ElectricalEnergyMeasurement.ID) then tags = tags.."-energy-powerConsumption" end
+  if associated_endpoint_ids ~= {} then
+    table.sort(associated_endpoint_ids)
+    local primary_associated_ep_id = type(associated_endpoint_ids[1]) == "table" and associated_endpoint_ids[1].value or associated_endpoint_ids[1]
```
ctowns: Can we define an expectation for the value of the `associated_endpoint_ids` parameter so that this check is unnecessary? The caller can ensure that the data is in the correct format, which essentially means that we ensure that since this is an internal helper function. It might be somewhat cleaner, and would clearly define the expected input.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -297,50 +318,86 @@ function utils.detect_bridge(device)
...
-      ["end"] = epoch_to_iso8601(current_time - 1),
-      deltaEnergy = energy_delta_wh,
-      energy = latest_total_imported_energy_wh
-    }))
+function utils.set_fields_for_electrical_sensor_endpoint(device, electrical_ep_info, associated_endpoint_ids)
+  local tags = ""
+  if utils.find_cluster_for_ep(electrical_ep_info, clusters.ElectricalPowerMeasurement.ID) then tags = tags.."-power" end
+  if utils.find_cluster_for_ep(electrical_ep_info, clusters.ElectricalEnergyMeasurement.ID) then tags = tags.."-energy-powerConsumption" end
+  if associated_endpoint_ids ~= {} then
+    table.sort(associated_endpoint_ids)
```
ctowns, line 367: It looks like we will just use the lowest numbered endpoint as the "main" electrical/power measurement endpoint. Is there any specific reason for this? And is it documented somewhere?

ctowns (commented):
An overall comment: with the increasing complexity of this driver, I think it could be beneficial to provide `LuaDoc` style comments on certain functions in order to improve readability and help create more defined function/parameter expectations. There is a lot of complexity required to handle these increasingly complex device types, but I think that the driver could benefit from more documentation to help with tracking the expect values of parameters, and also the expected information to be stored on the device table, which has newe fields that are needed to help us handle these more complex device types.

Here is an example of a function with a LuaDoc style comment from our lua libs:
```
--- Given the endpoint ID find the corresponding component for this device
---
--- This will use the function set by st.matter.Device:set_endpoint_to_component_fn to
--- return the appropriate component given the endpoint.  If the function is unset
--- it defaults to "main"
---
--- @param ep number the endpoint ID to find the component for
--- @return string the component ID the endpoint matches to
function MatterDevice:endpoint_to_component(ep)
  local ep_to_comp = self:get_field(matter_device.ENDPOINT_TO_COMPONENT_FUNC)
  if ep_to_comp ~= nil then
    return ep_to_comp(self, ep)
  else
    return "main"
  end
end
```

This is more of a general improvement to the driver and is something that I think could help with improving maintainability and reviews. I don't think it's neccessary to comment on every function, and it doesn't need to happen in this PR per se, but I think it could be an overall improvement to this driver.

What are your thoughts on something like this?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: yes, and will be set in the associated `availableEndpoints` and `partsList` handlers respectively, once all data for those topology types has been collected.

hcarter-775:
@ctowns I agree, and have thought about the LuaDoc style commenting. I also believe (like you noted) this is outside the scope of this particular PR, but it is something I've been thinking of doing for a while. [This PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2560) for example puts some effort into getting that format for a couple particular functions, but I'd like to do it for all of them.

My main sticking point is the (probably minor) memory impact, so I'd like to look into ways to define the luadoc perhaps in a different file that is never loaded, or in the longer term how we might strip comments when packing drivers.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: This is relatively arbitrary and done to create some amount of internal consistency. One actual use case would be that it would keep the electrical sensor functionality associated with the default OnOff endpoint. I can add a comment noting this arbitrariness.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: good idea, implemented in the latest

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: shouldn't be possible in this case, but agreed, technically could be. done

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: done!

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
hcarter-775: done

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
nickolas-deboom: Ah, gotcha thanks!

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -267,45 +270,82 @@ end
...
+      if is_periodic_report then
+        -- handle this report only if cumulative reports are not supported
+        if device:get_field(fields.CUMULATIVE_REPORTS_SUPPORTED) then return end
+        local state_device = switch_utils.find_child(device, ib.endpoint_id) or device
+        local energy_meter_latest_state = state_device:get_latest_state(
+          "main", capabilities.energyMeter.ID, capabilities.energyMeter.energy.NAME, 0 -- 0 as the default if state is nil
+        )
+        energy_imported_wh = energy_imported_wh + energy_meter_latest_state
+      end
+      device:emit_event_for_endpoint(ib.endpoint_id, capabilities.energyMeter.energy({ value = energy_imported_wh, unit = "Wh" }))
```
ctowns, line 288: Should this be `state_device` instead of just `device`? I'm not entirely sure I am following what `state_device` is supposed to be in this case

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -267,45 +270,82 @@ end
...
+--- AvailableEndpoints: This attribute SHALL indicate the list of endpoints capable of
+--- providing power to and/or consuming power from the endpoint hosting this server.
+---
+--- In the case there are multiple endpoints supporting the PowerTopology cluster with
+--- SET feature, all AvailableEndpoints responses must be handled before profiling.
+function AttributeHandlers.available_endpoints_handler(driver, device, ib, response)
+  local set_topology_eps = device:get_field(fields.ELECTRICAL_SENSOR_EPS)
+  for i, set_ep_info in pairs(set_topology_eps or {}) do
+    if ib.endpoint_id == set_ep_info.endpoint_id then
+      switch_utils.remove_field_index(device, fields.ELECTRICAL_SENSOR_EPS, i) -- ep found, remove from table
```
ctowns: I might just be missing context, but for both of these handlers (`available_endpoint_handler` and `parts_list_handler`) I don't understand what is being done by this function and why this endpoint is being removed. I think some additional documentation could be useful here for developers outside this scope of expertise. Maybe just a quick line explaining why these endpoints are being processed and why the source endpoint for this handler seems to be removed from the list that is later processed.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -297,50 +318,95 @@ function utils.detect_bridge(device)
 end
 
 function utils.detect_matter_thing(device)
```
nickolas-deboom: Looks like this function isn't being used anymore for a while. Could remove?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: comment added

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: removed

ctowns: Why is this function no longer being used? It was needed for handling a special case with On/Off devices that supported bindings even though we do not support bindings: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1082

Although it looks like it's use was removed prior to this PR - is this really no longer needed?

hcarter-775: yup, the logic to handle that case has been re-worked to be part of the generic device profiling flow, and so there is no gate anymore to do the check if the device profile is currently a `matter-thing`.

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -267,45 +270,87 @@ end
...
-    switch_utils.report_power_consumption_to_st_energy(device, device:get_field(fields.TOTAL_IMPORTED_ENERGY))
+
+-- [[ POWER TOPOLOGY CLUSTER ATTRIBUTES ]] --
+
+--- AvailableEndpoints: This attribute SHALL indicate the list of endpoints capable of
+--- providing power to and/or consuming power from the endpoint hosting this server.
+---
+--- In the case there are multiple endpoints supporting the PowerTopology cluster with
+--- SET feature, all AvailableEndpoints responses must be handled before profiling.
+function AttributeHandlers.available_endpoints_handler(driver, device, ib, response)
```
nickolas-deboom, line 304: just a nit but since this handler along with `parts_list_handler` are identical, you could have them share a handler, and perhaps just pass in `clusters.PowerTopology.types.Feature.SET_TOPOLOGY` or `clusters.PowerTopology.types.Feature.TREE_TOPOLOGY` in the driver template, since that's the only difference between the handlers as far as I can tell

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -70,20 +77,14 @@ function SwitchLifecycleHandlers.device_init(driver, device)
     switch_utils.check_field_name_updates(device)
     device:set_component_to_endpoint_fn(switch_utils.component_to_endpoint)
     device:set_endpoint_to_component_fn(switch_utils.endpoint_to_component)
+    device:extend_device("get_latest_state_for_endpoint", switch_utils.get_latest_state_for_endpoint)
```
ctowns: nit: Do we need to add this for all devices or just those that support electrical energy measurements? Also, if it is being added as a blanket statement for all devices, why don't we just call it directly? It is not overriding a library function and is just adding a different function pointer to the device object that points to other functions that we already call throughout the driver by using `switch_utils.function_name`

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
hcarter-775: we could, but to me the `partsList` handler is much more generic than just this. At the moment it's 1-1 with availableEndpoints but that's only due to its specific use case right now. Therefore I'd like to keep them separate.

drivers/SmartThings/matter-switch/src/embedded_clusters/Descriptor/server/attributes/init.lua
```diff
@@ -0,0 +1,27 @@
+-- Copyright © 2025 SmartThings, Inc.
```
nickolas-deboom, line 1: Have you tested using embedded clusters? Maybe not important for the power topology cluster but would be good to test the descriptor cluster since it was only introduced in lua libs 16

hcarter-775: yes I have! Since this v16 FW is not released, the device tested that used the TREE topology was on "older" aka current FW

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
ctowns: Discussed offline and more comments were added to the function :+1:

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: I agree that this is a semi-arbitrary decision. I personally think it makes more sense to handle it like so, leaving it a blanket call since there's not much harm in not doing that, and having a `device:get_latest_state_for_endpoint` looks more consistent to me as a developer than `switch_utils.get_latest_state_for_endpoint(device, ...)`.

Personally, I kinda want to do the same for some other functions like writing `device:get_field_for_endpoint` makes more "sense" to me than `switch_utils.get_field_for_endpoint(device, ...)`

Ultimately arbitrary though

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -285,51 +311,101 @@ function utils.detect_bridge(device)
...
+---
+--- @param endpoint_info number|table an endpoint id or an ib (the ib data includes endpoint_id, cluster_id, and attribute_id fields)
+function utils.get_latest_state_for_endpoint(device, endpoint_info, capability_id, attribute_id, default_value, default_state_table)
+  if type(endpoint_info) == "number" then
+    endpoint_info = { endpoint_id = endpoint_info }
   end
-  return device:supports_capability(capabilities.refresh)
+
+  local component = device:endpoint_to_component(endpoint_info)
+  local state_device = utils.find_child(device, endpoint_info.endpoint_id) or device
```
nickolas-deboom, line 326: Why is `find_child` used here?

drivers/SmartThings/matter-switch/src/init.lua
ctowns: I am not going to push back too much, but this decision seems arbitrary to me and inconsistent with how other functions in `switch_utils` that take the `device` table as a parameter are called throughout the driver.

This makes me think that we really need to make stronger coding/style guidelines for the drivers. I personally wouldn't use `extend_device` unless we were actually extending a library function, and I would push for consistency in following the behavior used by other switch util functions where the device is passed as a parameter where needed. But that is just my 2 cents :money_with_wings:

drivers/SmartThings/matter-switch/src/embedded_clusters/Descriptor/server/attributes/init.lua
nickolas-deboom: Ok perfect!

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: This is an expansion of what was already in the PR, recall that a device may be a child device for `get_latest_state` with these introduced changes.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: In the long run, my current "idea" (open to critique/iteration) is to create a separate file, something like `device_utils.lua`, that holds all of the `device:<blank>` function overrides and extensions. Then, this file would also have a setter function that could be called in the `init` function to set all this stuff and therefore make it clear what is/isn't extensible as part of the `device:function` framework. AKA is the function primarily acted on from the device, or is the device just a small part of the driver process.

ctowns (approved):

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -73,18 +73,20 @@ SwitchFields.device_type_profile_map = {
...
 -- button devices that require component mapping.
 SwitchFields.COMPONENT_TO_ENDPOINT_MAP = "__component_to_endpoint_map"
-SwitchFields.ENERGY_MANAGEMENT_ENDPOINT = "__energy_management_endpoint"
 SwitchFields.IS_PARENT_CHILD_DEVICE = "__is_parent_child_device"
+
+--- If the ASSIGNED_CHILD_KEY field is populated for an endpoint, it should be
+--- used as the key in the get_child_by_parent_assigned_key() function. This allows
+--- multiple endpoints to associate with the same child device, though right now child
+--- devices are keyed using only one endpoint id.
+SwitchFields.ASSIGNED_CHILD_KEY = "__ASSIGNED_CHILD_KEY"
```
nickolas-deboom: tiny nit but could you change these new field strings here and below to lowercase to match the others?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -258,12 +279,17 @@ function utils.matter_handler(driver, device, response_block)
...
   local dt_eps = {}
   for _, ep in ipairs(device.endpoints) do
     for _, dt in ipairs(ep.device_types) do
       if dt.device_type_id == device_type_id then
-        table.insert(dt_eps, ep.endpoint_id)
+        if opts.with_info then
+          table.insert(dt_eps, ep)
+        else
+          table.insert(dt_eps, ep.endpoint_id)
+        end
```
nickolas-deboom, line 292: add break here?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -33,6 +45,14 @@ function utils.set_field_for_endpoint(device, field, endpoint, value, additional
   device:set_field(string.format("%s_%d", field, endpoint), value, additional_params)
 end
 
+function utils.remove_field_index(device, field_name, index)
+  local new_table = device:get_field(field_name)
+  if type(new_table) == "table" then
+    new_table[index] = nil -- remove value associated with index from table
+  end
+  device:set_field(field_name, new_table)
```
nickolas-deboom: nit but could move this inside the if statement?

drivers/SmartThings/matter-switch/src/test/test_electrical_sensor_tree.lua
```diff
@@ -0,0 +1,393 @@
+-- Copyright © 2024 SmartThings, Inc.
```
nickolas-deboom: date

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: I don't think so, cause there might be >1

nickolas-deboom: But we only want to add the same endpoint one time right?

hcarter-775: oh I see what you mean- like the case where a device really breaks spec and has >1 of the same device type on a single endpoint. Or just to cut checks a little. Will add

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: @ctowns FYI after our conversation here I decided to remove this use of `extend_device` for now. I think that the more generic framework I described is kinda neat and worth exploring, but until I can do it "generally" I think this one-off use is more confusing than it is useful.

ctowns: Sounds good! I think the framework you proposed makes sense and would be a good candidate for making those changes as an "all-in-one" improvement similar to what you've done with the refactoring and reorganization of these drivers

PR #2445, opened 2025-10-02
zigbee-switch: Filter incorrect network type

tandres:
A small number of zwave devices failed to migrate correctly and ended up as zwave devices attached to the zigbee switch driver. This change adds a patched driver.lua which includes a mechanism to filter devices when building device info which don't match device type filters added to the driver.

https://smartthings.atlassian.net/browse/CHAD-16558

Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/non_zigbee_devices/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/non_zigbee_devices/init.lua
@@ -1,0 +1,57 @@
+-- Copyright 2025 SmartThings
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- This is a patch for the zigbee-switch driver to fix https://smartthings.atlassian.net/browse/CHAD-16558
+-- Several hubs were found that had zigbee switch drivers hosting zwave devices.
+-- This patch works around it until hubcore 0.59 is released with
+-- https://smartthings.atlassian.net/browse/CHAD-16552
+
+local st_device = require "st.device"
+local log = require "log"
+
+local function can_handle(opts, driver, device)
+  if device.network_type ~= st_device.NETWORK_TYPE_ZIGBEE and device.network_type ~= st_device.NETWORK_TYPE_CHILD then
+    return true, require("non_zigbee_devices")
+  end
+  return false
+end
+
+local function device_added(driver, device, event)
+    log.info(string.format("Non zigbee device added: %s", device))
+end
+
+local function device_init(driver, device, event)
+    log.info(string.format("Non zigbee device init: %s", device))
+end
+
+local function do_configure(driver, device)
+    log.info(string.format("Non zigbee do configure: %s", device))
+end
+
+local function info_changed(driver, device, event, args)
+    log.info(string.format("Non zigbee infoChanged: %s", device))
+end
+
+local non_zigbee_devices = {
+  NAME = "non zigbee devices filter",
+  lifecycle_handlers = {
+    init = device_init,
+    added = device_added,
+    doConfigure = do_configure,
+    infoChanged = info_changed
+  },
+  can_handle = can_handle
+}
+
+return non_zigbee_devices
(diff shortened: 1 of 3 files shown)


varzac:
I think rather than including the entire driver.lua file here (since we often have to leave workarounds like this in for a long time due to slow to update hubs), I would rather we just override the individual functions.

drivers/SmartThings/zigbee-switch/src/zigbee_driver_patch.lua
```diff
@@ -0,0 +1,160 @@
...
+    if self.device_cache ~= nil then
+      self.device_cache[device_uuid] = nil
+    end
+    device.thread:queue_event(device.deleted, device)
+  end
+end
+
+function Driver:_filter_network_type(raw_device_table)
+  if self.device_network_type_filter then
+    return self.device_network_type_filter[raw_device_table.network_type] ~= nil
```
cjswedes:
```suggestion
    return self.device_network_type_filter[raw_device_table.network_type] or false
```
Not critical, but in case we ever add an entry with `false`, this would still work.

drivers/SmartThings/zigbee-switch/src/test/test_bad_device_kind.lua
```diff
@@ -0,0 +1,29 @@
...
+local mock_device = test.mock_device.build_test_zwave_device({
+  profile = t_utils.get_profile_definition("on-off-level.yml"),
+})
+
+local function test_init()
+  test.mock_device.add_test_device(mock_device)
+end
+
+test.set_test_init_function(test_init)
+test.register_test("mismatched_prot_ignored", function() end, nil)
```
greens: I don't understand this test.

tandres: I pushed up some additional comments on the test. It's just making sure the driver doesn't crash when you add a ZWave device to it.

greens:
I agree with @varzac . These "temporary" fixes tend to stick around for a while. I think it would be better to write this as a sub-driver full of no-ops.

varzac:
> I agree with @varzac . These "temporary" fixes tend to stick around for a while. I think it would be better to write this as a sub-driver full of no-ops.

I hadn't considered a subdriver full of noops.  Originally I was thinking that may not be sufficient because any functions in the driver class may be assuming a specific device type.  But looking at the zigbee driver class, the only place the device object is referenced is if the `health_check` stuff is true, and we have that disabled.  So we might be able to consider this approach as a short term fix.  I don't necessarily have a problem with the approach taken here, but I do think a subdriver might be a neater solution.

drivers/SmartThings/zigbee-switch/src/non_zigbee_devices/init.lua
```diff
@@ -0,0 +1,52 @@
...
+-- This is a patch for the zigbee-switch driver to fix https://smartthings.atlassian.net/browse/CHAD-16558
+-- Several hubs were found that had zigbee switch drivers hosting zwave devices.
+-- This patch works around it until hubcore 0.59 is released with
+-- https://smartthings.atlassian.net/browse/CHAD-16552
+
+local st_device = require "st.device"
+local log = require "log"
+
+local function can_handle(opts, driver, device)
+  if device.network_type ~= st_device.NETWORK_TYPE_ZIGBEE then
```
varzac: Child devices shouldn't use this subdriver either.

drivers/SmartThings/zigbee-switch/src/test/test_bad_device_kind.lua
```diff
@@ -0,0 +1,37 @@
...
+  profile = t_utils.get_profile_definition("on-off-level.yml"),
+})
+
+-- Just validating that the driver doesn't crash is enough to validate
+-- that the work-around is effective in ignoring the incorrect device kind
+test.register_coroutine_test("mismatched_device_kind_ignored", function()
+    test.mock_device.add_test_device(mock_device)
+    test.socket.device_lifecycle:__queue_receive({ mock_device.id, "added" })
+    test.wait_for_events()
+    test.socket.device_lifecycle:__queue_receive({ mock_device.id, "init" })
```
varzac, line 38: include `doConfigure` as well

cjswedes (commented):
Can you add a test, where a non zigbee device has a capability command come in, to make sure that doesnt crash.

One thing I liked about monkey patching the driver, is that it more deeply prevented the device from being used in by anything in the zigbee driver than the subdriver approach does.

cjswedes (approved):

tandres:
> Can you add a test, where a non zigbee device has a capability command come in, to make sure that doesnt crash. I also think that `infoChanged` could cause problems, so please add that to the existing test.
> 
> One thing I liked about monkey patching the driver, is that it more deeply prevented the device from being used in by anything in the zigbee driver than the subdriver approach does.

Info changed added, also added a command test which doesn't seem to break anything. My driver skillz are weak so I appreciate the in-depth review!

PR #2446, opened 2025-10-03
Matter Lock: Allow subscription updates after modular-to-modular profile updates

hcarter-775:
# Description of Change
On top of checking whether a profile id has changed, this checks whether the stored `profile.components` has changed, specifically for hubs on 0.58 or greater. This would be useful for a modular profile update where the id has not changed, but new optional capabilities are included in the inventory sync. (For example, permitting a modular fingerprint in some cases).

# Summary of Completed Tests
Tested by adding a lock-modular fingerprint on a driver running on the hub 58 FW. When the profile update occurred, I confirmed the subscription updated in the `infoChanged` as expected.

Regression tested by onboarding a device on a static lock fingerprint and checking for the same results as the above.

### DIFF
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -285,9 +285,29 @@ local function match_profile_switch(driver, device)
   device:try_update_metadata({profile = profile_name})
 end
 
+local function compare_components(synced_components, prev_components)
+  if #synced_components ~= #prev_components then
+    return false
+  end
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-lock/capabilities/lockAlarm.yml
```diff
@@ -1,27 +0,0 @@
-id: lockAlarm
```
hcarter-775, line 1: this capability now exists and is used throughout the driver using an updated definition. Removing this embedded def that is only used in the test files.

PR #2454, opened 2025-10-07
WWSTCERT-8745 Add support to frient vibration sensor wiszb 137

marcintyminski:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
+++ b/drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
@@ -20,7 +20,6 @@ local battery_defaults = require "st.zigbee.defaults.battery_defaults"
 local device_management = require "st.zigbee.device_management"
 local data_types = require "st.zigbee.data_types"
 local threeAxis = capabilities.threeAxis
-local log = require "log"
 
 local TemperatureMeasurement = zcl_clusters.TemperatureMeasurement
 local IASZone = zcl_clusters.IASZone


drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
```diff
@@ -0,0 +1,260 @@
...
+
+  -- Process the attribute records
+  for _, attribute_record in ipairs(zb_rx.body.zcl_body.attr_records) do
+    local attribute_id = attribute_record.attr_id.value
+    local axis_value = attribute_record.data.value
+
+    if attribute_id == 0x0000 then
+      measured_x = axis_value
+      device:set_field("measured_x", measured_x)
+      log.trace("Updated X value: " .. axis_value)
```
greens: if you're satisfied with the behavior, please remove the logging

drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
```diff
@@ -0,0 +1,260 @@
...
+    end
+  end
+  return false
+end
+
+local Frient_AccelerationMeasurementCluster = {
+  ID = 0xFC04,
+  ManufacturerSpecificCode = 0x1015,
+  attributes = {
+    MeasuredValueX = { ID = 0x0000, data_type = data_types.name_to_id_map["Int16"] },
```
greens: you can just use `data_types.Int16`

drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
```diff
@@ -0,0 +1,260 @@
...
+  generate_event_from_zone_status(driver, device, zb_rx.body.zcl_body.zone_status, zb_rx)
+end
+
+local function device_init(driver, device)
+    log.trace "Initializing sensor"
+    battery_defaults.build_linear_voltage_init(2.3, 3.0)(driver, device)
+    --Add the manufacturer-specific attributes to generate their configure reporting and bind requests
+    for _, config in pairs(get_cluster_configurations()) do
+        device:add_configured_attribute(config)
+    end
```
greens: nit: we use 2 spaces

drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
```diff
@@ -0,0 +1,241 @@
...
+      end
+    elseif device.preferences.contactSensorAxis == "Z" then
+      local initial_position = device.preferences.sensorInitialPosition or 0
+      if math.abs(initial_position - measured_z) >= device.preferences.contactSensorValue - device.preferences.contactSensorValue * (device.preferences.tolerance / 100) then
+        device:emit_event(capabilities.contactSensor.contact.open())
+      else
+        device:emit_event(capabilities.contactSensor.contact.closed())
+      end
+    end
+  end
```
greens:
```suggestion
  if device.preferences.garageSensor == "Yes" then
    local garageAxis = measured_x
    if device.preferences.contactSensorAxis == "Y" then 
      garageAxis = measured_y 
    elseif device.preferences.contactSensorAxis == "Z" then
      garageAxis = measured_z
    end
    local initial_position = device.preferences.sensorInitialPosition or 0
    if math.abs(initial_position - garageAxis) >= device.preferences.contactSensorValue - device.preferences.contactSensorValue * (device.preferences.tolerance / 100) then
      device:emit_event(capabilities.contactSensor.contact.open())
    else
      device:emit_event(capabilities.contactSensor.contact.closed())
    end    
  end
```

marcintyminski: I have made a change and pushed the change to pr

drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
```diff
@@ -0,0 +1,231 @@
...
+      device:set_field("measured_y", measured_y)
+    elseif attribute_id == 0x0002 then
+      measured_z = axis_value
+      device:set_field("measured_z", measured_z)
+    end
+  end
+
+  -- Ensure all values are non-nil before emitting
+  if measured_x and measured_y and measured_z then
+    device:emit_event(threeAxis.threeAxis({measured_x, measured_y, measured_z}))
```
greens, line 70: Don't the `measured_{}` fields need to be reset after an event is sent so that a new event is only sent when you get three _new_ measurements?

greens: would be good to have a test case of this.

GAFfrient: @greens Each value shall update whenever any of them changes; therefore, not only when 3 of them change.

GAFfrient: Just to clarify on it, the body will always contains of three values, as these values are never nill or empty. Also, each of values shall update whenever there is a change on them.

GAFfrient: > would be good to have a test case of this.

I can see that we have some test cases for X, Y, Z axes cases - can you clarify on which additional test cases you're thinking about? Thanks!

greens: My comment was mostly if you had been expecting to always wait for all three attributes to be updated before sending an event, but if you're always expecting them to come in a group, you can use local variables rather than the device store, and in your tests, you can construct the message like we do here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-contact/src/test/test_smartthings_multi_sensor.lua#L158 to give you a more accurate representation of what you expect the actual device to send.

GAFfrient: @greens Just to be sure, since I don't understand you fully - what's wrong with the current way the driver behaves?

We want an update when any of the 3 axes changes, not only when 3 of them change.

Please let me know how to proceed, we'd like to close this PR.

greens: @GAFfrient You can correct me if that's not the case here, but usually when acceleration sensors move, they rotate on all three axes, and since it seems like you're sending all three attribute updates at the same time anyway I just worry about the case where, after the first update, every update will cause 3 events to be emitted in quick succession.

GAFfrient: It's not the case here. With this sensor, it is possible that one axis won't change when sensor moves.

greens: In that situation, would your device send a single attribute report for one axis or a bundled report of all three axes (with 2 unchanged values)?

marcintyminski: The device would send bundled report with the changed value, and last cached values for remaining axes.

greens: Right, so your driver would send three events. 2 of those are likely to get bounced for being duplicates, but it's not an ideal situation to have to rely on that.

Since you can always rely on the device sending all 3 attributes every time, it should be easy to update the driver to only send *one* event after it's received a report for all three attributes.

marcintyminski: done

greens: I think you can now omit the usage of `device:set_field` and `get_field`.

drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
```diff
@@ -0,0 +1,230 @@
...
+    elseif attribute_id == Frient_AccelerationMeasurementCluster.attributes.MeasuredValueZ.ID then
+      measured_z = axis_value
+      device:set_field("measured_z", measured_z)
+    end
+  end
+
+  if measured_x and measured_y and measured_z then
+    device:emit_event(threeAxis.threeAxis({measured_x, measured_y, measured_z}))
+
+    if device.preferences.garageSensor == "Yes" then
```
greens: a better check is probably to see if the device supports the `contactSensor` capability.

drivers/SmartThings/zigbee-contact/profiles/acceleration-motion-temperature-battery.yml
```diff
@@ -0,0 +1,50 @@
...
+  - id: temperatureMeasurement
+    version: 1
+  - id: battery
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: VibrationsSensor
```
greens: I am not certain this category is defined.

drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
marcintyminski: done

drivers/SmartThings/zigbee-contact/src/frient/frient-vibration/init.lua
marcintyminski: Changed

drivers/SmartThings/zigbee-contact/profiles/acceleration-motion-temperature-battery.yml
marcintyminski: You're right. Changed it to ContactSensor

PR #2455, opened 2025-10-07
Remove dkjson usages

greens:


cjswedes (approved):
Approving, but I think we need an on target test, particularly of the case where there are first 0 user codes, then 1 user code, and then 0 user codes again.

PR #2458, opened 2025-10-12
WWSTCERT-8471/8468 [Aqara] Wireless Remote Switch T1(Single/Double)

seojune79:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-button/profiles/aqara-double-buttons.yml
+++ b/drivers/SmartThings/zigbee-button/profiles/aqara-double-buttons.yml
@@ -6,6 +6,8 @@ components:
         version: 1
       - id: batteryLevel
         version: 1
+      - id: firmwareUpdate
+        version: 1
       - id: refresh
         version: 1
     categories:


drivers/SmartThings/zigbee-button/profiles/aqara-double-buttons.yml
```diff
@@ -0,0 +1,30 @@
...
+  - id: main
+    capabilities:
+      - id: button
+        version: 1
+      - id: batteryLevel
+        version: 1
+      - id: refresh
+        version: 1
+    categories:
+      - name: Button
```
Kwang-Hui: @seojune79 Could you change this to RemoteController also?

seojune79: @Kwang-Hui Yes, I've reflected that.

drivers/SmartThings/zigbee-button/src/aqara/init.lua
```diff
@@ -29,74 +29,113 @@ local MFG_CODE = 0x115F
...
-    if value.value == 1 then
-        device:emit_event(capabilities.button.button.pushed({state_change = true}))
-    elseif value.value == 2 then
-        device:emit_event(capabilities.button.button.double({state_change = true}))
-    elseif value.value == 0 then
-        device:emit_event(capabilities.button.button.held({state_change = true}))
-    end
+  local end_point = zb_rx.address_header.src_endpoint.value
+  local btn_evt_cnt = FINGERPRINTS[device:get_model()].btn_cnt or 1
+  local evt = capabilities.button.button.held({ state_change = true })
```
greens, line 62: should we be explicit about this only being the case when the value is `0`, just in case a future device uses values > 2?

drivers/SmartThings/zigbee-button/src/aqara/init.lua
```diff
@@ -29,74 +29,113 @@ local MFG_CODE = 0x115F
...
+  if value.value == 1 then
+    evt = capabilities.button.button.pushed({ state_change = true })
+  elseif value.value == 2 then
+    evt = capabilities.button.button.double({ state_change = true })
+  end
+  device:emit_event(evt)
+  if btn_evt_cnt > 1 then
+    device:emit_component_event(device.profile.components[COMP_LIST[end_point]], evt)
```
greens, line 70: It seems like button events will only be emitted from the `all` component when they come from endpoint 3, but they will always be emitted from the `main` endpoint. That's a bit confusing. 

In the past we've used the `main` component as an any/all component for button presses. It's a pattern with some precedent. Did you avoid that with your profiles because the name is not descriptive?

drivers/SmartThings/zigbee-button/profiles/one-button-batteryLevel.yml
greens, line 1: rename to `one-button-batteryLevel` and include the `firmwareUpdate` capability.

drivers/SmartThings/zigbee-button/src/aqara/init.lua
seojune79: I don't anticipate any issues, as it should work the same way with values of 2 or higher

drivers/SmartThings/zigbee-button/src/aqara/init.lua
seojune79: To monitor all button events on the card, they are set to be emitted from the main component. Endpoint 3 triggers an event when both button 1 and button 2 are pressed simultaneously, so the all component is used to indicate that both buttons must be pressed together.

drivers/SmartThings/zigbee-button/profiles/one-button-batteryLevel.yml
seojune79: I have updated.

drivers/SmartThings/zigbee-button/src/aqara/init.lua
Kwang-Hui: @greens This device appears to have two physical buttons, but it actually functions like a 3-button device. When both buttons are pressed simultaneously, it triggers the button event for the all component, but not for button1 or button2. However, when naming it button3, it would be difficult for users to associate it with pressing simultaneously, so it was suggested to create it as an all component.
The function of the main component is intended to update the main device simultaneously whenever an event occurs at any endpoint, just like other multi-button devices.

If you have better idea about `all` component naming, please let me know

greens: I'd suggest `both`, but it's not a big deal. I understand now.

drivers/SmartThings/zigbee-button/src/test/test_aqara_button.lua
greens, line 40: This device should still be using the `one-button-battery` profile, as you have not updated the fingerprint.

seojune79: I have updated the fingerprints for all Aqara Button Devices by implementing the batteryLevel capability.

drivers/SmartThings/zigbee-button/profiles/aqara-double-buttons.yml
```diff
@@ -0,0 +1,30 @@
+name: aqara-double-buttons
+components:
+  - id: main
+    capabilities:
+      - id: button
+        version: 1
+      - id: batteryLevel
+        version: 1
+      - id: refresh
+        version: 1
```
greens, line 12: this profile needs to include `firmwareUpdate`

seojune79: Okay, I have added the firmwareUpdate capability.

PR #2466, opened 2025-10-14
Matter Switch: Attempt re-profiling on device software updates

hcarter-775:
# Description of Change
We store data related to the device software version (saved from the BasicInformation cluster's softwareVersion attribute) on the driver st_store. This leverages that version to attempt a re-profiling if this is ever updated. The same data is what is leveraged by the hub to re-interview a device on a device update as well.

This is built on top of the following PR: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2481 since that allows for more generic re-profiling of the parent devices.
 
# Summary of Completed Tests

Tested that `try_update_metadata` works as expected for child devices.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -36,10 +36,8 @@ function SwitchLifecycleHandlers.device_added(driver, device)
   -- refresh child devices to get an initial attribute state for OnOff in case child device
   -- was created after the initial subscription report
   if device.network_type == device_lib.NETWORK_TYPE_CHILD then
-    print("???1")
     device:send(clusters.OnOff.attributes.OnOff:read(device))
   elseif device.network_type == device_lib.NETWORK_TYPE_MATTER then
-    print("???2")
     switch_utils.handle_electrical_sensor_info(device)
   end
 


drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
```diff
@@ -47,31 +47,31 @@ SwitchFields.CURRENT_HUESAT_ATTR_MAX = 254
...
-SwitchFields.DIMMABLE_PLUG_DEVICE_TYPE_ID = 0x010B
-SwitchFields.ON_OFF_SWITCH_ID = 0x0103
-SwitchFields.ON_OFF_DIMMER_SWITCH_ID = 0x0104
-SwitchFields.ON_OFF_COLOR_DIMMER_SWITCH_ID = 0x0105
+SwitchFields.AGGREGATOR_ID = 0x000E
+SwitchFields.ON_OFF_LIGHT_ID = 0x0100
+SwitchFields.DIMMABLE_LIGHT_ID = 0x0101
+SwitchFields.COLOR_TEMPERATURE_LIGHT_ID = 0x010C
+SwitchFields.EXTENDED_COLOR_LIGHT_ID = 0x010D
+SwitchFields.ON_OFF_PLUG_IN_UNIT_ID = 0x010A
```
hcarter-775: can be reverted, but I changed the device type fields to more closely match the spec-defined device types (`ON_OFF_PLUG` -> `ON_OFF_PLUG_IN_UNIT`) and I cut the `DEVICE_TYPE` out of the ones that had it, since it's kinda self-explanatory. If we think it's not, I might move them into their own file with a `DeviceType` header or something.

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
```diff
@@ -199,40 +198,27 @@ end
...
+
+  local server_onoff_eps = device:get_endpoints(clusters.OnOff.ID, { cluster_type = "SERVER" })
+  if #server_onoff_eps > 0 then
+    SwitchDeviceConfiguration.create_child_switch_devices(driver, device, server_onoff_eps, main_endpoint)
+    -- workaround: finds a profile for devices of the Light Switch device type set that break spec and implement OnOff as 'server' instead of 'client'.
+    -- note: since the Light Switch device set isn't supported, these devices join as a matter-thing.
+    if switch_utils.detect_matter_thing(device) then
+      profile_name = SwitchDeviceConfiguration.match_light_switch_device_profile(device, main_endpoint)
+    end
+  end
```
hcarter-775: This may look like a change in functionality but it is not.

When removing the `initialize_buttons_and_switches` helper function, it became clear there was some formatting that no longer made a lot of sense. Aka, the semi-arbitrary `profile_found = true` statements and the
```
  if profile_found then
    return
  end
```
statement based around those declarations that had been in place since more logic was hidden by the helper function.

So what I did was remove that profile_found logic, and now have `match_light_switch_device_profile` use the same `try_metadata_update` as everything else (at the bottom of match_profile), only returning the profile name instead of doing an actual update.

I would have done the same for the button logic, but it gets a little messy due to the electrical energy stuff. This will be fixed though in the powerTopology PR, so we can streamline that too. I think having one point of contact with `try_metadata_update` is clearer, catches more edge cases by default, and is easier to maintain, which is why I'm moving the logic in that direction.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -66,11 +66,16 @@ end
 function SwitchLifecycleHandlers.info_changed(driver, device, event, args)
   if device.profile.id ~= args.old_st_store.profile.id then
     device:subscribe()
-    local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
-    if #button_eps > 0 and device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    if device.network_type == device_lib.NETWORK_TYPE_MATTER then
```
greens, line 63: is the checking for button support done inside `button_cfg`?

hcarter-775: yes, this function will be a no-op if there are no button-mappable endpoints. So I could revert it, but it removes an extra check. Maybe this adds confusion though?

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -72,18 +72,23 @@ function SwitchDeviceConfiguration.create_child_devices(driver, device, server_o
...
           profile = child_profile,
           parent_device_id = device.id,
           parent_assigned_child_key = string.format("%d", ep_id),
-          vendor_provided_label = name
-        }
-      )
+          vendor_provided_label = child_device_name
+        })
+      else
+        existing_child_device:try_update_metadata({
```
nickolas-deboom, line 67: Does this function work correctly for child devices? I think I remember you asking about that at some point but I don't remember what came from that discussion

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: thought about this more, and I improved the api by having the caller pass in the endpoints that will be mapped, still cutting out the extra check while keeping this clarity

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
hcarter-775: yes it does. I tested it and the cloud event is triggered, followed by an expected `infoChanged` event.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -127,6 +105,10 @@ function SwitchLifecycleHandlers.device_init(driver, device)
   end
 end
 
+function SwitchLifecycleHandlers.device_removed(driver, device)
```
hcarter-775, line 99: I just moved this to the bottom of the file. We never touch this and I personally am always getting confused where device_init is when developing with this placed above it.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -60,16 +60,21 @@ end
...
-    local button_eps = device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
-    if #button_eps > 0 and device.network_type == device_lib.NETWORK_TYPE_MATTER then
-      button_cfg.configure_buttons(device)
+    if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+      device:subscribe()
+      button_cfg.configure_buttons(device,
+        device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
+      )
+    elseif device.network_type == device_lib.NETWORK_TYPE_CHILD then
+      switch_utils.update_subscriptions(device:get_parent_device()) -- parent device required to scan through EPs and update subscriptions
```
nickolas-deboom, line 69: Could `update_subscriptions` just be done in the `device.network_type == device_lib.NETWORK_TYPE_MATTER` arm, and continue ignoring child devices?

hcarter-775: What if the device whose profile is updated is only a child device, and the parent is unchanged?

nickolas-deboom: Oh yeah, good call

PR #2468, opened 2025-10-15
WWSTCERT-8461 Aqara dimmer controller t1

haedo-doo:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
@@ -15,12 +15,6 @@ local PRIVATE_CLUSTER_ID = 0xFCC0
 local PRIVATE_ATTRIBUTE_ID = 0x0009
 local MFG_CODE = 0x115F
 
-local FINGERPRINTS = {
-  { mfr = "LUMI", model = "lumi.light.acn004" },
-  { mfr = "Aqara", model = "lumi.light.acn014" },
-  { mfr = "LUMI", model = "lumi.light.cwacn1" }
-}
-
 local function do_refresh(self, device)
   device:send(OnOff.attributes.OnOff:read(device))
   device:send(Level.attributes.CurrentLevel:read(device))
--- a/drivers/SmartThings/zigbee-switch/src/aqara-light/can_handle.lua
+++ b/drivers/SmartThings/zigbee-switch/src/aqara-light/can_handle.lua
@@ -4,7 +4,8 @@
 return function(opts, driver, device)
   local FINGERPRINTS = {
     { mfr = "LUMI", model = "lumi.light.acn004" },
-    { mfr = "Aqara", model = "lumi.light.acn014" }
+    { mfr = "Aqara", model = "lumi.light.acn014" },
+    { mfr = "LUMI", model = "lumi.light.cwacn1" }
   }
 
   for _, fingerprint in ipairs(FINGERPRINTS) do


drivers/SmartThings/zigbee-switch/src/preferences.lua
greens: you should just be able to add the new model here rather than duplicating this map entirely

drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
```diff
@@ -54,9 +55,20 @@ local function set_level_handler(driver, device, cmd)
...
+local function init(self, device)
+  local min = 2700
+  local max = 6000
+
+  if device:get_model() == "lumi.light.cwacn1" then
+    max = 6500
+  end
+  device:emit_event(capabilities.colorTemperature.colorTemperatureRange({ minimum = min, maximum = max }))
+end
+
```
greens, line 64: I understand the instinct here so that existing devices get this change, but we do not want this event to fire every time the driver starts.

I would add this event to the `added` handler as well, and a follow-up PR a week later should be created to remove this code here in `init`. I'd also add gating to check for an existing `colorTemperatureRange` value before emitting this event.

drivers/SmartThings/zigbee-switch/src/preferences.lua
haedo-doo: Sorry for the late confirmation. It's been updated—please check again.

drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
haedo-doo: Sorry, I'm a bit confused. Should I add this event to the added handler and remove the one in init in a follow-up PR a week later—instead of doing it now?

greens: So we want to make sure that existing already-installed devices get this change, which is (I assumed) why you placed it in `init`, since that is called every time the driver starts, rather than in `added`, which is only called once in the lifecycle of the device.

In general, this code should only be in the `added` handler, but for a week, I recommend having it in both `added` and `init` so that already-installed devices will update. 

After that there should be a follow-up PR to remove the code from `init` since it will no longer be needed.

drivers/SmartThings/zigbee-switch/src/preferences.lua
```diff
@@ -75,11 +78,29 @@ preferences.sync_preferences = function(driver, device)
...
+    end
+    -- Array format: { { mfr = "...", model = "..." }, ... }
+    if type(matrix) == "table" then
+      for _, match in ipairs(matrix) do
+        if match.mfr == mfr and match.model == model then
+          return device.PARAMETERS
+        end
+      end
     end
+    ::continue::
```
greens: what was the bug here that required this rewrite?

haedo-doo: If I add the new model here rather than duplicating the entire map, ```preferences``` won't work properly because the existing ```get_device_parameters``` function can't handle the Array format for the model.

drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
haedo-doo: OK, got it. I'll add this event to the ```added``` handler first and remove the code from ```init``` in a follow-up PR later. Thanks!

drivers/SmartThings/zigbee-switch/src/preferences.lua
greens:
```suggestion
    MATCHING_MATRIX = {
      { mfr = "Aqara", model = "lumi.light.acn014" },
    }
```

drivers/SmartThings/zigbee-switch/src/preferences.lua
```diff
@@ -75,11 +78,29 @@ preferences.sync_preferences = function(driver, device)
 end
 
 preferences.get_device_parameters = function(zigbee_device)
+  local mfr   = zigbee_device:get_manufacturer()
+  local model = zigbee_device:get_model()
+
   for _, device in pairs(devices) do
```
greens, line 73:
```suggestion
  for _, device in pairs(devices) do
    for _, fingerprint in pairs(device.MATCHING_MATRIX) do
```

drivers/SmartThings/zigbee-switch/src/preferences.lua
greens: I've left some suggestions that seem a lot simpler than this rewrite, based on how this is handled elsewhere in the code base.

haedo-doo: Thanks for the suggestions, I've updated the code.

drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
```diff
@@ -54,9 +67,20 @@ local function set_level_handler(driver, device, cmd)
...
 end
 
+local function init(self, device)
+  local min = 2700
+  local max = 6000
+
+  if device:get_model() == "lumi.light.cwacn1" then
+    max = 6500
+  end
+  device:emit_event(capabilities.colorTemperature.colorTemperatureRange({ minimum = min, maximum = max }))
```
greens: Can you add the `emit_event_if_latest_state_missing` check for this code, since it runs every time the driver starts?

haedo-doo: OK, done.

greens:
You will need to rebase now, as we did a major reconfiguration of the zigbee switch driver code to reduce memory usage

haedo-doo:
> You will need to rebase now, as we did a major reconfiguration of the zigbee switch driver code to reduce memory usage

I've rebased.

PR #2469, opened 2025-10-15
Lazy loading test for zigbee switch

cjswedes:
We are considering adding a new way to lazy load subdrivers in 0.59.x, and this is what we would want to do to our zigbee switch driver to have it work with that functionality. The improvement is that we avoid requiring the base subdriver entirely to get access to the can handle and any nested subdrivers. So to convert our drivers subdrivers will all need to have a `can_handle.lua` file and any base driver or subdriver that has subdrivers will need a `sub_drivers.lua` file.

- [x] Undo name update from testing

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/aqara-light/can_handle.lua
+++ b/drivers/SmartThings/zigbee-switch/src/aqara-light/can_handle.lua
@@ -1,3 +1,6 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
 return function(opts, driver, device)
   local FINGERPRINTS = {
     { mfr = "LUMI", model = "lumi.light.acn004" },
--- a/drivers/SmartThings/zigbee-switch/src/lazy_load_subdriver.lua
+++ b/drivers/SmartThings/zigbee-switch/src/lazy_load_subdriver.lua
@@ -1,3 +1,6 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
 return function(sub_driver_name)
   -- gets the current lua libs api version
   local version = require "version"
--- a/drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/aqara-light/init.lua
@@ -1,3 +1,6 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
 local clusters = require "st.zigbee.zcl.clusters"
 local cluster_base = require "st.zigbee.cluster_base"
 local data_types = require "st.zigbee.data_types"
--- a/drivers/SmartThings/zigbee-switch/src/aqara/can_handle.lua
+++ b/drivers/SmartThings/zigbee-switch/src/aqara/can_handle.lua
@@ -1,3 +1,6 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
 return function(opts, driver, device)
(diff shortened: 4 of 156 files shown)


drivers/SmartThings/zigbee-switch/src/table_tracker.lua
greens, line 1: Where's this guy getting used?

cjswedes: ope, this is something sitting in my repo that I used with a different test. It isnt being used, but I will remove it.

greens:
Also you'll need to add the copyright disclaimer to all the new lua files.

cjswedes:
> Also you'll need to add the copyright disclaimer to all the new lua files.

Is this something that is definitely required for every single file for our license to apply, or is it just a practice that we have been following. I ask because adding these comments at the top of every file negatively affects memory of our driver processes. When a lua file is run it is loaded into memory and parsed into bytecode which is why comments affect the memory of a lua program. That load step brings in the comment data, which even though is not included in bytecode, it does affect the processes peak memory. We are already stripping out the comments from the lua libs we ship on target (and leaving newlines to preserve line numbers) to mitigate this somewhat for drivers. 

If we do need something in every file can we use a condensed version that is only one or 2 lines?

cjswedes:
> can we use a condensed version that is only one or 2 lines?

Maybe something like:
```
-- Copyright 2025 SmartThings. Licensed under the Apache License, Version 2.0:
-- http://www.apache.org/licenses/LICENSE-2.0
```

greens:
Good questions. 

In the meantime luacheck has beef with you.

cjswedes:
This is ready for a review now. Testing has been done on my home hub (uses about 2 subdrivers), and then unit test coverage provides good coverage; it was relied on heavily during development.

I also updated all the copyright comments in source files to match what we should be doing now as well as added it to the plethora of new files.

drivers/SmartThings/zigbee-switch/src/lazy_load_subdriver.lua
aleclorimer, line 1: Will this functionality be duplicated elsewhere? Looking like I'm going to use the same tool/logic for the other drivers (like zwave-switch) and I'm wondering if there's a more generally accessible place for this logic across drivers?

drivers/SmartThings/zigbee-switch/src/aqara-light/can_handle.lua
```diff
@@ -0,0 +1,17 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+return function(opts, driver, device)
+  local FINGERPRINTS = {
+    { mfr = "LUMI", model = "lumi.light.acn004" },
+    { mfr = "Aqara", model = "lumi.light.acn014" }
+  }
```
aleclorimer, line 8: nit: Move these into a `fingerprints.lua` file. I do agree through, if its in the function, or really small, not necessarily needed. But just thinking in terms of uniformity. nitpick of nitpicks just for discussion.

drivers/SmartThings/zigbee-switch/src/lazy_load_subdriver.lua
cjswedes: We did not include such a thing in the feature because we wanted to be explicit about having drivers specify how they are doing the lazy loading. So I think it is fine to reproduce across the drivers that we are updating to use this.

drivers/SmartThings/zigbee-switch/src/aqara-light/can_handle.lua
cjswedes: Yeah, I only made these moves if it was larger than a few entries. Its nice to have them where they are used if its not gonna save that much memory imo

greens (approved):

PR #2479, opened 2025-10-21
Matter Switch: remove 3 unused profiles

hcarter-775:
# Description of Change
Remove 3 profiles in Matter Switch that are cannot be navigated to, and that are not used at all.

# Summary of Completed Tests
Internal query ensured these profiles are used 0 times across all users.


hcarter-775:
Failing duplicate profile check is expected when removing profiles.

greens (approved):
I believe (but am not positive) that deleting profiles from a deployed driver will require direct HCS intervention.

hcarter-775:
> I believe (but am not positive) that deleting profiles from a deployed driver will require direct HCS intervention.

Talked about elsewhere, but closing the loop here- that is no longer the case, and can be done directly on our end.

gharveymn (approved):

PR #2480, opened 2025-10-21
Matter Switch: Update Vendor Overrides

hcarter-775:
# Description of Change
I updated/consolidated the vendor override logic so it is cleaner.

This could be removed, but I also added a `get_endpoints_by_device_type` helper function in this PR. It could be done separately, but I think the changed logic here is very obvious.

# Summary of Completed Tests
Since this only affects device behavior that is specifically gated by an override, unit tests properly show that all cases are still supported. 

Piggy-backing the `get_endpoints_by_dt` function addition, I have updated the `test_matter_bridge` function since the original version did not catch a bug that I later patched.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
@@ -50,8 +50,8 @@ function SwitchDeviceConfiguration.assign_child_profile(device, child_ep)
     end
   end
 
-  -- workaround: Aqara Light Switch H2: add Electrical Sensor profile only to the first EDGE_CHILD
-  if child_ep == 1 or profile == switch_utils.get_product_override_field(device, "initial_profile") then
+  -- vendor override checks
+  if child_ep == switch_utils.get_product_override_field(device, "ep_id") or profile == switch_utils.get_product_override_field(device, "initial_profile") then
     profile = switch_utils.get_product_override_field(device, "target_profile") or profile
   end
 
--- a/drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
+++ b/drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
@@ -112,13 +112,13 @@ SwitchFields.vendor_overrides = {
     [0x000D] = { target_profile = "switch-binary", initial_profile = "plug-binary" },
   },
   [0x115F] = { -- AQARA_MANUFACTURER_ID
-    [0x1003] = { target_profile = "light-power-energy-powerConsumption" },       -- 2 Buttons(Generic Switch), 1 Channel(On/Off Light)
-    [0x1004] = { target_profile = "light-power-energy-powerConsumption" },       -- 2 Buttons(Generic Switch), 2 Channels(On/Off Light)
-    [0x1005] = { target_profile = "light-power-energy-powerConsumption" },       -- 4 Buttons(Generic Switch), 3 Channels(On/Off Light)
-    [0x1008] = { target_profile = "light-power-energy-powerConsumption" },       -- 2 Buttons(Generic Switch), 1 Channel(On/Off Light)
-    [0x1009] = { target_profile = "light-power-energy-powerConsumption" },       -- 4 Buttons(Generic Switch), 2 Channels(On/Off Light)
-    [0x1006] = { ignore_combo_switch_button = true, target_profile = "light-level-power-energy-powerConsumption" }, -- 3 Buttons(Generic Switch), 1 Channels(Dimmable Light)
-    [0x100A] = { ignore_combo_switch_button = true, target_profile = "light-level-power-energy-powerConsumption" }, -- 1 Buttons(Generic Switch), 1 Channels(Dimmable Light)
+    [0x1003] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 2 Buttons(Generic Switch), 1 Channel(On/Off Light)
+    [0x1004] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 2 Buttons(Generic Switch), 2 Channels(On/Off Light)
+    [0x1005] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 4 Buttons(Generic Switch), 3 Channels(On/Off Light)
+    [0x1008] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 2 Buttons(Generic Switch), 1 Channel(On/Off Light)
+    [0x1009] = { target_profile = "light-power-energy-powerConsumption", ep_id = 1 },       -- 4 Buttons(Generic Switch), 2 Channels(On/Off Light)
+    [0x1006] = { ignore_combo_switch_button = true, target_profile = "light-level-power-energy-powerConsumption", ep_id = 1 }, -- 3 Buttons(Generic Switch), 1 Channels(Dimmable Light)
+    [0x100A] = { ignore_combo_switch_button = true, target_profile = "light-level-power-energy-powerConsumption", ep_id = 1 }, -- 1 Buttons(Generic Switch), 1 Channels(Dimmable Light)
     [0x2004] = { is_climate_sensor_w100 = true }, -- Climate Sensor W100, requires unique profile
   }
 }


drivers/SmartThings/matter-switch/src/test/test_multi_switch_parent_child_lights.lua
```diff
@@ -274,11 +274,10 @@ local function test_init_parent_child_endpoints_non_sequential()
     parent_assigned_child_key = string.format("%d", child2_ep_non_sequential)
   })
 
-  -- switch-binary will be selected as an overridden child device profile
   mock_device_parent_child_endpoints_non_sequential:expect_device_create({
     type = "EDGE_CHILD",
     label = "Matter Switch 3",
-    profile = "switch-binary",
+    profile = "plug-binary",
```
nickolas-deboom: What's the reason for this change? Edit: nvm, I understand after looking at the mock device structure

drivers/SmartThings/matter-switch/src/utils/switch_utils.lua
```diff
@@ -63,6 +63,14 @@ function utils.mired_to_kelvin(value, minOrMax)
   end
 end
 
+function utils.check_vendor_overrides(device, override_key)
+  for product_id, overrides in pairs(fields.vendor_overrides[device.manufacturer_info.vendor_id] or {}) do
+    if product_id == device.manufacturer_info.product_id and overrides[override_key] ~= nil then
+      return true
```
nickolas-deboom:
```suggestion
    if product_id == device.manufacturer_info.product_id then
      return overrides[override_key]
```
One suggestion I was thinking of would be return this so it can be checked/used at the callsite

drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
```diff
@@ -106,19 +106,16 @@ SwitchFields.HUE_SAT_COLOR_MODE = clusters.ColorControl.types.ColorMode.CURRENT_
 SwitchFields.X_Y_COLOR_MODE = clusters.ColorControl.types.ColorMode.CURRENTX_AND_CURRENTY
 
 
-SwitchFields.child_device_profile_overrides_per_vendor_id = {
-  [0x1321] = {
```
nickolas-deboom, line 110: I might be missing something but what happened to these overrides?

hcarter-775: Quoted from the PR description:
```
In response to this comment: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1872/files#r1944660330

we should remove this sonoff vendor override. In doing this, I updated/consolidated the vendor override logic so it is cleaner.
```

hcarter-775: Though re-reading the comment I quoted, I realize I misread this, oops. We do in fact still need this override

drivers/SmartThings/matter-switch/src/utils/switch_utils.lua
hcarter-775: good idea! will try this out

drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
hcarter-775: wait, sorry. I actually think this initial reading is correct, and that we should remove this from the handling. Please review your comment here: 
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/1872/files#r1945009286

drivers/SmartThings/matter-switch/src/utils/switch_utils.lua
hcarter-775: implemented

drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
```diff
@@ -106,19 +106,16 @@ SwitchFields.HUE_SAT_COLOR_MODE = clusters.ColorControl.types.ColorMode.CURRENT_
...
-  },
-  [0x115F] = {
-    { product_id = 0x1003, target_profile = "light-power-energy-powerConsumption" },       -- 2 Buttons(Generic Switch), 1 Channel(On/Off Light)
-    { product_id = 0x1004, target_profile = "light-power-energy-powerConsumption" },       -- 2 Buttons(Generic Switch), 2 Channels(On/Off Light)
-    { product_id = 0x1005, target_profile = "light-power-energy-powerConsumption" },       -- 4 Buttons(Generic Switch), 3 Channels(On/Off Light)
-    { product_id = 0x1006, target_profile = "light-level-power-energy-powerConsumption" }, -- 3 Buttons(Generic Switch), 1 Channels(Dimmable Light)
-    { product_id = 0x1008, target_profile = "light-power-energy-powerConsumption" },       -- 2 Buttons(Generic Switch), 1 Channel(On/Off Light)
-    { product_id = 0x1009, target_profile = "light-power-energy-powerConsumption" },       -- 4 Buttons(Generic Switch), 2 Channels(On/Off Light)
-    { product_id = 0x100A, target_profile = "light-level-power-energy-powerConsumption" }, -- 1 Buttons(Generic Switch), 1 Channels(Dimmable Light)
+SwitchFields.vendor_overrides = {
```
ctowns, line 109: I like the idea of having all vendor overrides in one place so it's clear exactly what devices have overrides in the main driver as opposed to using a sub driver.

That brings up a different question though - are any of these overrides worthy of a sub driver. I am guessing not since a lot of them are profile/configuration things that would require a lot of duplicated code in the sub driver. But just wanted to pose the question here to get your thoughts as you keep making improvements to the switch driver.

hcarter-775: Yes, that is a very good point. So yes, all of these overrides are for profiling like you said, which we have generally left in the main driver in the past. However, since I've moved all of the helper functions into switch_utils, it is possible we could move even small vendor-specific fixes into sub-drivers without too much work. On the other hand, I believe all the overrides we currently have might still be more trouble than they're worth, especially given that in [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2444), I will be removing more than half of them, and that'll leave us with just 3.

The climate sensor one _could_ be moved, but if you look closely at how it's actually being used, I _think_ I'd rather just improve the main driver handling as a way to get rid of it.

drivers/SmartThings/matter-switch/src/utils/switch_utils.lua
```diff
@@ -63,6 +63,14 @@ function utils.mired_to_kelvin(value, minOrMax)
   end
 end
 
+function utils.check_vendor_overrides(device, override_key)
```
gharveymn: This doesn't necessarily return a bool, so maybe prefixing this with `check_` could be misleading in the future. Also, we are keying the product inside of a vendor, so perhaps`get_product_override_field` or similar?

drivers/SmartThings/matter-switch/src/utils/switch_utils.lua
```diff
@@ -171,6 +168,19 @@ function utils.matter_handler(driver, device, response_block)
   device.log.info(string.format("Fallback handler for %s", response_block))
 end
 
+-- get a list of endpoints for a specified device type.
+function utils.get_endpoints_by_dt(device, device_type_id)
```
gharveymn: Is `dt` used very commonly as an abbreviation of "device type" in this repo? Otherwise, I would suggest just `device_type`.

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
```diff
@@ -50,20 +50,9 @@ function SwitchDeviceConfiguration.assign_child_profile(device, child_ep)
...
-  --   2. The selected profile for the child device matches the initial profile defined in
-  --      child_device_profile_overrides
-  for id, vendor in pairs(fields.child_device_profile_overrides_per_vendor_id) do
-    for _, fingerprint in ipairs(vendor) do
-      if device.manufacturer_info.product_id == fingerprint.product_id and
-         ((device.manufacturer_info.vendor_id == fields.AQARA_MANUFACTURER_ID and child_ep == 1) or profile == fingerprint.initial_profile) then
-         return fingerprint.target_profile
-      end
-    end
+  -- workaround: Aqara Light Switch H2: add Electrical Sensor profile only to the first EDGE_CHILD
```
gharveymn: Is this comment saying that this workaround requires that all future overrides like this do the same?

gharveymn (approved):
Nice job

drivers/SmartThings/matter-switch/src/utils/switch_utils.lua
hcarter-775: dt is commonly used by our matter drivers, but I kinda agree that removing the abbreviation may be preferable either way.

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
hcarter-775: yes, I suppose it is. The comment was mostly meant to clarify what the seemingly random bit of code did in the first place.

FYI, this workaround will be removed altogether in later PRs, specifically here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2444

drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
nickolas-deboom: Sorry for the late reply, I believe that was in reference to the Shin Dong switch from that PR and not to the Sonoff devices in this list

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
```diff
@@ -50,20 +50,9 @@ function SwitchDeviceConfiguration.assign_child_profile(device, child_ep)
...
+  if child_ep == 1 then
+    profile = switch_utils.get_product_override_field(device, "target_profile") or profile
```
nickolas-deboom, line 55: Shouldn't we check the override table in all cases when determining the child profile? It seems to me that this is the only use of the `target_profile` field so we could probably just have this line without the if clause.

hcarter-775: nope, this should only be done for the first child endpoint. It is a workaround for a very specific device layout for the Aqara Switch H2 device

drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
hcarter-775: oh, apologies. yeah, I reverted this change,

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
hcarter-775: However, I did update the override tables to specify the child_ep check should be directly associated with the Aqara device

nickolas-deboom: Ah yeah, I see! The changes look good to me

PR #2481, opened 2025-10-21
Add greater dynamic profiling to matter switch

hcarter-775:
# Description of Change

Right now, nearly all Light / PlugIn / Switch fingerprints are static in the Matter Switch driver. This update makes it so that more dynamic ability to re-profiling these devices. It also attempts to streamline some of this handling.

Though there are a few cases where parent/child device composition may cause a parent device to mis-profile (which this would fix- ex. a child with more support than its parent will cause the parent to misprofile with the more robust child support), this change mostly aims at making dynamic profiling more straightforward in later work.

# Summary of Completed Tests

All Light superset devices were onboarded with correct profile.
All Plug devices were onboarded with correct profile.
Specialty devices show continued correct profiling through unit tests.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/test/test_matter_switch.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_matter_switch.lua
@@ -77,6 +77,52 @@ local mock_device_no_hue_sat = test.mock_device.build_test_matter_device({
   }
 })
 
+local mock_device_color_temp = test.mock_device.build_test_matter_device({
+  profile = t_utils.get_profile_definition("light-level-colorTemperature.yml"),
+  manufacturer_info = {
+    vendor_id = 0x0000,
+    product_id = 0x0000,
+  },
+  endpoints = {
+    {
+      endpoint_id = 1,
+      clusters = {
+        {cluster_id = clusters.OnOff.ID, cluster_type = "SERVER"},
+        {cluster_id = clusters.ColorControl.ID, cluster_type = "BOTH", feature_map = 30},
+        {cluster_id = clusters.LevelControl.ID, cluster_type = "SERVER"}
+      },
+      device_types = {
+        {device_type_id = 0x0100, device_type_revision = 1}, -- On/Off Light
+        {device_type_id = 0x010C, device_type_revision = 1} -- Color Temperature Light
+      }
+    }
+  }
+})
+
+local mock_device_extended_color = test.mock_device.build_test_matter_device({
+  profile = t_utils.get_profile_definition("light-color-level.yml"),
+  manufacturer_info = {
+    vendor_id = 0x0000,
+    product_id = 0x0000,
+  },
+  endpoints = {
+    {
+      endpoint_id = 1,
+      clusters = {
+        {cluster_id = clusters.OnOff.ID, cluster_type = "SERVER"},
+        {cluster_id = clusters.ColorControl.ID, cluster_type = "BOTH", feature_map = 30},
+        {cluster_id = clusters.LevelControl.ID, cluster_type = "SERVER", feature_map = 2}
+      },
+      device_types = {
+        {device_type_id = 0x0100, device_type_revision = 1}, -- On/Off Light
+        {device_type_id = 0x0101, device_type_revision = 1}, -- Dimmable Light
+        {device_type_id = 0x010C, device_type_revision = 1}, -- Color Temperature Light
+        {device_type_id = 0x010D, device_type_revision = 1}, -- Extended Color Light
+      }
+    }
+  }
+})
+
 local cluster_subscribe_list = {
   clusters.OnOff.attributes.OnOff,
   clusters.LevelControl.attributes.CurrentLevel,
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
```diff
@@ -31,103 +31,69 @@ local DeviceConfiguration = {}
...
-          }
-        )
-        parent_child_device = true
-        if idx == 1 and string.find(child_profile, "energy") then
-          -- when energy management is defined in the root endpoint(0), replace it with the first switch endpoint and process it.
-          device:set_field(fields.ENERGY_MANAGEMENT_ENDPOINT, ep, {persist = true})
+function SwitchDeviceConfiguration.assign_profile_for_onoff_ep(device, onoff_ep_id, is_child_device)
+  local ep = switch_utils.get_endpoint_info(device, onoff_ep_id)
+  local primary_dt_id = switch_utils.find_max_subset_device_type(ep, fields.DEVICE_TYPE_ID.LIGHT)
+    or (switch_utils.detect_matter_thing(device) and switch_utils.find_max_subset_device_type(ep, fields.DEVICE_TYPE_ID.SWITCH))
```
ctowns: This could use a comment on why it is being done. I assume this is meant to handle the workaround for bindings that is explained [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/cc30d984e28f35cf972901e9981f002d72265a30/drivers/SmartThings/matter-switch/src/utils/device_configuration.lua#L218)? Otherwise this seems confusing

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
```diff
@@ -199,75 +165,60 @@ end
...
+        if #embedded_cluster_utils.get_endpoints(device, clusters.ElectricalPowerMeasurement.ID) > 0 then electrical_tags = electrical_tags .. "-power" end
+        if #embedded_cluster_utils.get_endpoints(device, clusters.ElectricalEnergyMeasurement.ID) > 0 then electrical_tags = electrical_tags .. "-energy-powerConsumption" end
+        if electrical_tags ~= "" then updated_profile = string.gsub(updated_profile, "-binary", "") .. electrical_tags end
+      end
+    elseif generic_profile("light-color-level") and #device:get_endpoints(clusters.FanControl.ID) > 0 then
+      updated_profile = "light-color-level-fan"
+    elseif generic_profile("light-level") and #device:get_endpoints(clusters.OccupancySensing.ID) > 0 then
+      updated_profile = "light-level-motion"
+    elseif generic_profile("light-level-colorTemperature") or generic_profile("light-color-level") then
+      -- ignore attempts to dynamically profile light-level-colorTemperature and light-color-level devices for now, since
```
ctowns, line 194: Do any of these devices also have button components? In that case wouldn't we _not_ want to return from this conditional branch so that the buttons could be handled below?

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
hcarter-775: updated this with a comment in the latest commit, and reworked the ordering a little to hopefully better clarify what is happening.

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
hcarter-775: that's true, in that case we would not want that. However, we do not support any profiles at the moment that would include an extended color light or a color temp light on the same profile as a button device.

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
```diff
@@ -31,92 +31,57 @@ local DeviceConfiguration = {}
...
-      profile = fields.device_type_profile_map[id]
-      break
-    end
-  end
-
-  -- vendor override checks
-  if child_ep == switch_utils.get_product_override_field(device, "ep_id") or profile == switch_utils.get_product_override_field(device, "initial_profile") then
-    profile = switch_utils.get_product_override_field(device, "target_profile") or profile
-  end
+  -- per spec, the Switch device types support OnOff as CLIENT, though some vendors break spec and support it as SERVER.
```
nickolas-deboom, line 37: I think the context from the old comment is still useful to have here

hcarter-775: what context? the stuff about the light device types? If so, that comment has been moved to the description for the `find_max_subset_device_type` function

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
```diff
@@ -192,75 +157,61 @@ end
...
+    updated_profile = SwitchDeviceConfiguration.assign_profile_for_onoff_ep(device, main_endpoint_id)
+    local generic_profile = function(s) return string.find(updated_profile or "", s, 1, true) end
+    if generic_profile("plug-binary") or generic_profile("plug-level") then
+      if switch_utils.check_switch_category_vendor_overrides(device) then
+        updated_profile = string.gsub(updated_profile, "plug", "switch")
+      else
+        local electrical_tags = ""
+        if #embedded_cluster_utils.get_endpoints(device, clusters.ElectricalPowerMeasurement.ID) > 0 then electrical_tags = electrical_tags .. "-power" end
+        if #embedded_cluster_utils.get_endpoints(device, clusters.ElectricalEnergyMeasurement.ID) > 0 then electrical_tags = electrical_tags .. "-energy-powerConsumption" end
+        if electrical_tags ~= "" then updated_profile = string.gsub(updated_profile, "-binary", "") .. electrical_tags end
```
nickolas-deboom, line 187: Is `updated_profile` guaranteed to be a valid profile name at this point?

drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
```diff
@@ -123,6 +127,31 @@ SwitchFields.vendor_overrides = {
   }
 }
 
+SwitchFields.switch_category_vendor_overrides = {
```
nickolas-deboom, line 130: Where is this list coming from?

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
nickolas-deboom: Right, I see it now, thanks

drivers/SmartThings/matter-switch/src/utils/switch_utils.lua
```diff
@@ -82,17 +83,44 @@ function utils.check_field_name_updates(device)
...
-  local dimmable_eps = utils.get_endpoints_by_device_type(device, fields.DIMMABLE_LIGHT_DEVICE_TYPE_ID)
+  local dimmable_eps = utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.LIGHT.DIMMABLE)
   return utils.tbl_contains(dimmable_eps, endpoint_id)
 end
 
+-- Some devices report multiple device types which are a subset of
+-- a superset device type (Ex. Dimmable Light is a superset of On/Off Light).
+-- We should map to the largest superset device type supported.
+-- This can be done by matching to the device type with the highest ID
+function utils.find_max_subset_device_type(ep, device_type_set)
```
nickolas-deboom, line 109: Has the functionality of this function changed? I was comparing them and their usages but I wasn't able to tell for sure

nickolas-deboom:
Have you tested with plugs that support energy clusters?

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
hcarter-775: yes it is- see that this exists in the context that generic_profile() has evaluated to true for a profile

drivers/SmartThings/matter-switch/src/utils/switch_fields.lua
hcarter-775: these are all the fingerprints that use switch

drivers/SmartThings/matter-switch/src/utils/switch_utils.lua
hcarter-775: it has not, the logic has just become more stringent to keep the max within the subset

hcarter-775:
> Have you tested with plugs that support energy clusters?

Unit tests catch the case that anything has changed in that regard, and no functionality would be changed as far as this path goes. That is to say, if a plug correctly profiles then the electrical sensor portion would not be breaking (and thus can be confirmed in unit tests)

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
nickolas-deboom: Nice!

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
```diff
@@ -192,75 +157,61 @@ end
...
+      end
+    elseif generic_profile("light-color-level") and #device:get_endpoints(clusters.FanControl.ID) > 0 then
+      updated_profile = "light-color-level-fan"
+    elseif generic_profile("light-level") and #device:get_endpoints(clusters.OccupancySensing.ID) > 0 then
+      updated_profile = "light-level-motion"
+    elseif generic_profile("light-level-colorTemperature") or generic_profile("light-color-level") then
```
nickolas-deboom, line 193: Could you add a test for this case?

drivers/SmartThings/matter-switch/src/utils/switch_utils.lua
nickolas-deboom: I see now, thanks!

nickolas-deboom:
I took another pass through this PR and it's looking good from my perspective :+1:  I'm approving now but I had one more comment about a unit test that would be good to add before merging

drivers/SmartThings/matter-switch/src/utils/device_configuration.lua
hcarter-775: added one for both device types!

nickolas-deboom: Thanks!

PR #2493, opened 2025-10-23
Fix some cases where we could crash trying to iterate over `nil`

dljsjr:
Check all that apply

# Type of Change

- [x] Bug fix

# Checklist

# Description of Change

The recent refactor to reduce memory usage also performed a normalization of variable casing; all values that are "our" data structures use `snake_case` names, and all values that are "their" data structures (i.e. tables that are created directly from JSON or XML bodies) use `camelCase`.

Since we're no longer storing the REST response bodies as-is, but just a subset of them, there are a few places where variables that used to be `camelCase` are now `snakeCase`; we addressed all of these direct accesses, but we missed a few places where we created fallback values for iteratees that might be `nil`.

While fixing the above, we also add empty table fallbacks to almost all iteration sites to avoid this writ-large.

# Summary of Completed Tests

Tested on my personal setup, however, I believe that the crash this fixes is only reproducible on S1 bonded sets/stereo pairs, which I don't have the personal inventory to test (my only bonded-set eligible speakers are also S2 only speakers).

### DIFF
--- a/drivers/SmartThings/sonos/src/api/sonos_connection.lua
+++ b/drivers/SmartThings/sonos/src/api/sonos_connection.lua
@@ -259,7 +259,9 @@ local function _oauth_reconnect_task(sonos_conn)
 
       if unauthorized then
         sonos_conn.driver:alert_unauthorized()
-        local token, channel_error = token_receive_handle:receive()
+        local token, channel_error =
+          (token_receive_handle and token_receive_handle:receive()) or nil,
+          "no token receive handle"
         if not token then
           log.warn(string.format("Error requesting token: %s", channel_error))
           local _, get_token_err = sonos_conn.driver:get_oauth_token()


drivers/SmartThings/sonos/src/api/sonos_connection.lua
```diff
@@ -405,8 +405,8 @@ function SonosConnection.new(driver, device)
             )
             return
           end
-          local group = household.groups[header.groupId] or { playerIds = {} }
-          for _, player_id in ipairs(group.player_ids) do
+          local group = household.groups[header.groupId] or { player_ids = {} }
+          for _, player_id in ipairs(group.player_ids or {}) do
```
dljsjr: The empty table fallback is redundant here I suppose…

drivers/SmartThings/sonos/src/ssdp.lua
```diff
@@ -197,7 +197,7 @@ end
 --- the search end time will be pushed out based on the `mx` parameter, extending the amount of
 --- time that `receive_m_search_response` will return results with a timeout.
 function _ssdp_mt:multicast_m_search()
-  for term, _ in pairs(self.search_terms) do
+  for term, _ in pairs(self.search_terms or {}) do
```
cjswedes, line 200: Seems like we may want to crash here if this is nil, since it would mean potentially not discovering devices when we should be.

dljsjr:
> Seems like all pairs/ipairs were updated with protections against nil input. Im fine with this, but if were doing it for fields we know should never be nil when accessed, it could potentially be hiding future errors that we would only discover with user feedback rather than driver crashes.

Yeah, I went back and forth on this a bit. Here's my take on it:

We've learned that crashes in LAN drivers like this can lead to unintended hot crash loops. While we have a variety of mitigations on place on the hub to limit the impact of these situations, they often still end up causing undue burden on the hub's resources. And we usually end up getting bug reports for this stuff anyway.

That said there's definitely still value in the early canary of the driver crash metrics, so I could see undoing this in some places being valuable.

hcarter-775 (approved):

PR #2494, opened 2025-10-23
Matter Lock: Use Door Lock Device type in Lock Modular logic

hcarter-775:
# Description of Change
Slightly reduces number of checks in the lock modular logic by using the device type as a first check, before searching for the door lock cluster to map a profile to.

# Summary of Completed Tests
Unit tests continue to pass, and VDA onboarding indicated a correct modular profile was selected.

### DIFF
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -217,6 +217,7 @@ local function match_profile_modular(driver, device)
   for _, ep_data in ipairs(device.endpoints) do
     if ep_data.device_types[1].device_type_id == DEVICE_TYPE_ID.DOOR_LOCK then
       door_lock_ep_data = ep_data
+      break
     end
   end
   for _, cluster in pairs(door_lock_ep_data.clusters) do


drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -209,35 +213,39 @@ local function match_profile_modular(driver, device)
...
-          device:emit_event(capabilities.lock.supportedLockValues({"locked", "unlocked", "not fully locked"}, {visibility = {displayed = false}}))
-          device:emit_event(capabilities.lock.supportedLockCommands({"lock", "unlock"}, {visibility = {displayed = false}}))
-        end
-        if clus_has_feature(DoorLock.types.Feature.ALIRO_PROVISIONING) then
-          table.insert(main_component_capabilities, capabilities.lockAliro.ID)
-        end
-        break
+  local door_lock_ep_data
+  for _, ep_data in ipairs(device.endpoints) do
+    if ep_data.device_types[1].device_type_id == DEVICE_TYPE_ID.DOOR_LOCK then
```
gharveymn, line 218: Could the endpoint have multiple device types?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -209,35 +213,39 @@ local function match_profile_modular(driver, device)
...
-        else
-          device:emit_event(capabilities.lock.supportedLockValues({"locked", "unlocked", "not fully locked"}, {visibility = {displayed = false}}))
-          device:emit_event(capabilities.lock.supportedLockCommands({"lock", "unlock"}, {visibility = {displayed = false}}))
-        end
-        if clus_has_feature(DoorLock.types.Feature.ALIRO_PROVISIONING) then
-          table.insert(main_component_capabilities, capabilities.lockAliro.ID)
-        end
-        break
+  local door_lock_ep_data
+  for _, ep_data in ipairs(device.endpoints) do
```
gharveymn, line 217: This got changed from `pairs` to `ipairs`. Intentional?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -209,35 +213,39 @@ local function match_profile_modular(driver, device)
...
+    if ep_data.device_types[1].device_type_id == DEVICE_TYPE_ID.DOOR_LOCK then
+      door_lock_ep_data = ep_data
+    end
+  end
```
gharveymn, line 222: What happens if there are multiple endpoints with the DoorLock device type? We only care about the last one?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: per spec, application device types cannot have >1 device type on an endpoint

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: yes. Not super important either way but device.endpoints is a numeric list

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: This was already how the handling supported devices, that is, with the assumption there would only be one DoorLock device type endpoint per device. At the moment, if there were >1, we would not have the infrastructure to support it anyway, so creating better handling for that would be better suited for a separate PR.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: Also, application device types cannot be supported on another endpoint as a secondary.

gharveymn: Discussed offline. Here's the spec wording:

> 9.2.1. Endpoint Requirements
[9.18] Each [Simple endpoint](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2494#ref_SimpleEndpoint) SHALL support exactly one [Application device type](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2494#ref_ApplicationEndpoint) with the exception that, the endpoint MAY support additional device types which are subsets of the [Application device](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2494#ref_ApplicationEndpoint) type (the superset). See [Section 9.2.10, “Superset Device Types”](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2494#ref_SupersetDeviceTypes) for cluster requirements of superset devices.

where the complement of simple endpoints are dynamic endpoints, which aren't used by us (now? ever?).

hcarter-775:
insignificant change, closing for now

PR #2496, opened 2025-10-23
Matter Switch: Fix Fingerprint

hcarter-775:
# Description of Change

The switch-4 profile is not currently supported by the Matter Switch driver, and so any devices being onboarded with this profile will not work as expected. The current expected handling for this device should be 4-button, since it supports 4 generic switches. 

# Summary of Completed Tests


gharveymn:
Harrison and I talked offline and he will be doing a separate PR to add comments to the numeric profiles to indicate that they are deprecated and not to be used by new fingerprinting.

nickolas-deboom (approved):

PR #2513, opened 2025-10-31
Matter Smoke Alarm: add smoke-temp-humidity-battery profile

hcarter-775:
# Description of Change
Add a new smoke profile.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-sensor/fingerprints.yml
+++ b/drivers/SmartThings/matter-sensor/fingerprints.yml
@@ -277,6 +277,13 @@ matterGeneric:
     deviceTypes:
       - id: 0x0076 # Smoke CO Alarm
     deviceProfileName: smoke-co
+  - id: "matter/smoke-temp-humidity/sensor"
+    deviceLabel: Matter Smoke/Temp/Humidity Sensor
+    deviceTypes:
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-sensor/profiles/smoke-temp-humidity-battery.yml
```diff
@@ -0,0 +1,27 @@
+name: smoke-temp-humidity-battery
```
nickolas-deboom, line 1: Do we want a corresponding `batteryLevel` profile?

nickolas-deboom (approved):

drivers/SmartThings/matter-sensor/profiles/smoke-temp-humidity-battery.yml
hcarter-775: what I'd prefer to do is add modular handling for this device type I think. I don't think the lift on that would be extremely high

nickolas-deboom: Yeah I agree that'd be nice

PR #2514, opened 2025-10-31
Matter Switch: Add switch-fan profile

hcarter-775:
# Description of Change
Add new profile. Generically choose it always whether the device is a plug or a light binary device.

# Summary of Completed Tests
Added a test case to show correct update logic.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/test/test_matter_light_fan.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_matter_light_fan.lua
@@ -112,8 +112,6 @@ local CLUSTER_SUBSCRIBE_LIST ={
 
 local SWITCH_CLUSTER_SUBSCRIBE_LIST ={
   clusters.OnOff.attributes.OnOff,
-  clusters.FanControl.attributes.FanModeSequence,
-  clusters.FanControl.attributes.FanMode,
   clusters.FanControl.attributes.PercentCurrent,
 }
 


drivers/SmartThings/matter-switch/profiles/switch-fan.yml
```diff
@@ -0,0 +1,14 @@
...
+  - id: switch
+    version: 1
+  - id: fanSpeedPercent
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Switch
```
ctowns, line 14: Why are we using the `Switch` if this could also be a light? Just want to make sure we don't get pinged about this later. And why not use `Fan` instead?

ctowns (approved):
Looks good, let's have a discussion offline about the category selection, but that can also be changed later if needed.

nickolas-deboom (approved):

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -190,7 +190,9 @@ function DeviceConfiguration.match_profile(driver, device)
   if switch_utils.tbl_contains(server_onoff_ep_ids, default_endpoint_id) then
     updated_profile = SwitchDeviceConfiguration.assign_profile_for_onoff_ep(device, default_endpoint_id)
     local generic_profile = function(s) return string.find(updated_profile or "", s, 1, true) end
-    if generic_profile("light-color-level") and #device:get_endpoints(clusters.FanControl.ID) > 0 then
+    if (generic_profile("plug-binary") or generic_profile("light-binary")) and #device:get_endpoints(clusters.FanControl.ID) > 0 then
```
nickolas-deboom, line 193: Similar to Cooper's comment, should we split this up and have a separate `light-fan` profile with the `Light` category?

hcarter-775:
generically using the following PR instead: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2653

PR #2525, opened 2025-11-04
Matter Switch: Lazy load subdrivers if possible

nickolas-deboom:
This change utilizes the new `lazy_load_sub_driver_v2` api to lazy load subdrivers in the matter switch driver.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -297,11 +297,11 @@ local matter_driver_template = {
   },
   supported_capabilities = fields.supported_capabilities,
   sub_drivers = {
-    require("sub_drivers.aqara_cube"),
+    switch_utils.lazy_load_if_possible("sub_drivers.aqara_cube"),
     switch_utils.lazy_load("sub_drivers.camera"),
-    require("sub_drivers.eve_energy"),
-    require("sub_drivers.ikea_scroll"),
-    require("sub_drivers.third_reality_mk1")
+    switch_utils.lazy_load_if_possible("sub_drivers.eve_energy"),
+    switch_utils.lazy_load_if_possible("sub_drivers.ikea_scroll"),
+    switch_utils.lazy_load_if_possible("sub_drivers.third_reality_mk1")
   }
 }
 
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/aqara_cube/can_handle.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/aqara_cube/can_handle.lua
@@ -1,0 +1,14 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local device_lib = require "st.device"
+
+return function(opts, driver, device)
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+    local name = string.format("%s", device.manufacturer_info.product_name)
+    if string.find(name, "Aqara Cube T1 Pro") then
+      return true, require("sub_drivers.aqara_cube")
+    end
+  end
+  return false
+end
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
@@ -1,0 +1,18 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local device_lib = require "st.device"
+local fields = require "switch_utils.fields"
+local switch_utils = require "switch_utils.utils"
+
+return function(opts, driver, device)
+  local EVE_MANUFACTURER_ID = 0x130A
+  -- this sub driver does NOT support child devices, and ONLY supports Eve devices
+  -- that do NOT support the Electrical Sensor device type
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
+    device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID and
+    #switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.ELECTRICAL_SENSOR) == 0 then
+    return true, require("sub_drivers.eve_energy")
+  end
+  return false
+end
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/third_reality_mk1/can_handle.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/third_reality_mk1/can_handle.lua
@@ -1,0 +1,14 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local device_lib = require "st.device"
+
+return function(opts, driver, device)
+  local THIRD_REALITY_MK1_FINGERPRINT = { vendor_id = 0x1407, product_id = 0x1388 }
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
+    device.manufacturer_info.vendor_id == THIRD_REALITY_MK1_FINGERPRINT.vendor_id and
+    device.manufacturer_info.product_id == THIRD_REALITY_MK1_FINGERPRINT.product_id then
+    return true, require("sub_drivers.third_reality_mk1")
+  end
+  return false
+end
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/aqara_cube/init.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/aqara_cube/init.lua
@@ -21,16 +21,6 @@ local INITIAL_PRESS_ONLY = "__initial_press_only" -- for devices that support MS
 local CUBEACTION_TIMER = "__cubeAction_timer"
 local CUBEACTION_TIME = 3
 
-local function is_aqara_cube(opts, driver, device)
-  if device.network_type == device_lib.NETWORK_TYPE_MATTER then
-    local name = string.format("%s", device.manufacturer_info.product_name)
-    if string.find(name, "Aqara Cube T1 Pro") then
-      return true
-    end
-  end
-  return false
-end
-
 local callback_timer = function(device)
   return function()
     device:emit_event(cubeAction.cubeAction("noAction"))
@@ -240,7 +230,7 @@ local aqara_cube_handler = {
       }
     },
   },
-  can_handle = is_aqara_cube
+  can_handle = require("sub_drivers.aqara_cube.can_handle")
 }
 
 return aqara_cube_handler
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/init.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/init.lua
@@ -11,14 +11,11 @@ local cluster_base = require "st.matter.cluster_base"
 local st_utils = require "st.utils"
 local data_types = require "st.matter.data_types"
 local device_lib = require "st.device"
-local switch_utils = require "switch_utils.utils"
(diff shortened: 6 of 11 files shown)


drivers/SmartThings/matter-switch/src/utils/lazy_load_subdriver.lua
```diff
@@ -0,0 +1,14 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+return function(sub_driver_name)
+  local version = require "version"
+  local MatterDriver = require "st.matter.driver"
```
hcarter-775: I think this should probably just go in `switch_utils`? I guess I don't see why it would have its own file necessarily, relative to the other functions we have.

I also am not sure if we should load MatterDriver in the function itself, I think our garbage collection is sturdy enough where we can just load it at the top of the file and the garbage collection will properly optimize this for us- especially since it will already be loaded by default in the driver

nickolas-deboom: I had it in a separate file to match the zigbee implementation, but yeah I think I'll just move it to switch_utils

nickolas-deboom: Updated in latest commit.

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
```diff
@@ -0,0 +1,11 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+return function(opts, driver, device )
+  local EVE_MANUFACTURER_ID = 0x130A
+  if device.network_type == require("st.device").NETWORK_TYPE_MATTER and
```
hcarter-775: similar to my other comment, I generally believe the require's should be outside the functions in the case it is a static require like this, where we know what we're requiring ahead of time. same goes for the other `can_handle` functions

nickolas-deboom: I believe that limiting the scope of `require`'s like this was brought up as part of edge drivers api 2.0, to avoid holding onto references unless needed as well as avoiding pulling in files at the top level when possible.

hcarter-775: Hmmm, that makes sense I think. The thing is though, this `st.device` is already being pulled in at the top level in `init.lua`, so this is kinda just an extra operation, right?

gharveymn: We probably need to better structure some stuff to make it more clear what is actually loaded at a given time, but in this case, `st.device` will always already be loaded in, so it's fine if you do a static require. Alternatively, maybe you can access it through the `device` argument of the function?

drivers/SmartThings/matter-switch/src/sub_drivers/third_reality_mk1/can_handle.lua
```diff
@@ -0,0 +1,12 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+return function(opts, driver, device)
+  local THIRD_REALITY_MK1_FINGERPRINT = { vendor_id = 0x1407, product_id = 0x1388 }
+  if device.network_type == require("st.device").NETWORK_TYPE_MATTER and
```
hcarter-775: Separately but related, I was thinking we could pull this device-specific sub-driver stuff into the `SwitchFields.vendor_overrides` table? Then here we could call something like `if switch_utils.get_product_override_field(device, "is_third_reality_mk1") then ... end` or something like that?

We could do that for all these `can_handle` functions, since this may be good for moving all this "same" data into one central place

hcarter-775: It's technically a separate idea, but it's related enough where I may suggest implementing it here?

drivers/SmartThings/matter-switch/src/sub_drivers/aqara_cube/can_handle.lua
```diff
@@ -0,0 +1,12 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+return function(opts, driver, device )
```
hcarter-775:
```suggestion
return function(opts, driver, device)
```

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
```diff
@@ -0,0 +1,11 @@
+-- Copyright 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+return function(opts, driver, device )
```
hcarter-775:
```suggestion
return function(opts, driver, device)
```

drivers/SmartThings/matter-switch/src/sub_drivers/third_reality_mk1/can_handle.lua
nickolas-deboom: That would definitely be nice, but I might favor doing that in a separate PR because currently, only the 3rd reality can_handle uses an exact vid and pid while the other two have slightly different checks, and I'm not sure if I want to change those here.

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
nickolas-deboom: This is just comparing the network type retrieved from the `device` argument against a string value from the lua libs - so I could replace `require("st.device").NETWORK_TYPE_MATTER` with the string, but not sure if that's worth it considering this module is already loaded at this point as Harrison pointed out.

drivers/SmartThings/matter-switch/src/sub_drivers/aqara_cube/can_handle.lua
nickolas-deboom: fix in latest commit

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
nickolas-deboom: fixed in latest commit

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
cjswedes: IMO the `require("module").some_field` is not the easiest to read and would prefer a local variable for the require; whether that is local to the module or function isnt a huge deal, but it seems like we are favoring making it local to the function at this point.

cjswedes (approved):
Can you update the copywrite on all the files? Its nice to get that done when there are actual code changes for a driver in a release cycle. You can do in a follow up PR too, if you get it in this release cycle.

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
hcarter-775: I am of the opinion we should do the `requires` at the module level for now, if only for general consistency across the Matter drivers. To me, making things "the same" creates a more cohesive environment with more clearly set expectations.

hcarter-775:
Also, with the PR I just merged, the switch_utils file now has a different name and is in a different directory (thus a conflict). Should be a simple fix, but I'm not sure how nicely `git rebase` will understand this change. Maybe well

cjswedes:
Please make sure tests are run against ~~59~~ 58 lua libs locally before merging to exercise these changes with `lazy_load_sub_driver_v2`. edit: to 58 since now 59 is what is being used as the latest release, but we need to check tests on both.

aleclorimer:
What is the plan/status of this?

nickolas-deboom:
> What is the plan/status of this?

This will most likely land this week.

nickolas-deboom:
> Can you update the copywrite on all the files? Its nice to get that done when there are actual code changes for a driver in a release cycle. You can do in a follow up PR too, if you get it in this release cycle.

Sure thing. Should the dates be updated to 2025 or left as they are? I've heard conflicting direction on that so not too sure

hcarter-775:
@nickolas-deboom I'd say leave the year as-is. However in this case I think all the files are already 2025

aleclorimer:
> > Can you update the copywrite on all the files? Its nice to get that done when there are actual code changes for a driver in a release cycle. You can do in a follow up PR too, if you get it in this release cycle.
> 
> Sure thing. Should the dates be updated to 2025 or left as they are? I've heard conflicting direction on that so not too sure

What I've been told and have been doing is updating everything to the newer shorter copyright notice. I think we should go for that

drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
nickolas-deboom: updated as suggested.

nickolas-deboom:
> Please make sure tests are run against ~59~ 58 lua libs locally before merging to exercise these changes with `lazy_load_sub_driver_v2`. edit: to 58 since now 59 is what is being used as the latest release, but we need to check tests on both.

I checked on 58 and everything passes except for the camera tests, since they use clusters not available below 59.

drivers/SmartThings/matter-switch/src/sub_drivers/third_reality_mk1/can_handle.lua
```diff
@@ -0,0 +1,14 @@
...
+-- Licensed under the Apache License, Version 2.0
+
+local device_lib = require "st.device"
+
+return function(opts, driver, device)
+  local THIRD_REALITY_MK1_FINGERPRINT = { vendor_id = 0x1407, product_id = 0x1388 }
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and
+    device.manufacturer_info.vendor_id == THIRD_REALITY_MK1_FINGERPRINT.vendor_id and
+    device.manufacturer_info.product_id == THIRD_REALITY_MK1_FINGERPRINT.product_id then
+    return true, require("sub_drivers.third_reality_mk1")
```
gharveymn, line 11: Not sure if it really matters, but you lost a log here.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -297,10 +297,10 @@ local matter_driver_template = {
   },
   supported_capabilities = fields.supported_capabilities,
   sub_drivers = {
-    require("sub_drivers.aqara_cube"),
+    switch_utils.lazy_load_if_possible("sub_drivers.aqara_cube"),
     switch_utils.lazy_load("sub_drivers.camera"),
```
gharveymn, line 301: This one can't use `lazy_load_if_possible`?

gharveymn (approved):

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: It can't, because it's not supported at all for lower lua libs versions, rather only supported for 16 and above where `lazy_load_sub_driver_v2` is available

drivers/SmartThings/matter-switch/src/sub_drivers/third_reality_mk1/can_handle.lua
nickolas-deboom: Yeah that was intentional- to match the other subdrivers which don't emit a log

nickolas-deboom:
FYI I rebased following the changes from #2579, and so in the latest force push I added lazy loading for the bilresa subdriver.

PR #2534, opened 2025-11-05
Matter Switch: default Eve devices to main driver

hcarter-775:
# Description of Change
In the case an Eve device supports the Matter 1.3 Electrical Sensor device type, we should use the main driver handling of the cluster instead of the unique sub-driver handling.

To note, the profile being used by default for energy eve devices is `power-energy-powerConsumption`, but the new profile (using the main driver) will be `plug-power-energy-powerConsumption`. These are identical but are differently named. This should not affect anything relevant to the driver, and even in the case an Eve device is re-profiled for whatever reason in the future, this should not break any existing rules as the profile composition remains the same.

# Summary of Completed Tests
Unit test added to confirm sub-driver is not used in the case that the device supports the Electrical Sensor device type.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
@@ -294,13 +294,6 @@ end
 
 function AttributeHandlers.energy_imported_factory(is_cumulative_report)
   return function(driver, device, ib, response)
-    -- workaround: ignore devices supporting Eve's private energy cluster AND the ElectricalEnergyMeasurement cluster
-    local EVE_MANUFACTURER_ID, EVE_PRIVATE_CLUSTER_ID = 0x130A, 0x130AFC01
-    local eve_private_energy_eps = device:get_endpoints(EVE_PRIVATE_CLUSTER_ID)
-    if device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID and #eve_private_energy_eps > 0 then
-      return
-    end
-
     if is_cumulative_report then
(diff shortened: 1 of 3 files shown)


greens (approved):

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -294,13 +294,6 @@ end
 
 function AttributeHandlers.energy_imported_factory(is_cumulative_report)
   return function(driver, device, ib, response)
-    -- workaround: ignore devices supporting Eve's private energy cluster AND the ElectricalEnergyMeasurement cluster
-    local EVE_MANUFACTURER_ID, EVE_PRIVATE_CLUSTER_ID = 0x130A, 0x130AFC01
-    local eve_private_energy_eps = device:get_endpoints(EVE_PRIVATE_CLUSTER_ID)
-    if device.manufacturer_info.vendor_id == EVE_MANUFACTURER_ID and #eve_private_energy_eps > 0 then
-      return
-    end
-
```
nickolas-deboom, line 303: It seems to me that devices supporting both the matter clusters and the manufacturer-specific clusters will now default to use the matter handlers whereas they previously would have used the manufacturer-specific clusters. Is that the case?

hcarter-775: yep, that's the idea!

nickolas-deboom: Ah gotcha. In that case this looks good to me!

PR #2535, opened 2025-11-06
Matter Switch: Write EXECUTE_IF_OFF bit to Control Cluster Options attribute

hcarter-775:
# Description of Change
During device configuration (and only during commissioning per spec), we can write the EXECUTE_IF_OFF bit to the Options attribute of the Level Control and Color Control clusters (that are supported as SERVER) to indicate to all devices that level controlability should be allowed even when the device is off.

This is the norm for most light devices, so for consistency across all user experiences on the ST platform, we should attempt to make all devices functional like this.

# Summary of Completed Tests
Unit tests updated. Tested on previously failing devices. Regression tested on VDA- makes no difference in device functionality.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
@@ -193,13 +193,13 @@ function utils.ep_supports_cluster(ep_info, cluster_id, opts)
     return (cluster.feature_map & checked_feature) == checked_feature
   end
   for _, cluster in ipairs(ep_info.clusters) do
-      if ((cluster.cluster_id == cluster_id)
-            and (opts.feature_bitmap == nil or clus_has_features(cluster, opts.feature_bitmap))
-            and ((opts.cluster_type == nil and cluster.cluster_type == "SERVER" or cluster.cluster_type == "BOTH")
-              or (opts.cluster_type == cluster.cluster_type))
-            or (cluster_id == nil)) then
-              return true
-      end
+    if ((cluster.cluster_id == cluster_id)
+      and (opts.feature_bitmap == nil or clus_has_features(cluster, opts.feature_bitmap))
+      and ((opts.cluster_type == nil and cluster.cluster_type == "SERVER" or cluster.cluster_type == "BOTH")
+      or (opts.cluster_type == cluster.cluster_type))
+      or (cluster_id == nil)) then
+        return true
+    end
   end
 end
 


mocelet:
Out of curioisity, is there an scenario where it is needed? The driver already sets the Options Mask to override the ExecuteIfOff bit and set it to 1 when setting a colour or temperature and for setting the brightness level it uses the MoveToLevelWithOnOff command and ExecuteIfOff does not apply to WithOnOff commands.

ldeora:
> Out of curioisity, is there an scenario where it is needed? The driver already sets the Options Mask to override the ExecuteIfOff bit and set it to 1 when setting a colour or temperature and for setting the brightness level it uses the MoveToLevelWithOnOff command and ExecuteIfOff does not apply to WithOnOff commands.

Multi-Admin?

My interpretation is, that this ensures consistent behavior in multi-admin Matter setups, where the device may be controlled by more than one controller. Not all controllers reliably include OptionsMask = 1 and OptionsOverride = 1 in their command payloads for the Color Control or Level Control clusters. If the server’s Options attribute has the bit set to 0, those external commands may fail when the device is off. This can lead to inconsistent user experiences across platforms - for example, attempting to adjust hue or brightness from a third-party app may not take effect unless the device is manually turned on first. Setting the bit to 1 during commissioning ensures that commands execute by default across all clients, without depending on overrides in each command.

Additionally, this helps future-proof the device against driver updates or new command patterns where overrides may not be applied, as well as scenarios that involve reading or adjusting level/color states while the device is off. If your setup uses only a single controller (SmartThings only), the driver’s overrides already handle the behavior, so the attribute change may be redundant, but it remains a low-cost improvement for broader compatibility.

drivers/SmartThings/matter-switch/src/test/test_matter_switch_device_types.lua
```diff
@@ -472,6 +480,14 @@ local function test_init_color_dimmer()
...
   test.socket.device_lifecycle:__queue_receive({ mock_device_color_dimmer.id, "init" })
   test.socket.device_lifecycle:__queue_receive({ mock_device_color_dimmer.id, "doConfigure" })
+  -- test.socket.matter:__expect_send({
+  --   mock_device_color_dimmer.id,
+  --   clusters.LevelControl.attributes.Options:write(mock_device_color_dimmer, 7, clusters.LevelControl.types.OptionsBitmap.EXECUTE_IF_OFF)
+  -- })
+  -- test.socket.matter:__expect_send({
+  --   mock_device_color_dimmer.id,
+  --   clusters.ColorControl.attributes.Options:write(mock_device_color_dimmer, 7, clusters.ColorControl.types.OptionsBitmap.EXECUTE_IF_OFF)
+  -- })
```
cjswedes: Seeing this commented out makes me wonder if these expectations are working being set in the test init function. Is this device not supposed to set these options bits?

hcarter-775: whoops, gotta remove those. I had those lines originally but then made the code more stringent, to only do the write on devices supporting these clusters as server rather than client, so this test should (correctly) not have these writes occur.

cjswedes (approved):

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -187,6 +187,22 @@ function utils.get_endpoint_info(device, endpoint_id)
   return {}
 end
 
+function utils.ep_supports_cluster(ep_info, cluster_id, opts)
+  opts = opts or {}
+  local clus_has_features = function(cluster, checked_feature)
+    return (cluster.feature_map & checked_feature) == checked_feature
+  end
+  for _, cluster in ipairs(ep_info.clusters) do
+      if ((cluster.cluster_id == cluster_id)
```
greens: spacing

greens (approved):

nickolas-deboom (approved):

ctowns (approved):

PR #2544, opened 2025-11-10
Matter Sensor->Switch: Move Bosch Contact Button support to Matter Switch

hcarter-775:
# Description of Change
A commit to the Matter Sensor driver a few months ago introduced a sub-driver with button functionality- we generally elect to add sensor logic into Matter Switch rather than the other way around, to promote greater structural clarity and more centralized development.

Therefore, this commit attempts to rectify this organizational error and move the introduced logic into Matter Switch and away from Matter Sensor.

This should _not_ affect currently onboarded Contact Button devices in any way, though their support will be limited to what was introduced at the time of the WWST certification, and future changes will be supported only for devices attached to the Matter Switch driver.

# Summary of Completed Tests
Unit tests were added in Matter Switch to ensure proper support continued as expected.


hcarter-775:
Closing, moving in a different direction here

PR #2549, opened 2025-11-11
Add fanSpeedPercent to Thermostat device types

nickolas-deboom:
This PR adds the `fanSpeedPercent` capability for thermostats. These changes were originally in #2425 but were extracted to a separate PR in order to support a faster turnaround, since by itself this is a very low risk change.

Note that a range of [1, 100] is enforced because the Off mode is not supported for thermostats.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/thermostat_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-thermostat/src/thermostat_handlers/attribute_handlers.lua
@@ -367,22 +367,20 @@ function AttributeHandlers.fan_mode_sequence_handler(driver, device, ib, respons
 
   -- remove 'off' as a supported fan mode for thermostat device types, unless the
   -- device previously had 'off' as a supported fan mode to avoid breaking routines
-  if thermostat_utils.get_device_type(device) == fields.THERMOSTAT_DEVICE_TYPE_ID then
+  if thermostat_utils.get_device_type(device) == fields.THERMOSTAT_DEVICE_TYPE_ID and
+    device:supports_capability_by_id(capabilities.fanMode.ID) then
     local prev_supported_fan_modes = device:get_latest_state(
       device:endpoint_to_component(ib.endpoint_id),
       supported_fan_modes_capability.ID,
       supported_fan_modes_attribute.NAME
     ) or {}
-    if not thermostat_utils.tbl_contains(prev_supported_fan_modes, "off") then
-      local off_mode_present, off_mode_idx = thermostat_utils.tbl_contains(supported_fan_modes, "off")
-      if off_mode_present then
-        table.remove(supported_fan_modes, off_mode_idx)
-      end
+    -- per the definitions set above, the first index always contains "off"
+    if prev_supported_fan_modes[1] ~= "off" then
+      table.remove(supported_fan_modes, 1)
     end
   end
 
-  local event = supported_fan_modes_attribute(supported_fan_modes, {visibility = {displayed = false}})
-  device:emit_event_for_endpoint(ib.endpoint_id, event)
+  device:emit_event_for_endpoint(ib.endpoint_id, supported_fan_modes_attribute(supported_fan_modes, {visibility = {displayed = false}}))
 end
 
 function AttributeHandlers.percent_current_handler(driver, device, ib, response)
--- a/drivers/SmartThings/matter-thermostat/src/thermostat_utils/utils.lua
+++ b/drivers/SmartThings/matter-thermostat/src/thermostat_utils/utils.lua
@@ -10,9 +10,9 @@ local ThermostatUtils = {}
 
 function ThermostatUtils.tbl_contains(array, value)
   if value == nil then return false end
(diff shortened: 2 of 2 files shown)


hcarter-775:
Since we're adding the [1,100] gating here for the fan, let's merge the remove "off" for thermostats first. Then everything will be cohesive.

nickolas-deboom:
> Since we're adding the [1,100] gating here for the fan, let's merge the remove "off" for thermostats first. Then everything will be cohesive.

I agree, that PR is here for reference: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2428

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1538,39 +1542,59 @@ local function fan_mode_handler(driver, device, ib, response)
...
     end
   else
-    supported_fan_modes_attribute = capabilities.fanMode.supportedFanModes
+    supported_fan_modes_capability = capabilities.fanMode
+    supported_fan_modes_attribute = supported_fan_modes_capability.supportedFanModes
+  end
+
+  -- remove 'off' as a supported fan mode for thermostat device types, unless the
+  -- device previously had 'off' as a supported fan mode to avoid breaking routines
+  if get_device_type(device) == THERMOSTAT_DEVICE_TYPE_ID then
```
hcarter-775: have you checked if any routines use `off`? want to make sure this isn't superfluous at this point.

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1538,39 +1542,59 @@ local function fan_mode_handler(driver, device, ib, response)
...
+      device:endpoint_to_component(ib.endpoint_id),
+      supported_fan_modes_capability.ID,
+      supported_fan_modes_attribute.NAME
+    ) or {}
+    if not tbl_contains(prev_supported_fan_modes, "off") then
+      local _, off_idx = tbl_contains(supported_fan_modes, "off")
+      if off_idx then
+        table.remove(supported_fan_modes, off_idx)
+      end
+    end
```
hcarter-775: I'm starting to wonder, we don't expect this value to change, do we? I almost wonder if it makes sense to just return early if some `prev_supported_fan_modes` is found, or to just use this table rather than dynamically remove something off the newly made `supported_fan_modes` table.

However, if we don't go that route, then:

```suggestion
    -- per the definitions set above, the first index always contains "off"
    if prev_supported_fan_modes[1] ~= "off" then
      table.remove(supported_fan_modes, 1)
    end
```

drivers/SmartThings/matter-thermostat/src/init.lua
```diff
@@ -1538,39 +1542,59 @@ local function fan_mode_handler(driver, device, ib, response)
...
   end
 
-  local event = supported_fan_modes_attribute(supportedFanModes, {visibility = {displayed = false}})
+  local event = supported_fan_modes_attribute(supported_fan_modes, {visibility = {displayed = false}})
   device:emit_event_for_endpoint(ib.endpoint_id, event)
```
hcarter-775:
```suggestion
device:emit_event_for_endpoint(ib.endpoint_id, supported_fan_modes_attribute(supported_fan_modes, {visibility = {displayed = false}}))
```
nit

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: I see this rule wouldn't work for `thermostatFanMode`. To solve this, I'd say do the following:
```
if get_device_type(device) == THERMOSTAT_DEVICE_TYPE_ID and device:supports_capability_by_id(capabilities.fanMode.ID) then
```

nickolas-deboom: Updated as suggested!

drivers/SmartThings/matter-thermostat/src/init.lua
nickolas-deboom: done in latest

nickolas-deboom:
@hcarter-775 I rebased this following the directory restructure and also addressed the feedback you left previously

drivers/SmartThings/matter-thermostat/src/init.lua
hcarter-775: saw you rebased @nickolas-deboom, but I am still curious about this. Cause honestly even if there are like 2, it might be  easier/better not to maintain this and ignore this whole code block

nickolas-deboom: Not sure about how many use `off` specifically, but I checked and there are 30 rules existing rules that use `fanMode` in some way, on two total devices.
Edit: actually there are 89 total rules in place across 31 devices.

nickolas-deboom:
Closing in favor of https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2670 and https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2673

PR #2556, opened 2025-11-14
sonos: Retry ssdp task creation to fix log spam and spinning

NoahCornell:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

We would try to create the SSDP task a single time and if the UDP socket creation failed then we would spin in the init lifecycle loop. This has us retry the task creation if it fails and adds a sleep in the loop so we don't spin anymore.

# Summary of Completed Tests

I made it so the UDP socket creation returns nil for the first 5 attempts and reproduced the behavior, then with this change we eventually recover.

### DIFF
--- a/drivers/SmartThings/sonos/src/lifecycle_handlers.lua
+++ b/drivers/SmartThings/sonos/src/lifecycle_handlers.lua
@@ -165,6 +165,7 @@ function SonosDriverLifecycleHandlers.initialize_device(driver, device)
             { hub_logs = true },
             string.format("Driver wasn't able to spin up SSDP task, cannot initialize devices.")
           )
+          cosock.socket.sleep(30)
         end
       end
     end,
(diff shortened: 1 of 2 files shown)


cjswedes:
@NoahCornell I am not really sure how, but I think this has caused https://smartthings.atlassian.net/browse/CHAD-17218. The timing of the release to beta lines up with the first report of the issue, and the 30 sleeps in this PR also line up with the reported 30 second delays. Another potential cause is related to the oauth functionality where our wait for oauth timeout is also 30 seconds.

NoahCornell:
> @NoahCornell I am not really sure how, but I think this has caused https://smartthings.atlassian.net/browse/CHAD-17218. The timing of the release to beta lines up with the first report of the issue, and the 30 sleeps in this PR also line up with the reported 30 second delays. Another potential cause is related to the oauth functionality where our wait for oauth timeout is also 30 seconds.

I'll take a look at the logs. I would expect this to be caused by the latter because this should really only be an issue at driver start and only if the hub doesn't have a connection which causes us to fail opening a UDP socket.

PR #2503, opened 2025-10-28
Add driver support for Matter Cameras

nickolas-deboom:
Check all that apply

# Type of Change

- [ ] WWST Certification Request
     - If this is your first time contributing code:
          - [ ] I have reviewed the README.md file
          - [ ] I have reviewed the CODE_OF_CONDUCT.md file
          - [ ] I have signed the CLA
     - [ ] I plan on entering a WWST Certification Request or have entered a request through the WWST Certification console at developer.smartthings.com
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have verified my changes by testing with a device or have communicated a plan for testing
- [ ] I am adding new behavior, such as adding a sub-driver, and have added and run new unit tests to cover the new behavior

# Description of Change

# Summary of Completed Tests

PR #2560, opened 2025-11-17
Matter Switch: Simplify component to endpoint mapping logic

hcarter-775:
# Description of Change
This cuts down the amount of checks and reformatting handled in the `endpoint_to_component` and `emit_event_for_endpoint` functions.

This logic was just added by me in the recent [Camera PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2503) and it was a bit rushed. This cleans up some unnecessary logic that was introduced.

# Summary of Completed Tests
Tested by Unit Test and with VDA. Expanded functionality is currently only utilized by unreleased Camera subdriver

### DIFF
--- a/drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
@@ -173,19 +173,19 @@ end
 --- to a single endpoint.
 ---
 --- @param device any a Matter device object
---- @param opts number|table either an ep_id or a table { endpoint_id, optional(cluster_id), optional(attribute_id) }
+--- @param ep_info number|table either an ep_id or a table { endpoint_id, optional(cluster_id), optional(attribute_id) }
 --- where cluster_id is required for an attribute_id to be handled.
 --- @return string component
-function utils.endpoint_to_component(device, opts)
-  if type(opts) == "number" then
-    opts = { endpoint_id = opts }
+function utils.endpoint_to_component(device, ep_info)
+  if type(ep_info) == "number" then
+    ep_info = { endpoint_id = ep_info }
   end
   for component, map_info in pairs(device:get_field(fields.COMPONENT_TO_ENDPOINT_MAP) or {}) do
-    if type(map_info) == "number" and map_info == opts.endpoint_id then
+    if type(map_info) == "number" and map_info == ep_info.endpoint_id then
       return component
-    elseif type(map_info) == "table" and map_info.endpoint_id == opts.endpoint_id
-      and (not map_info.cluster_id or (map_info.cluster_id == opts.cluster_id
-      and (not map_info.attribute_ids or utils.tbl_contains(map_info.attribute_ids, opts.attribute_id)))) then
+    elseif type(map_info) == "table" and map_info.endpoint_id == ep_info.endpoint_id
+      and (not map_info.cluster_id or (map_info.cluster_id == ep_info.cluster_id
+      and (not map_info.attribute_ids or utils.tbl_contains(map_info.attribute_ids, ep_info.attribute_id)))) then
         return component
     end
   end


drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -214,24 +202,21 @@ end
...
   if type(ep_info) == "number" then
-    ep_info = { endpoint_id = ep_info }
+    formatted_info = { endpoint_id = ep_info }
   elseif type(ep_info) == "table" then
-    ep_info = {
-      endpoint_id = ep_info.endpoint_id,
-      cluster_id = ep_info.cluster_id,
-      attribute_id = ep_info.attribute_id
-    }
+    formatted_info = ep_info
```
gharveymn: Effectively, can't this just be

```lua
  if type(ep_info) == "number" then
    ep_info = { endpoint_id = ep_info }
  end
```

? There isn't any transform done to the table form, and other types are not transformed.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -214,24 +202,21 @@ end
 --- @param ep_info number|table endpoint_id or ib (includes endpoint_id, cluster_id, attribute_id)
```
gharveymn: Are `cluster_id` and `attribute_id` required when in table form? Is `capability_id` required as well?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -176,29 +176,17 @@ end
 --- @param opts number|table either is an ep_id or a table { endpoint_id, capability_id }
```
gharveymn: I think this doc should include mention of the required fields of `cluster_id` and `attribute_id` when in table form.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -176,29 +176,17 @@ end
...
-    end
+    opts = { endpoint_id = opts }
   end
   for component, map_info in pairs(device:get_field(fields.COMPONENT_TO_ENDPOINT_MAP) or {}) do
-    if type(map_info) == "number" and map_info == ep_info.endpoint_id then
+    if type(map_info) == "number" and map_info == opts.endpoint_id then
+      return component
+    elseif type(map_info) == "table" and map_info.endpoint_id == opts.endpoint_id
+      and (not map_info.cluster_id or (map_info.cluster_id == opts.cluster_id
+      and utils.tbl_contains(map_info.attribute_ids, opts.attribute_id)))
```
gharveymn: We are using the logic that existence of the `cluster_id` field implies the existence of a non-nil `attribute_id`. Is this always the case?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -214,24 +202,21 @@ end
...
-    local child = utils.find_child(device, ep_info.endpoint_id)
+    local child = utils.find_child(device, formatted_info.endpoint_id)
     if child ~= nil then
       child:emit_event(event)
       return
     end
   end
-  local opts = { endpoint_info = ep_info, capability_id = event.capability.ID }
-  local comp_id = utils.endpoint_to_component(device, opts)
+  formatted_info.capability_ID = event.capability.ID
```
gharveymn: Typo:

```suggestion
  formatted_info.capability_id = event.capability.ID
```

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: oh it can. I just thought that might be weird/wrong cause the capability event id is going to get appended later and that's not technically "ep info"

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: It is at the time (and I can't see when/why it wouldn't be), though I can see how it can seem a little arbitrary. fixed in next commit (cannot push now due to github outage)

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: again, they are at the time but I suppose don't need to be. changed the wording in the upcoming commit

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: but I suppose it's kinda equally weird to break it apart like I did, so I'll switch it

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: well, it actually reads `endpoint_id or ib`, and I note that the `ib` includes that data. I agree this wording is a little confused

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: ended up removing capabilities from the mapping, so this makes even more sense now

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: updated

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: updated

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: updated

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -168,60 +168,40 @@ function utils.component_to_endpoint(device, component)
...
---- An extension of the library function endpoint_to_component, to support a mapping scheme
---- that includes cluster and attribute id's so that we can use multiple components for a
---- single endpoint.
+--- An extension of the library function endpoint_to_component, used to support a mapping scheme
+--- that optionally includes cluster and attribute ids so that multiple components can be mapped
+--- to a single endpoint.
 ---
 --- @param device any a Matter device object
---- @param opts number|table either is an ep_id or a table { endpoint_id, capability_id }
+--- @param ep_info number|table either an ep_id or a table { endpoint_id, optional(cluster_id), optional(attribute_id) }
```
ctowns, line 176: Can we choose a single parameter type and make the the caller provide the additional info? I know lua doesn't have strong parameter typing but I think this is a practice we should try to avoid. If you need to handle additional optional values, they can be passed as optional parameters:
```
function utils.endpoint_to_component(device, ep, cluster_id, attribute_id)
```

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -168,60 +168,40 @@ function utils.component_to_endpoint(device, component)
...
---- differentiated based on a simple endpoint id to component mapping, but we can extend
---- this mapping to include the cluster and attribute id's so that we know which component
---- to route events to.
+--- An extension of the library function emit_event_for_endpoint, used to support devices with
+--- multiple components mapped to the same endpoint. This is handled by extending the parameters to optionally
+--- include a cluster id and attribute id for more specific routing
 ---
 --- @param device any a Matter device object
---- @param ep_info number|table endpoint_id or ib (includes endpoint_id, cluster_id, attribute_id)
+--- @param ep_info number|table endpoint_id or an ib (the ib data includes endpoint_id, cluster_id, and attribute_id fields)
```
ctowns, line 200: Same comment as above, I think we should try to make parameters a known single type whenever possible

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: We cannot. Since we want to have this be compatible with the lua libs, we cannot add extra arguments. This was a compromise I figured was relatively clear given the limitation.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: See other response above

hcarter-775: Once nicety as well with this (and the other def) is that it allows the caller to simply give an `ib` directly, which is a very common datatype that we handle.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
ctowns: I don't understand the reasoning for why we cannot add extra arguments - why is backwards compatibility a concern if we are creating a new function that overrides the lua libs? I think we should only use this override when needed, and it shouldn't be a blanket override for all devices in the switch driver, unless they specifically require this advanced mapping, right? 

If this is a function override, I assume we can make it whatever we want. Even using `extend_device` allows us to override the function pointer from the lua libs and just point to this new function, even though I don't neccessarily think using `extend_device` in this case is the way to go.

ctowns: Also just a note that this discussion isn't a blocker if this needs to go in urgently, but I think it's worth having more discussion around how we should be designing these APIs for this complex mapping.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -168,60 +168,40 @@ function utils.component_to_endpoint(device, component)
...
-    if opts.endpoint_info then
-      ep_info = opts.endpoint_info
-    else
-      ep_info = {
-        endpoint_id = opts.endpoint_id,
-        cluster_id = opts.cluster_id,
-        attribute_id = opts.attribute_id
-      }
-    end
+function utils.endpoint_to_component(device, ep_info)
```
nickolas-deboom, line 179: ep_info not being passed in at the callsite doesn't cause an issue with the `type` checks right?

nickolas-deboom (approved):
LGTM, I tested with the camera example app and things are looking good with these changes.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: yep, everything would turn out fine with a nil `ep_info`. In that case we'd just end up returning the main component

PR #2564, opened 2025-11-18
Matter Switch: Add more reliability to device type checking

hcarter-775:
# Description of Change
Because not all devices match spec exactly, device type checking should not directly call the "first" device type, since even Application scoped endpoints may have other non-spec compliant device types attached as well (such as bridged node). This has been seen in the field.

# Summary of Completed Tests
Tested on a device breaking spec by putting Bridged Node on all endpoints

### DIFF
--- a/drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
@@ -93,12 +93,12 @@ function utils.device_type_supports_button_switch_combination(device, endpoint_i
   return utils.tbl_contains(dimmable_eps, endpoint_id)
 end
 
---- Some devices report multiple device types which are a subset of
---- a superset device type (Ex. Dimmable Light is a superset of On/Off Light).
---- We should map to the largest superset device type supported.
+--- Some devices report multiple device types which are a subset of a superset
+--- device type (Ex. Dimmable Light is a superset of On/Off Light). We should map
+--- to the largest superset device type supported.
 --- This can be done by matching to the device type with the highest ID
---- note: that superset device types have a higher ID than their subset device types is heuristic
---- and could therefore break in the future if the spec were expanded
+--- note: that superset device types have a higher ID than those of their subset
+--- is heuristic and could therefore break in the future, were the spec expanded
 function utils.find_max_subset_device_type(ep, device_type_set)
   if ep.endpoint_id == 0 then return end -- EP-scoped device types not permitted on Root Node
   local primary_dt_id = -1


drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -93,23 +93,31 @@ function utils.device_type_supports_button_switch_combination(device, endpoint_i
...
 end
 
--- Some devices report multiple device types which are a subset of
--- a superset device type (Ex. Dimmable Light is a superset of On/Off Light).
--- We should map to the largest superset device type supported.
--- This can be done by matching to the device type with the highest ID
+--- Some devices report multiple device types which are a subset of
+--- a superset device type (Ex. Dimmable Light is a superset of On/Off Light).
+--- We should map to the largest superset device type supported.
+--- This can be done by matching to the device type with the highest ID
```
gharveymn, line 99: Can we cite this?

hcarter-775: done https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2564/commits/b11bf3f534b98b216464617aaeb215e58f970b65, how's this look?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -93,23 +93,33 @@ function utils.device_type_supports_button_switch_combination(device, endpoint_i
...
-      end
+  local primary_dt_id = -1
+  for _, dt in ipairs(ep.device_types) do
+    -- only device types in the subset should be considered.
+    if utils.tbl_contains(device_type_set, dt.device_type_id) then
+      primary_dt_id = math.max(primary_dt_id, dt.device_type_id)
+    end
+  end
+  return (primary_dt_id > 0) and primary_dt_id or nil
+end
```
nickolas-deboom, line 112:
```suggestion
  local primary_dt_id
  for _, dt in ipairs(ep.device_types) do
    -- only device types in the subset should be considered.
    if utils.tbl_contains(device_type_set, dt.device_type_id) then
      primary_dt_id = math.max(primary_dt_id, dt.device_type_id)
    end
  end
  return primary_dt_id
end
```
nit- minor suggestion feel free to ignore

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -93,23 +93,33 @@ function utils.device_type_supports_button_switch_combination(device, endpoint_i
...
+
+--- Lights and Switches are Device Types that have Superset-style functionality
+--- For all other device types, this function should be used to identify the primary device type
+function utils.find_primary_device_type(ep_info)
+  for _, dt in ipairs(ep_info.device_types) do
```
nickolas-deboom, line 117: Sort this list first maybe?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: I believe this would crash, since the first `math.max(primary_dt_id, dt.device_type_id)` would compare a nil with a number?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: what would it be sorting? Further, the device type ids aren't necessarily "ordered" in a meaningful way, adding to my question of what "sorting" would be wanted per se?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
gharveymn: Sorry, I wasn't being clear. I was actually referring to the assertion that the superset IDs are always greater than the IDs of subsets. I think what you had before for the rest is actually more clear than after the change.

gharveymn: If it's not explicitly specified, and it isn't feasible to switch to an alternative algorithm, we should explicitly point out that this is just a heuristic, and could break in the future.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
nickolas-deboom: Oh yeah good call 👍

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
nickolas-deboom: Is there anything special about the 'first' device type in `ep_info.device_types` that would be returned by this function in that case?

hcarter-775: well, the idea is we need to return something. Generally, there should only _be_ one. And before, it was just the first one. Now it's the first one that isn't bridged node. It's kinda jank cause it handles a spec-breaking system.

gharveymn: > the device type ids aren't necessarily "ordered" in a meaningful way

does that fact add additional context to [this discussion](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2564#discussion_r2539319889), or is this referring to something else?

hcarter-775: @gharveymn kinda, but not necessarily. It does speak to the fact that for "superset" scoped device types, the fact that the ids increment is heuristic and may not always be true (though at present it is). On the other hand, this comment also speaks to the fact that the ordering of device types on an endpoint is arbitrary as well

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: revrerted, and added a note

PR #2565, opened 2025-11-18
Matter Window Covering: Remove default tilt preset

nickolas-deboom:
Remove the `GoToTiltPercentage` command sent when the `presetPosition` capability command is received, which previously would set the tilt preset level to a default level of 50%. A new capability will be introduced later to support setting the tilt preset level.

### DIFF
--- a/drivers/SmartThings/matter-window-covering/src/init.lua
+++ b/drivers/SmartThings/matter-window-covering/src/init.lua
@@ -29,7 +29,7 @@ local battery_support = {
 }
 local REVERSE_POLARITY = "__reverse_polarity"
 local PRESET_LEVEL_KEY = "__preset_level_key"
-local PRESET_LEVEL = 50
+local DEFAULT_PRESET_LEVEL = 50
 
 local function find_default_endpoint(device, cluster)
   local res = device.MATTER_DEFAULT_ENDPOINT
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -131,25 +131,15 @@ local function device_removed(driver, device) log.info("device removed") end
...
 -- capability handlers
 local function handle_preset(driver, device, cmd)
+  local lift_value = device:get_latest_state(
+    "main", capabilities.windowShadePreset.ID, capabilities.windowShadePreset.position.NAME
+  ) or DEFAULT_PRESET_LEVEL
+  local hundredths_lift_percent = (100 - lift_value) * 100
   local endpoint_id = device:component_to_endpoint(cmd.component)
-  local lift_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.LIFT})
-  local tilt_eps = device:get_endpoints(clusters.WindowCovering.ID, {feature_bitmap = clusters.WindowCovering.types.Feature.TILT})
-  if #lift_eps > 0 then
```
nickolas-deboom, line 137: Note that this check isn't needed since any device using a profile with the `windowShadePreset`capability supports the LIFT feature.

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -131,25 +131,15 @@ local function device_removed(driver, device) log.info("device removed") end
...
-  if #tilt_eps > 0 then
-    local req = clusters.WindowCovering.server.commands.GoToTiltPercentage(
-      device, endpoint_id, PRESET_LEVEL * 100
-    )
-    device:send(req)
-  end
+  local req = clusters.WindowCovering.server.commands.GoToLiftPercentage(
+    device, endpoint_id, hundredths_lift_percent
+  )
+  device:send(req)
```
hcarter-775:
```suggestion
  device:send(clusters.WindowCovering.server.commands.GoToLiftPercentage(
    device, endpoint_id, hundredths_lift_percent
  ))
```
nit?

hcarter-775 (approved):

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: changed in latest

PR #2575, opened 2025-11-24
WWSTCERT-9051 eufy FamiLock E32

greens:


drivers/SmartThings/matter-lock/fingerprints.yml
```diff
@@ -51,6 +51,11 @@ matterManufacturer:
     vendorId: 0x1533
     productId: 0x0012
     deviceProfileName: lock-user-pin-battery
+  - id: "5427/22"
+    deviceLabel: eufy FamiLock E32
+    vendorId: 0x1533
+    productId: 0x0016
```
nickolas-deboom, line 57: I don't see this device in the DCL, should it be?

PR #2578, opened 2025-11-24
WWSTCERT-9105 Meross Smart Presence Sensor (Thread)

greens:


drivers/SmartThings/matter-sensor/fingerprints.yml
```diff
@@ -128,6 +128,11 @@ matterManufacturer:
     vendorId: 0x1345
     productId: 0x4201
     deviceProfileName: motion-illuminance
+  - id: "4933/16898"
+    deviceLabel: Smart Presence Sensor (Thread)
+    vendorId: 0x1345
+    productId: 0x4202
+    deviceProfileName: matter-motion-battery-illuminance
```
nickolas-deboom: I believe that `motion-illuminance-battery` would be preferred, since this is what the driver will actually profile to. However, it doesn't really make a difference since they have the same capabilities. Do you have an opinion @hcarter-775 ?

hcarter-775: yeah let's not use this profile. I missed this in my earlier audit of profiles marked for deprecation

hcarter-775 (approved):

PR #2579, opened 2025-11-24
Matter Switch Subdriver: Add Bilresa Support

hcarter-775:
# Description of Change
Add initial subdriver support for the Bilresa device.

# Summary of Completed Tests
Unit test added to ensure proper onboarding. Tested on-device.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/ikea_scroll/scroll_utils/fields.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/ikea_scroll/scroll_utils/fields.lua
@@ -11,7 +11,8 @@ IkeaScrollFields.ENDPOINT_POWER_SOURCE = 0
 -- Switch Endpoints used for basic press functionality
 IkeaScrollFields.ENDPOINTS_PRESS = {3, 6, 9}
 
--- Required Events for the ENDPOINTS_PRESS. Ignore InitialPress since this slows handling.
+-- Required Events for the ENDPOINTS_PRESS. Remove the default subscription to
+-- InitialPress since this is a MultiPress device and InitialPress will be ignored.
 IkeaScrollFields.switch_press_subscribed_events = {
   clusters.Switch.events.MultiPressComplete.ID,
   clusters.Switch.events.LongPress.ID,


tpmanley (approved):
What's here looks good to me. I do have a question about the capabilities on the `main` component which we can discuss offline.

drivers/SmartThings/matter-switch/profiles/ikea-scroll.yml
```diff
@@ -0,0 +1,33 @@
...
+      - id: firmwareUpdate
+        version: 1
+      - id: refresh
+        version: 1
+    categories:
+    - name: RemoteController
+  - id: group1
+    label: Group 1
+    capabilities:
+      - id: button
```
tpmanley: For consistency with other multi-button Matter devices lets put the first group `button` on the `main` component

mocelet: Also, if main only has the battery it displays as an active tile with the battery percentage instead of the typical "Standby" state.

I'm going to publish a custom driver to use the wheel and add some features but I'd like to use the same IDs of stock drivers so people don't have to re-create automations and can switch back and forth at will.

drivers/SmartThings/matter-switch/src/sub_drivers/ikea_scroll/scroll_utils/fields.lua
```diff
@@ -0,0 +1,20 @@
...
+
+local clusters = require "st.matter.clusters"
+
+local IkeaScrollFields = {}
+
+-- PowerSource supported on Root Node
+IkeaScrollFields.ENDPOINT_POWER_SOURCE = 0
+
+-- Switch Endpoints used for basic press functionality
+IkeaScrollFields.ENDPOINTS_PRESS = {3, 6, 9}
```
nickolas-deboom, line 12: Would it be better to read the endpoints rather than use hard coded values? I'm just thinking about a scenario where they get changed following a fw update or something

nickolas-deboom: Now that I think about it that seems pretty unlikely and would break a lot more, so probably not worth worrying about

drivers/SmartThings/matter-switch/src/sub_drivers/ikea_scroll/scroll_utils/fields.lua
```diff
@@ -0,0 +1,20 @@
...
+
+-- Required Events for the ENDPOINTS_PRESS. Ignore InitialPress since this slows handling.
```
nickolas-deboom: Could you clarify why InitialPress isn't subscribed to?

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -117,7 +117,10 @@ SwitchFields.vendor_overrides = {
     [0x1006] = { ignore_combo_switch_button = true, target_profile = "light-level-power-energy-powerConsumption", ep_id = 1 }, -- 3 Buttons(Generic Switch), 1 Channels(Dimmable Light)
     [0x100A] = { ignore_combo_switch_button = true, target_profile = "light-level-power-energy-powerConsumption", ep_id = 1 }, -- 1 Buttons(Generic Switch), 1 Channels(Dimmable Light)
     [0x2004] = { is_climate_sensor_w100 = true }, -- Climate Sensor W100, requires unique profile
-  }
+  },
+  [0x117C] = { -- IKEA_MANUFACTURER_ID
+    [0x8000] = { is_ikea_scroll = true }
```
nickolas-deboom, line 117: I like this approach for `can_handle`, maybe we should implement this for other subdrivers eventually

drivers/SmartThings/matter-switch/src/test/test_ikea_scroll.lua
```diff
@@ -0,0 +1,223 @@
...
+        )
+      },
+    },
+    {
+      channel = "capability",
+      direction = "send",
+      message = mock_ikea_scroll:generate_test_message("main",
+        capabilities.button.supportedButtonValues({"pushed", "double", "held", "pushed_3x"}, {visibility = {displayed = false}}))
+    },
+        {
```
nickolas-deboom: Theres an extra tab here and on line 204 😿

nickolas-deboom (approved):
This looks good from my POV. Btw is there only one version / pid of this device? Do we want to add a fingerprint for it?

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
hcarter-775: yeah that's the idea!

drivers/SmartThings/matter-switch/src/sub_drivers/ikea_scroll/scroll_utils/fields.lua
hcarter-775: `InitialPress` is always ignored if the device supports multipress, since we wait for multi press complete. Therefore, since this is a single device that we know supports multipress, we don't even need to handle it.

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
hcarter-775: Generally, my thought is to put all device overrides, of any type, into this table in some form or another. I think that would be nice for keeping everything organized.

drivers/SmartThings/matter-switch/src/sub_drivers/ikea_scroll/scroll_utils/fields.lua
nickolas-deboom: Ah I see now, thank you!

ctowns: nit: should this say something like this?
```suggestion
-- Required Events for the ENDPOINTS_PRESS. Ignore InitialPress since this is a multi press device and initial press is ignored.
```

ctowns (approved):
LGTM! It's very well organized, nice job Harrison!

drivers/SmartThings/matter-switch/src/sub_drivers/ikea_scroll/scroll_utils/fields.lua
hcarter-775: yeah, changed a bit following this suggestion!

hcarter-775:
> This looks good from my POV. Btw is there only one version / pid of this device? Do we want to add a fingerprint for it?
@nickolas-deboom There is, but I think the fingerprint should only be added when it is WWST certified.

PR #2590, opened 2025-12-01
Support additional Camera device types

nickolas-deboom:
This adds support for the following device types:
 *  0x0140 Intercom
 *  0x0141 Audio Doorbell
 *  0x0142 Camera
 *  0x0143 Video Doorbell
 *  0x0144 Floodlight Camera
 *  0x0145 Snapshot Camera
 *  0x0146 Chime
 *  0x0148 DoorBell

[ ](https://smartthings.atlassian.net/browse/CHAD-17567)


samadDotDev (approved):
Will need to re-evaluate after rebase

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -257,13 +257,19 @@ function utils.matter_handler(driver, device, response_block)
...
 end
 
--- get a list of endpoints for a specified device type.
+-- get a list of endpoints for a specified device type or list of device types
 function utils.get_endpoints_by_device_type(device, device_type_id)
   local dt_eps = {}
   for _, ep in ipairs(device.endpoints) do
     for _, dt in ipairs(ep.device_types) do
-      if dt.device_type_id == device_type_id then
+      if type(device_type_id) == "table" then
```
hcarter-775: this function will have to be rebased a bit carefully- there have been updates here since this PR was created.

nickolas-deboom: Thanks for the callout - I updated this function a bit in the last force push

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -295,18 +295,28 @@ function utils.matter_handler(driver, device, response_block)
   device.log.info(string.format("Fallback handler for %s", response_block))
 end
 
--- get a list of endpoints for a specified device type.
+-- get a list of endpoints for a specified device type or list of device types
```
hcarter-775: can we add a proper lua doc here, like:
``` 
--- @param device ...
--- @param device_type_id number|table a specified device type or list of device types
--- @param opts ....
--- @return table ...
```

I think there's enough variation in return and input here to warrant proper use docs.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -295,18 +295,28 @@ function utils.matter_handler(driver, device, response_block)
...
+  end
+
   for _, ep in ipairs(device.endpoints) do
     for _, dt in ipairs(ep.device_types) do
-      if dt.device_type_id == device_type_id then
-        if opts.with_info then
-          table.insert(dt_eps, ep)
-        else
-          table.insert(dt_eps, ep.endpoint_id)
+      if type(device_type_id) == "table" then
```
hcarter-775: Rather than check for a table in the loop body, maybe at the top do something like:

```
if device_type_id == nil then return end
device_type_id = device_type_id and type(device_type_id) == "table" or { device_type_id }
``` 
Then we can ensure the type + an added nil check before we event start the loop?

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -310,6 +310,14 @@ function CameraUtils.subscribe(device)
     }
   }
 
+  -- If using the standalone Chime profile, change the subscribed attribute list for audioMute appropriately.
+  if #switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.CAMERA.CHIME) > 0 and
+    #switch_utils.get_endpoints_by_device_type(device, camera_fields.camera_profile_device_types) == 0 then
+    camera_subscribed_attributes[capabilities.audioMute.ID] = {
+      clusters.Chime.attributes.Enabled
+    }
```
hcarter-775: can we just add this to the table?

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/fields.lua
```diff
@@ -47,4 +48,15 @@ CameraFields.ABS_ZOOM_MIN = 1
 CameraFields.ABS_VOL_MAX = 254.0
 CameraFields.ABS_VOL_MIN = 0.0
 
+-- Subset of camera device types that always use the camera profile, excluding
+-- DoorBells and Chimes as they can be standalone devices.
```
hcarter-775: If this is true, we should just remove those from the camera subset right? Then we can use that directly instead of redefining it here?

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -25,10 +25,17 @@ SwitchFields.CURRENT_HUESAT_ATTR_MAX = 254
...
-  CHIME = 0x0146,
+  CAMERA = {
+    INTERCOM = 0x0140,
+    AUDIO_DOORBELL = 0x0141,
+    CAMERA = 0x0142,
+    VIDEO_DOORBELL = 0x0143,
+    FLOODLIGHT_CAMERA = 0x0144,
+    SNAPSHOT_CAMERA = 0x0145,
+    CHIME = 0x0146,
+    DOORBELL = 0x0148,
```
hcarter-775: mentioned in another comment, these are not camera-specific, correct?

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/fields.lua
nickolas-deboom: I did it this way to reduce the number of `get_endpoints_by_device_type` calls in can_handle.lua, but decided to move them out of the camera list like you suggested, because I think that it makes more sense.

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
nickolas-deboom: Done in 5e52d8a651715fc6676043f3ae1998ab07ad014b!

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
nickolas-deboom: Done in 5e52d8a651715fc6676043f3ae1998ab07ad014b

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
nickolas-deboom: They are listed under the Camera device type in the spec, but since they are also able to be implemented as standalone devices I moved them out of this list as suggested!

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: Adding this to the table revealed an issue in `populate_subscribe_request_for_device`. I noticed that the standalone Chime mock device in test_matter_camera.lua was subscribing to the other attributes listed for `audioMute` in the `camera_subscribed_attributes` table (`clusters.CameraAvStreamManagement.attributes.SpeakerMuted` &       `clusters.CameraAvStreamManagement.attributes.MicrophoneMuted`), even though the mock device does not support the CAVSM cluster.

The issue is that  `populate_subscribe_request_for_device` does not perform a check to make sure the device supports a given cluster before subscribing to the attributes like `add_subscribed_attribute` does. I added a check in 0542fb3a423366726073aa9e31c3214a08306772. Let me know if you have thoughts on this approach!

Note that adding this check lead to failures in test_matter_sensor_offset_preferences.lua because it was missing some endpoint definitions in its mock device and therefore no longer subscribing to certain attributes.

drivers/SmartThings/matter-switch/src/test/test_matter_sensor_offset_preferences.lua
```diff
@@ -6,116 +6,166 @@ local dkjson = require "dkjson"
 local clusters = require "st.matter.clusters"
 
 local mock_device = test.mock_device.build_test_matter_device({
```
nickolas-deboom, line 11: Mentioned elsewhere, but after making the changes in 0542fb3a423366726073aa9e31c3214a08306772 the tests here began to fail because they expect subscriptions to the button events yet no endpoints with the Switch cluster were present in the mock device. I added the endpoints and also fixed the tabbing in this file which is why the diff looks so large

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
hcarter-775: Yeah this works! Maybe let's also add an else log like `else log.warn_with({ hub_logs = true }, string.format("Device does not support cluster 0x%04X, not adding subscribed event", cluster_id)) end`

nickolas-deboom: Done in fb72b212!

drivers/SmartThings/matter-switch/profiles/av-doorbell.yml
```diff
@@ -0,0 +1,159 @@
...
+      capability: imageCapture
+      version: 1
+  dpInfo:
+    - os: ios
+      dpUri: "storyboard://HMVSController/HMVSViewController"
+    - os: android
+      dpUri: "plugin://com.samsung.android.plugin.camera"
+metadata:
+  mnmn: SmartThingsEdge
+  vid: matter-camera
```
samadDotDev:
```suggestion
  vid: matter-av-doorbell
```

drivers/SmartThings/matter-switch/src/test/test_matter_camera.lua
```diff
@@ -114,7 +114,68 @@ local mock_device = test.mock_device.build_test_matter_device({
...
+        },
+      },
+      device_types = {
+        {device_type_id = 0x0146, device_type_revision = 1} -- Chime
+      }
+    }
+  }
+})
+
+local mock_device_doorbell = test.mock_device.build_test_matter_device({
```
samadDotDev, line 151: Probably worth creating one for video doorbell too due to the superset device type nuances

samadDotDev (approved):

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -115,40 +117,48 @@ function CameraDeviceConfiguration.match_profile(device, status_light_enabled_pr
...
-  if #doorbell_endpoints > 0 then
-    table.insert(doorbell_component_capabilities, capabilities.button.ID)
-  end
-  if status_light_enabled_present then
-    table.insert(status_led_component_capabilities, capabilities.switch.ID)
-  end
-  if status_light_brightness_present then
-    table.insert(status_led_component_capabilities, capabilities.mode.ID)
-  end
+	  if #device:get_endpoints(clusters.Chime.ID, {cluster_type = "SERVER"}) > 0 then
```
samadDotDev: `SERVER` or `BOTH`? Basically reuse logic similar to `has_server_cluster_type()`

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/fields.lua
hcarter-775: we should remove this table if it's unused, right?

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
nickolas-deboom: Done in b082796b6366cec6be8b2c57d377b303d3a8ba45, does that look alright? It uses the first endpoint containing the Chime cluster, but I wasn't sure if the client and server clusters would necessarily be on the same endpoint or not

drivers/SmartThings/matter-switch/profiles/av-doorbell.yml
nickolas-deboom: Done in b082796b6366cec6be8b2c57d377b303d3a8ba45

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/fields.lua
nickolas-deboom: Oop, removed!

drivers/SmartThings/matter-switch/src/test/test_matter_camera.lua
nickolas-deboom: Done in 5a1631dff8a6256840247bfc73f9ef0fa1d4f070!

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -447,34 +458,43 @@ end
...
+--- @param parent_device any the parent Matter device object; could be the same as checked_device if checked_device is the parent device
 --- @param subscribe_request table a subscribe request that will be appended to as needed for the device
 --- @param capabilities_seen table a list of capabilities that have already been checked by previously handled devices
 --- @param attributes_seen table a list of attributes that have already been checked
 --- @param events_seen table a list of events that have already been checked
 --- @param subscribed_attributes table key-value pairs mapping capability ids to subscribed attributes
 --- @param subscribed_events table key-value pairs mapping capability ids to subscribed events
-function utils.populate_subscribe_request_for_device(checked_device, subscribe_request, capabilities_seen, attributes_seen, events_seen, subscribed_attributes, subscribed_events)
- for _, component in pairs(checked_device.st_store.profile.components) do
+function utils.populate_subscribe_request_for_device(checked_device, parent_device, subscribe_request, capabilities_seen, attributes_seen, events_seen, subscribed_attributes, subscribed_events)
```
hcarter-775: Before merging, let's document a little testing here, since this change does affect our subscription logic. LGTM, but I think it's still worth ensuring we haven't missed something

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -25,10 +25,17 @@ SwitchFields.CURRENT_HUESAT_ATTR_MAX = 254
 SwitchFields.DEVICE_TYPE_ID = {
   AGGREGATOR = 0x000E,
   BRIDGED_NODE = 0x0013,
-  CAMERA = 0x0142,
+  CAMERA = {
+    INTERCOM = 0x0140,
```
hcarter-775: This isn't a camera really?

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -25,10 +25,17 @@ SwitchFields.CURRENT_HUESAT_ATTR_MAX = 254
...
+    AUDIO_DOORBELL = 0x0141,
+    CAMERA = 0x0142,
+    VIDEO_DOORBELL = 0x0143,
+    FLOODLIGHT_CAMERA = 0x0144,
```
hcarter-775: Do we need to define Floodlight camera now? Since this is a composite device type is there any unique handling that would require this device type?

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -25,10 +25,17 @@ SwitchFields.CURRENT_HUESAT_ATTR_MAX = 254
 SwitchFields.DEVICE_TYPE_ID = {
   AGGREGATOR = 0x000E,
   BRIDGED_NODE = 0x0013,
-  CAMERA = 0x0142,
+  CAMERA = {
+    INTERCOM = 0x0140,
+    AUDIO_DOORBELL = 0x0141,
```
hcarter-775: This also isn't really a camera?

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
hcarter-775: Similar question for VIDEO_DOORBELL? It appears that all current handling exist by separating out CAMERA and DOORBELL explicitly?

I guess I'm not sure what should be done, but to me it feels like it would be good to clarify and either 1. just remove this/these references to device types that we don't use, or 2. add explicit handling for these device type configurations instead of handling a "if one exists but not the other, elseif " case-by-case basis kind of thing.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/fields.lua
```diff
@@ -2,6 +2,7 @@
 -- Licensed under the Apache License, Version 2.0
 
 local clusters = require "st.matter.clusters"
+local switch_fields = require "switch_utils.fields"
```
hcarter-775: remove, unused

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
nickolas-deboom: For video doorbell, we need the device type listed here in order to differentiate between CAMERA and VIDEO_DOORBELL so we match to the proper profile (av-doorbell). And same with Audio DoorBell. We don't necessarily need Floodlight Camera since it would currently use the camera profile rather than a distinct one, but I don't see harm in leaving it here since we may create a new category/icon and profile for it eventually

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
nickolas-deboom: Intercoms are grouped with Camera in the spec and since intercoms create an audio connection they use CAVSM and other camera related clusters I think it makes sense to leave it here. (unlike doorbell and chime, neither of which include camera related clusters)

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
samadDotDev: Due to the device type requirements being different between floodlight camera (composed of on/off light and camera device type) and the snapshot camera (a standalone device type - doesn't contain camera device type), I think we should just avoid clustering these together. This also makes it less confusing when trying to access `CAMERA.CAMERA` for the base camera device type id

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/fields.lua
nickolas-deboom: done

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -48,21 +48,26 @@ function CameraDeviceConfiguration.match_profile(device, status_light_enabled_pr
...
   end
 
   local camera_endpoints = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.CAMERA)
+  local chime_endpoints = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.CHIME)
+  local doorbell_endpoints = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.DOORBELL)
+  local av_stream_ep_ids = device:get_endpoints(clusters.CameraAvStreamManagement.ID)
+
   if #camera_endpoints > 0 then
-    local camera_ep = switch_utils.get_endpoint_info(device, camera_endpoints[1])
+    local camera_ep = switch_utils.get_endpoint_info(device, av_stream_ep_ids[1] or camera_endpoints[1])
```
samadDotDev: Unless a separate category is needed based on device types (which is already covered later by keyeing off doorbell), the endpoints look up here should be specific to the cluster itself to avoid undefined scenarios where device type doesn't match and the cluster could be on a separate endpoint.
```suggestion
    local av_stream_ep = switch_utils.get_endpoint_info(device, av_stream_ep_ids[1])
```
You'll also need to do the same for CameraAvSettingsUserLevelManagement.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -48,21 +48,26 @@ function CameraDeviceConfiguration.match_profile(device, status_light_enabled_pr
...
     for _, ep_cluster in pairs(camera_ep.clusters or {}) do
       if ep_cluster.cluster_id == clusters.CameraAvStreamManagement.ID and has_server_cluster_type(ep_cluster) then
         local clus_has_feature = function(feature_bitmap)
           return clusters.CameraAvStreamManagement.are_features_supported(feature_bitmap, ep_cluster.feature_map)
         end
         if clus_has_feature(clusters.CameraAvStreamManagement.types.Feature.VIDEO) then
-          if switch_utils.find_cluster_on_ep(camera_ep, clusters.PushAvStreamTransport.ID, "SERVER") then
+          if #device:get_endpoints(clusters.PushAvStreamTransport.ID, {cluster_type = "SERVER"}) > 0 then
```
samadDotDev: Same thing is needed for the PushAV gating of `audioRecording`

nickolas-deboom: Done!

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
nickolas-deboom: Done, see 17d5bb3988f4d45389faba1592cefcb58ab9373e. This helped clean up some of the profiling logic as well

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
nickolas-deboom: Done in 17d5bb3988f4d45389faba1592cefcb58ab9373e

drivers/SmartThings/matter-switch/src/sub_drivers/camera/can_handle.lua
```diff
@@ -2,13 +2,16 @@
...
 return function(opts, driver, device, ...)
+  local clusters = require "st.matter.clusters"
   local device_lib = require "st.device"
   local fields = require "switch_utils.fields"
   local switch_utils = require "switch_utils.utils"
   if device.network_type == device_lib.NETWORK_TYPE_MATTER then
     local version = require "version"
     if version.rpc >= 10 and version.api >= 16 and
-      #switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.CAMERA) > 0 then
+      #device:get_endpoints(clusters.CameraAvStreamManagement.ID, {cluster_type = "SERVER"}) > 0 or
```
ctowns: We would want to use the version gating for any of these device types, so you'd need to add parentesis around the string of `or` conditionals so that the logic is something like if version > X and (has_camera_cluster or is_chime or is_doorbell):
```
if version.rpc >= 10 and version.api >= 16 and
    (#device:get_endpoints(clusters.CameraAvStreamManagement.ID, { cluster_type = "SERVER" }) > 0 or
      #switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.CHIME) > 0 or
      #switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.DOORBELL) > 0) then
  return true, require("sub_drivers.camera")
end
```

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -201,107 +201,123 @@ function CameraDeviceConfiguration.match_profile(device)
...
+      table.insert(main_component_capabilities, capabilities.imageControl.ID)
+    end
+    if clus_has_feature(clusters.CameraAvStreamManagement.types.Feature.HIGH_DYNAMIC_RANGE) then
+      table.insert(main_component_capabilities, capabilities.hdr.ID)
+    end
+    if clus_has_feature(clusters.CameraAvStreamManagement.types.Feature.NIGHT_VISION) then
+      table.insert(main_component_capabilities, capabilities.nightVision.ID)
+    end
+    if #chime_endpoints > 0 then
+      table.insert(main_component_capabilities, capabilities.sounds.ID)
```
ctowns, line 254: We would also need to include `audioMute` for handling the `Enabled` attribute on the chime cluster, which is mandatory

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -492,34 +503,43 @@ end
...
```
ctowns: Could we just update this function to get the parent device of the `checked_device` as needed rather than passing on the parent device in the function parameters? If the device doesn't have a parent device, then we would know that it is the parent device.

drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -4251,6 +4246,50 @@ matterGeneric:
...
+    deviceLabel: Matter Floodlight Camera
+    deviceTypes:
+      - id: 0x0144 # Floodlight Camera
+    deviceProfileName: camera
+  - id: "matter/snapshot/camera"
+    deviceLabel: Matter Snapshot Camera
+    deviceTypes:
+      - id: 0x0145 # Snapshot Camera
+    deviceProfileName: camera
+  - id: "matter/chime"
```
ctowns, line 4370: A device that supports Camera and Chime would equally match to the generic fingerprint for both `Matter Camera` and `Matter Chime`, so it is ambiguous which profile would be used initially. I know we would expect this to be fixed by the match profile logic, but could you add an additional fingerprint that matches on both the Camera and Chime device types and uses the camera profile? Also, this would allow us to set the default name to "Matter Camera" for these devices.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/can_handle.lua
nickolas-deboom: fixed in 42bfb6a393d4ed284e22dad5690e966c152448fb

drivers/SmartThings/matter-switch/profiles/av-doorbell.yml
```diff
@@ -0,0 +1,159 @@
...
+                operator: EQUALS
+                operand: "active"
+  detailView:
+    - component: main
+      capability: webrtc
+      version: 1
+    - component: main
+      capability: mechanicalPanTiltZoom
+      version: 1
+    - component: main
```
ctowns, line 137: We will need to add `videoCapture2` here so that the presentation shows up correctly for iOS, similar to the fix that was made here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3018

nickolas-deboom: done in afca26fcddc2145c090b906712644524bd32414e

drivers/SmartThings/matter-switch/fingerprints.yml
nickolas-deboom: done in afca26fcddc2145c090b906712644524bd32414e

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
nickolas-deboom: Good idea, changed in afca26fcddc2145c090b906712644524bd32414e as suggested

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
nickolas-deboom: done in afca26fcddc2145c090b906712644524bd32414e

ctowns (approved):
I think this looks good to me, thanks for making those changes @nickolas-deboom !

Before merging, is there any testing that can be documented in the top comment of the PR? Even if we don't have any of these devices yet, it might be good to run a quick regression test on an existing camera device just to make sure things still work since the init logic is being changed here.

PR #2601, opened 2025-12-02
CHAD-17043: Created pre-commit script to check for newline and copyright on all staged lua files

aleclorimer:
# Description of Change

Created a script to check all staged lua files that they end with a newline character and begin with the current year's copyright header. This file should be made executable and copied to `.git/hooks/pre-commit`.

# Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor

### DIFF
--- a/tools/pre-commit
+++ b/tools/pre-commit
@@ -6,13 +6,17 @@
 # pre-commit
 #
 # - Soft link into .git/hooks/pre-commit
-#   - $ ln -s tools/pre-commit .git/hooks/pre-commit
+#   - $ ln -s ../../tools/pre-commit .git/hooks/pre-commit
 # - Ensure executable
 
 set -e
 
 # Get changed files that are Added or Modified
-files=$(git diff --staged --name-only --diff-filter=AM)
+if [[ "$1" ]]; then
+    files=$(find $1 -type f -follow -print)
+else
+    files=$(git diff --staged --name-only --diff-filter=AM)
+fi
 
 fail=0
 for file in $files; do
@@ -23,7 +27,7 @@ for file in $files; do
             [ -f "$file" ] || continue
 
             # Check if file is not empty, is a SmartThings driver and has the copyright header
-            if [ -s "$file" ] && [[ "$file" =~ "drivers/SmartThings" ]] 
+            if [ -s "$file" ] && [[ "$file" =~ "drivers/SmartThings" ]] \
                 && [[ ! $(grep $file -P -e "-{2,} Copyright \d{4}(?:-\d{4})? SmartThings") ]]; then
                 echo "$file: SmartThings Copyright missing from file"
                 fail=1


tools/pre-commit
```diff
@@ -0,0 +1,41 @@
...
+
+fail=0
+for file in $files; do
+    # Only process .lua files
+    case "$file" in
+        *.lua)
+            # Skip if not a regular file
+            [ -f "$file" ] || continue
+
+            # Check if file is non-empty and does NOT contain the current year copyright header
```
tandres: We do not want to error on year != current_year. Overwriting the year with any modification means that we are erasing some (very minor) historical data.

aleclorimer: Huh... This is a good point. I/we have done a lot of changes to some drivers with the lazy loading v2 work changing the copyright to 2025 from their previously listed year. Is this anything of concern even though you said "very minor"?

tandres: No nothing to lose sleep over the already changed files but we should not continue this practice. It's one of those things that is vanishingly unlikely to occur but this can have ramifications. For this PR we should just omit checking that the year is correct but that we do still have the header.

tools/pre-commit
```diff
@@ -0,0 +1,39 @@
...
+fail=0
+for file in $files; do
+    # Only process .lua files
+    case "$file" in
+        *.lua)
+            # Skip if not a regular file
+            [ -f "$file" ] || continue
+
+            # Check if file is non-empty and does NOT contain the copyright header
+            if [ -s "$file" ] && [[ ! $(grep $file -P -e "-{2,} Copyright \d{4}(?:-\d{4})? SmartThings") ]]; then
```
gharveymn: What if the author isn't SmartThings?

aleclorimer: I should probably constrain it to just the `drivers/SmartThings` directory

tools/pre-commit
```diff
@@ -0,0 +1,39 @@
...
+            # Check if file is non-empty and does NOT end with a newline
+            if [ -s "$file" ] && [ -n "$(tail -c 1 "$file")" ]; then
+                echo "$file: Missing newline at end of file"
+                fail=1
+            fi
+            ;;
+    esac
+done
+
+exit $fail
```
cjswedes: nit: newline

aleclorimer: The irony is palpable for me...

tools/pre-commit
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+#
+# Copyright 2025 SmartThings, Inc.
+# Licensed under the Apache License, Version 2.0
+#
+# pre-commit
+#
+# - Copy to .git/hooks/pre-commit
+# - Ensure executable
```
gharveymn: I would suggest just making sure this is committed as executable to avoid the hassle.

Also I would suggest to use a softlink instead of copying, so that you don't need to re-copy it later (`ln -s ../../tools/pre-commit .git/hooks/pre-commit` (use a relative link so you can move the directory later)).

tools/pre-commit
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+#
+# Copyright 2025 SmartThings, Inc.
+# Licensed under the Apache License, Version 2.0
+#
+# pre-commit
+#
+# - Soft link into .git/hooks/pre-commit
+#   - $ ln -s tools/pre-commit .git/hooks/pre-commit
```
gharveymn: Maybe it's different with your version of `ln`, but usually the first argument is assigned to the symlink verbatim, rather than computed from where you invoked the command. So, it should be `../../tools/pre-commit`.

aleclorimer: You were correct... Good idea and thank you. Guess I needed to brush up on my soft linking knowledge!

PR #2606, opened 2025-12-05
WWSTCERT-9293 eufy FamiLock E40

greens:


drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -75,6 +75,7 @@ local NEW_MATTER_LOCK_PRODUCTS = {
   {0x1533, 0x0010}, -- eufy, FamiLock S3
   {0x1533, 0x0011}, -- eufy, FamiLock E34
   {0x1533, 0x0012}, -- eufy, FamiLock E35
+  {0x1533, 0x0014}, -- eufy, FamiLock E40
```
hcarter-775, line 79: we should wait for this guy to enter the DCL as well, probably

PR #2616, opened 2025-12-10
CHAD-17062: zigbee-bed lazy loading subdrivers

aleclorimer:
Lazy loading of `zigbee-bed` sub-drivers and copyright updates.


drivers/SmartThings/zigbee-bed/src/init.lua
```diff
@@ -20,9 +10,7 @@ local zigbee_bed_template = {
   supported_capabilities = {
     capabilities.refresh,
   },
-  sub_drivers = {
-    require("shus-mattress"),
-  },
+  sub_drivers = require("sub_drivers"),
```
varzac, line 13: ~This looks incorrect~

I see the file now

PR #2621, opened 2025-12-10
CHAD-17067: zigbee-fan lazy load subdrivers

aleclorimer:
Lazy loading of `zigbee-fan` sub-drivers and copyright updates.


drivers/SmartThings/zigbee-fan/src/init.lua
cjswedes, line 1: We can move the `configurations` into the function where it is used, since it is only used at startup. That way it gets gc'd

PR #2646, opened 2025-12-12
CHAD-17092: zwave-sensor lazy loading of sub-drivers

aleclorimer:
`zwave-sensor` lazy loading of sub-drivers and copyright updates


drivers/SmartThings/zwave-sensor/src/apiv6_bugfix/init.lua
greens, line 1: note to self: probably safe to remove this sub-driver (does not need to be done in this PR)

PR #2653, opened 2025-12-12
Matter Switch: Support fan/light devices as parent/child

hcarter-775:
# Description of Change
Add support for fan/light devices as parent/child

# Summary of Completed Tests
Unit tests updated, tested on-device

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -67,7 +67,7 @@ function SwitchLifecycleHandlers.info_changed(driver, device, event, args)
         device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
       )
     elseif device.network_type == device_lib.NETWORK_TYPE_CHILD then
-      switch_utils.update_subscriptions(device:get_parent_device()) -- parent device required to scan through EPs and update subscriptions      
+      switch_utils.update_subscriptions(device:get_parent_device()) -- parent device required to scan through EPs and update subscriptions
     end
   end
 


drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -164,26 +165,31 @@ function utils.find_default_endpoint(device)
...
   end
 
-  -- If both switch and button endpoints are present, check the device type on the main switch
+  -- Return the first momentary switch endpoint as the default endpoint if no onoff endpoints are present
+  if #onoff_ep_ids == 0 and #momentary_switch_ep_ids > 0 then
+    return get_first_non_zero_endpoint(momentary_switch_ep_ids)
+  end
+
+  -- If both onoff and momentary switch endpoints are present, check the device type on the main switch
   -- endpoint. If it is not a supported device type, return the first button endpoint as the
```
nickolas-deboom:
```suggestion
  -- endpoint. If it is not a supported device type, return the first momentary switch endpoint as the
```

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -16,8 +16,59 @@ if version.api < 11 then
...
+function ChildConfiguration.create_child_devices(driver, device, server_cluster_ep_ids, default_endpoint_id, assign_profile_fn)
+  if #server_cluster_ep_ids == 1 and server_cluster_ep_ids[1] == default_endpoint_id then -- no children will be created
+   return
+  end
+
+  table.sort(server_cluster_ep_ids)
+  for device_num, ep_id in ipairs(server_cluster_ep_ids) do
+    if ep_id ~= default_endpoint_id then -- don't create a child device that maps to the main endpoint
+      local label_and_name = string.format("%s %d", device.label, device_num)
+      local child_profile, _ = assign_profile_fn(device, ep_id, true)
```
nickolas-deboom, line 33: Looks like create_child_devices is currently only called with `SwitchDeviceConfiguration.assign_profile_for_onoff_ep` passed in for `assign_profile_fn`. Is the idea to make this generic in case we want to allow other types of child devices?

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -205,6 +227,12 @@ function DeviceConfiguration.match_profile(driver, device)
     end
   end
 
+  local fan_device_type_ep_ids = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.FAN)
+  if #fan_device_type_ep_ids > 0 then
+    updated_profile, optional_component_capabilities = FanDeviceConfiguration.assign_profile_for_fan_ep(device, default_endpoint_id)
+    device:set_field(fields.MODULAR_PROFILE_UPDATED, true)
+  end
```
nickolas-deboom, line 271: For now this is perfectly fine, but it might be worth discussing alternate approaches for the default endpoint logic at some point. Because at this point we already know the default endpoint contains a fan device type so it seems like we shouldn't need the `get_endpoints_by_device_type` call.

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Just wondering if we might want to directly call SwitchDeviceConfiguration.assign_profile_for_onoff_ep here and make it generic in the future if needed

drivers/SmartThings/matter-switch/src/test/test_matter_light_fan.lua
```diff
@@ -66,31 +66,51 @@ local CLUSTER_SUBSCRIBE_LIST ={
   clusters.ColorControl.attributes.CurrentSaturation,
   clusters.ColorControl.attributes.CurrentX,
   clusters.ColorControl.attributes.CurrentY,
-  clusters.ColorControl.attributes.ColorMode,
```
nickolas-deboom, line 69: Why is ColorMode no longer subscribed to for this device?

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
hcarter-775: agreed

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
hcarter-775: I think we could definitely leave it and change it as-needed later (the og idea had the fan as a child device, which is why I had done this) but since we already have a couple cases we know will be using this sooner rather than later, I'd prefer to leave it generic, since it's already done here.

drivers/SmartThings/matter-switch/src/test/test_matter_light_fan.lua
hcarter-775: I see why, and this is an error. `ColorMode` was never added to that child table (which is getting removed elsewhere haha), so the subscription no longer adds it.

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Sounds good!

nickolas-deboom:
One more thing I thought of - wasn't there a switch/fan device out there as well? Should we add support for that in this PR?

drivers/SmartThings/matter-switch/src/test/test_matter_light_fan.lua
nickolas-deboom: Ahh, I see

hcarter-775:
@nickolas-deboom any WWST request would be done in a separate PR in my mind. Also, I believe that manufacturer re-tested the device with this current PR and liked the result, category and all

drivers/SmartThings/matter-switch/src/test/test_matter_light_fan.lua
hcarter-775: fixed!

hcarter-775: had to update some unit tests that weren't checking for color mode, that got caught due to this

nickolas-deboom: Nice, thanks!

PR #2664, opened 2025-12-16
CHAD-17161: lazy loading of matter-sensor sub-drivers

aleclorimer:
Lazy loading of `matter-sensor` sub drivers and copyright updates.


aleclorimer:
Looking to get this in the next driver release if I could get some reviews

PR #2665, opened 2025-12-16
CHAD-17156: Lazy loading of matter-lock sub-drivers

aleclorimer:
Lazy loading of `matter-lock` sub drivers and copyright updates.


drivers/SmartThings/matter-lock/src/new-matter-lock/can_handle.lua
```diff
@@ -0,0 +1,19 @@
+-- Copyright 2025 SmartThings, Inc.
```
nickolas-deboom, line 1: Should probably update these to 2026 for any new files

aleclorimer: Ah... Though I wonder if it is from the time that we put it into main or from the time that the file(s) were created?

cjswedes (approved):

drivers/SmartThings/matter-lock/src/new-matter-lock/can_handle.lua
nickolas-deboom: Hmm, good question. Based on some googling it seems to be a little unclear which one is preferred so I would say what you have now is fine!

PR #2667, opened 2025-12-16
Matter Switch: update subscribe logic

hcarter-775:
# Description of Change
Override generic matter switch handling of device:subscribe() to more properly and generically handle child devices.

# Summary of Completed Tests
Tested multiple parent/child devices where the children had unique subscriptions from the parent device.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -49,6 +49,10 @@ function SwitchLifecycleHandlers.do_configure(driver, device)
   if device.network_type == device_lib.NETWORK_TYPE_MATTER and not switch_utils.detect_bridge(device) then
     switch_cfg.set_device_control_options(device)
     device_cfg.match_profile(driver, device)
+  elseif device.network_type == device_lib.NETWORK_TYPE_CHILD then
+    -- because get_parent_device() may cause race conditions if used in init, an initial child subscribe is handled in doConfigure.
+    -- all future calls to subscribe will be handled by the parent device in init
+    device:subscribe()
   end
 end
 
@@ -68,7 +72,7 @@ function SwitchLifecycleHandlers.info_changed(driver, device, event, args)
         device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH})
       )
     elseif device.network_type == device_lib.NETWORK_TYPE_CHILD then
-      switch_utils.update_subscriptions(device:get_parent_device()) -- parent device required to scan through EPs and update subscriptions
+      device:get_parent_device():subscribe() -- parent device required to send subscription requests
     end
   end
 
@@ -87,7 +91,8 @@ function SwitchLifecycleHandlers.device_init(driver, device)
     if device:get_field(fields.IS_PARENT_CHILD_DEVICE) then
       device:set_find_child(switch_utils.find_child)
     end
-    switch_utils.update_subscriptions(device)
+    device:extend_device("subscribe", switch_utils.subscribe)
+    device:subscribe()
 
     -- device energy reporting must be handled cumulatively, periodically, or by both simulatanously.
     -- To ensure a single source of truth, we only handle a device's periodic reporting if cumulative reporting is not supported.
@@ -290,7 +295,40 @@ local matter_driver_template = {
       [capabilities.valve.commands.open.NAME] = capability_handlers.handle_valve_open,
     },
   },
-  supported_capabilities = fields.supported_capabilities,
+  supported_capabilities = {
+    capabilities.audioMute,
+    capabilities.audioRecording,
+    capabilities.audioVolume,
+    capabilities.battery,
+    capabilities.batteryLevel,
+    capabilities.button,
+    capabilities.cameraPrivacyMode,
+    capabilities.cameraViewportSettings,
+    capabilities.colorControl,
+    capabilities.colorTemperature,
+    capabilities.energyMeter,
+    capabilities.fanMode,
+    capabilities.fanSpeedPercent,
+    capabilities.hdr,
+    capabilities.illuminanceMeasurement,
+    capabilities.imageControl,
+    capabilities.level,
+    capabilities.localMediaStorage,
+    capabilities.mechanicalPanTiltZoom,
+    capabilities.motionSensor,
+    capabilities.nightVision,
+    capabilities.powerMeter,
+    capabilities.powerConsumptionReport,
+    capabilities.relativeHumidityMeasurement,
+    capabilities.sounds,
+    capabilities.switch,
+    capabilities.switchLevel,
+    capabilities.temperatureMeasurement,
+    capabilities.valve,
+    capabilities.videoStreamSettings,
+    capabilities.webrtc,
+    capabilities.zoneManagement
+  },
   sub_drivers = {
     switch_utils.lazy_load_if_possible("sub_drivers.aqara_cube"),
     switch_utils.lazy_load("sub_drivers.camera"),
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
@@ -193,7 +193,8 @@ function CameraUtils.subscribe(device)
       clusters.CameraAvStreamManagement.attributes.StatusLightBrightness
     },
     [capabilities.switch.ID] = {
-      clusters.CameraAvStreamManagement.attributes.StatusLightEnabled
+      clusters.CameraAvStreamManagement.attributes.StatusLightEnabled,
+      clusters.OnOff.attributes.OnOff
     },
     [capabilities.videoStreamSettings.ID] = {
       clusters.CameraAvStreamManagement.attributes.RateDistortionTradeOffPoints,
@@ -223,8 +224,26 @@ function CameraUtils.subscribe(device)
     },
     [capabilities.motionSensor.ID] = {
       clusters.OccupancySensing.attributes.Occupancy
-    }
+    },
+    [capabilities.switchLevel.ID] = {
+      clusters.LevelControl.attributes.CurrentLevel,
+      clusters.LevelControl.attributes.MaxLevel,
+      clusters.LevelControl.attributes.MinLevel,
+    },
+    [capabilities.colorControl.ID] = {
+      clusters.ColorControl.attributes.ColorMode,
+      clusters.ColorControl.attributes.CurrentHue,
+      clusters.ColorControl.attributes.CurrentSaturation,
+      clusters.ColorControl.attributes.CurrentX,
+      clusters.ColorControl.attributes.CurrentY,
+    },
+    [capabilities.colorTemperature.ID] = {
+      clusters.ColorControl.attributes.ColorTemperatureMireds,
+      clusters.ColorControl.attributes.ColorTempPhysicalMaxMireds,
+      clusters.ColorControl.attributes.ColorTempPhysicalMinMireds,
+    },
   }
+
   local camera_subscribed_events = {
     [capabilities.zoneManagement.ID] = {
       clusters.ZoneManagement.events.ZoneTriggered,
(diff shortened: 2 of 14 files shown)


drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -427,4 +427,62 @@ function utils.lazy_load_if_possible(sub_driver_name)
...
 
+--- helper for the switch subscribe override, which adds to a subscribed request for a checked device
+---
+--- @param checked_device any a Matter device object, either a parent or child device, so not necessarily the same as device
+--- @param subscribe_request table a subscribe request that will be appended to as needed for the device
+--- @param capabilities_seen any a list of capabilities that have already been checked by already checked devices
+-- @param matter_elements_seen table a list of attribute and event ids that have been added already
+--- @param subscribed_attributes table key-value pairs mapping capability ids to subscribed attributes
+--- @param subscribed_events table key-value pairs mapping capability ids to subscribed events
+function utils.populate_subscribe_request_for_device(checked_device, subscribe_request, capabilities_seen, attributes_seen, events_seen, subscribed_attributes, subscribed_events)
```
cjswedes, line 445: Nice impl!

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua
```diff
@@ -409,15 +396,15 @@ test.register_message_test(
 test.register_coroutine_test(
   "Test MCD configuration not including switch for unsupported switch device type, create child device instead",
   function()
+    local unsup_mock_device = mock_device_mcd_unsupported_switch_device_type
     -- added sets a bunch of fields on the device, and calls init
     local cluster_subscribe_list = {
-      clusters.OnOff.attributes.OnOff,
+      -- clusters.OnOff.attributes.OnOff,
```
cjswedes: should this be commented out?

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua
```diff
@@ -436,12 +423,23 @@ test.register_coroutine_test(
...
     unsup_mock_device:expect_metadata_update({ profile = "2-button" })
     unsup_mock_device:expect_metadata_update({ provisioning_state = "PROVISIONED" })
+
     test.wait_for_events()
 
     local updated_device_profile = t_utils.get_profile_definition("2-button.yml")
     test.socket.device_lifecycle:__queue_receive(unsup_mock_device:generate_info_changed({ profile = updated_device_profile }))
+
+    local cluster_subscribe_list = {
+      -- clusters.OnOff.attributes.OnOff,
```
cjswedes: Why is this no longer expected?

cjswedes (commented):
Some of the test changes need a bit of explanation, but otherwise LGTM.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -223,8 +224,26 @@ function CameraUtils.subscribe(device)
...
       clusters.OccupancySensing.attributes.Occupancy
-    }
+    },
+    [capabilities.switchLevel.ID] = {
+      clusters.LevelControl.attributes.CurrentLevel,
+      clusters.LevelControl.attributes.MaxLevel,
+      clusters.LevelControl.attributes.MinLevel,
+    },
+    [capabilities.colorControl.ID] = {
+      clusters.ColorControl.attributes.ColorMode,
```
nickolas-deboom, line 234: Why is ColorMode being subscribed to?

hcarter-775: `ColorMode` is in the base subscription list in init.lua. If it should be subscribed to there, it should here as well, no?

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua
hcarter-775: true, this should just be removed altogether here

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: Ah my bad, I forgot we added that subscription.

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua
hcarter-775: This is no longer expected because now, the subscription is based on the capability->attribute/event maps for devices that currently exist.

Before, we were doing a device type check on all endpoints, which was catching that this device would need to   subscribe to onOff eventually, when its child device was created. Now that logic has been pushed to when the child device _is_ actually created (which is not now). Thus the attribute is no longer subscribed to at this point

hcarter-775: Also, I reworked this file a bit to get the structure tested more thoroughly by only initializing the `mock_child` in the tests where it's directly needed. This way, I showed that before the device is created, only some of the subscriptions are handled, but after the device is initialized, the more complete subscribe table is used.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -80,25 +84,7 @@ function SwitchLifecycleHandlers.device_init(driver, device)
...
-             attr ~= clusters.PowerSource.attributes.BatPercentRemaining and
-             attr ~= clusters.PowerSource.attributes.BatChargeLevel then
-            device:add_subscribed_event(attr)
-          else
-            device:add_subscribed_attribute(attr)
-          end
-        end
-      end
-    end
+    device:extend_device("subscribe", switch_utils.subscribe)
```
nickolas-deboom, line 94: Should this call also be used before the subscribe in `init`? can't recall if extend_device persists or not

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -427,4 +427,61 @@ function utils.lazy_load_if_possible(sub_driver_name)
...
+--- helper for the switch subscribe override, which adds to a subscribed request for a checked device
+---
+--- @param checked_device any a Matter device object, either a parent or child device, so not necessarily the same as device
+--- @param subscribe_request table a subscribe request that will be appended to as needed for the device
+--- @param capabilities_seen any a list of capabilities that have already been checked by already checked devices
+--- @param subscribed_attributes table key-value pairs mapping capability ids to subscribed attributes
+--- @param subscribed_events table key-value pairs mapping capability ids to subscribed events
+function utils.populate_subscribe_request_for_device(checked_device, subscribe_request, capabilities_seen, attributes_seen, events_seen, subscribed_attributes, subscribed_events)
+ for _, component in pairs(checked_device.st_store.profile.components) do
+    for _, capability in pairs(component.capabilities) do
```
nickolas-deboom, line 447: It might be a good idea to have null checks for these tables, cuz I'm not sure if they are guaranteed to be populated or not? (Although, I believe this code wouldn't run until after the profile update occurs so it might not matter)

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: `extend_device` does not persist, so yes, we'll have to call this before the init subscribe (as this PR currently does)

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: These tables will never be nil I don't think. If a component has no capabilities on it, it will not be here at all, and all our profiles have at least one component on them (aka `main`), which is required to have at least one capability on it (`refresh`)

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Ah my bad, I didn't have the code expanded in the web gui and thought it was located in info_changed!

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua
```diff
@@ -350,63 +355,45 @@ test.register_coroutine_test(
...
-      }
-    },
-    {
-      channel = "capability",
-      direction = "send",
-      message = mock_child:generate_test_message("main", capabilities.colorTemperature.colorTemperature(1800))
-    },
-  }
+  function()
+    test.mock_device.add_test_device(mock_child)
```
nickolas-deboom, line 361: I'm a little confused on how these tests work. This test for instance looks like it initially subscribed to `CLUSTER_SUBSCRIBE_LIST_NO_CHILD`, and then wait to add the child device until right here. When do the additional attributes from `CLUSTER_SUBSCRIBE_LIST_WITH_CHILD` get subscribed to?

drivers/SmartThings/matter-switch/src/sub_drivers/camera/init.lua
```diff
@@ -49,10 +49,10 @@ end
 function CameraLifecycleHandlers.info_changed(driver, device, event, args)
   if camera_utils.profile_changed(device.profile.components, args.old_st_store.profile.components) then
     camera_cfg.initialize_camera_capabilities(device)
+    device:subscribe()
```
nickolas-deboom, line 52: Just curious but why make this change?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -433,27 +433,63 @@ function utils.lazy_load_if_possible(sub_driver_name)
...
+      if not capabilities_seen[capability.id] then
+        for _, attr in ipairs(subscribed_attributes[capability.id] or {}) do
+          local cluster_id = attr.cluster or attr._cluster.ID
+          local attr_id = attr.ID or attr.attribute
+          if not attributes_seen[cluster_id..attr_id] then
+            local ib = im.InteractionInfoBlock(nil, cluster_id, attr_id)
+            subscribe_request:with_info_block(ib)
+            attributes_seen[cluster_id..attr_id] = true
+          end
         end
```
nickolas-deboom, line 457: nit - might be nice to extract this to a function to reduce repeated code

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button_switch_mcd.lua
hcarter-775: for ease, the main test path doesn't actually test the child subscription logic. That path is tested in the particular test that uses `CLUSTER_SUBSCRIBE_LIST_WITH_CHILD`

drivers/SmartThings/matter-switch/src/sub_drivers/camera/init.lua
hcarter-775: no reason at all, would've changed it back but then I'd have to fix the unit tests....

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -238,56 +257,26 @@ function CameraUtils.subscribe(device)
...
-    end
-  end
-  for capability, event_list in pairs(camera_subscribed_events) do
-    if device:supports_capability_by_id(capability) then
-      for _, event in pairs(event_list) do
-        device:add_subscribed_event(event)
-      end
-    end
-  end
+  local im = require "st.matter.interaction_model"
```
nickolas-deboom, line 260: Could we replace the block below with a call to the `subscribe` function implemented in `switch_utils`?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
hcarter-775: since it's only used twice, won't be used more than twice, and the functions look the same but actually have way more variable differences when you start to look more closely at it, I think I might disagree.

nickolas-deboom: Yeah I suppose there are more variable differences than I saw at first glance

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
hcarter-775: the issue there is how do we add the subscription to `clusters.CameraAvStreamManagement.attributes.AttributeList.ID` if we do it like that? I couldn't really find a way around that at this time, so I left it like this

nickolas-deboom: Ah yeah I see what you mean. I think it's easiest to leave it as-is in that case :+1:

nickolas-deboom (approved):

PR #2673, opened 2025-12-17
Exclude 'off' from fanMode capability for Thermostats

nickolas-deboom:
This change removes Off as a default value from supportedFanModes attribute for the Thermostat device type. Thermostats typically never support turning off the fan, but after switching from `thermostatFanMode` to `fanMode` it became an option.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/thermostat_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-thermostat/src/thermostat_handlers/attribute_handlers.lua
@@ -329,7 +329,7 @@ function AttributeHandlers.fan_mode_handler(driver, device, ib, response)
 end
 
 function AttributeHandlers.fan_mode_sequence_handler(driver, device, ib, response)
-  local supported_fan_modes, supported_fan_modes_attribute, supported_fan_modes_capability
+  local supported_fan_modes, supported_fan_modes_attribute
   if ib.data.value == clusters.FanControl.attributes.FanModeSequence.OFF_LOW_MED_HIGH then
     supported_fan_modes = { "off", "low", "medium", "high" }
   elseif ib.data.value == clusters.FanControl.attributes.FanModeSequence.OFF_LOW_HIGH then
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/matter-thermostat/src/thermostat_handlers/attribute_handlers.lua
```diff
@@ -325,40 +329,58 @@ function AttributeHandlers.fan_mode_handler(driver, device, ib, response)
...
-    supported_fan_modes_attribute = capabilities.airPurifierFanMode.supportedAirPurifierFanModes
+    supported_fan_modes_capability = capabilities.airPurifierFanMode
+    supported_fan_modes_attribute = supported_fan_modes_capability.supportedAirPurifierFanModes
   elseif device:supports_capability_by_id(capabilities.airConditionerFanMode.ID) then
-    supported_fan_modes_attribute = capabilities.airConditionerFanMode.supportedAcFanModes
+    supported_fan_modes_capability = capabilities.airConditionerFanMode
+    supported_fan_modes_attribute = supported_fan_modes_capability.supportedAcFanModes
   elseif device:supports_capability_by_id(capabilities.thermostatFanMode.ID) then
-    supported_fan_modes_attribute = capabilities.thermostatFanMode.supportedThermostatFanModes
+    supported_fan_modes_capability = capabilities.thermostatFanMode
```
hcarter-775: I'm not sure I understand why this variable was introduced. Seems like it might be adding extra complexity without reason?

nickolas-deboom: It was needed for the `get_latest_state` call in the logic down below. As we discussed, some of this logic won't be needed, so this variable can be removed

nickolas-deboom: removed in latest commit

hcarter-775 (approved):

PR #2677, opened 2025-12-29
Matter Switch: Add StatelessStep capability support for SwitchLevel and ColorTemperature

hcarter-775:
# Description of Change
Update all profiles to support statelessStep capabilities as needed. Include handlers for these step capabilities using the Matter-provided step commands.

This breaks apart the work found in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2669 so that the stateless capability support can be reviewed as an isolated unit.

# Summary of Completed Tests
Unit tests added. Tested on-device.

### DIFF
--- a/drivers/SmartThings/matter-switch/profiles/light-color-level.yml
+++ b/drivers/SmartThings/matter-switch/profiles/light-color-level.yml
@@ -10,12 +10,16 @@ components:
       values:
         - key: "level.value"
           range: [1, 100]
+  - id: statelessSwitchLevelStep
+    version: 1
   - id: colorTemperature
     version: 1
     config:
       values:
         - key: "colorTemperature.value"
           range: [ 2200, 6500 ]
+  - id: statelessColorTemperatureStep
+    version: 1
   - id: colorControl
     version: 1
   - id: firmwareUpdate
--- a/drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
@@ -47,6 +47,17 @@ function CapabilityHandlers.handle_switch_set_level(driver, device, cmd)
 end
 
 
+-- [[ STATELESS SWITCH LEVEL STEP CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_step_level(driver, device, cmd)
+  local step_size = st_utils.round((cmd.args and cmd.args.stepSize or 0)/100.0 * 254)
+  if step_size == 0 then return end
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local step_mode = step_size > 0 and clusters.LevelControl.types.StepMode.UP or clusters.LevelControl.types.StepMode.DOWN
+  device:send(clusters.LevelControl.server.commands.Step(device, endpoint_id, step_mode, math.abs(step_size), fields.TRANSITION_TIME_FAST, fields.OPTIONS_MASK, fields.IGNORE_COMMAND_IF_OFF))
+end
+
+
 -- [[ COLOR CONTROL CAPABILITY COMMANDS ]] --
 
 function CapabilityHandlers.handle_set_color(driver, device, cmd)
@@ -56,10 +67,10 @@ function CapabilityHandlers.handle_set_color(driver, device, cmd)
   if switch_utils.tbl_contains(huesat_endpoints, endpoint_id) then
     local hue = switch_utils.convert_huesat_st_to_matter(cmd.args.color.hue)
     local sat = switch_utils.convert_huesat_st_to_matter(cmd.args.color.saturation)
-    req = clusters.ColorControl.server.commands.MoveToHueAndSaturation(device, endpoint_id, hue, sat, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.OPTIONS_OVERRIDE)
+    req = clusters.ColorControl.server.commands.MoveToHueAndSaturation(device, endpoint_id, hue, sat, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.HANDLE_COMMAND_IF_OFF)
   else
     local x, y, _ = st_utils.safe_hsv_to_xy(cmd.args.color.hue, cmd.args.color.saturation)
-    req = clusters.ColorControl.server.commands.MoveToColor(device, endpoint_id, x, y, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.OPTIONS_OVERRIDE)
+    req = clusters.ColorControl.server.commands.MoveToColor(device, endpoint_id, x, y, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.HANDLE_COMMAND_IF_OFF)
   end
   device:send(req)
 end
@@ -69,7 +80,7 @@ function CapabilityHandlers.handle_set_hue(driver, device, cmd)
   local huesat_endpoints = device:get_endpoints(clusters.ColorControl.ID, {feature_bitmap = clusters.ColorControl.FeatureMap.HUE_AND_SATURATION})
   if switch_utils.tbl_contains(huesat_endpoints, endpoint_id) then
     local hue = switch_utils.convert_huesat_st_to_matter(cmd.args.hue)
-    local req = clusters.ColorControl.server.commands.MoveToHue(device, endpoint_id, hue, 0, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.OPTIONS_OVERRIDE)
+    local req = clusters.ColorControl.server.commands.MoveToHue(device, endpoint_id, hue, 0, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.HANDLE_COMMAND_IF_OFF)
     device:send(req)
   else
     device.log.warn("Device does not support huesat features on its color control cluster")
@@ -81,7 +92,7 @@ function CapabilityHandlers.handle_set_saturation(driver, device, cmd)
   local huesat_endpoints = device:get_endpoints(clusters.ColorControl.ID, {feature_bitmap = clusters.ColorControl.FeatureMap.HUE_AND_SATURATION})
   if switch_utils.tbl_contains(huesat_endpoints, endpoint_id) then
     local sat = switch_utils.convert_huesat_st_to_matter(cmd.args.saturation)
-    local req = clusters.ColorControl.server.commands.MoveToSaturation(device, endpoint_id, sat, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.OPTIONS_OVERRIDE)
+    local req = clusters.ColorControl.server.commands.MoveToSaturation(device, endpoint_id, sat, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.HANDLE_COMMAND_IF_OFF)
     device:send(req)
   else
     device.log.warn("Device does not support huesat features on its color control cluster")
@@ -103,9 +114,24 @@ function CapabilityHandlers.handle_set_color_temperature(driver, device, cmd)
   elseif max_temp_kelvin ~= nil and temp_in_kelvin >= max_temp_kelvin then
     temp_in_mired = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MIN, endpoint_id)
   end
-  local req = clusters.ColorControl.server.commands.MoveToColorTemperature(device, endpoint_id, temp_in_mired, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.OPTIONS_OVERRIDE)
+  local req = clusters.ColorControl.server.commands.MoveToColorTemperature(device, endpoint_id, temp_in_mired, fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.HANDLE_COMMAND_IF_OFF)
   device:set_field(fields.MOST_RECENT_TEMP, cmd.args.temperature, {persist = true})
   device:send(req)
+end
+
+
+-- [[ STATELESS COLOR TEMPERATURE STEP CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_step_color_temperature_by_percent(driver, device, cmd)
+  local step_percent_change = cmd.args and cmd.args.stepSize or 0
+  if step_percent_change == 0 then return end
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  -- before the Matter 1.3 lua libs update (HUB FW 55), there was no ColorControl StepModeEnum type defined
+  local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.DOWN or 3) or (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.UP or 1)
+  local min_mireds = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MIN, endpoint_id) or fields.COLOR_TEMPERATURE_MIRED_MIN -- default min mireds
+  local max_mireds = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MAX, endpoint_id) or fields.COLOR_TEMPERATURE_MIRED_MAX -- default max mireds
+  local step_size_in_mireds = st_utils.round((max_mireds - min_mireds) * (math.abs(step_percent_change)/100.0))
+  device:send(clusters.ColorControl.server.commands.StepColorTemperature(device, endpoint_id, step_mode, step_size_in_mireds, fields.TRANSITION_TIME_FAST, min_mireds, max_mireds, fields.OPTIONS_MASK, fields.IGNORE_COMMAND_IF_OFF))
 end
 
 
--- a/drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
@@ -1,6 +1,8 @@
 -- Copyright © 2025 SmartThings, Inc.
 -- Licensed under the Apache License, Version 2.0
 
+local st_utils = require "st.utils"
+
 local SwitchFields = {}
 
 SwitchFields.MOST_RECENT_TEMP = "mostRecentTemp"
@@ -13,8 +15,8 @@ SwitchFields.MIRED_KELVIN_CONVERSION_CONSTANT = 1000000
 -- These values are a "sanity check" to check that values we are getting are reasonable
 local COLOR_TEMPERATURE_KELVIN_MAX = 15000
 local COLOR_TEMPERATURE_KELVIN_MIN = 1000
-SwitchFields.COLOR_TEMPERATURE_MIRED_MAX = SwitchFields.MIRED_KELVIN_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN
-SwitchFields.COLOR_TEMPERATURE_MIRED_MIN = SwitchFields.MIRED_KELVIN_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX
+SwitchFields.COLOR_TEMPERATURE_MIRED_MAX = st_utils.round(SwitchFields.MIRED_KELVIN_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN)
+SwitchFields.COLOR_TEMPERATURE_MIRED_MIN = st_utils.round(SwitchFields.MIRED_KELVIN_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX)
 
 SwitchFields.SWITCH_LEVEL_LIGHTING_MIN = 1
 SwitchFields.CURRENT_HUESAT_ATTR_MIN = 0
@@ -186,10 +188,13 @@ SwitchFields.TEMP_BOUND_RECEIVED = "__temp_bound_received"
 SwitchFields.TEMP_MIN = "__temp_min"
 SwitchFields.TEMP_MAX = "__temp_max"
 
-SwitchFields.TRANSITION_TIME = 0 --1/10ths of a second
--- When sent with a command, these options mask and override bitmaps cause the command
--- to take effect when the switch/light is off.
+SwitchFields.TRANSITION_TIME = 0 -- number of 10ths of a second
+SwitchFields.TRANSITION_TIME_FAST = 3 -- 0.3 seconds
+
+-- For Level/Color Control cluster commands, this field indicates which bits in the OptionsOverride field are valid. In this case, we specify that the ExecuteIfOff option (bit 1) may be overridden.
 SwitchFields.OPTIONS_MASK = 0x01
-SwitchFields.OPTIONS_OVERRIDE = 0x01
+-- the OptionsOverride field's first bit overrides the ExecuteIfOff option, defining whether the command should take effect when the device is off.
+SwitchFields.HANDLE_COMMAND_IF_OFF = 0x01
+SwitchFields.IGNORE_COMMAND_IF_OFF = 0x00
 
 return SwitchFields
--- a/drivers/SmartThings/matter-switch/src/test/test_stateless_step.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_stateless_step.lua
@@ -1,0 +1,163 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local test = require "integration_test"
+local t_utils = require "integration_test.utils"
+local clusters = require "st.matter.clusters"
+
+local mock_device_color_temp = test.mock_device.build_test_matter_device({
+  profile = t_utils.get_profile_definition("light-level-colorTemperature.yml"),
+  manufacturer_info = {
+    vendor_id = 0x0000,
+    product_id = 0x0000,
+  },
+  endpoints = {
+    {
+      endpoint_id = 1,
+      clusters = {
+        {cluster_id = clusters.OnOff.ID, cluster_type = "SERVER"},
+        {cluster_id = clusters.ColorControl.ID, cluster_type = "BOTH", feature_map = 30},
+        {cluster_id = clusters.LevelControl.ID, cluster_type = "SERVER"}
+      },
+      device_types = {
+        {device_type_id = 0x0100, device_type_revision = 1}, -- On/Off Light
+        {device_type_id = 0x010C, device_type_revision = 1} -- Color Temperature Light
+      }
+    }
+  }
+})
+
+local cluster_subscribe_list = {
+  clusters.OnOff.attributes.OnOff,
+  clusters.LevelControl.attributes.CurrentLevel,
+  clusters.LevelControl.attributes.MaxLevel,
+  clusters.LevelControl.attributes.MinLevel,
+  clusters.ColorControl.attributes.ColorTemperatureMireds,
+  clusters.ColorControl.attributes.ColorTempPhysicalMaxMireds,
+  clusters.ColorControl.attributes.ColorTempPhysicalMinMireds,
+}
+
+local function test_init()
+  local subscribe_request = cluster_subscribe_list[1]:subscribe(mock_device_color_temp)
+  for i, cluster in ipairs(cluster_subscribe_list) do
+    if i > 1 then
+      subscribe_request:merge(cluster:subscribe(mock_device_color_temp))
+    end
+  end
+  test.socket.matter:__expect_send({mock_device_color_temp.id, subscribe_request})
+  test.mock_device.add_test_device(mock_device_color_temp)
+end
+test.set_test_init_function(test_init)
+
+local fields = require "switch_utils.fields"
+
+test.register_message_test(
+  "Color Temperature Step Command Test",
+  {
+    {
+      channel = "capability",
+      direction = "receive",
+      message = {
+        mock_device_color_temp.id,
+        { capability = "statelessColorTemperatureStep", component = "main", command = "stepColorTemperatureByPercent", args = { 20 } }
+      }
+    },
+    {
+      channel = "matter",
+      direction = "send",
+      message = {
+        mock_device_color_temp.id,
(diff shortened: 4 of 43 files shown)


drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -109,6 +120,21 @@ function CapabilityHandlers.handle_set_color_temperature(driver, device, cmd)
...
 
+-- [[ STATELESS COLOR TEMPERATURE STEP CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_step_color_temperature_by_percent(driver, device, cmd)
+  local step_percent_change = cmd.args and cmd.args.stepSize or 0
+  if step_percent_change == 0 then return end
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  -- before the Matter 1.3 lua libs update (HUB FW 55), there was no ColorControl StepModeEnum type defined
+  local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.StepModeEnum.DOWN or 3) or (clusters.ColorControl.types.StepModeEnum.UP or 1)
+  local min_mireds = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MIN, endpoint_id) or 2200 -- default min mireds
```
mocelet: The default value for `min_mireds` is `2200` which is a kelvin value instead of mired. Also, mired and kelvin have an inverse relationship so the default minimum mired should match the default maximum kelvin value.

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -109,6 +120,21 @@ function CapabilityHandlers.handle_set_color_temperature(driver, device, cmd)
...
 
 
+-- [[ STATELESS COLOR TEMPERATURE STEP CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_step_color_temperature_by_percent(driver, device, cmd)
+  local step_percent_change = cmd.args and cmd.args.stepSize or 0
+  if step_percent_change == 0 then return end
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  -- before the Matter 1.3 lua libs update (HUB FW 55), there was no ColorControl StepModeEnum type defined
+  local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.StepModeEnum.DOWN or 3) or (clusters.ColorControl.types.StepModeEnum.UP or 1)
```
nickolas-deboom: Given this comment, does this line need to be something like this?
```suggestion
  -- before the Matter 1.3 lua libs update (HUB FW 55), there was no ColorControl StepModeEnum type defined
  local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.DOWN or 3) or (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.UP or 1)
```

drivers/SmartThings/matter-switch/capabilities/statelessColorTemperatureStep.yml
```diff
@@ -0,0 +1,15 @@
+name: Stateless Color Temperature Step
```
nickolas-deboom: These capabilities are available in the capabilities repo - can they be removed here?

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
hcarter-775: good catch :+1:

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
hcarter-775: good catch :+1:

drivers/SmartThings/matter-switch/capabilities/statelessColorTemperatureStep.yml
hcarter-775: yeah, I think so.

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -109,6 +120,21 @@ function CapabilityHandlers.handle_set_color_temperature(driver, device, cmd)
...
+
+function CapabilityHandlers.handle_step_color_temperature_by_percent(driver, device, cmd)
+  local step_percent_change = cmd.args and cmd.args.stepSize or 0
+  if step_percent_change == 0 then return end
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  -- before the Matter 1.3 lua libs update (HUB FW 55), there was no ColorControl StepModeEnum type defined
+  local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.DOWN or 3) or (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.UP or 1)
+  local min_mireds = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MIN, endpoint_id) or fields.COLOR_TEMPERATURE_MIRED_MIN -- default min mireds
+  local max_mireds = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MAX, endpoint_id) or fields.COLOR_TEMPERATURE_MIRED_MAX -- default max mireds
+  local step_size_in_mireds = (max_mireds - min_mireds) * st_utils.round((math.abs(step_percent_change)/100))
```
nickolas-deboom: I think we wouldn't want to round this value right? Since abs(step_percent_change) is between 0 and 100 so this expression would round to 0 or 1

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -47,6 +47,17 @@ function CapabilityHandlers.handle_switch_set_level(driver, device, cmd)
...
 
 
+-- [[ STATELESS SWITCH LEVEL STEP CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_step_level(driver, device, cmd)
+  local step_size = math.floor((cmd.args and cmd.args.stepSize or 0)/100.0 * 254)
+  if step_size == 0 then return end
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local step_mode = step_size > 0 and clusters.LevelControl.types.StepMode.UP or clusters.LevelControl.types.StepMode.DOWN
+  device:send(clusters.LevelControl.server.commands.Step(device, endpoint_id, step_mode, math.abs(step_size), fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.OPTIONS_OVERRIDE))
```
nickolas-deboom: Should `MinLevel` & `MaxLevel` be taken into account here like they are in `handle_step_color_temperature_by_percent`?

mocelet: `Step` command for brightness level has no arguments for min and max level, it stops when reaching the min or the max automatically. Assuming good Matter compliance, of course, I have a Matter-bridged IKEA light with a `Step` bug that will go past 100% and continue from 0% like if it was a circular level instead of linear but, again, a bug.

`StepColorTemperature` command, however, has mandatory arguments for min and max. According to spec both can be set to 0 so the light clips to the min and max supported but pretty much every Matter light I have fails if you use 0 as max so explicitly passing the minimum and maximum like in this PR seems like a better approach to avoid Matter implementation bugs.

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
hcarter-775: Oh shoot, when making some changes last week I dropped the "*100". Will add that back in

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
nickolas-deboom: Ah, sorry for the confusion, I wasn't referring to min and max arguments for the Step command, but rather to scaling the requested percentage step change using the MinLevel and MaxLevel attributes. For example if MinLevel is 50 and MaxLevel is 200, then a percentage change of 10% would be equivalent to `(200-50)*10%=15`, whereas the current implementation would send a step change of `254*10%=25`.

mocelet: Got it! It is an interesting question since the rest of the driver assumes 100% in `switchLevel` capability is 254 in Matter units so in that example (min 50, max 200) the user would see a light that cannot go below 19% or above 78%.

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/9e159e3de19f6820f4c84e3d26aedfe6c89233dd/drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua#L43

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/9e159e3de19f6820f4c84e3d26aedfe6c89233dd/drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua#L45

When being in the minimum (19%), a step of 10% should result in 29% to be intuitive. Even if it is not a 10% of the actual range.

hcarter-775: @nickolas-deboom The way we use `MaxLevel` and `MinLevel` does not lend itself to your suggestion at the moment. We only use those values to gate the displayed range of values for our `switchLevel` capability, see:
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/9e159e3de19f6820f4c84e3d26aedfe6c89233dd/drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua#L75

So actually, these values do not scale anything, and so a 10 step on a device with a more limited range (19-78, for example) should have the same output as a device with a larger range of (0-100), aka a step of 10. The only difference would be that the more limited device would hit its max/min earlier, and so some of these commands would be no-ops.

This distinction can be seen even in the capability command names (stepLevel vs stepColorTemperatureByPercent). The former does not account for any scaling due to our current handling.

mocelet: Is it really an issue though? I find it completely fine to map 100% to 254 always, both in `switchLevel` and the step.

For starters it is stateless so, in case the device reports wrong min-max values or a report was missed for whatever reason, it will not affect the dimming capabilities. Also, if a manufacturer limits the range for any reason I believe it is fine to let the user know.

nickolas-deboom: Ah yeah makes sense to me!

nickolas-deboom (approved):

hcarter-775:
@nickolas-deboom thanks for the catch, got those removed

drivers/SmartThings/matter-switch/profiles/light-color-level.yml
```diff
@@ -6,12 +6,16 @@ components:
     version: 1
   - id: switchLevel
     version: 1
+  - id: statelessSwitchLevelStep
```
tpmanley: This should be after the `config` here and in some of the other profiles

drivers/SmartThings/matter-switch/src/test/test_stateless_step.lua
```diff
@@ -0,0 +1,163 @@
+-- Copyright © 2022 SmartThings, Inc.
```
tpmanley:
```suggestion
-- Copyright © 2026 SmartThings, Inc.
```

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -27,8 +27,8 @@ SwitchFields.MIRED_KELVIN_CONVERSION_CONSTANT = 1000000
 -- These values are a "sanity check" to check that values we are getting are reasonable
 local COLOR_TEMPERATURE_KELVIN_MAX = 15000
 local COLOR_TEMPERATURE_KELVIN_MIN = 1000
-SwitchFields.COLOR_TEMPERATURE_MIRED_MAX = SwitchFields.MIRED_KELVIN_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN
-SwitchFields.COLOR_TEMPERATURE_MIRED_MIN = SwitchFields.MIRED_KELVIN_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX
+SwitchFields.COLOR_TEMPERATURE_MIRED_MAX = math.floor(SwitchFields.MIRED_KELVIN_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN)
```
tpmanley: Why is this changing?

nickolas-deboom: If these default values got used (which normally they would not be), `COLOR_TEMPERATURE_MIRED_MAX` specifically would be a decimal value while commands are expecting ints, which a unit test was complaining about. I should also note that the min/max kelvin values here were arbitrarily chosen, and I mistakenly assumed these conversions would be a clean integer division

hcarter-775: the step function requires that these are numbers, not floats. I could have done the floor-ing in the handler, but since they are just arbitrary constants, I figured why not do it in the base def

drivers/SmartThings/matter-switch/profiles/light-color-level.yml
hcarter-775: good catch!

hcarter-775: fixed

drivers/SmartThings/matter-switch/src/test/test_stateless_step.lua
hcarter-775: fixed

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
hcarter-775: discussed offline, changed to st_utils.round

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -47,6 +47,17 @@ function CapabilityHandlers.handle_switch_set_level(driver, device, cmd)
 end
 
 
+-- [[ STATELESS SWITCH LEVEL STEP CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_step_level(driver, device, cmd)
+  local step_size = math.floor((cmd.args and cmd.args.stepSize or 0)/100.0 * 254)
```
tpmanley: Let's use `round` here like we do in the other level handling functions

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -47,6 +47,17 @@ function CapabilityHandlers.handle_switch_set_level(driver, device, cmd)
...
+  if step_size == 0 then return end
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local step_mode = step_size > 0 and clusters.LevelControl.types.StepMode.UP or clusters.LevelControl.types.StepMode.DOWN
```
tpmanley, line 56: This enum existing before the Matter 1.3 lua libs update (HUB FW 55)?

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -109,6 +120,21 @@ function CapabilityHandlers.handle_set_color_temperature(driver, device, cmd)
...
 
+-- [[ STATELESS COLOR TEMPERATURE STEP CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_step_color_temperature_by_percent(driver, device, cmd)
+  local step_percent_change = cmd.args and cmd.args.stepSize or 0
+  if step_percent_change == 0 then return end
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  -- before the Matter 1.3 lua libs update (HUB FW 55), there was no ColorControl StepModeEnum type defined
+  local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.DOWN or 3) or (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.UP or 1)
+  local min_mireds = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MIN, endpoint_id) or fields.COLOR_TEMPERATURE_MIRED_MIN -- default min mireds
```
tpmanley, line 131: This could be correct but just want to confirm, this should be `COLOR_TEMP_MIN` here and not `COLOR_TEMP_MAX`? The min mired value is the max kelvin value and vice versa

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
hcarter-775: the switchLevel capability handler defined right above this uses `floor` as well. The attribute handlers use round though, apparently.

The break in same behavior occurred here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/b52b3c082f55739bf94e88877b4ba5fa2f5ee214. This is not in beta yet

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
hcarter-775: yes. There are a few places where ColorControl copied LevelControl after the fact, and LevelControl had stuff setup ahead of time

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
hcarter-775: Good thing to check, but pretty sure this is OK. See this logic in the bounds handler:
```
    switch_utils.set_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_KELVIN..minOrMax, ib.endpoint_id, temp_in_kelvin)
    -- the minimum color temp in kelvin corresponds to the maximum temp in mireds
    if minOrMax == fields.COLOR_TEMP_MIN then
      switch_utils.set_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MAX, ib.endpoint_id, temp_in_mired)
    else
      switch_utils.set_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MIN, ib.endpoint_id, temp_in_mired)
    end
```
so we set `fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MIN` as the min value for mireds, and ` fields.COLOR_TEMP_BOUND_RECEIVED_KELVIN..fields.COLOR_TEMP_MIN` as the min value for kelvin

tpmanley: It makes me head hurt so if you've tested and confirmed the right fields are getting set correctly that's good enough for me :)

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
hcarter-775: Made a PR to change the normal switchLevel handler as well: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2690

If this one is merged soon, it can go in alongside the other commit I mentioned for the release

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -47,6 +47,18 @@ function CapabilityHandlers.handle_switch_set_level(driver, device, cmd)
...
 
+-- [[ STATELESS SWITCH LEVEL STEP CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_step_level(driver, device, cmd)
+  local step_size = st_utils.round((cmd.args and cmd.args.stepSize or 0)/100.0 * 254)
+  if step_size == 0 then return end
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local step_mode = step_size > 0 and clusters.LevelControl.types.StepMode.UP or clusters.LevelControl.types.StepMode.DOWN
+  print("level:", step_size)
+  device:send(clusters.LevelControl.server.commands.Step(device, endpoint_id, step_mode, math.abs(step_size), fields.TRANSITION_TIME, fields.OPTIONS_MASK, fields.OPTIONS_OVERRIDE))
```
inasail: @tpmanley @hcarter-775 What about using StepWithOnOff to turn on the light while it's off state?
Do we plan to support the level change during the 'off' state?

and Does transition(0) is best choice for the smooth transition? I wonder this is the best value from the test with real devices.
Now, all have '0' value.
lua_libs\st\matter\defaults\colorControl.lua
local TRANSITION_TIME = 0 --1/10ths of a second
lua_libs\st\matter\defaults\colorTemperature.lua
local TRANSITION_TIME = 0 --1/10ths of a second

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -109,6 +121,22 @@ function CapabilityHandlers.handle_set_color_temperature(driver, device, cmd)
...
+  -- before the Matter 1.3 lua libs update (HUB FW 55), there was no ColorControl StepModeEnum type defined
+  local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.DOWN or 3) or (clusters.ColorControl.types.StepModeEnum and clusters.ColorControl.types.StepModeEnum.UP or 1)
+  local min_mireds = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MIN, endpoint_id) or fields.COLOR_TEMPERATURE_MIRED_MIN -- default min mireds
+  local max_mireds = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_MIRED..fields.COLOR_TEMP_MAX, endpoint_id) or fields.COLOR_TEMPERATURE_MIRED_MAX -- default max mireds
+  local step_size_in_mireds = st_utils.round((max_mireds - min_mireds) * (math.abs(step_percent_change)/100.0))
+  print("percent:", step_size_in_mireds)
+  device:send(clusters.ColorControl.server.commands.StepColorTemperature(device, endpoint_id, step_mode, step_size_in_mireds, fields.TRANSITION_TIME, min_mireds, max_mireds, fields.OPTIONS_MASK, fields.OPTIONS_OVERRIDE))
```
inasail: @hcarter-775 @tpmanley 
Don't we need to turn on the light if the light is off state?

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
mocelet: It really depends on the scenario, using the normal `Step` and a transition time of 0 is the most sensible default.

I have been playing with all those parameters for months with my custom driver for Matter lights and multiple buttons (including the new BILRESA scroll) so I will share the experience in case it is useful.

`StepWithOnOff` is great for dimmers that do not have on/off dedicated buttons since otherwise you would have no way to control the on/off state. It has two main drawbacks though:

- Cannot set the lights to the minimum brightness level. Most of the time you end up with the light turned off because it's virtually impossible to stop at the minimum. Or just impossible if the minimum is 1% and the steps are, let's say, 10%, it will go from 10% to off with the final step.
- If you are controlling multiple lights at the same time, they will all turn on or off. As a user, most of the time I do not want that behaviour and expect to control the brightness of the lights that are on, not turn on everything.

Transition time is a great feature of Matter that should have more presence in the capabilities. `setLevel` of `switchLevel` has it but not the commands to set colours or colour temperature. It is something requested time ago and the main reason for me to use a custom driver.

When it comes to stepping the brightness level, using transitions is tricky:

- Some lights flood the hub with brightness level reports while stepping, especially before Matter 1.4 made quiet reporting mandatory. That makes the hub sluggish for a moment and the user would feel the dimming is not responsive. With some lights a transition time of 1 (tenth of a second) the experience is fine, with others it's better to leave it at 0 since they already apply some default transition anyway.
- If two step commands are sent too close in time and the first one did not end the transition yet, the first step is essentially killed. That's easy to experience with a BILRESA scroll since the steps can happen too fast as you scroll.

For added flexibility, the capability should have optional second and third command arguments for the transition time and the withOnOff behaviour. That would make the default 0 and normal Step if not provided. Maybe it is not late to add them being a proposed capability not yet used elsewhere. 

While unrelated to the steps, adding the transition time argument to the commands to set colour and temperature would be great too, like in `setLevel` of `switchLevel` capability. It should not break anything since the default is 0 anyway.

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
hcarter-775: @inasail We could do this, so it depends how we expect the capability to be used. In my testing, I had one button routine to turn the device on/off, and then I separately used the scroll functionality.

Since it doesn't make sense to turn the device off given the scroll functionality, I wonder if it makes as much sense to turn the device on with this capability?

I also wonder how nice this user experience would be, since the scrolling would change the color temperature the device is on while it turns it on. This seems like it may create an awkward experience. What do you think? cc: @tpmanley

inasail: @hcarter-775 I think we need to consider how often users will actually need to change the switch level and color temperature while the device is turned off. If we don't turn on the light. Users should always turn on by themselves first and rotate the switch. It needs double actions all the time.

But on the other hand, like you said, if there are multiple lights in a group where some are on and others are off(_Even though controlling multiple devices via a knob switch isn't supported yet_)and we may want to adjust the switchLevel and colorTemperatureLevel only for the lights that are already 'on' while keeping the others 'off', then it seems we shouldn't necessarily force them to turn on.

For that 'doing nothing' if the light is 'off' state could be another option. Let's think more.
cc: @tpmanley @greens @varzac

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
inasail: @hcarter-775 Thank you for your detailed explanation.
It could be linked to the below [conversion](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2677/files#r2674413692), so we need decide more beneficial UX for the user.
1) 'turning on' before changing the level
2) keep 'off' state and just change level silently.
3) doing nothing while 'off' state.

and Regarding the 'transition time', what about adding the preference setting in edge driver for this?
We can't expect same result for different models, so we can give an option to the user. (default:'0')
Definitely, we need to support multi-language for the new setting menu.
cc: @tpmanley

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
mocelet: > For that 'doing nothing' if the light is 'off' state could be another option

Alternatively, just send the `Step` like it is now without checking the on/off state and let the light handle it. When the normal `Step` is received, a Matter light that is off will change the level/temperature without turning on.

> Even though controlling multiple devices via a knob switch isn't supported yet

By the way, I use the knob with Rules API to control multiple devices and the popcorn effect using an Aeotec v3 hub is quite noticeable. Hopefully a native handler is on the way like for the absolute brightness level.

Since soon there will be no going back, please consider adding a couple extra arguments to the stateless step capability commands: an integer for the transition time and a boolean for the withOnOff behaviour. There is no one size fits all and, even if the app UI only offers a sensible default with no configuration, having the option to customize the behaviour with a rule is more than welcome.

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
mocelet: > @hcarter-775 Thank you for your detailed explanation

That was my comment actually, I hope I didn't create any confusion! I'm the developer of the Matter Playground custom driver and thought it was useful to share my feedback.

inasail: @mocelet Oh, sorry my mistake. Thank you for sharing your thought and experience.

hcarter-775: @inasail, after some internal discussion, I have updated the implementation to ignore step updates if the device is turned off, but without turning the device "on" if it is off, or vice versa. This decision was made in part from the idea that multiple lights may be connected to the knob, though brightening some should not necessarily turn them all on, if they were previously off. However, in this same vein, I also do not think the off devices should be affected while they are off.
 
 Let me know if you disagree with this implementation.

Also, since this conversation is very similar to the one below, I am going to resolve that conversation for the moment. Feel free to re-open as needed.

mocelet: @hcarter-775 The downside of the Step not executing when the light is off is that the brightness used to turn on cannot be changed and that would lead to undesired "flashes" of the last brightness.

Let's say you last used a light at 100% brightness during the afternoon. But at night you want it at a low brightness. If you can't change the brightness while it is off, it will turn on at 100% which may not be pleasing. If the brightness could be changed before switching on you would be able to lower the brightness, then turn it on and avoid the 100% brightness "flash".

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -172,10 +174,13 @@ SwitchFields.TEMP_BOUND_RECEIVED = "__temp_bound_received"
...
--- When sent with a command, these options mask and override bitmaps cause the command
--- to take effect when the switch/light is off.
+SwitchFields.TRANSITION_TIME = 0 -- number of 10ths of a second
+SwitchFields.TRANSITION_TIME_FAST = 3 -- 0.3 seconds
+
+-- For Level/Color Control cluster commands, this field indicates which bits in the OptionsOverride field are valid. In this case, we specify that the ExecuteIfOff option (bit 1) may be overridden.
 SwitchFields.OPTIONS_MASK = 0x01
-SwitchFields.OPTIONS_OVERRIDE = 0x01
+-- the OptionsOverride field's first bit overrides the ExecuteIfOff option, defining whether the command should take effect when the device is off.
+SwitchFields.HANDLE_COMMAND_IF_OFF = 0x01
```
tpmanley, line 197: I like this, easier to understand

PR #2682, opened 2026-01-02
Support keypair generation for Aliro

HunsupJung:
# Type of Change
          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change
- https://smartthings.atlassian.net/browse/REQ-23925
- https://smartthings.atlassian.net/wiki/spaces/STHK/pages/4707090966/Support+Keypair+generation

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -27,7 +27,6 @@ end
 
 local DoorLock = clusters.DoorLock
 local PowerSource = clusters.PowerSource
-local DoorLockFeatureMapAttr = {ID = 0xFFFC, cluster = DoorLock.ID}
 
 local INITIAL_CREDENTIAL_INDEX = 1
 local ALL_INDEX = 0xFFFE
@@ -98,9 +97,11 @@ local profiling_data = {
   BATTERY_SUPPORT = "__BATTERY_SUPPORT",
 }
 
+local DoorLockFeatureMapAttr = {ID = 0xFFFC, cluster = DoorLock.ID}
 local subscribed_attributes = {
   [capabilities.lock.ID] = {
-    DoorLock.attributes.LockState
+    DoorLock.attributes.LockState,
+    DoorLockFeatureMapAttr
   },
   [capabilities.remoteControlStatus.ID] = {
     DoorLock.attributes.OperatingMode
@@ -195,7 +196,6 @@ local function device_init(driver, device)
       end
     end
   end
-  device:add_subscribed_attribute(DoorLockFeatureMapAttr)
   device:subscribe()
  end
 
--- a/drivers/SmartThings/matter-lock/src/test/test_matter_lock_modular.lua
+++ b/drivers/SmartThings/matter-lock/src/test/test_matter_lock_modular.lua
@@ -15,6 +15,7 @@
 local test = require "integration_test"
 local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
+local cluster_base = require "st.matter.cluster_base"
 local t_utils = require "integration_test.utils"
 local uint32 = require "st.matter.data_types.Uint32"
 
@@ -211,6 +212,7 @@ local mock_device_modular = test.mock_device.build_test_matter_device({
 })
 
 
+local DoorLockFeatureMapAttr = {ID = 0xFFFC, cluster = DoorLock.ID}
 local function test_init()
   test.disable_startup_messages()
   -- subscribe request
@@ -218,6 +220,7 @@ local function test_init()
   subscribe_request:merge(DoorLock.attributes.OperatingMode:subscribe(mock_device))
   subscribe_request:merge(DoorLock.events.LockOperation:subscribe(mock_device))
   subscribe_request:merge(DoorLock.events.DoorLockAlarm:subscribe(mock_device))
+  subscribe_request:merge(cluster_base.subscribe(mock_device, nil, DoorLockFeatureMapAttr.cluster, DoorLockFeatureMapAttr.ID))
   -- add test device
   test.mock_device.add_test_device(mock_device)
   -- actual onboarding flow
@@ -239,6 +242,7 @@ local function test_init_unlatch()
   subscribe_request:merge(DoorLock.attributes.OperatingMode:subscribe(mock_device_unlatch))
   subscribe_request:merge(DoorLock.events.LockOperation:subscribe(mock_device_unlatch))
   subscribe_request:merge(DoorLock.events.DoorLockAlarm:subscribe(mock_device_unlatch))
+  subscribe_request:merge(cluster_base.subscribe(mock_device, nil, DoorLockFeatureMapAttr.cluster, DoorLockFeatureMapAttr.ID))
   -- add test device, handle initial subscribe
   test.mock_device.add_test_device(mock_device_unlatch)
   -- actual onboarding flow
@@ -266,6 +270,7 @@ local function test_init_user_pin()
   subscribe_request:merge(DoorLock.events.LockOperation:subscribe(mock_device_user_pin))
   subscribe_request:merge(DoorLock.events.DoorLockAlarm:subscribe(mock_device_user_pin))
   subscribe_request:merge(DoorLock.events.LockUserChange:subscribe(mock_device_user_pin))
+  subscribe_request:merge(cluster_base.subscribe(mock_device, nil, DoorLockFeatureMapAttr.cluster, DoorLockFeatureMapAttr.ID))
   -- add test device
   test.mock_device.add_test_device(mock_device_user_pin)
   -- actual onboarding flow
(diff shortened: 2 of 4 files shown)


Kwang-Hui:
@tpmanley Could you please review this PR?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -207,10 +215,118 @@ local function device_added(driver, device)
...
+        digital_signature = true
+      },
+      certificate_policies = {
+        critical = true,
+        policy_2030_5_self_signed_client = true
+      }
+    }
+  }
+  local status = security.generate_self_signed_cert(request_opts)
+  local privKey = string.sub(utils.bytes_to_hex_string(status.key_der), PRIV_KEY_START, PRIV_KEY_END)
```
tpmanley: If I'm not mistaken, this won't always work because it doesn't take into account that DER uses variable length integers. I think we may need to parse the data in order to accurately extract the desired components.

HunsupJung: Okay, I will change it to extract the desired components accurately.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -26,14 +27,15 @@ end
 
 local DoorLock = clusters.DoorLock
 local PowerSource = clusters.PowerSource
+local DoorLockFeatureMapAttr = {ID = 0xFFFC, cluster = 0x0101}
```
tpmanley:
```suggestion
local DoorLockFeatureMapAttr = {ID = 0xFFFC, cluster = DoorLock.ID}
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -198,6 +196,9 @@ local function device_init(driver, device)
       end
     end
   end
+  if device.manufacturer_info.vendor_id == 0x135D then
```
tpmanley: We should try to avoid vendor specific checks like this. Any reason we can't just subscribe to the attribute on all devices?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -211,10 +212,155 @@ local function device_added(driver, device)
...
+  local total_len = string.byte(status.key_der, pos)
+  local pubKey = nil
+  local privKey = nil
+  while pos < total_len + ASN1_TOTAL_LEN_POSITION do
+    tag = string.byte(status.key_der, pos + 1)
+    len = string.byte(status.key_der, pos + 2)
+    pos = pos + 2
+    -- Ignore Version(0x02) and Curve parameters field(0xa0)
+    if tag == 0x04 then -- Private key field
+      pubKey = utils.bytes_to_hex_string(string.sub(status.key_der, pos + 1, pos + len))
```
tpmanley: Should this be privKey?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -211,10 +212,155 @@ local function device_added(driver, device)
...
+    }
+  }
+  local status = security.generate_self_signed_cert(request_opts)
+
+  local tag, len
+  local pos = ASN1_TOTAL_LEN_POSITION
+  local total_len = string.byte(status.key_der, pos)
+  local pubKey = nil
+  local privKey = nil
+  while pos < total_len + ASN1_TOTAL_LEN_POSITION do
```
tpmanley: I think the parsing code is a bit fragile still in that it doesn't handle multi-byte lengths. I haven't tried this code below, but I think it handles the multi-byte lengths and should be more resilient to any changes in the response of `generate_self_signed_cert`. Could you take a look, maybe test it out and see what you think?

```
local status = security.generate_self_signed_cert(request_opts)
  if not status or not status.key_der then
    device.log.error("generate_self_signed_cert returned no data")
    return nil, nil
  end

  local der = status.key_der
  local privKey, pubKey = nil, nil

  -- Helper: Parse ASN.1 length (handles 1-byte and multi-byte lengths)
  local function get_length(data, start_pos)
    local b = string.byte(data, start_pos)
    if not b then return nil, start_pos end
    
    if b < 0x80 then
      return b, start_pos + 1
    else
      local num_bytes = b - 0x80
      local len = 0
      for i = 1, num_bytes do
        len = (len * 256) + string.byte(data, start_pos + i)
      end
      return len, start_pos + 1 + num_bytes
    end
  end

  -- Start parsing after the initial SEQUENCE tag (0x30)
  -- Most keys start: [0x30][Length]. We find the first length to find the start of content.
  local _, pos = get_length(der, 2) 

  while pos < #der do
    local tag = string.byte(der, pos)
    local len, content_start = get_length(der, pos + 1)
    if not len then break end

    if tag == 0x02 then 
      -- Version field: Skip it
    elseif tag == 0x04 then 
      -- PRIVATE KEY: Octet String
      privKey = utils.bytes_to_hex_string(string.sub(der, content_start, content_start + len - 1))
    
    elseif tag == 0xA1 then 
      -- PUBLIC KEY Wrapper: Explicit Tag [1]
      -- Inside 0xA1 is a BIT STRING (0x03)
      local inner_tag = string.byte(der, content_start)
      if inner_tag == 0x03 then
        local bit_len, bit_start = get_length(der, content_start + 1)
        
        -- BIT STRINGS have a "leading null byte" (unused bits indicator)
        -- We skip that byte (bit_start) and the 0x04 EC prefix to get the raw X/Y coordinates
        local actual_key_start = bit_start + 2 
        local actual_key_len = bit_len - 2
        
        pubKey = PUB_KEY_PREFIX .. utils.bytes_to_hex_string(string.sub(der, actual_key_start, actual_key_start + actual_key_len - 1))
      end
    end

    -- Move pointer to the next tag
    pos = content_start + len
  end

  if not privKey or not pubKey then
    device.log.error("Failed to extract keys from DER")
  end

  return privKey, pubKey
end
```

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: I coded it because I thought @ctowns wanted it to be applied only to specific devices. If I misunderstood, I will delete it.
https://samsung.slack.com/archives/C063J01GHR6/p1765382793593279?thread_ts=1732583132.684279&cid=C063J01GHR6

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: pubKey and privKey have changed.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: It works well. I will update it~!

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
ctowns: After further discussion, this attribute would be useful for other locks than just this device, so I think we should subscribe to the attribute for all relevant devices by adding it to the subscribed attributes list like this:

```
DoorLockFeatureMapAttr = {ID = 0xFFFC, cluster = 0x0101}
local subscribed_attributes = {
  [capabilities.lock.ID] = {
    DoorLock.attributes.LockState,
    DoorLockFeatureMapAttr -- added here
  },
  [capabilities.remoteControlStatus.ID] = {
    DoorLock.attributes.OperatingMode
  },
  [capabilities.lockUsers.ID] = {
    DoorLock.attributes.NumberOfTotalUsersSupported
  },
    DoorLock.attributes.MinPINCodeLength,
    DoorLock.attributes.RequirePINforRemoteOperation
  },
  [capabilities.lockSchedules.ID] = {
    DoorLock.attributes.NumberOfWeekDaySchedulesSupportedPerUser,
    DoorLock.attributes.NumberOfYearDaySchedulesSupportedPerUser
  },
  [capabilities.lockAliro.ID] = {
    DoorLock.attributes.AliroReaderVerificationKey,
    DoorLock.attributes.AliroReaderGroupIdentifier,
    DoorLock.attributes.AliroReaderGroupSubIdentifier,
    DoorLock.attributes.AliroExpeditedTransactionSupportedProtocolVersions,
    DoorLock.attributes.AliroGroupResolvingKey,
    DoorLock.attributes.AliroSupportedBLEUWBProtocolVersions,
    DoorLock.attributes.AliroBLEAdvertisingVersion,
    DoorLock.attributes.NumberOfAliroCredentialIssuerKeysSupported,
    DoorLock.attributes.NumberOfAliroEndpointKeysSupported,
  },
  [capabilities.battery.ID] = {
    PowerSource.attributes.BatPercentRemaining
  },
  [capabilities.batteryLevel.ID] = {
    PowerSource.attributes.BatChargeLevel
  }
}
```

Sorry for suggesting the other way originally! I think you can remove this vendor-specific check and make the change above so that this attribute will be available for add devices now :+1:

tpmanley:
Thanks for the making the requested changes @HunsupJung . The driver code is looking pretty good to me now. In addition to what @ctowns suggested, can you also take a look at the unit tests failures reported by CI and try to add tests for the new code?

HunsupJung:
@tpmanley 
Thank you for the review. (and I'm sorry for the late reply. I was on vacation.)
I will check the test code and If there is a problem, I will let you know.

PR #2684, opened 2026-01-05
Add support for Irrigation System device type

nickolas-deboom:
This adds support for the Irrigation System device type, introduced with Matter 1.5.

Tested with VDA Irrigation System
[ ](https://smartthings.atlassian.net/browse/CHAD-17520)

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -8,9 +8,9 @@ local clusters = require "st.matter.clusters"
 local log = require "log"
 local version = require "version"
 local cfg = require "switch_utils.device_configuration"
+local button_cfg = cfg.ButtonCfg
 local device_cfg = cfg.DeviceCfg
 local switch_cfg = cfg.SwitchCfg
-local button_cfg = cfg.ButtonCfg
 local fields = require "switch_utils.fields"
 local switch_utils = require "switch_utils.utils"
 local attribute_handlers = require "switch_handlers.attribute_handlers"
@@ -47,6 +47,7 @@ function SwitchLifecycleHandlers.do_configure(driver, device)
     switch_cfg.set_device_control_options(device)
     device_cfg.match_profile(driver, device)
   elseif device.network_type == device_lib.NETWORK_TYPE_CHILD then
+    device_cfg.match_child_profile(driver, device)
     -- because get_parent_device() may cause race conditions if used in init, an initial child subscribe is handled in doConfigure.
     -- all future calls to subscribe will be handled by the parent device in init
     device:subscribe()
@@ -148,6 +149,11 @@ local matter_driver_template = {
         [clusters.FanControl.attributes.FanModeSequence.ID] = attribute_handlers.fan_mode_sequence_handler,
         [clusters.FanControl.attributes.PercentCurrent.ID] = attribute_handlers.percent_current_handler
       },
+      [clusters.FlowMeasurement.ID] = {
+        [clusters.FlowMeasurement.attributes.MeasuredValue.ID] = attribute_handlers.flow_attr_handler,
+        [clusters.FlowMeasurement.attributes.MinMeasuredValue.ID] = attribute_handlers.flow_attr_handler_factory(fields.FLOW_MIN),
+        [clusters.FlowMeasurement.attributes.MaxMeasuredValue.ID] = attribute_handlers.flow_attr_handler_factory(fields.FLOW_MAX)
+      },
       [clusters.IlluminanceMeasurement.ID] = {
         [clusters.IlluminanceMeasurement.attributes.MeasuredValue.ID] = attribute_handlers.illuminance_measured_value_handler
       },
@@ -158,6 +164,11 @@ local matter_driver_template = {
       },
       [clusters.OccupancySensing.ID] = {
         [clusters.OccupancySensing.attributes.Occupancy.ID] = attribute_handlers.occupancy_handler,
+      },
+      [clusters.OperationalState.ID] = {
+        [clusters.OperationalState.attributes.AcceptedCommandList.ID] = attribute_handlers.operational_state_accepted_command_list_attr_handler,
+        [clusters.OperationalState.attributes.OperationalState.ID] = attribute_handlers.operational_state_attr_handler,
+        [clusters.OperationalState.attributes.OperationalError.ID] = attribute_handlers.operational_error_attr_handler
       },
       [clusters.OnOff.ID] = {
         [clusters.OnOff.attributes.OnOff.ID] = attribute_handlers.on_off_attr_handler,
@@ -226,23 +237,33 @@ local matter_driver_template = {
     [capabilities.fanSpeedPercent.ID] = {
       clusters.FanControl.attributes.PercentCurrent
     },
+    [capabilities.flowMeasurement.ID] = {
+      clusters.FlowMeasurement.attributes.MeasuredValue,
+      clusters.FlowMeasurement.attributes.MinMeasuredValue,
+      clusters.FlowMeasurement.attributes.MaxMeasuredValue
+    },
     [capabilities.illuminanceMeasurement.ID] = {
       clusters.IlluminanceMeasurement.attributes.MeasuredValue
-    },
-    [capabilities.motionSensor.ID] = {
-      clusters.OccupancySensing.attributes.Occupancy
     },
     [capabilities.level.ID] = {
       clusters.ValveConfigurationAndControl.attributes.CurrentLevel
     },
-    [capabilities.switch.ID] = {
-      clusters.OnOff.attributes.OnOff
+    [capabilities.motionSensor.ID] = {
+      clusters.OccupancySensing.attributes.Occupancy
+    },
+    [capabilities.operationalState.ID] = {
+      clusters.OperationalState.attributes.AcceptedCommandList,
+      clusters.OperationalState.attributes.OperationalState,
+      clusters.OperationalState.attributes.OperationalError
     },
     [capabilities.powerMeter.ID] = {
       clusters.ElectricalPowerMeasurement.attributes.ActivePower
     },
     [capabilities.relativeHumidityMeasurement.ID] = {
       clusters.RelativeHumidityMeasurement.attributes.MeasuredValue
+    },
+    [capabilities.switch.ID] = {
+      clusters.OnOff.attributes.OnOff
     },
     [capabilities.switchLevel.ID] = {
       clusters.LevelControl.attributes.CurrentLevel,
@@ -287,6 +308,10 @@ local matter_driver_template = {
     [capabilities.level.ID] = {
       [capabilities.level.commands.setLevel.NAME] = capability_handlers.handle_set_level
     },
+    [capabilities.operationalState.ID] = {
+      [capabilities.operationalState.commands.pause.NAME] = capability_handlers.handle_operational_state_pause,
+      [capabilities.operationalState.commands.resume.NAME] = capability_handlers.handle_operational_state_resume
+    },
     [capabilities.statelessColorTemperatureStep.ID] = {
       [capabilities.statelessColorTemperatureStep.commands.stepColorTemperatureByPercent.NAME] = capability_handlers.handle_step_color_temperature_by_percent,
     },
@@ -319,6 +344,7 @@ local matter_driver_template = {
     capabilities.energyMeter,
     capabilities.fanMode,
     capabilities.fanSpeedPercent,
+    capabilities.flowMeasurement,
     capabilities.hdr,
     capabilities.illuminanceMeasurement,
     capabilities.imageControl,
@@ -327,6 +353,7 @@ local matter_driver_template = {
     capabilities.mechanicalPanTiltZoom,
     capabilities.motionSensor,
     capabilities.nightVision,
+    capabilities.operationalState,
     capabilities.powerMeter,
     capabilities.powerConsumptionReport,
     capabilities.relativeHumidityMeasurement,
--- a/drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
@@ -550,4 +550,79 @@ function AttributeHandlers.percent_current_handler(driver, device, ib, response)
   device:emit_event_for_endpoint(ib.endpoint_id, capabilities.fanSpeedPercent.percent(ib.data.value))
 end
 
+
+-- [[ OPERATIONAL STATE CLUSTER ATTRIBUTES ]] --
+
+function AttributeHandlers.operational_state_accepted_command_list_attr_handler(driver, device, ib, response)
+  if ib.data.elements == nil then return end
+  local accepted_command_list = {}
+  for _, accepted_command in ipairs(ib.data.elements) do
+    local accepted_command_id = accepted_command.value
+    if fields.operational_state_command_map[accepted_command_id] ~= nil then
+      table.insert(accepted_command_list, fields.operational_state_command_map[accepted_command_id])
+    end
+  end
+  local event = capabilities.operationalState.supportedCommands(accepted_command_list, {visibility = {displayed = false}})
+  device:emit_event_for_endpoint(ib.endpoint_id, event)
+end
+
+function AttributeHandlers.operational_state_attr_handler(driver, device, ib, response)
+  if ib.data.value == clusters.OperationalState.types.OperationalStateEnum.STOPPED then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.stopped())
+  elseif ib.data.value == clusters.OperationalState.types.OperationalStateEnum.RUNNING then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.running())
+  elseif ib.data.value == clusters.OperationalState.types.OperationalStateEnum.PAUSED then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.paused())
+  end
+end
+
+function AttributeHandlers.operational_error_attr_handler(driver, device, ib, response)
+  if ib.data.elements == nil or ib.data.elements.error_state_id == nil or ib.data.elements.error_state_id.value == nil then return end
+  if version.api < 10 then
+    clusters.OperationalState.types.ErrorStateStruct:augment_type(ib.data)
+  end
+  local operationalError = ib.data.elements.error_state_id.value
+  if operationalError == clusters.OperationalState.types.ErrorStateEnum.UNABLE_TO_START_OR_RESUME then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.unableToStartOrResume())
+  elseif operationalError == clusters.OperationalState.types.ErrorStateEnum.UNABLE_TO_COMPLETE_OPERATION then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.unableToCompleteOperation())
+  elseif operationalError == clusters.OperationalState.types.ErrorStateEnum.COMMAND_INVALID_IN_STATE then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.commandInvalidInCurrentState())
+  end
+end
+
+
+-- [[ FLOW MEASUREMENT CLUSTER ATTRIBUTES ]] --
+
+function AttributeHandlers.flow_attr_handler(driver, device, ib, response)
+  local measured_value = ib.data.value
+  if measured_value ~= nil then
+    local flow = measured_value / 10.0
+    local unit = "m^3/h"
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.flowMeasurement.flow({value = flow, unit = unit}))
+  end
+end
+
+function AttributeHandlers.flow_attr_handler_factory(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value == nil then
+      return
+    end
+    local flow_bound = ib.data.value / 10.0
+    local unit = "m^3/h"
+    switch_utils.set_field_for_endpoint(device, fields.FLOW_BOUND_RECEIVED..minOrMax, ib.endpoint_id, flow_bound)
+    local min = switch_utils.get_field_for_endpoint(device, fields.FLOW_BOUND_RECEIVED..fields.FLOW_MIN, ib.endpoint_id)
+    local max = switch_utils.get_field_for_endpoint(device, fields.FLOW_BOUND_RECEIVED..fields.FLOW_MAX, ib.endpoint_id)
+    if min ~= nil and max ~= nil then
+      if min < max then
+        device:emit_event_for_endpoint(ib.endpoint_id, capabilities.flowMeasurement.flowRange({ value = { minimum = min, maximum = max }, unit = unit }))
+        switch_utils.set_field_for_endpoint(device, fields.FLOW_BOUND_RECEIVED..fields.FLOW_MIN, ib.endpoint_id, nil)
+        switch_utils.set_field_for_endpoint(device, fields.FLOW_BOUND_RECEIVED..fields.FLOW_MAX, ib.endpoint_id, nil)
+      else
+        device.log.warn_with({hub_logs = true}, string.format("Device reported a min flow measurement %d that is not lower than the reported max flow measurement %d", min, max))
+      end
+    end
+  end
+end
+
 return AttributeHandlers
--- a/drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
@@ -232,4 +232,21 @@ function CapabilityHandlers.handle_reset_energy_meter(driver, device, cmd)
   end
 end
 
+
+-- [[ OPERATIONAL STATE CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_operational_state_resume(driver, device, cmd)
+  local endpoint_id = device:get_endpoints(clusters.OperationalState.ID)[1]
+  device:send(clusters.OperationalState.server.commands.Resume(device, endpoint_id))
+  device:send(clusters.OperationalState.attributes.OperationalState:read(device, endpoint_id))
+  device:send(clusters.OperationalState.attributes.OperationalError:read(device, endpoint_id))
+end
+
+function CapabilityHandlers.handle_operational_state_pause(driver, device, cmd)
+  local endpoint_id = device:get_endpoints(clusters.OperationalState.ID)[1]
+  device:send(clusters.OperationalState.server.commands.Pause(device, endpoint_id))
+  device:send(clusters.OperationalState.attributes.OperationalState:read(device, endpoint_id))
+  device:send(clusters.OperationalState.attributes.OperationalError:read(device, endpoint_id))
+end
+
 return CapabilityHandlers
--- a/drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
@@ -20,6 +20,7 @@ local ChildConfiguration = {}
 local SwitchDeviceConfiguration = {}
 local ButtonDeviceConfiguration = {}
 local FanDeviceConfiguration = {}
+local ValveDeviceConfiguration = {}
 
 function ChildConfiguration.create_or_update_child_devices(driver, device, server_cluster_ep_ids, default_endpoint_id, assign_profile_fn)
   if #server_cluster_ep_ids == 1 and server_cluster_ep_ids[1] == default_endpoint_id then -- no children will be created
@@ -30,7 +31,7 @@ function ChildConfiguration.create_or_update_child_devices(driver, device, serve
   for device_num, ep_id in ipairs(server_cluster_ep_ids) do
     if ep_id ~= default_endpoint_id then -- don't create a child device that maps to the main endpoint
       local label_and_name = string.format("%s %d", device.label, device_num)
-      local child_profile, _ = assign_profile_fn(device, ep_id, true)
+      local child_profile, optional_component_capabilities = assign_profile_fn(device, ep_id, true)
       local existing_child_device = device:get_field(fields.IS_PARENT_CHILD_DEVICE) and switch_utils.find_child(device, ep_id)
       if not existing_child_device then
         driver:try_create_device({
@@ -43,7 +44,8 @@ function ChildConfiguration.create_or_update_child_devices(driver, device, serve
         })
       else
         existing_child_device:try_update_metadata({
-          profile = child_profile
+          profile = child_profile,
+          optional_component_capabilities = optional_component_capabilities
         })
       end
     end
@@ -73,7 +75,6 @@ function FanDeviceConfiguration.assign_profile_for_fan_ep(device, server_fan_ep_
   table.insert(optional_supported_component_capabilities, {"main", main_component_capabilities})
   return "fan-modular", optional_supported_component_capabilities
 end
-
 
 function SwitchDeviceConfiguration.assign_profile_for_onoff_ep(device, server_onoff_ep_id, is_child_device)
   local ep_info = switch_utils.get_endpoint_info(device, server_onoff_ep_id)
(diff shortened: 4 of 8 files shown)


drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -299,7 +308,16 @@ SwitchFields.device_type_attribute_map = {
...
+  },
+  [SwitchFields.DEVICE_TYPE_ID.WATER_VALVE] = {
+    clusters.ValveConfigurationAndControl.attributes.CurrentState,
+    clusters.ValveConfigurationAndControl.attributes.CurrentLevel
   }
 }
 
+SwitchFields.operational_state_command_map = {
+  [clusters.OperationalState.commands.Pause.ID] = "pause",
+  [clusters.OperationalState.commands.Resume.ID] = "resume"
```
nickolas-deboom: Note that only pause and resume (and not start or stop) are supported for Irrigation System by the spec

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -156,6 +157,48 @@ function ButtonDeviceConfiguration.configure_buttons(device)
   end
 end
 
+function ValveDeviceConfiguration.assign_profile_for_valve_ep(device, valve_ep_id)
+  local profile = "water-valve"
+
+  for _, ep in ipairs(device.endpoints) do
+    if ep.endpoint_id == valve_ep_id then
```
hcarter-775: we can use something like  `local ep_info = switch_utils.get_endpoint_info(valve_ep_id)` helper to do this a little more easily

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -156,6 +157,48 @@ function ButtonDeviceConfiguration.configure_buttons(device)
...
+      if switch_utils.find_cluster_on_ep(ep, clusters.ValveConfigurationAndControl.ID) then
```
hcarter-775: this helper can take the feature map as a third parameter, might simplify this a bit more

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -174,7 +217,61 @@ function DeviceConfiguration.match_profile(driver, device)
   local default_endpoint_id = switch_utils.find_default_endpoint(device)
   local updated_profile
 
-  if #embedded_cluster_utils.get_endpoints(device, clusters.ValveConfigurationAndControl.ID) > 0 then
+  local is_irrigation_system = false
+  for _, ep in ipairs(device.endpoints) do
+    for _, dt in ipairs(ep.device_types) do
+      if dt.device_type_id == fields.DEVICE_TYPE_ID.IRRIGATION_SYSTEM then
```
hcarter-775: maybe use the `get_endpoints_by_device_type` helper here?

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -174,7 +217,61 @@ function DeviceConfiguration.match_profile(driver, device)
...
+
+        device:set_field(fields.SUPPORTED_COMPONENT_CAPABILITIES, total_supported_capabilities, { persist = true })
+      end
+    else
+      device:try_update_metadata({profile = profile_name})
+    end
+
+    if #valve_ep_ids > 1 then
+      table.remove(valve_ep_ids, 1) -- the first valve ep is accounted for in the main irrigation system profile
+      ValveDeviceConfiguration.create_child_devices(driver, device, valve_ep_ids, default_endpoint_id)
```
hcarter-775: I might move this logic into the create_child_devices function itself- see the create_child_devices for onOff eps to see what I'm suggesting (since we do the same behavior there as well)

hcarter-775: We could also plug this logic within the second check, rather than doing an if-else.

Like:
```
if #valve_ep_ids > 0 then
  ... logic for default ep ...
  if #valve_ep_ids > 1 then
    ... child stuff ...
  end
end
```

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -174,7 +217,61 @@ function DeviceConfiguration.match_profile(driver, device)
...
+      if version.api < 15 or version.rpc < 9 then
+        -- add mandatory capabilities for subscription
+        local total_supported_capabilities = optional_supported_component_capabilities
+        table.insert(total_supported_capabilities[MAIN_COMPONENT_IDX][CAPABILITIES_LIST_IDX], capabilities.refresh.ID)
+        table.insert(total_supported_capabilities[MAIN_COMPONENT_IDX][CAPABILITIES_LIST_IDX], capabilities.firmwareUpdate.ID)
+
+        device:set_field(fields.SUPPORTED_COMPONENT_CAPABILITIES, total_supported_capabilities, { persist = true })
+      end
+    else
+      device:try_update_metadata({profile = profile_name})
```
hcarter-775: I don't think this api should have 2 calls in 2 different places for `try_update_metadata`. I think we should work towards there being a single "path" that this logic follows.

I haven't looked at it super closely, but most of this could/should just be handled within the` if #valve_ep_ids > 1 then` logic block below?

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -174,7 +217,61 @@ function DeviceConfiguration.match_profile(driver, device)
...
+        is_irrigation_system = true
+        break
+      end
+    end
+  end
+
+  local valve_ep_ids = device:get_endpoints(clusters.ValveConfigurationAndControl.ID)
+
+  if is_irrigation_system then
+    updated_profile = "irrigation-system"
```
hcarter-775: I think I'm a little bit unsure why we even have to make an `irrigation-system` profile? Why not just use a single modular water valve profile? This can also be used for the child devices as well?

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -538,4 +538,74 @@ function AttributeHandlers.percent_current_handler(driver, device, ib, response)
...
+
+function AttributeHandlers.flow_attr_handler_factory(minOrMax)
+  return function(driver, device, ib, response)
+    if ib.data.value == nil then
+      return
+    end
+    local flow_bound = ib.data.value / 10.0
+    local unit = "m^3/h"
+    set_field_for_endpoint(device, FLOW_BOUND_RECEIVED..minOrMax, ib.endpoint_id, flow_bound)
+    local min = get_field_for_endpoint(device, FLOW_BOUND_RECEIVED..FLOW_MIN, ib.endpoint_id)
```
hcarter-775: I don't think these fields are actually mapped to anything as they are now, since these are defined in `fields.lua`. Can you add some tests for flow measurement?

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
hcarter-775: Also, if we're adding flow measurement, the spec defines that all water valves might have flow support optionally... may want to add that as well?

This would mean that as far as the profile, the only difference we'd see between irrigation-system and water-valve would be the operational state capability at this time. Since that can be handled modularly,

hcarter-775: However, I see you're using a different category for IrrigationSystem. So I like the idea of having the separate profile. We still may want to add flow to water valve though

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -174,7 +217,61 @@ function DeviceConfiguration.match_profile(driver, device)
...
+
+      -- earlier modular profile gating (min api v14, rpc 8) ensures we are running >= 0.57 FW.
+      -- This gating specifies a workaround required only for 0.57 FW, which is not needed for 0.58 and higher.
+      if version.api < 15 or version.rpc < 9 then
+        -- add mandatory capabilities for subscription
+        local total_supported_capabilities = optional_supported_component_capabilities
+        table.insert(total_supported_capabilities[MAIN_COMPONENT_IDX][CAPABILITIES_LIST_IDX], capabilities.refresh.ID)
+        table.insert(total_supported_capabilities[MAIN_COMPONENT_IDX][CAPABILITIES_LIST_IDX], capabilities.firmwareUpdate.ID)
+
+        device:set_field(fields.SUPPORTED_COMPONENT_CAPABILITIES, total_supported_capabilities, { persist = true })
```
hcarter-775: If we're going to add support for pre-api 15, we'll also have to extend in init and info_changed some logic like this:
```
  if device:get_field(fields.SUPPORTED_COMPONENT_CAPABILITIES) and (version.api < 15 or version.rpc < 9) then
    -- assume that device is using a modular profile on 0.57 FW, override supports_capability_by_id
    -- library function to utilize optional capabilities
    device:extend_device("supports_capability_by_id", aqs_utils.supports_capability_by_id_modular)
  end
```
I'm not totally convinced this is a worthwhile path to take though... but that may just be me

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Done in 07eba5141bab3d4dab3653c551bc303c7106cb90

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Nice! done in 07eba5141bab3d4dab3653c551bc303c7106cb90

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Doen in 07eba5141bab3d4dab3653c551bc303c7106cb90

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Since the default endpoint would be the one with the Irrigation System endpoint, but the parent device would also include the first Valve endpoint, I don't know if the logic in create_child_devices would work, so I implemented the second suggestion you provided. Lemme know how that looks!

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Yeah I agree, and re-did a lot of the logic in match_profile to match the flow for other device types in the latest commit. I think it's a lot more cohesive now, see this commit: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/07eba5141bab3d4dab3653c551bc303c7106cb90

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: I removed this block, because I also don't really think it's worth supporting. Plus, it's not being done for other modular devices such as Fan so it probably makes more sense to keep it consistent

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
nickolas-deboom: Nice catch, I fixed the issues with these fields and added in some test cases!

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: I agree, I'll make a ticket for that

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -542,4 +542,74 @@ function AttributeHandlers.percent_current_handler(driver, device, ib, response)
...
+-- [[ OPERATIONAL STATE CLUSTER ATTRIBUTES ]] --
+
+function AttributeHandlers.operational_state_accepted_command_list_attr_handler(driver, device, ib, response)
+  local accepted_command_list = {}
+  for _, accepted_command in ipairs(ib.data.elements) do
+    local accepted_command_id = accepted_command.value
+    if fields.operational_state_command_map[accepted_command_id] ~= nil then
+      table.insert(accepted_command_list, fields.operational_state_command_map[accepted_command_id])
+    end
+  end
```
hcarter-775, line 564:
```suggestion
  if ib.data.elements == nil then return end
  local accepted_command_list = {}
  for _, accepted_command in ipairs(ib.data.elements) do
    local accepted_command_id = accepted_command.value
    if fields.operational_state_command_map[accepted_command_id] ~= nil then
      table.insert(accepted_command_list, fields.operational_state_command_map[accepted_command_id])
    end
  end
```

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -542,4 +542,74 @@ function AttributeHandlers.percent_current_handler(driver, device, ib, response)
...
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.stopped())
+  elseif ib.data.value == clusters.OperationalState.types.OperationalStateEnum.RUNNING then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.running())
+  elseif ib.data.value == clusters.OperationalState.types.OperationalStateEnum.PAUSED then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.paused())
+  end
+end
+
+function AttributeHandlers.operational_error_attr_handler(driver, device, ib, response)
+  if version.api < 10 then
```
hcarter-775, line 581:
```suggestion
  if ib.data.elements == nil or ib.data.elements.error_state_id.vaue == nil then return end
  if version.api < 10 then
```

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
```diff
@@ -542,4 +542,74 @@ function AttributeHandlers.percent_current_handler(driver, device, ib, response)
...
+  local operationalError = ib.data.elements.error_state_id.value
+  if operationalError == clusters.OperationalState.types.ErrorStateEnum.UNABLE_TO_START_OR_RESUME then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.unableToStartOrResume())
+  elseif operationalError == clusters.OperationalState.types.ErrorStateEnum.UNABLE_TO_COMPLETE_OPERATION then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.unableToCompleteOperation())
+  elseif operationalError == clusters.OperationalState.types.ErrorStateEnum.COMMAND_INVALID_IN_STATE then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.operationalState.operationalState.commandInvalidInCurrentState())
+  end
+end
+
```
hcarter-775, line 593:
```suggestion
-- [[ FLOW MEASUREMENT CLUSTER ATTRIBUTES ]] --
```
kinda like the others

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -38,9 +39,11 @@ SwitchFields.DEVICE_TYPE_ID = {
   ELECTRICAL_SENSOR = 0x0510,
   FAN = 0x002B,
   GENERIC_SWITCH = 0x000F,
+  IRRIGATION_SYSTEM = 0x0040,
   MOUNTED_ON_OFF_CONTROL = 0x010F,
   MOUNTED_DIMMABLE_LOAD_CONTROL = 0x0110,
   ON_OFF_PLUG_IN_UNIT = 0x010A,
+  WATER_VALVE = 0x0042,
```
hcarter-775: this was added in a separate PR

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -90,6 +93,11 @@ SwitchFields.LEVEL_BOUND_RECEIVED = "__level_bound_received"
 SwitchFields.LEVEL_MIN = "__level_min"
 SwitchFields.LEVEL_MAX = "__level_max"
 SwitchFields.COLOR_MODE = "__color_mode"
+SwitchFields.FLOW_BOUND_RECEIVED = "__flow_bound_received"
+SwitchFields.FLOW_MIN = "__flow_min"
+SwitchFields.FLOW_MAX = "__flow_max"
+
+SwitchFields.SUPPORTED_COMPONENT_CAPABILITIES = "__supported_component_capabilities"
```
hcarter-775: remove?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -47,6 +50,11 @@ function SwitchLifecycleHandlers.do_configure(driver, device)
     switch_cfg.set_device_control_options(device)
     device_cfg.match_profile(driver, device)
   elseif device.network_type == device_lib.NETWORK_TYPE_CHILD then
+    local valve_ep_ids = embedded_cluster_utils.get_endpoints(device:get_parent_device(), clusters.ValveConfigurationAndControl.ID)
+    if #valve_ep_ids > 0 then
+      local default_endpoint_id = switch_utils.find_default_endpoint(device:get_parent_device())
+      child_cfg.create_or_update_child_devices(driver, device:get_parent_device(), valve_ep_ids, default_endpoint_id, valve_cfg.assign_profile_for_valve_ep)
+    end
```
hcarter-775: what is going on here? I don't think we'd be creating a child if it is a child? Either way, I think logic like this should be taking place in `match_profile`, no?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -193,6 +193,11 @@ function utils.find_default_endpoint(device)
     end
   end
 
+  local water_valve_eps = utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.WATER_VALVE)
```
hcarter-775: this has also been added elsewhere, note to watch for on rebase

drivers/SmartThings/matter-switch/src/test/test_matter_irrigation_system.lua
```diff
@@ -0,0 +1,426 @@
...
+      clusters.ValveConfigurationAndControl.server.commands.Open(mock_irrigation_system, endpoints.VALVE_1_EP)
+    })
+  end
+)
+
+test.register_coroutine_test(
+  "Parent device: Close command should send the appropriate commands",
+  function()
+    update_device_profile()
+    test.wait_for_events()
```
hcarter-775: Similarly mentioned in the Soil Sensor PR- I don't think we should necessarily be testing the doconfigure flow for all unit tests if it is not explicitly required.

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -209,12 +248,10 @@ function DeviceConfiguration.match_profile(driver, device)
...
-  if #embedded_cluster_utils.get_endpoints(device, clusters.ValveConfigurationAndControl.ID) > 0 then
-    updated_profile = "water-valve"
-    if #embedded_cluster_utils.get_endpoints(device, clusters.ValveConfigurationAndControl.ID,
-      {feature_bitmap = clusters.ValveConfigurationAndControl.types.Feature.LEVEL}) > 0 then
-      updated_profile = updated_profile .. "-level"
-    end
+  local valve_ep_ids = embedded_cluster_utils.get_endpoints(device, clusters.ValveConfigurationAndControl.ID)
+  if #valve_ep_ids > 0 then
+    updated_profile, optional_component_capabilities = ValveDeviceConfiguration.assign_profile_for_valve_ep(device, default_endpoint_id, false)
+    ChildConfiguration.create_or_update_child_devices(driver, device, valve_ep_ids, default_endpoint_id, ValveDeviceConfiguration.assign_profile_for_valve_ep)
```
hcarter-775: Kind of a generic suggestion, I think the flow should be more irrigation system centered. Since water valve and irrigation system are different device types, I don't think we necessarily need to merge them like this.

For example, I'm kind of thinking we should have an `assign_profile_for_irrigation_system_ep` which is distinct from the `valve` one, to keep things more modular.

Then, we can have a check like `if irrigation system endpoints are found, create it with the first water valve stuff tacked on, and try to create all other children afterwards. Else, do basic water valve device creation`

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: The idea here was that the child devices already exist at this point, but `try_update_metadata` has not been called yet for them with the list of enabled optional capabilities. So this causes it to go through that logic again, but this time through the [`try_update_metadata`](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/ad681612df818c2a9592f8c7e58e24339620add1/drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua#L46) path rather than `try_create_device`. Is there a better way of doing this that you know of?

hcarter-775: I see, I think this speaks to my suggestion [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2684#discussion_r3119935365), since if we break down irrigation system and water valve endpoints, we wouldn't even need an update for the children, since the water valves don't have any modular functionality.

However, even if we don't break that down, I don't think the children would need to be updated after the fact in this way, since they don't include any optional capabilities to enable

nickolas-deboom: `level` is an optional capability for the child devices. We could opt to use water-valve and water-valve-level for the children, which I think is part of your suggestion, but in one of our discussions we decided to have both the parent and child devices use the same irrigation system profile so that they all have the IrrigationSystem category and icon. With that in mind, should we still create a separate `assign_profile_for_irrigation_system_ep`?

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
nickolas-deboom: Done in latest force push

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
nickolas-deboom: Thanks, fixed in rebase in last force push

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
nickolas-deboom: Thanks, updated

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
nickolas-deboom: Done in a5d4748f0417393a8b3713d78d5567162beb058d

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
nickolas-deboom: Done in a5d4748f0417393a8b3713d78d5567162beb058d

drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
nickolas-deboom: Done in a5d4748f0417393a8b3713d78d5567162beb058d

drivers/SmartThings/matter-switch/src/test/test_matter_irrigation_system.lua
nickolas-deboom: I cleaned it up a bit and removed the profiling update call for all tests that don't need it, see 84eb7ecf860249028ab6dc98895061749408ded0

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I ended up implemented your idea for a separate `assign_profile_for_irrigation_system_ep` in 6ea6e0abb7fac4471394999556a2458fd8f5b40b. I think we still may need the update call here, if we want the child devices to continue using `irrigation-system.yml` like i mentioned

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Done in 6ea6e0abb7fac4471394999556a2458fd8f5b40b, let me know if that looks alright to you

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -230,12 +275,16 @@ function DeviceConfiguration.match_profile(driver, device)
...
+  local irrigation_system_ep_ids = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.IRRIGATION_SYSTEM)
+  local valve_ep_ids = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.WATER_VALVE)
+  if #irrigation_system_ep_ids > 0 then
+    updated_profile, optional_component_capabilities = ValveDeviceConfiguration.assign_profile_for_irrigation_system_ep(device, irrigation_system_ep_ids[1], false)
+    if #valve_ep_ids > 0 then
+      ChildConfiguration.create_or_update_child_devices(driver, device, valve_ep_ids, default_endpoint_id, ValveDeviceConfiguration.assign_profile_for_irrigation_system_ep)
     end
+  elseif #valve_ep_ids > 0 then
+    updated_profile = ValveDeviceConfiguration.assign_profile_for_valve_ep(device, default_endpoint_id, false)
+    ChildConfiguration.create_or_update_child_devices(driver, device, valve_ep_ids, default_endpoint_id, ValveDeviceConfiguration.assign_profile_for_valve_ep)
```
hcarter-775: I don't think we should add child water valves in this PR, if only to keep the scope only to irrigation system support

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -230,12 +275,16 @@ function DeviceConfiguration.match_profile(driver, device)
...
-  if #switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.WATER_VALVE) > 0 then
-    updated_profile = "water-valve"
-    if #embedded_cluster_utils.get_endpoints(device, clusters.ValveConfigurationAndControl.ID,
-      {feature_bitmap = clusters.ValveConfigurationAndControl.types.Feature.LEVEL}) > 0 then
-      updated_profile = updated_profile .. "-level"
+  local irrigation_system_ep_ids = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.IRRIGATION_SYSTEM)
+  local valve_ep_ids = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.WATER_VALVE)
+  if #irrigation_system_ep_ids > 0 then
+    updated_profile, optional_component_capabilities = ValveDeviceConfiguration.assign_profile_for_irrigation_system_ep(device, irrigation_system_ep_ids[1], false)
+    if #valve_ep_ids > 0 then
```
hcarter-775: technically we shouldn't need this check, since per spec there will be at least 1 water valve endpoint in an irrigation system

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -190,6 +190,51 @@ function ButtonDeviceConfiguration.configure_buttons(device, momentary_switch_ep
...
+function ValveDeviceConfiguration.assign_profile_for_irrigation_system_ep(device, irrigation_system_ep_id, is_child_device)
+  local main_component_capabilities = {}
+
+  local valve_ep_ids = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.WATER_VALVE)
+  if #valve_ep_ids > 0 then
+    local supports_level = switch_utils.find_cluster_on_ep(
+      switch_utils.get_endpoint_info(device, valve_ep_ids[1]),
+      clusters.ValveConfigurationAndControl.ID,
+      {feature_bitmap = clusters.ValveConfigurationAndControl.types.Feature.LEVEL}
+    )
```
hcarter-775: I don't think we should necessarily check the first valve endpoint for the LEVEL feature and assign that to all devices. I think this makes sense for the actual irrigation system ep, but the children should just check their own endpoint configurations. Also, to this end we may want to sort the eps first for the irrigation system ep to ensure we're mapping the right one

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Ok yeah, that makes sense- I'd forgotten about the idea that we wanted to use the irrigation category for the children, that does change things.

But looking at it again, I kinda think we can just use device_cfg.match_profile() again here, since if you look at the [get_endpoints](https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/main/lua_libs/st/matter/child.lua#L41) def for children, it will limit the scope of what's checked to literally just the endpoint we care about, and so everything should theoretically just go smoothly.

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -222,4 +222,21 @@ function CapabilityHandlers.handle_reset_energy_meter(driver, device, cmd)
   end
 end
 
+
+-- [[ OPERATIONAL STATE CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.handle_operational_state_resume(driver, device, cmd)
+  local endpoint_id = device:component_to_endpoint(cmd.component)
```
hcarter-775: I don't think this will work as expected- I don't think there are any unit test for these capabilities, so I think we should add some, and I think by adding one we will catch that the endpoint id returned here will be for the first water valve ep, not the first irrigation system ep.

This makes me wonder, we may need to use a COMPONENT_TO_ENDPOINT map here to be able to map both the water valve _and_ the irrigation system eps to the main component. Maybe not though.

Same issue for the pause state

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: ah yeah good call

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: restored in 9de1938386e2ce57c526fd3cce4e473cd9a1c9ef

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
nickolas-deboom: What do you think about this, implemented in 6fdbef7b2cdca5f5b29f0e38375b8addf8bb732c
```
local endpoint_id = device:get_endpoints(clusters.OperationalState.ID)[1]
```
Since the presence operationalState is gated by the existence of this cluster we shouldn't need any other checks here I think.

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Yeah good point. in 59cd5969b76e6363f5ab69ac0fef81970477cd13 I added the sort and changed it so that it will use the correct endpoint depending on whether it's the child or parent device

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: I played around with that, but I think the issue is that since the scope of the get_endpoints check would be limited to just the child device ep, match_profile won't detect the irrigation system ep and will just go through the standalone valve logic instead and assign the 'water-valve' profile to the child device. So I think that it might still make sense to call `create_or_update_child_devices` here in do_configure - do you agree?

Also, I realized I was using the wrong assign_profile function here - it was using `assign_profile_for_valve_ep` when it should have been using `assign_profile_for_irrigation_system_ep`. I also just removed `assign_profile_for_valve_ep` since it's not needed

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
hcarter-775: works for me, we have some precedence there via the fan control handlers.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Hm, yeah that's true. Then I might suggest we create a `ChildConfiguration.update_profile` function or something similar here in `device_configuration.lua`, and plop some of this logic in there. First, I think this will end up being cleaner and clearer for the future, but also, the current implementation tries to re-update every child device, which I don't think we want, so I might suggest only inputting the actual endpoint we care about in that create_or_update children function. You could maybe use the [get_endpoint](https://github.ecodesamsung.com/iot-hub/scripting-engine/blob/main/lua_libs/st/matter/child.lua#L28) function and input like { device:get_endpoint() } as the input. Also, to cut down on calls we could probably just put nil for the default endpoint id, since it's not really useful in this context

nickolas-deboom: Rather than using device:get_endpoint I think we can reference `device.parent_assigned_child_key`, as in: f08c21714dfd06a7bf690e62a0a07ef566e9abda

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -190,9 +191,57 @@ function ButtonDeviceConfiguration.configure_buttons(device, momentary_switch_ep
...
 -- [[ PROFILE MATCHING AND CONFIGURATIONS ]] --
 
+function DeviceConfiguration.match_child_profile(driver, device)
+  local parent_device = device:get_parent_device()
+  local irrigation_system_ep_ids = switch_utils.get_endpoints_by_device_type(
+    parent_device,
+    fields.DEVICE_TYPE_ID.IRRIGATION_SYSTEM
+  )
+  if #irrigation_system_ep_ids > 0 then
+    local child_ep_id = device.parent_assigned_child_key
```
hcarter-775: I think this is a string as-is, so we have to convert that to a number. Let's also add a unit test for this flow to catch this in the future
To wrap this we may want to use the get_endpoint() function for clarity

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -255,7 +309,11 @@ function DeviceConfiguration.match_profile(driver, device)
...
 return {
+  ButtonCfg = ButtonDeviceConfiguration,
+  ChildCfg = ChildConfiguration,
   DeviceCfg = DeviceConfiguration,
   SwitchCfg = SwitchDeviceConfiguration,
-  ButtonCfg = ButtonDeviceConfiguration
+  ValveCfg = ValveDeviceConfiguration
 }
+
+
```
hcarter-775: nit: remove extra newline

nickolas-deboom: fixed

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Switched to use get_endpoint and added a UT in 759e8f83901c9a2abcd093c6556f8835e1cec509

hcarter-775 (approved):

ctowns (approved):

PR #2686, opened 2026-01-06
Matter Camera: Only update profile if capabilities have changed

nickolas-deboom:
The `optional_capabilities_list_changed` function was not properly comparing the new set of supported capabilities with the current set, meaning that a profile update was occurring each time `camera_av_stream_management_attribute_list_handler` runs. This commit fixes `optional_capabilities_list_changed` and provides new test cases to ensure that the extraneous profile updates no longer occur.

Tested with Matter SDK Camera example.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
@@ -121,8 +121,6 @@ function CameraDeviceConfiguration.match_profile(device, status_light_enabled_pr
   local doorbell_endpoints = switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.DOORBELL)
   if #doorbell_endpoints > 0 then
     table.insert(doorbell_component_capabilities, capabilities.button.ID)
-    CameraDeviceConfiguration.update_doorbell_component_map(device, doorbell_endpoints[1])
-    button_cfg.configure_buttons(device, device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH}))
   end
   if status_light_enabled_present then
     table.insert(status_led_component_capabilities, capabilities.switch.ID)
@@ -147,6 +145,10 @@ function CameraDeviceConfiguration.match_profile(device, status_light_enabled_pr
 
   if camera_utils.optional_capabilities_list_changed(optional_supported_component_capabilities, device.profile.components) then
     device:try_update_metadata({profile = "camera", optional_component_capabilities = optional_supported_component_capabilities})
+    if #doorbell_endpoints > 0 then
+      CameraDeviceConfiguration.update_doorbell_component_map(device, doorbell_endpoints[1])
+      button_cfg.configure_buttons(device, device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH}))
+    end
   end
 end
 
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
@@ -120,28 +120,49 @@ function CameraUtils.profile_changed(synced_components, prev_components)
   return false
 end
 
-function CameraUtils.optional_capabilities_list_changed(optional_capabilities, prev_component_list)
-  local prev_optional_capabilities = {}
-  for idx, comp in pairs(prev_component_list or {}) do
-    local cap_list = {}
-    for _, capability in pairs(comp.capabilities or {}) do
-      table.insert(cap_list, capability.id)
+function CameraUtils.optional_capabilities_list_changed(new_component_capability_list, previous_component_capability_list)
+  local previous_capability_map = {}
+  local component_sizes = {}
+
+  local previous_component_count = 0
+  for component_name, component in pairs(previous_component_capability_list or {}) do
+    previous_capability_map[component_name] = {}
+    component_sizes[component_name] = 0
+    for _, capability in pairs(component.capabilities or {}) do
+      if capability.id ~= "firmwareUpdate" and capability.id ~= "refresh" then
+        previous_capability_map[component_name][capability.id] = true
+        component_sizes[component_name] = component_sizes[component_name] + 1
+      end
     end
-    table.insert(prev_optional_capabilities, {idx, cap_list})
+    previous_component_count = previous_component_count + 1
   end
-  if #optional_capabilities ~= #prev_optional_capabilities then
+
+  local number_of_components_counted = 0
+  for _, new_component_capabilities in pairs(new_component_capability_list or {}) do
+    local component_name = new_component_capabilities[1]
+    local capability_list = new_component_capabilities[2]
+
+    number_of_components_counted = number_of_components_counted + 1
+
+    if previous_capability_map[component_name] == nil then
+      return true
+    end
+
+    for _, capability in ipairs(capability_list) do
+      if previous_capability_map[component_name][capability] == nil then
+        return true
+      end
+    end
+
+    if #capability_list ~= component_sizes[component_name] then
+      return true
+    end
+  end
+
+  if number_of_components_counted ~= previous_component_count then
     return true
   end
-  for _, capability in pairs(optional_capabilities or {}) do
-    if not switch_utils.tbl_contains(prev_optional_capabilities, capability) then
-      return true
-    end
-  end
-  for _, capability in pairs(prev_optional_capabilities or {}) do
-    if not switch_utils.tbl_contains(optional_capabilities, capability) then
-      return true
-    end
-  end
+
   return false
 end
 
--- a/drivers/SmartThings/matter-switch/src/test/test_matter_camera.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_matter_camera.lua
@@ -5,6 +5,7 @@ local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
 local t_utils = require "integration_test.utils"
 local test = require "integration_test"
+local uint32 = require "st.matter.data_types.Uint32"
 
 test.disable_startup_messages()
 
@@ -284,7 +285,6 @@ local expected_metadata = {
 }
 
 local function update_device_profile()
-  local uint32 = require "st.matter.data_types.Uint32"
   test.socket.matter:__queue_receive({
     mock_device.id,
     clusters.CameraAvStreamManagement.attributes.AttributeList:build_test_report_data(mock_device, CAMERA_EP, {
@@ -292,8 +292,9 @@ local function update_device_profile()
       uint32(clusters.CameraAvStreamManagement.attributes.StatusLightBrightness.ID)
     })
   })
-  test.socket.matter:__expect_send({mock_device.id, clusters.Switch.attributes.MultiPressMax:read(mock_device, DOORBELL_EP)})
   mock_device:expect_metadata_update(expected_metadata)
+  test.socket.matter:__expect_send({mock_device.id, clusters.Switch.attributes.MultiPressMax:read(mock_device, DOORBELL_EP)})
+  test.wait_for_events()
   local updated_device_profile = t_utils.get_profile_definition(
     "camera.yml", {enabled_optional_capabilities = expected_metadata.optional_component_capabilities}
   )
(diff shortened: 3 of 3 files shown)


drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -145,6 +143,10 @@ function CameraDeviceConfiguration.match_profile(device, status_light_enabled_pr
 
   if camera_utils.optional_capabilities_list_changed(optional_supported_component_capabilities, device.profile.components) then
     device:try_update_metadata({profile = "camera", optional_component_capabilities = optional_supported_component_capabilities})
+    if #doorbell_endpoints > 0 then
```
nickolas-deboom, line 148: button config can be put behind the `optional_capabilities_list_changed` gate so it doesn't run each time `CameraAvStreamManagement.AttributeList` is received.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -121,27 +121,69 @@ function CameraUtils.profile_changed(synced_components, prev_components)
...
+  -- Build a normalized structure from prev_component_list: {component_name = {cap1, cap2, ...}}
   local prev_optional_capabilities = {}
-  for idx, comp in pairs(prev_component_list or {}) do
+  for comp_name, comp in pairs(prev_component_list or {}) do
     local cap_list = {}
     for _, capability in pairs(comp.capabilities or {}) do
       table.insert(cap_list, capability.id)
     end
-    table.insert(prev_optional_capabilities, {idx, cap_list})
+    table.sort(cap_list)
```
hcarter-775: why are we sorting on each iteration?

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -121,27 +121,69 @@ function CameraUtils.profile_changed(synced_components, prev_components)
 end
 
 function CameraUtils.optional_capabilities_list_changed(optional_capabilities, prev_component_list)
```
hcarter-775: can we make the parameter names for this function more similar?

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -121,27 +121,69 @@ function CameraUtils.profile_changed(synced_components, prev_components)
...
+  local prev_count = 0
+  for _ in pairs(prev_optional_capabilities) do
+    prev_count = prev_count + 1
+  end
+  local new_count = 0
+  for _ in pairs(new_optional_capabilities) do
+    new_count = new_count + 1
+  end
+  if prev_count ~= new_count then
     return true
```
hcarter-775: I'd cut all this, right? why check the number if we're just looping through everything again anyway?

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -121,27 +121,69 @@ function CameraUtils.profile_changed(synced_components, prev_components)
...
-  for _, capability in pairs(prev_optional_capabilities or {}) do
-    if not switch_utils.tbl_contains(optional_capabilities, capability) then
+
+  -- Check if all components in prev exist in new_optional_capabilities
+  for comp_name, _ in pairs(prev_optional_capabilities) do
+    if new_optional_capabilities[comp_name] == nil then
       return true
     end
   end
+
```
hcarter-775: It might make sense to consolidate this... like create a size count when first initializing one of these "normalized tables", and then when checking one of these paths, make sure the size of the other was fully checked. If not, this means the other was bigger/smaller than the other and thus not everything is the same.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -121,27 +121,69 @@ function CameraUtils.profile_changed(synced_components, prev_components)
...
+    for _, cap_name in pairs(component_entry[2] or {}) do
+      table.insert(cap_list, cap_name)
+    end
+    if comp_name == "main" then
+      table.insert(cap_list, "firmwareUpdate")
+      table.insert(cap_list, "refresh")
+    end
+    table.sort(cap_list)
+    new_optional_capabilities[comp_name] = cap_list
+  end
```
hcarter-775: I don't think we necessarily need to normalize both tables explicitly. I like the idea of doing it for one and then having that nil table check, though. That is a nice trick 👍

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: Yeah, looking at it again this isn't needed

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: When I was testing this it appeared to be needed, even though it pretty much puts the new list of capabilities in the exact same format it was already in. I'll take another look because I believe you're right that this isn't needed

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: Oops, yeah that's leftover from testing and not needed

nickolas-deboom: removed in latest commit

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: done in latest commit

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: removed in latest commit

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: In the latest commit I decided to format the components and capabilities from the profile from the device object into the same format as our optional component/capabilities list that gets formed in match_profile. The downside of this is that we can't index into these lists as easily and instead have to iterate through them. So basically the new algorithm is:

1. Reformat `previous_component_capability_list`
2. Iterate through each list and check for any components and capabilities that don't exist in the other list.

I think that overall it's easier to follow now but let me know what you think!

hcarter-775: Hm, I actually think I like the first version better, since this significantly cuts down checks ( O(n) vs O(n^2), give or take). At the same time, the lists we're looking through aren't all that big I suppose. 

Still, I may suggest changing the formatting back to creating an explicit table and then just checking whether the key is nil or not.

Also, though the code is much cleaner now with your helper function (nice) this still includes a lot of double checking, since we check in one direction and then the other. Cutting the n^2 down to n will help with this, but I think we can go further. Specifically, I think we can do something like this:

```
  local component_sizes  = {} 
  for component_name, component in pairs(previous_component_capability_list or {}) do
      component_sizes[component_name] = 0
      for _, capability in pairs(component.capabilities or {}) do
          ... make the table that we can key off of later ...
          component_sizes[component_name] = component_sizes[component_name] + 1 -- also, increment the number of capabilities found 
      end
  end
  
  local number_of_components_counted = 0
  for component_name, component in pairs(new_component_capability_list or {}) do
    number_of_components_counted = number_of_components_counted + 1 -- see how many components are in this new list
    local capabilities_counted_for_component = 0
    for _, capability in pairs(component.capabilities or {}) do
        .... do the checking of keys ....
        capabilities_counted_for_component = capabilities_counted_for_component + 1
    end
    if capabilities_counted_for_component ~= #component_sizes[component_name] then return false end -- make sure the number of capabilities just checked matches the number of capabilities in this list. If it doesn't, these aren't the same.
  end
  if number_of_components_counted ~= #component_sizes then return false end -- same idea as above for caps on a comp, but for the components.
```
I think something like this can keep the total looping down to once per list, by using a hash map + a use of the comp/cap lengths

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
hcarter-775: responded to this idea over [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2686#discussion_r2677215789)

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: The algorithm you suggested is quite nice, however I think we need to check the actual capability IDs rather than just the number of capabilities per component, right? (and maybe we can do that with the code you suggested?). However, given the small size of the lists as well as the infrequency that this code will run, the difference in the computation complexity seems pretty negligible and my preference leans toward the current implementation, since it feels a bit simpler and takes the cap ID's into account. Although, when I think about it I kind of doubt that the capability list would change but stay at the same number of capabilities, so either implementation is probably just fine.

hcarter-775: @nickolas-deboom , this idea does take capability ids into account.  I have that part in the little comment here `        .... do the checking of keys ....` where I meant check the capability we're looking at now exists in the table we just made, aka the key. So the primary thing it does is check capabilities. 

The added thing this version does is that it also checks sizes to make sure everything was actually checked, so we don't have to check both sides explicitly.

nickolas-deboom: Ah, thank you! I was misreading it. I will go ahead and implement this change in that case.

nickolas-deboom: Ok, made this change in 784195cfce13c3c1f185b617569498b27de9945f. Lemme know if that looks ok to you!

hcarter-775: Thanks for making these changes Nick! Looks really nice, gonna take a last pass at it

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -120,28 +120,51 @@ function CameraUtils.profile_changed(synced_components, prev_components)
...
   end
-  for _, capability in pairs(optional_capabilities or {}) do
-    if not switch_utils.tbl_contains(prev_optional_capabilities, capability) then
+
+  local number_of_components_counted = 0
+  for _, new_component_capabilities in pairs(new_component_capability_list or {}) do
+    local component_name = new_component_capabilities[1]
+    local capability_list = new_component_capabilities[2]
+
+    number_of_components_counted = number_of_components_counted + 1
```
hcarter-775, line 145: I think, since this list is actually just a simple table that we know the structure of, we can just do
```
local new_list_num_components = #new_component_capability_list
```
and similar for the capabilities:
```
local new_list_num_capabilities_in_component = #capability_list
```

hcarter-775: or maybe we don't even need to define them as variables in this case, and just call this directly as needed

nickolas-deboom: I tried that but apparently it doesn't work for keyed tables. I didn't try very hard tho, so lemme see if it'll work

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -120,28 +120,51 @@ function CameraUtils.profile_changed(synced_components, prev_components)
...
+
+    if not previous_capability_map[component_name] then
```
hcarter-775: nit but can we do ` if <something> == nil then` in these cases? Idk, since we're talking about whether these were found this structuring seems more natural? Just throwing it out there, what do you think

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
hcarter-775: ah my bad, that's true. I forgot about that.

hcarter-775: actually, ins't the `capability_list` a basic table? That one might work

nickolas-deboom: Ah yeah good call, updated in b5204a5599ddca8d7992e0595126c6c6b366530e

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
nickolas-deboom: Sure thing, updated in b5204a5599ddca8d7992e0595126c6c6b366530e

hcarter-775 (approved):

PR #2689, opened 2026-01-07
Matter Switch: Add new RGBW profile with a fixed W

hcarter-775:
# Description of Change
Add new profile, `light-color-level-noTemp`, to handle devices supporting color but no temperature. Adds new default logic, gating on the ColorTemeperature Feature, to define this.

To make this work, I had to work around the current deprecated profiles used by fingerprints, so that generic color lights could be re-profiled. To do this, I did 2 things:
1. Made all `2200K-6500K` fingerprints the generic counterpart. Since this is the default in the generic profile, this changes nothing on the user end.
2. For all remaining non-default fingerprints, I added a new table that will be checked during onboarding to ignore re-profiling if it is found. This list can (hopefully) be whittled down further with time, as we move away from any deprecated profiles. 

# Summary of Completed Tests
Unit tests updated to ensure proper updates, unit test added for new functionality.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
@@ -29,6 +29,12 @@ function SwitchDeviceConfiguration.assign_profile_for_onoff_ep(device, server_on
 
   local generic_profile = fields.device_type_profile_map[primary_dt_id]
 
+  -- remove colorTemperature capability from profile if device doesn't support it
+  if generic_profile == "light-color-level" and
+    not switch_utils.find_cluster_on_ep(ep_info, clusters.ColorControl.ID, {feature_bitmap=clusters.ColorControl.types.Feature.COLOR_TEMPERATURE}) then
+    generic_profile = "light-color-level-noTemp"
+  end
+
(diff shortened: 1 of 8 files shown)


drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -194,13 +200,13 @@ function DeviceConfiguration.match_profile(driver, device)
...
-      if switch_utils.check_switch_category_vendor_overrides(device) then
-        updated_profile = string.gsub(updated_profile, "plug", "switch")
-      end
-    elseif generic_profile("light-level-colorTemperature") or generic_profile("light-color-level") then
-      -- ignore attempts to dynamically profile light-level-colorTemperature and light-color-level devices for now, since
-      -- these may lose fingerprinted Kelvin ranges when dynamically profiled.
+    elseif switch_utils.check_switch_category_vendor_overrides(device) then
+      -- check whether the overwrite should be over "plug" or "light" based on the current profile
+      local overwrite_category = string.find(updated_profile, "plug") and "plug" or "light"
+      updated_profile = string.gsub(updated_profile, overwrite_category, "switch")
```
hcarter-775, line 206: This change is technically unrelated. When looking this over, I didn't think it was generic to assume all devices using a "Switch" category override have a plug device type. That does make the most since I think, but should not be a prerequisite in this logic.

greens (approved):

PR #2691, opened 2026-01-08
Matter Switch: Handle nullable case for ActivePower

hcarter-775:
# Description of Change
If no power is currently being used when the driver is starting or restarting, devices generally report Null as their value. This should be taken as 0.0 for a better user experience that is more consistent with other protocols.

# Summary of Completed Tests
Tested on-device. Since we do not support the TLV-encoded null type in lua right now, we cannot make a unit test for this case.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_handlers/attribute_handlers.lua
@@ -251,6 +251,9 @@ function AttributeHandlers.active_power_handler(driver, device, ib, response)
   if ib.data.value then
     local watt_value = ib.data.value / 1000 -- convert received milliwatt to watt
     device:emit_event_for_endpoint(ib.endpoint_id, capabilities.powerMeter.power({ value = watt_value, unit = "W"}))
+  else
+    -- note: we don't support nullable types, so this handles the Null case for ActivePower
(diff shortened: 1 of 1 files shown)


nickolas-deboom:
In the spec I'm seeing that a device would report NULL for ActivePower if the power cannot be measured, not necessarily if no power is being used. Since you mentioned the issue occurs primarily on driver start/restart, should we emit 0 in device_init rather than in this NULL case?

hcarter-775:
> In the spec I'm seeing that a device would report NULL for ActivePower if the power cannot be measured, not necessarily if no power is being used. Since you mentioned the issue occurs primarily on driver start/restart, should we emit 0 in device_init rather than in this NULL case?

@nickolas-deboom I think keeping actual device actions and capability events coupled is an good principle, which is why I lean against the idea of adding this to init. 

And, in that same vein, I wonder whether emitting a 0 event for a null event is really what we should do either? Yes, it is an unexpected value, but it is what the device reported. I know I opened this PR (lol) but the possible unintended consequences here seem important to keep in mind...

nickolas-deboom:
Plus there's the fact that if it was emitted during init it could cause offline devices to be marked online... definitely not a good idea :sweat_smile: 
>And, in that same vein, I wonder whether emitting a 0 event for a null event is really what we should do either? Yes, it is an unexpected value, but it is what the device reported.

Yeah I think you have a point. Are you thinking we continue to ignore Null values then?

PR #2702, opened 2026-01-13
Matter Switch: Add garage door profile and device override

hcarter-775:
# Description of Change
Add a new garage door profile for the custom behavior of the 3R Garage Door, which uses a On Off Light device typed endpoint to trigger its garage door functionality. This PR works to handle that more clearly in the ST app.

# Summary of Completed Tests
See unit tests. Tested on-target and device functioned as expected, using the override driver.


drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -91,7 +91,8 @@ function SwitchDeviceConfiguration.assign_profile_for_onoff_ep(device, server_on
     generic_profile = string.gsub(generic_profile, "-binary", "") .. static_electrical_tags
   end
 
-  if is_child_device and generic_profile == switch_utils.get_product_override_field(device, "initial_profile") then
+  if switch_utils.get_product_override_field(device, "on_children_only") == is_child_device and
```
gharveymn: This is also true when `on_children_only` and `is_child_device` are both false. Is that expected?

hcarter-775: Discussed offline, and though this specific behavior was desired, the conditional was edited to catch more cases.

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -91,7 +91,8 @@ function SwitchDeviceConfiguration.assign_profile_for_onoff_ep(device, server_on
     generic_profile = string.gsub(generic_profile, "-binary", "") .. static_electrical_tags
   end
 
-  if is_child_device and generic_profile == switch_utils.get_product_override_field(device, "initial_profile") then
+  if not switch_utils.get_product_override_field(device, "on_children_only") or is_child_device and
```
gharveymn: I think you'll need parentheses around the first two conditions here. Here's what'll happen otherwise:

```
> if true or false and false then print "hi" end
hi
> if (true or false) and false then print "hi" end
>
```

drivers/SmartThings/matter-switch/profiles/garage-door.yml
```diff
@@ -0,0 +1,12 @@
+name: garage-door-binary
+components:
+- id: main
+  capabilities:
+  - id: switch
```
nickolas-deboom: Should `button` be used instead?

nickolas-deboom: Sorry, nvm it uses OnOff

nickolas-deboom (approved):

drivers/SmartThings/matter-switch/src/sub_drivers/third_reality_garage_door/init.lua
```diff
@@ -0,0 +1,91 @@
...
+
+-- Prevent any of the main driver's logic from running
+local function info_changed(driver, device, event, args) end
+
+local function do_configure(driver, device)
+  match_profile(driver, device)
+end
+
+local function driver_switched(driver, device)
+  match_profile(driver, device)
```
ctowns, line 38: Do we need to set the device state to `PROVISIONED` here?

ctowns (approved):

drivers/SmartThings/matter-switch/src/sub_drivers/third_reality_garage_door/init.lua
hcarter-775: Yes we do, added!

nickolas-deboom (approved):

PR #2703, opened 2026-01-13
WWSTCERT-9799 Eve Thermostat (Europe)

greens:


drivers/SmartThings/matter-thermostat/fingerprints.yml
```diff
@@ -33,6 +33,11 @@ matterManufacturer:
     vendorId: 0x130A
     productId: 0x007D
     deviceProfileName: thermostat-heating-only-nostate
+  - id: "4874/127"
+    deviceLabel: Eve Thermostat (Europe)
+    vendorId: 0x130A
+    productId: 0x007F
+    deviceProfileName: thermostat-heating-only-nostate-nobattery
```
nickolas-deboom, line 40: The profile will be changed to thermostat-modular when the device is added to the driver so this doesn't matter much, but it's interesting they would select a heating only profile, I looked this product up and it says it will support heating and cooling

PR #2705, opened 2026-01-14
WWSTCERT-9895 WWSTCERT-9936 OREiN Matter smart Bathroom Fan

greens:


drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -161,6 +161,16 @@ matterManufacturer:
     vendorId: 0x1396
     productId: 0x1076
     deviceProfileName: light-color-level
+  - id: "5014/4246"
```
hcarter-775, line 164: the implementation of these devices has changed and will be going to prod in the new format starting monday (aka as parent/child). These should be changed to `fan-modular`

hcarter-775 (approved):

PR #2707, opened 2026-01-14
Matter Camera: Include all supported resolutions

nickolas-deboom:
This change adds logic for including the resolutions exposed by the `VideoSensorParams` and `MinViewport` attributes in `supportedResolutions`, rather than only looking at the resolution embedded within `RateDistortionTradeOffPoints`.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
@@ -113,9 +113,6 @@ end
 function CameraAttributeHandlers.rate_distortion_trade_off_points_handler(driver, device, ib, response)
   if not ib.data.elements then return end
   local resolutions = {}
-  local max_encoded_pixel_rate = device:get_field(camera_fields.MAX_ENCODED_PIXEL_RATE)
-  local max_fps = device:get_field(camera_fields.MAX_FRAMES_PER_SECOND)
-  local emit_capability = max_encoded_pixel_rate ~= nil and max_fps ~= nil
   for _, v in ipairs(ib.data.elements) do
     local rate_distortion_trade_off_points = v.elements
     local width = rate_distortion_trade_off_points.resolution.elements.width.value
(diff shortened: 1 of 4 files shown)


drivers/SmartThings/matter-switch/src/test/test_matter_camera.lua
```diff
@@ -719,6 +719,11 @@ local function emit_supported_resolutions()
         width = 3840,
         height = 2160,
         fps = 15
+      },
+      {
+        width = 7360,
+        height = 4912,
+        fps = 0
```
samadDotDev, line 747: Probably worth adding a resolution entry in the rate distortion stub that matches either the min or max, to show that we don't see duplicate entries in the supported resolutions capability here.

nickolas-deboom: In the latest commit I added a `receive_min_viewport` test function which has the same resolution as one within this list. Note that this commit also adds logic to ensure that `supportedResolutions` is only emitted one time, after MinViewportResolution, VideoSensorParams, RateDistortionTradeOffPoints, and MaxEncodedPixelRate are all received (previously, it would be been emitted twice)

hcarter-775 (approved):

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
```diff
@@ -115,7 +115,7 @@ function CameraAttributeHandlers.rate_distortion_trade_off_points_handler(driver
   local resolutions = {}
   local max_encoded_pixel_rate = device:get_field(camera_fields.MAX_ENCODED_PIXEL_RATE)
   local max_fps = device:get_field(camera_fields.MAX_FRAMES_PER_SECOND)
-  local emit_capability = max_encoded_pixel_rate ~= nil and max_fps ~= nil
+  local emit_capability = max_encoded_pixel_rate ~= nil and max_fps ~= nil and device:get_field(camera_fields.MIN_RESOLUTION) and device:get_field(camera_fields.MAX_RESOLUTION)
```
hcarter-775: just a nit, but if this is only being used to gate that if statement, maybe we should just put this in the statement itself? Like `if <all this boolean stuff> then`, and remove this variable altogether? Confused me a bit reading this initially.

Also, if you go with this suggestion, maybe make all of these booleans either use the explicit "~= nil" syntax or make them all implicit, rather than a mix of the 2?

nickolas-deboom: Sure, made them all implicit in the latest commit. Also, I realized I could clean up some logic in `video_sensor_parameters_handler` (I developed that handler while using an embedded version of the CAVSM cluster and for some reason I wasn't able to directly index into `VideoSensorParamsStruct`. Now with the actual cluster it works just fine)

PR #2714, opened 2026-01-20
Migrate Window Covering handling to Matter Switch

nickolas-deboom:
To support devices using the WindowCovering cluster in addition to light/switch/button endpoints, this moves the handling for window coverings into a subdriver within the switch driver. Note that the subdriver is called `closures` since it will be expanded to cover more Matter 1.5 closure types.

[ ](https://smartthings.atlassian.net/browse/CHAD-17448)


drivers/SmartThings/matter-switch/profiles/window-covering.yml
```diff
@@ -0,0 +1,20 @@
+name: window-covering
```
hcarter-775, line 1: If we're moving the window logic into switch, let's make this modular, no?

nickolas-deboom: I would definitely like to, but might not work given that window coverings can be created as child devices

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -135,13 +136,20 @@ SwitchFields.switch_category_vendor_overrides = {
...
 --- used in tandem with an EP ID. Stores the required electrical tags "-power", "-energy-powerConsumption", etc.
---- for an Electrical Sensor EP with a "primary" endpoint, used during device profling.
+--- for an Electrical Sensor EP with a "primary" endpoint, used during device profiling.
 SwitchFields.ELECTRICAL_TAGS = "__electrical_tags"
 
 SwitchFields.MODULAR_PROFILE_UPDATED = "__modular_profile_updated"
 
 SwitchFields.profiling_data = {
   POWER_TOPOLOGY = "__power_topology",
+  BATTERY_SUPPORT = "__battery_support",
```
hcarter-775, line 156: Hey @nickolas-deboom , would you mind moving this battery support logic into a separate PR? I believe it will be easier to review this separately, since it should be done for other work and so is not directly related to the window work.

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
```diff
@@ -0,0 +1,90 @@
...
+    -- These should only ever be nil once (and at the same time) for already-installed devices
+    -- It can be removed after migration is complete
+    device:emit_event(capabilities.windowShadePreset.supportedCommands({"presetPosition", "setPresetPosition"}, {visibility = {displayed = false}}))
+    local preset_position = device:get_field(closure_fields.PRESET_LEVEL_KEY) or
+      (device.preferences ~= nil and device.preferences.presetPosition) or
+      closure_fields.DEFAULT_PRESET_LEVEL
+    device:emit_event(capabilities.windowShadePreset.position(preset_position, {visibility = {displayed = false}}))
+    device:set_field(closure_fields.PRESET_LEVEL_KEY, preset_position, {persist = true})
+  end
+  device:subscribe()
```
hcarter-775, line 36: We should probably use the extended subscribe logic, right? Since it is possible this will be a part of a parent/child device?

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
```diff
@@ -0,0 +1,90 @@
...
+
+local ClosureLifecycleHandlers = {}
+
+function ClosureLifecycleHandlers.device_init(driver, device)
+  device:set_component_to_endpoint_fn(switch_utils.component_to_endpoint)
+  device:set_endpoint_to_component_fn(switch_utils.endpoint_to_component)
+  if device:supports_capability_by_id(capabilities.windowShadePreset.ID) and
+    device:get_latest_state("main", capabilities.windowShadePreset.ID, capabilities.windowShadePreset.position.NAME) == nil then
+    -- These should only ever be nil once (and at the same time) for already-installed devices
+    -- It can be removed after migration is complete
```
hcarter-775: We don't need logic related to this, correct? There is no migration for devices who will now be using the switch driver?

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
```diff
@@ -0,0 +1,90 @@
...
+  end
+  device:subscribe()
+end
+
+function ClosureLifecycleHandlers.device_added(driver, device)
+  device:emit_event(capabilities.windowShade.supportedWindowShadeCommands({"open", "close", "pause"}, {visibility = {displayed = false}}))
+  device:set_field(closure_fields.REVERSE_POLARITY, false, { persist = true })
+end
+
+function ClosureLifecycleHandlers.info_changed(driver, device, event, args)
```
hcarter-775: The software version update logic has been lost here, as well as the expanded subscribe logic. I am unsure how much we want to repeat this kind of logic in a subdriver, or if we want to conditionally move the preference logic into the main driver.

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button.lua
```diff
@@ -128,6 +128,9 @@ local function test_init()
   test.disable_startup_messages()
   test.mock_device.add_test_device(mock_device) -- make sure the cache is populated
 
+  local fields = require "switch_utils.fields"
```
hcarter-775: why wouldn't we let this be set in the test framework itself? Similar question for the other tests. However, this work should, like suggested previously, be moved to a separate PR for clarity.

drivers/SmartThings/matter-switch/src/test/test_aqara_climate_sensor_w100.lua
```diff
@@ -140,21 +141,14 @@ local function test_init()
   test.socket.device_lifecycle:__queue_receive({ aqara_mock_device.id, "doConfigure" })
   local read_attribute_list = clusters.PowerSource.attributes.AttributeList:read()
   test.socket.matter:__expect_send({aqara_mock_device.id, read_attribute_list})
-  configure_buttons()
   aqara_mock_device:expect_metadata_update({ provisioning_state = "PROVISIONED" })
-
-  local device_info_copy = utils.deep_copy(aqara_mock_device.raw_st_data)
-  device_info_copy.profile.id = "3-button-battery-temperature-humidity"
-  local device_info_json = dkjson.encode(device_info_copy)
-  test.socket.device_lifecycle:__queue_receive({ aqara_mock_device.id, "infoChanged", device_info_json })
```
hcarter-775: We might want to keep logic testing this infoChanged event, no?

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
nickolas-deboom: I kept this because I wasn't sure if the comment is accurate or not, but I will check

drivers/SmartThings/matter-switch/profiles/window-covering.yml
hcarter-775: Modular profiles can be used on child devices. It just adds some complexity in that it has to first be created before being updated. So the create event will not be able to include the modular metadata, but a following update will work as expected.

drivers/SmartThings/matter-switch/src/test/test_matter_multi_button.lua
nickolas-deboom: It felt simpler this way, and there are already test cases that receive the AttributeList and update the profile accordingly. But, maybe it would be better to test this in the init function. I'll try it out.

drivers/SmartThings/matter-switch/src/test/test_aqara_climate_sensor_w100.lua
nickolas-deboom: I'm pretty sure this wasn't doing anything, since the profile didn't change. So all this was doing was testing that infoChanged causes a new subscribe and configure_buttons call, which didn't seem very necessary...

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
nickolas-deboom: I thought about that, but it felt too complex to keep it separate because the `power_source_attribute_list_handler` was previously calling `try_metadata_update` directly rather than `match_profile`, and that wouldn't work for window coverings since they could either be the main device or a child device, which means there would probably need to be separate handling within `power_source_attribute_list_handler` based on whether the device is a button or a window covering. So, I think it's cleaner to do both of these changes at the same time. I realize it's a pretty big diff though. What do you think?

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
nickolas-deboom: Yes definitely. I'll update this

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
nickolas-deboom: Update: new PR for battery handling is https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2724

drivers/SmartThings/matter-switch/profiles/window-covering.yml
nickolas-deboom: Ok nice, I implemented modular profiles for parent and child window coverings in the previous 2 commits

nickolas-deboom: Note that I left a few static profiles because there are existing fingerprints pointing to them. Devices will profile switch to the modular profile though

nickolas-deboom: Updated fingerprints in latest commit

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
nickolas-deboom: Done in latest. As discussed, we will need to add a generic refresh test case and come up with a solution for that

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
nickolas-deboom: Similar to the discussion from https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2723, I think we will want to emit these capabilities once per device, which the current check should be doing. So I just removed this comment.

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
nickolas-deboom: I moved the preference logic to the main driver in the latest commit.

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
nickolas-deboom: as a POC I made a fix in the latest commit, but we can discuss if that's the solution we want to go with or not (and let's make it a separate PR ofc)

nickolas-deboom: Here's that PR: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2727

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -54,6 +56,11 @@ function SwitchLifecycleHandlers.do_configure(driver, device)
     switch_cfg.set_device_control_options(device)
     device_cfg.match_profile(driver, device)
   elseif device.network_type == device_lib.NETWORK_TYPE_CHILD then
+    local window_covering_ep_ids = switch_utils.get_endpoints_by_device_type(device:get_parent_device(), fields.DEVICE_TYPE_ID.WINDOW_COVERING)
+    if #window_covering_ep_ids > 0 then
+      local default_endpoint_id = switch_utils.find_default_endpoint(device:get_parent_device())
+      child_cfg.create_or_update_child_devices(driver, device:get_parent_device(), window_covering_ep_ids, default_endpoint_id, window_covering_cfg.assign_profile_for_window_covering_ep)
+    end
```
hcarter-775, line 58: can't this be done directly in `match_profile`?

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -85,6 +92,16 @@ function SwitchLifecycleHandlers.info_changed(driver, device, event, args)
       device_cfg.match_profile(driver, device)
     end
   end
+
+  if args.old_st_store.preferences.reverse ~= nil and device.preferences.reverse ~= nil and
+    args.old_st_store.preferences.reverse ~= device.preferences.reverse then
+    local closure_fields = require "sub_drivers.closures.closure_utils.fields"
+    if device.preferences.reverse then
```
hcarter-775: isn't this check done just done before?
edit: I see it is actually slightly different. Can we make this explicit that it's "== true", since we do such a similar check earlier?

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: or rather, isn't it kind of already defined there?

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
```diff
@@ -0,0 +1,77 @@
...
+    device:set_field(closure_fields.PRESET_LEVEL_KEY, preset_position, {persist = true})
+  end
+  device:extend_device("subscribe", switch_utils.subscribe)
+  device:subscribe()
+end
+
+function ClosureLifecycleHandlers.device_added(driver, device)
+  device:emit_event(capabilities.windowShade.supportedWindowShadeCommands({"open", "close", "pause"}, {visibility = {displayed = false}}))
+  device:set_field(closure_fields.REVERSE_POLARITY, false, { persist = true })
+  switch_utils.handle_electrical_sensor_info(device)
```
hcarter-775, line 41: I don't think a window covering needs to handle this at the moment.

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: This is meant to run `create_or_update_child_devices` a second time to update the child devices with optional capabilities since aren't included by `try_create_device`, since match_profile won't be called a second time. Did you have something else in mind for updating the child profiles with the optional capabilities metadata?

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
nickolas-deboom: It's needed to set `fields.profiling_data.POWER_TOPOLOGY` so that `match_profile` can proceed. We could do something like 
```
  local electrical_sensor_eps = utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.ELECTRICAL_SENSOR, { with_info = true })
  if #electrical_sensor_eps == 0 then
    device:set_field(fields.profiling_data.POWER_TOPOLOGY, false, {persist=true})
  end
```
in `device_init` if you prefer? Similar to setting `profiling_data.BATTERY _SUPPORT`

drivers/SmartThings/matter-switch/src/init.lua
nickolas-deboom: Sure thing, done in latest force-push

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
hcarter-775: ah, I see! My bad, forgot about that. Yeah this is fine then. It is a little confusing though... maybe it would be better to port this logic into an if block of the main driver, rather than pulling the main driver logic into a subdriver? That, or maybe we should move this function to doConfigure so we don't have to worry about overrides at all?

Thinking a little more, we could maybe even do both if that makes sense... I know this is a 1-1 port from the old window driver but do these steps really have to occur in device_added rather than init for example? Of course, with init it would require a little extra logic to avoid extra calls but that is relatively common and may be worth it here.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: Ah, I see. Yeah this makes sense then. One other small nit that I'll mention here, now that we're pulling `create_or_update_child_devices()` into other places, I kinda regret putting it into its own ChildConfiguration table. I think it would be more natural to just put that function in DeviceConfiguration, so we don't have to pull this table just for the single function.

hcarter-775: Thinking some more about this, I'm not this will run as you're expecting, because the child device won't necessarily have experienced an info_changed event on create, it will have just been created so it will do the added, init, doConfigure flow. Also, this logic would have the create_or_update logic occur on every infoChanged event, irrespective of whether that should be happening.

nickolas-deboom: I'm a little bit confused- why would the child device not necessarily having an info_changed event matter exactly? This is implemented in doConfigure, which a child device should always go through, right?
(Did I used to have this logic in infoChanged maybe? I can't remember :sweat_smile:)

drivers/SmartThings/matter-switch/src/sub_drivers/closures/init.lua
nickolas-deboom: I tried a few different things out, but realized that we need to override the added handler in the subdriver anyways, since the added handler in the main driver calls the main driver's `device_init`, which we want to avoid. I do think it makes sense to move the capability initialization to device_init to be consistent with the init's for the preset capability, which I did in the last commit.

I could move the remaining code out of the added handler, but since we have to override it anyway we might as well make use of it I guess? Lemme know what you think about that, I'm good either way

drivers/SmartThings/matter-switch/src/sub_drivers/closures/handlers/attribute_handlers.lua
```diff
@@ -0,0 +1,68 @@
+-- Copyright © 2026 SmartThings, Inc.
```
hcarter-775, line 1: nit: For these "generic" sub-drivers, I think the new directory name meta should be the generic words "utils" and "handlers" instead of names like "closure_handlers" and "closure_utils" for example.

The reason is that the pathing should already be unique, since these are in a subdriver, so the require would look like ``` require "sub_drivers.closures.utils.fields" ```. Therefore, no information is really lost (the subdriver name is still visible), and we don't have to worry about any other paths in the driver matching this.

drivers/SmartThings/matter-switch/src/sub_drivers/closures/can_handle.lua
```diff
@@ -0,0 +1,15 @@
...
+return function(opts, driver, device, ...)
+  local device_lib = require "st.device"
+  local fields = require "switch_utils.fields"
+  local switch_utils = require "switch_utils.utils"
+  if device.network_type == device_lib.NETWORK_TYPE_CHILD then
+    device = device:get_parent_device()
+  end
+  if #switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.WINDOW_COVERING) > 0 then
+    return true, require("sub_drivers.closures")
+  end
```
hcarter-775: At the moment this is not really an issue I don't think, but what about the case where there is >1 child device and one of them is a window covering? The current logic might overwrite the lifecycle event and capability event handling for the other device, non window covering device.

drivers/SmartThings/matter-switch/src/test/test_matter_window_covering.lua
```diff
@@ -0,0 +1,1126 @@
+-- Copyright © 2026 SmartThings, Inc.
```
hcarter-775, line 1: Can we add some tests for the case that there is a onOff switch as a parent device and a window covering as a child device?

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -20,17 +20,18 @@ local ChildConfiguration = {}
 local SwitchDeviceConfiguration = {}
 local ButtonDeviceConfiguration = {}
 local FanDeviceConfiguration = {}
+local WindowCoveringDeviceConfiguration = {}
 
 function ChildConfiguration.create_or_update_child_devices(driver, device, server_cluster_ep_ids, default_endpoint_id, assign_profile_fn)
```
hcarter-775: I know I introduced this type so this is my bad... but now that we're returning ChildConfiguration to use elsewhere, I think we should just remove this and use the standard DeviceConfiguration table to house this function.

drivers/SmartThings/matter-switch/src/sub_drivers/closures/can_handle.lua
hcarter-775: Specifically, I think we might want to do a little extra gating based on `opts.dispatcher_class`.

As in, if the dispatcher class name is `MatterMessageDispatcher`, we'd likely want to follow something similar to this logic, since all matter messages will come through the parent driver. Therefore, no EDGE_CHILD check would be necessary.

And then in the case of a `DeviceLifecycleDispatcher` or a `CapabilityCommandDispatcher` name, it might be safer to do a check based on the particular device's supported capabilities, agnostic of whether it is a parent or a child.

drivers/SmartThings/matter-switch/src/sub_drivers/closures/handlers/attribute_handlers.lua
nickolas-deboom: Done in latest!

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
nickolas-deboom: Done in latest

drivers/SmartThings/matter-switch/src/sub_drivers/closures/can_handle.lua
nickolas-deboom: Done in latest force-push leveraging the can_handle function from #2770

PR #2721, opened 2026-01-20
CHAD-17218: Fix Sonos augmented driver store upserts

NoahCornell:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

As part of the 59 IPC rework between hubcore and the runner, the environment info events for augmented driver store changed erroneously and so the parsing being done by the default driver handler is no longer correct. The values in the store will be correct after a driver/hub restart but upserts outside of a restart will not be applied.

This mostly effects the sonos driver (other drivers may not immediately apply some configs) since the driver will not be able to update its oauth token once it expires (every 24 hours). When the driver receives a capability command and the token is expired, there is a 30 second period where the driver will wait for the token to be refreshed so that the capability command will not fail.

Users have Sonos linked in the ST app will see either:

- Speakers be uncontrollable, if 3rd party oauth is enabled in the Sonos app
- 30 seconds delays, if 3rd party oauth is disabled in the Sonos app

I will make a similar change in scripting-engine to handle both before and after the changes in 59.

# Summary of Completed Tests

I have tested that upserts are now applied with these changes and the token refresher task is able to update invalid tokens again.

### DIFF
--- a/drivers/SmartThings/sonos/src/sonos_driver.lua
+++ b/drivers/SmartThings/sonos/src/sonos_driver.lua
@@ -1,4 +1,6 @@
-local api_version = require "version".api
+local version = require "version"
+local api_version = version.api
+local rpc_version = version.rpc
 local capabilities = require "st.capabilities"
 local cosock = require "cosock"
 local json = require "st.json"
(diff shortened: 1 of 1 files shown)


drivers/SmartThings/sonos/src/sonos_driver.lua
```diff
@@ -651,6 +653,59 @@ local function do_refresh(driver, device, cmd)
...
 
+--- In RPC version 100, the events for augmented driver store were changed
+--- to no longer match the parsing done in API version 18 and below.
+---
+--- API version 19 will handle both before and after RPC 100 changes so this only needs
+--- to be applied for RPC version >= 100 and API version <= 18.
+if rpc_version >= 100 and api_version <= 18 then
+  log.info("Overriding environment info handler for RPC >= 100 and API <= 18")
+  function SonosDriver:environment_info_handler(channel)
+    log.info("Starting environment info handler for RPC >= 100 and API <= 18")
```
cjswedes: I think it might make sense to have this be `{hub_logs = true}`

PR #2757, opened 2026-02-05
Support Soil Sensor device type

nickolas-deboom:
This adds support for the Soil Sensor device type, introduced with Matter 1.5.

Tested with VDA Soil Sensor

### DIFF
--- a/drivers/SmartThings/matter-sensor/src/embedded_clusters/SoilMeasurement/server/attributes/SoilMoistureMeasurementLimits.lua
+++ b/drivers/SmartThings/matter-sensor/src/embedded_clusters/SoilMeasurement/server/attributes/SoilMoistureMeasurementLimits.lua
@@ -1,0 +1,67 @@
+local cluster_base = require "st.matter.cluster_base"
+local data_types = require "st.matter.data_types"
+local TLVParser = require "st.matter.TLV.TLVParser"
+
+local SoilMoistureMeasurementLimits = {
+  ID = 0x0000,
+  NAME = "SoilMoistureMeasurementLimits",
+  base_type = require "embedded_clusters.Global.types.MeasurementAccuracyStruct",
+}
+
+function SoilMoistureMeasurementLimits:new_value(...)
+  local o = self.base_type(table.unpack({...}))
+  self:augment_type(o)
+  return o
+end
+
+function SoilMoistureMeasurementLimits:read(device, endpoint_id)
+  return cluster_base.read(
+    device,
+    endpoint_id,
+    self._cluster.ID,
+    self.ID,
+    nil
+  )
+end
+
+function SoilMoistureMeasurementLimits:subscribe(device, endpoint_id)
+  return cluster_base.subscribe(
+    device,
+    endpoint_id,
+    self._cluster.ID,
+    self.ID,
+    nil
+  )
+end
+
+function SoilMoistureMeasurementLimits:set_parent_cluster(cluster)
+  self._cluster = cluster
+  return self
+end
+
+function SoilMoistureMeasurementLimits:build_test_report_data(
+  device,
+  endpoint_id,
+  value,
+  status
+)
+  local data = data_types.validate_or_build_type(value, self.base_type)
+  self:augment_type(data)
+  return cluster_base.build_test_report_data(
+    device,
+    endpoint_id,
+    self._cluster.ID,
+    self.ID,
+    data,
+    status
+  )
+end
+
+function SoilMoistureMeasurementLimits:deserialize(tlv_buf)
+  local data = TLVParser.decode_tlv(tlv_buf)
+  self:augment_type(data)
+  return data
+end
+
+setmetatable(SoilMoistureMeasurementLimits, {__call = SoilMoistureMeasurementLimits.new_value, __index = SoilMoistureMeasurementLimits.base_type})
+return SoilMoistureMeasurementLimits
--- a/drivers/SmartThings/matter-sensor/src/init.lua
+++ b/drivers/SmartThings/matter-sensor/src/init.lua
@@ -37,6 +37,12 @@ if version.api < 11 then
   clusters.BooleanStateConfiguration = require "embedded_clusters.BooleanStateConfiguration"
 end
 
+-- Include driver-side definitions when lua libs api version is < 21
+-- TODO: change this to < 20 once the lua libs have been updated for hub-core 61
+if version.api < 21 then
+  clusters.SoilMeasurement = require "embedded_clusters.SoilMeasurement"
+end
+
 local SensorLifecycleHandlers = {}
 
 function SensorLifecycleHandlers.do_configure(driver, device)
@@ -117,6 +123,10 @@ local matter_driver_template = {
       [clusters.RelativeHumidityMeasurement.ID] = {
         [clusters.RelativeHumidityMeasurement.attributes.MeasuredValue.ID] = attribute_handlers.humidity_measured_value_handler
       },
+      [clusters.SoilMeasurement.ID] = {
+        [clusters.SoilMeasurement.attributes.SoilMoistureMeasuredValue.ID] = attribute_handlers.soil_moisture_measured_value_handler,
+        [clusters.SoilMeasurement.attributes.SoilMoistureMeasurementLimits.ID] = attribute_handlers.soil_moisture_measurement_limits_handler
+      },
       [clusters.TemperatureMeasurement.ID] = {
         [clusters.TemperatureMeasurement.attributes.MeasuredValue.ID] = attribute_handlers.temperature_measured_value_handler,
         [clusters.TemperatureMeasurement.attributes.MinMeasuredValue.ID] = attribute_handlers.temperature_measured_value_bounds_factory(fields.TEMP_MIN),
@@ -164,7 +174,9 @@ local matter_driver_template = {
       clusters.BooleanState.attributes.StateValue,
     },
     [capabilities.relativeHumidityMeasurement.ID] = {
-      clusters.RelativeHumidityMeasurement.attributes.MeasuredValue
+      clusters.RelativeHumidityMeasurement.attributes.MeasuredValue,
+      clusters.SoilMeasurement.attributes.SoilMoistureMeasuredValue,
+      clusters.SoilMeasurement.attributes.SoilMoistureMeasurementLimits
     },
     [capabilities.temperatureAlarm.ID] = {
       clusters.BooleanState.attributes.StateValue,
@@ -242,9 +254,6 @@ local matter_driver_template = {
     [capabilities.radonMeasurement.ID] = {
       clusters.RadonConcentrationMeasurement.attributes.MeasuredValue,
       clusters.RadonConcentrationMeasurement.attributes.MeasurementUnit,
-    },
-    [capabilities.relativeHumidityMeasurement.ID] = {
-      clusters.RelativeHumidityMeasurement.attributes.MeasuredValue
     },
     [capabilities.tvocHealthConcern.ID] = {
       clusters.TotalVolatileOrganicCompoundsConcentrationMeasurement.attributes.LevelValue
--- a/drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
@@ -9,6 +9,10 @@ local fields = require "sensor_utils.fields"
 local device_cfg = require "sensor_utils.device_configuration"
 local version = require "version"
 
+if version.api < 13 then
+  clusters.Global = require "embedded_clusters.Global"
+end
+
 local AttributeHandlers = {}
 
 
@@ -69,16 +73,39 @@ function AttributeHandlers.humidity_measured_value_handler(driver, device, ib, r
 end
 
 
+-- [[ SOIL MEASUREMENT CLUSTER ATTRIBUTES ]] --
+
+function AttributeHandlers.soil_moisture_measured_value_handler(driver, device, ib, response)
+  if ib.data.value == nil then return end
+  local min = sensor_utils.get_field_for_endpoint(device, fields.SOIL_LIMIT_MIN, ib.endpoint_id) or sensor_utils.SOIL_MOISTURE_MIN
+  local max = sensor_utils.get_field_for_endpoint(device, fields.SOIL_LIMIT_MAX, ib.endpoint_id) or sensor_utils.SOIL_MOISTURE_MAX
+  local soil_moisture = st_utils.clamp_value(ib.data.value, min, max)
+  device:emit_event_for_endpoint(ib.endpoint_id, capabilities.relativeHumidityMeasurement.humidity(soil_moisture))
+end
+
+function AttributeHandlers.soil_moisture_measurement_limits_handler(driver, device, ib, response)
+  local MeasurementAccuracyStruct = require "embedded_clusters.Global.types.MeasurementAccuracyStruct"
+  MeasurementAccuracyStruct:augment_type(ib.data)
+  local min_val = ib.data.elements and ib.data.elements.min_measured_value and ib.data.elements.min_measured_value.value
+  local max_val = ib.data.elements and ib.data.elements.max_measured_value and ib.data.elements.max_measured_value.value
+  if not (min_val and max_val) or (min_val >= max_val) or (min_val < sensor_utils.SOIL_MOISTURE_MIN) or (max_val > sensor_utils.SOIL_MOISTURE_MAX) then
+    device.log.warn_with({hub_logs = true}, string.format("Device reported invalid soil moisture limits: min=%d, max=%d", min_val, max_val))
+  end
+  sensor_utils.set_field_for_endpoint(device, fields.SOIL_LIMIT_MIN, ib.endpoint_id, min_val)
+  sensor_utils.set_field_for_endpoint(device, fields.SOIL_LIMIT_MAX, ib.endpoint_id, max_val)
+end
+
+
 -- [[ BOOLEAN STATE CLUSTER ATTRIBUTES ]] --
 
 function AttributeHandlers.boolean_state_value_handler(driver, device, ib, response)
   local name
   for dt_name, _ in pairs(fields.BOOLEAN_DEVICE_TYPE_INFO) do
-      local dt_ep_id = device:get_field(dt_name)
-      if ib.endpoint_id == dt_ep_id then
-          name = dt_name
-          break
-      end
+    local dt_ep_id = device:get_field(dt_name)
+    if ib.endpoint_id == dt_ep_id then
+      name = dt_name
+      break
+    end
   end
   if name then
     device:emit_event_for_endpoint(ib.endpoint_id, fields.BOOLEAN_CAP_EVENT_MAP[ib.data.value][name])
--- a/drivers/SmartThings/matter-sensor/src/sensor_utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-sensor/src/sensor_utils/device_configuration.lua
@@ -11,18 +11,22 @@ if version.api < 11 then
   clusters.BooleanStateConfiguration = require "embedded_clusters.BooleanStateConfiguration"
 end
 
+if version.api < 21 then
+  clusters.SoilMeasurement = require "embedded_clusters.SoilMeasurement"
+end
+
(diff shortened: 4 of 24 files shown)


drivers/SmartThings/matter-sensor/src/embedded_clusters/SoilMeasurement/server/attributes/AcceptedCommandList.lua
```diff
@@ -0,0 +1,74 @@
+local cluster_base = require "st.matter.cluster_base"
```
hcarter-775: we cab remove most of these embedded files I think

drivers/SmartThings/matter-sensor/src/embedded_clusters/SoilMeasurement/server/attributes/AttributeList.lua
```diff
@@ -0,0 +1,74 @@
+local cluster_base = require "st.matter.cluster_base"
```
hcarter-775: like this one

drivers/SmartThings/matter-sensor/src/embedded_clusters/SoilMeasurement/server/attributes/SoilMoistureMeasurementLimits.lua
```diff
@@ -0,0 +1,67 @@
+local cluster_base = require "st.matter.cluster_base"
```
hcarter-775, line 1: and this one, unless you're planning on adding a handler for this?

drivers/SmartThings/matter-sensor/src/init.lua
hcarter-775, line 53: I think we need to add some kind of handling to the `match_profile` function so that it can support battery? Along with a couple unit tests?

nickolas-deboom: That should already be supported, no?

drivers/SmartThings/matter-sensor/src/init.lua
```diff
@@ -243,9 +253,6 @@ local matter_driver_template = {
       clusters.RadonConcentrationMeasurement.attributes.MeasuredValue,
       clusters.RadonConcentrationMeasurement.attributes.MeasurementUnit,
     },
-    [capabilities.relativeHumidityMeasurement.ID] = {
```
nickolas-deboom, line 246: This was a duplicate entry

drivers/SmartThings/matter-sensor/src/sensor_utils/device_configuration.lua
```diff
@@ -117,8 +123,7 @@ function DeviceConfiguration.match_profile(driver, device, battery_supported)
   -- remove leading "-"
   profile_name = string.sub(profile_name, 2)
 
-  device.log.info_with({hub_logs=true}, string.format("Updating device profile to %s.", profile_name))
```
nickolas-deboom, line 120: I removed this log because I realized this is now logged by the lua libs

drivers/SmartThings/matter-sensor/src/embedded_clusters/SoilMeasurement/server/attributes/SoilMoistureMeasurementLimits.lua
nickolas-deboom: Yep, I'll remove it

drivers/SmartThings/matter-sensor/src/embedded_clusters/SoilMeasurement/server/attributes/AcceptedCommandList.lua
nickolas-deboom: Yeah good call

nickolas-deboom: I removed them in the latest commit

drivers/SmartThings/matter-sensor/src/init.lua
hcarter-775: how would it be?

edit: nevermind, I forgot how the Matter Sensor match profile logic works.

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
```diff
@@ -69,16 +71,44 @@ function AttributeHandlers.humidity_measured_value_handler(driver, device, ib, r
...
+  device:emit_event_for_endpoint(ib.endpoint_id, capabilities.relativeHumidityMeasurement.humidity(soil_moisture))
+end
+
+function AttributeHandlers.soil_moisture_measurement_limits_handler(driver, device, ib, response)
+  clusters.Global.types.MeasurementAccuracyStruct:augment_type(ib.data)
+  if ib.data.elements == nil or
+    ib.data.elements.min_measured_value == nil or ib.data.elements.min_measured_value.value == nil or
+    ib.data.elements.max_measured_value == nil or ib.data.elements.max_measured_value.value == nil or
+    ib.data.elements.min_measured_value.value >= ib.data.elements.max_measured_value.value then
+    device.log.warn_with({hub_logs = true}, "Device reported invalid soil moisture limits")
```
hcarter-775: Can you expand on this warning to give slightly more info?

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
```diff
@@ -69,16 +71,44 @@ function AttributeHandlers.humidity_measured_value_handler(driver, device, ib, r
...
+    return
+  end
+  sensor_utils.set_field_for_endpoint(device, fields.SOIL_LIMIT_MIN, ib.endpoint_id, ib.data.elements.min_measured_value.value)
+  sensor_utils.set_field_for_endpoint(device, fields.SOIL_LIMIT_MAX, ib.endpoint_id, ib.data.elements.max_measured_value.value)
```
hcarter-775: Should we have sanity checks here you think?

drivers/SmartThings/matter-sensor/src/sensor_utils/device_configuration.lua
```diff
@@ -58,7 +58,16 @@ function DeviceConfiguration.match_profile(driver, device, battery_supported)
   end
 
   if device:supports_capability(capabilities.relativeHumidityMeasurement) then
-    profile_name = profile_name .. "-humidity"
+    -- Soil Sensor fingerprints to the humidity profile, so we should also check for
+    -- TemperatureMeasurement, which is an optional cluster for this device type.
```
hcarter-775: might be worth adding a TODO comment about having a new category for soil sensor. Otherwise this kinda doesn't make sense why there'd be a separate profile

drivers/SmartThings/matter-sensor/src/test/test_matter_soil_sensor.lua
```diff
@@ -0,0 +1,237 @@
...
+    subscribe_request:merge(attr:subscribe(mock_device))
+  end
+  test.socket.matter:__expect_send({mock_device.id, subscribe_request})
+end
+
+test.register_coroutine_test(
+  "Relative humidity reports should generate correct messages",
+  function()
+    update_device_profile()
+    test.wait_for_events()
```
hcarter-775: I suggest we break this into a separate test. I don't think we need to test the do configure profile update in a relative humidity report test. I think this makes changing tests later more difficult. Same comment for other tests.

Basically I think we should have this profile update be its own test, separate from everything else.

drivers/SmartThings/matter-sensor/src/sensor_utils/device_configuration.lua
nickolas-deboom: Added

drivers/SmartThings/matter-sensor/src/test/test_matter_soil_sensor.lua
nickolas-deboom: I moved it into test_init in the latest commit rather than a separate test so that the test cases would use the correct profile (since the mock device fingerprints to a different profile

hcarter-775: I think what I'm suggesting is that, if that's the case, it might be worth breaking the tests into more de-coupled pieces. If that means having more than 1 mock device or maybe just doing the extra updates in the test that rely on that, then we should probably do that. A lot of our current mock tests do a lot of this repeat testing of things to get to a particular profile, when I think we should often just test the configuration and then also test the capabilities and attributes as two distinct things. There's not a huge reason as far as the test is concerned for why the mock device has to have the incorrect profile for this test is my thinking

nickolas-deboom: It's not an incorrect profile but rather the profile that the device would fingerprint to based on device type. Tbh I think it makes more sense to setup the mock device the same way for each test case in test_init like this, especially since it's closer to the onboarding flow that an actual device would go through.

nickolas-deboom: After thinking about this more, I think you're right that it makes sense to break this into a separate UT. Let me know if that looks alright a279798e916110e960daf3366810daf1b878f632

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
nickolas-deboom: Done in 960b149ad62679121a6890c8b8105b8b3a325eca

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
nickolas-deboom: Done in 960b149ad62679121a6890c8b8105b8b3a325eca

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
```diff
@@ -9,6 +9,8 @@ local fields = require "sensor_utils.fields"
 local device_cfg = require "sensor_utils.device_configuration"
 local version = require "version"
 
+clusters.SoilMeasurement = require "embedded_clusters.SoilMeasurement"
```
hcarter-775: this should be version gated I think. Also we should gate the "global" type as well here

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
```diff
@@ -69,16 +71,48 @@ function AttributeHandlers.humidity_measured_value_handler(driver, device, ib, r
...
+function AttributeHandlers.soil_moisture_measured_value_handler(driver, device, ib, response)
+  if ib.data.value == nil then return end
+  local min = sensor_utils.get_field_for_endpoint(device, fields.SOIL_LIMIT_MIN, ib.endpoint_id)
+  local max = sensor_utils.get_field_for_endpoint(device, fields.SOIL_LIMIT_MAX, ib.endpoint_id)
+  if min == nil or max == nil then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.relativeHumidityMeasurement.humidity(ib.data.value))
+    return
+  end
+  local soil_moisture = st_utils.round((ib.data.value - min) / (max - min) * 100.0)
+  device:emit_event_for_endpoint(ib.endpoint_id, capabilities.relativeHumidityMeasurement.humidity(soil_moisture))
```
hcarter-775:
```suggestion
  local min = sensor_utils.get_field_for_endpoint(device, fields.SOIL_LIMIT_MIN, ib.endpoint_id) or sensor_utils.SOIL_MOISTURE_MIN
  local max = sensor_utils.get_field_for_endpoint(device, fields.SOIL_LIMIT_MAX, ib.endpoint_id) or sensor_utils.SOIL_MOISTURE_MAX
  local soil_moisture = st_utils.round((ib.data.value - min) / (max - min) * 100.0)
  device:emit_event_for_endpoint(ib.endpoint_id, capabilities.relativeHumidityMeasurement.humidity(soil_moisture))
```
I think we can use the default bounds to get the same result here

drivers/SmartThings/matter-sensor/src/sensor_utils/device_configuration.lua
```diff
@@ -53,12 +53,22 @@ function DeviceConfiguration.match_profile(driver, device, battery_supported)
     profile_name = profile_name .. "-illuminance"
   end
 
-  if device:supports_capability(capabilities.temperatureMeasurement) then
+  -- TemperatureMeasurement is an optional cluster for the Soil Sensor device type,
+  -- and since Soil Sensors fingerprint to the humidity profile (which does not
+  -- include the temperatureMeasurement capability), we should also check for the
+  -- presence of the TemperatureMeasurement cluster here.
+  -- TODO: Update soil sensor profiles to use the SoilSensor category once it is available.
```
hcarter-775: This comment should probably go by the soil sensor part

drivers/SmartThings/matter-sensor/src/sensor_utils/device_configuration.lua
```diff
@@ -53,12 +53,22 @@ function DeviceConfiguration.match_profile(driver, device, battery_supported)
     profile_name = profile_name .. "-illuminance"
   end
 
-  if device:supports_capability(capabilities.temperatureMeasurement) then
+  -- TemperatureMeasurement is an optional cluster for the Soil Sensor device type,
+  -- and since Soil Sensors fingerprint to the humidity profile (which does not
+  -- include the temperatureMeasurement capability), we should also check for the
+  -- presence of the TemperatureMeasurement cluster here.
```
hcarter-775: I'm not actually sure we need this comment? Checking the temperature endpoints seems like a fairly standard piece of logic to me?

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
```diff
@@ -69,16 +71,48 @@ function AttributeHandlers.humidity_measured_value_handler(driver, device, ib, r
...
+    ib.data.elements.min_measured_value == nil or ib.data.elements.min_measured_value.value == nil or
+    ib.data.elements.max_measured_value == nil or ib.data.elements.max_measured_value.value == nil or
+    ib.data.elements.min_measured_value.value >= ib.data.elements.max_measured_value.value then
+    local min_val = ib.data.elements and ib.data.elements.min_measured_value and ib.data.elements.min_measured_value.value
+    local max_val = ib.data.elements and ib.data.elements.max_measured_value and ib.data.elements.max_measured_value.value
+    device.log.warn_with({hub_logs = true}, string.format("Device reported invalid soil moisture limits: min=%d, max=%d", min_val, max_val))
+    return
+  end
+  local min_val = math.max(ib.data.elements.min_measured_value.value, sensor_utils.SOIL_MOISTURE_MIN)
+  local max_val = math.min(ib.data.elements.max_measured_value.value, sensor_utils.SOIL_MOISTURE_MAX)
```
hcarter-775:
```suggestion
    local min_val = ib.data.elements and ib.data.elements.min_measured_value and ib.data.elements.min_measured_value.value
    local max_val = ib.data.elements and ib.data.elements.max_measured_value and ib.data.elements.max_measured_value.value
    if not (min_val and max_val) or (min_val >= max_val) or (min_val < sensor_utils.SOIL_MOISTURE_MIN) or (max_val > sensor_utils.SOIL_MOISTURE_MAX) then
      device.log.warn_with({hub_logs = true}, string.format("Device reported invalid soil moisture limits: min=%d, max=%d", min_val, max_val))
    end
```

I think we can simplify this like so. Also, I added in the default value checking in the early return, because as-is we set the defaults to the field if an actual value is returned at all,though I think we really only want to set the field if a device-specific value is given. Else, we can just use the defaults

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
nickolas-deboom: This cluster is actually not in the lua libs currently...

nickolas-deboom: But agreed about Global

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
```diff
@@ -69,16 +71,48 @@ function AttributeHandlers.humidity_measured_value_handler(driver, device, ib, r
...
+
+function AttributeHandlers.soil_moisture_measured_value_handler(driver, device, ib, response)
+  if ib.data.value == nil then return end
+  local min = sensor_utils.get_field_for_endpoint(device, fields.SOIL_LIMIT_MIN, ib.endpoint_id)
+  local max = sensor_utils.get_field_for_endpoint(device, fields.SOIL_LIMIT_MAX, ib.endpoint_id)
+  if min == nil or max == nil then
+    device:emit_event_for_endpoint(ib.endpoint_id, capabilities.relativeHumidityMeasurement.humidity(ib.data.value))
+    return
+  end
+  local soil_moisture = st_utils.round((ib.data.value - min) / (max - min) * 100.0)
```
hcarter-775: Per spec:

```
2.15.4.2. SoilMoistureMeasuredValue Attribute
[2.593] This attribute SHALL indicate the water content of the soil in percentage.
```
 and:
```
2.15.4.1. SoilMoistureMeasurementLimits Attribute
[2.590] This attribute SHALL indicate the limits for the SoilMoistureMeasuredValue attribute.
```

So I don't think we should be re-defining the input according to the bounds, but rather just ensuring that the measured value is actually within the max/min bounds.

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
hcarter-775: Ah, I see. then I still think we might want to add pcall to this, and add Soil Measurement to the lua libs in the bug fix window. Can you make a ticket for that and put up the change?

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
nickolas-deboom: Done in 87f21e2e8ecb181985b4d6af8191b2a9d81f37d6

drivers/SmartThings/matter-sensor/src/sensor_utils/device_configuration.lua
nickolas-deboom: Agreed, removed

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
nickolas-deboom: Done in 87f21e2e8ecb181985b4d6af8191b2a9d81f37d6

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
nickolas-deboom: Agreed after rereading and updated in 87f21e2e8ecb181985b4d6af8191b2a9d81f37d6

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
```diff
@@ -74,30 +78,19 @@ end
 -- [[ SOIL MEASUREMENT CLUSTER ATTRIBUTES ]] --
 
 function AttributeHandlers.soil_moisture_measured_value_handler(driver, device, ib, response)
-  if ib.data.value == nil then return end
```
hcarter-775: keep nil check?

nickolas-deboom: Oops, restored

drivers/SmartThings/matter-sensor/src/sensor_handlers/attribute_handlers.lua
nickolas-deboom: I put up the change and added a gate for SoilMeasurement for api v20 and below

ctowns (approved):

hcarter-775 (approved):

PR #2787, opened 2026-02-18
Spot checking for lazy loading issues with drivers in beta and production for 2/17/26

aleclorimer:
Created a test_\<driver\>_v10.lua file from the generic driver test file, but set the version.api to 10 to test lazy loading v1 method


hcarter-775:
@aleclorimer can this be closed?

PR #2791, opened 2026-02-20
WWSTCERT-8045 - Aeotec Home Energy Meter Gen8 (Revert + fixes)

KKlimczukS:
WWSTCERT-8045 - Aeotec Home Energy Meter Gen8
cleanup + fixes


greens (commented):
@KKlimczukS are you open to making changes to this PR beyond the updates?

greens (commented):
There seems to be a lot of overlap between the three sub-drivers that could be pulled up into the parent. As far as I can tell the only difference between 2-phase and 3-phase are the endpoints used for meter reports and the fingerprints. I didn't check 1-phase but assume it is similar. Also please shorten the copyright/license statement to the new format.

KKlimczukS:
> There seems to be a lot of overlap between the three sub-drivers that could be pulled up into the parent. As far as I can tell the only difference between 2-phase and 3-phase are the endpoints used for meter reports and the fingerprints. I didn't check 1-phase but assume it is similar. Also please shorten the copyright/license statement to the new format.

@greens 
- I moved several duplicated functions to the parent, 
- supporting powerConsumptionEvents was moved to a separate file
- The rest of the code relies on HEM8_DEVICES tables - those keep data strictly related to specific version of the device, so those need to stay in sub-sub-drivers, otherwise further refactoring might basically flatten the whole structure

drivers/SmartThings/zwave-electric-meter/src/preferences.lua
```diff
@@ -0,0 +1,95 @@
...
+
+preferences.to_numeric_value = function(new_value)
+  local numeric = tonumber(new_value)
+  if numeric == nil then -- in case the value is boolean
+    numeric = new_value and 1 or 0
+  end
+  return numeric
+end
+
+return preferences
```
cjswedes: nit: newline

cjswedes (approved):

drivers/SmartThings/zwave-electric-meter/src/preferences.lua
KKlimczukS: fixed

aleclorimer (approved):

PR #2803, opened 2026-02-24
Matter Switch: Update ST Energy Reporting Interval for Multi-Plugs

hcarter-775:
# Description of Change
At present, the minimum 15 minute ST Energy interval is handled as a single value for the entire Matter device. However, multi-plug devices may be represented on the platform as 2+ devices, and perhaps this 15 minute interval should be evaluated differently for each of these devices since ST Energy evaluates them as completely separate devices, rather than uniquely.

# Summary of Completed Tests
Unit tests updated, tested on-device.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
@@ -92,6 +92,7 @@ SwitchFields.updated_fields = {
   { current_field_name = "__switch_intialized", updated_field_name = nil },
   { current_field_name = "__energy_management_endpoint", updated_field_name = nil },
   { current_field_name = "__total_imported_energy", updated_field_name = nil },
+  { current_field_name = "__last_imported_report_timestamp", updated_field_name = nil },
 }
 
 SwitchFields.vendor_overrides = {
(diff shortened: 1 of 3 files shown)


drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -92,6 +92,7 @@ SwitchFields.updated_fields = {
   { current_field_name = "__switch_intialized", updated_field_name = nil },
   { current_field_name = "__energy_management_endpoint", updated_field_name = nil },
   { current_field_name = "__total_imported_energy", updated_field_name = nil },
+  { current_field_name = "__last_imported_report_timestamp", updated_field_name = nil },
```
greens, line 95: this is just to clear out the storage of no-longer-used fields?

greens (approved):

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
hcarter-775: yes

gharveymn (approved):

PR #2818, opened 2026-03-02
Zigbee Switch: Add support/handling for Stateless Step Capabilities

wkhenon:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/color_temp_range_handlers/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/color_temp_range_handlers/init.lua
@@ -3,56 +3,59 @@
 
 local capabilities = require "st.capabilities"
 local clusters = require "st.zigbee.zcl.clusters"
-local utils = require "st.utils"
-local KELVIN_MAX = "_max_kelvin"
-local KELVIN_MIN = "_min_kelvin"
-local MIREDS_CONVERSION_CONSTANT = 1000000
-local COLOR_TEMPERATURE_KELVIN_MAX = 15000
-local COLOR_TEMPERATURE_KELVIN_MIN = 1000
-local COLOR_TEMPERATURE_MIRED_MAX = utils.round(MIREDS_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN) -- 1000
-local COLOR_TEMPERATURE_MIRED_MIN = utils.round(MIREDS_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX) -- 67
+local switch_utils = require "switch_utils"
+
+-- These values are a "sanity check" to ensure that max/min values we are getting are reasonable
+local COLOR_TEMPERATURE_MIRED_MAX = 1000 -- 1000 Kelvin
+local COLOR_TEMPERATURE_MIRED_MIN = 67 -- 15000 Kelvin
 
 local function color_temp_min_mireds_handler(driver, device, value, zb_rx)
-  local temp_in_mired = value.value
-  local endpoint = zb_rx.address_header.src_endpoint.value
-  if temp_in_mired == nil then
+  -- if mired value is nil or outside of sane bounds, log and ignore. Else, save value
+  local min_mired_bound = value.value
+  if min_mired_bound == nil then
+    return
+  elseif (min_mired_bound < COLOR_TEMPERATURE_MIRED_MIN or min_mired_bound > COLOR_TEMPERATURE_MIRED_MAX) then
+    device.log.warn_with({hub_logs = true}, string.format("Device reported a color temperature %d mired outside of sane range of %.2f-%.2f", min_mired_bound, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX))
     return
   end
-  if (temp_in_mired < COLOR_TEMPERATURE_MIRED_MIN or temp_in_mired > COLOR_TEMPERATURE_MIRED_MAX) then
-    device.log.warn_with({hub_logs = true}, string.format("Device reported a color temperature %d mired outside of sane range of %.2f-%.2f", temp_in_mired, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX))
+  device:set_field(switch_utils.MIRED_MIN_BOUND, min_mired_bound, {persist = true})
+
+  -- if we have already received a valid max mired bound, emit a colorTemperatureRange event
+  local max_mired_bound = device:get_field(switch_utils.MIRED_MAX_BOUND)
+  if max_mired_bound == nil then
     return
-  end
-  local temp_in_kelvin = utils.round(MIREDS_CONVERSION_CONSTANT / temp_in_mired)
-  device:set_field(KELVIN_MAX..endpoint, temp_in_kelvin)
-  local min = device:get_field(KELVIN_MIN..endpoint)
-  if min ~= nil then
-    if temp_in_kelvin > min then
-      device:emit_event_for_endpoint(endpoint, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = min, maximum = temp_in_kelvin}}))
-    else
-      device.log.warn_with({hub_logs = true}, string.format("Device reported a max color temperature %d K that is not higher than the reported min color temperature %d K", min, temp_in_kelvin))
-    end
+  elseif min_mired_bound < max_mired_bound then
+    local endpoint = zb_rx.address_header.src_endpoint.value
+    local max_kelvin_bound = switch_utils.convert_mired_to_kelvin(min_mired_bound)
+    local min_kelvin_bound = switch_utils.convert_mired_to_kelvin(max_mired_bound)
+    device:emit_event_for_endpoint(endpoint, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = min_kelvin_bound, maximum = max_kelvin_bound}}))
+  else
+    device.log.warn_with({hub_logs = true}, string.format("Device reported a max color temperature %d Mireds that is not higher than the reported min color temperature %d Mireds", max_mired_bound, min_mired_bound))
   end
 end
 
 local function color_temp_max_mireds_handler(driver, device, value, zb_rx)
-  local temp_in_mired = value.value
-  local endpoint = zb_rx.address_header.src_endpoint.value
-  if temp_in_mired == nil then
+  -- if mired value is nil or outside of sane bounds, log and ignore. Else, save value
+  local max_mired_bound = value.value
+  if max_mired_bound == nil then
+    return
+  elseif (max_mired_bound < COLOR_TEMPERATURE_MIRED_MIN or max_mired_bound > COLOR_TEMPERATURE_MIRED_MAX) then
+    device.log.warn_with({hub_logs = true}, string.format("Device reported a color temperature %d mired outside of sane range of %.2f-%.2f", max_mired_bound, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX))
     return
   end
-  if (temp_in_mired < COLOR_TEMPERATURE_MIRED_MIN or temp_in_mired > COLOR_TEMPERATURE_MIRED_MAX) then
-    device.log.warn_with({hub_logs = true}, string.format("Device reported a color temperature %d mired outside of sane range of %.2f-%.2f", temp_in_mired, COLOR_TEMPERATURE_MIRED_MIN, COLOR_TEMPERATURE_MIRED_MAX))
+  device:set_field(switch_utils.MIRED_MAX_BOUND, max_mired_bound, {persist = true})
+
+  -- if we have already received a valid min mired bound, emit a colorTemperatureRange event
+  local min_mired_bound = device:get_field(switch_utils.MIRED_MIN_BOUND)
+  if min_mired_bound == nil then
     return
-  end
-  local temp_in_kelvin = utils.round(MIREDS_CONVERSION_CONSTANT / temp_in_mired)
-  device:set_field(KELVIN_MIN..endpoint, temp_in_kelvin)
-  local max = device:get_field(KELVIN_MAX..endpoint)
-  if max ~= nil then
-    if temp_in_kelvin < max then
-      device:emit_event_for_endpoint(endpoint, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = temp_in_kelvin, maximum = max}}))
-    else
-      device.log.warn_with({hub_logs = true}, string.format("Device reported a min color temperature %d K that is not lower than the reported max color temperature %d K", temp_in_kelvin, max))
-    end
+  elseif max_mired_bound > min_mired_bound then
+    local endpoint = zb_rx.address_header.src_endpoint.value
+    local max_kelvin_bound = switch_utils.convert_mired_to_kelvin(min_mired_bound)
+    local min_kelvin_bound = switch_utils.convert_mired_to_kelvin(max_mired_bound)
+    device:emit_event_for_endpoint(endpoint, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = min_kelvin_bound, maximum = max_kelvin_bound}}))
+  else
+    device.log.warn_with({hub_logs = true}, string.format("Device reported a min color temperature %d Mireds that is not lower than the reported max color temperature %d Mireds", min_mired_bound, max_mired_bound))
   end
 end
 
--- a/drivers/SmartThings/zigbee-switch/src/stateless_handlers/can_handle.lua
+++ b/drivers/SmartThings/zigbee-switch/src/stateless_handlers/can_handle.lua
@@ -1,0 +1,14 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
+
+return function(opts, driver, device)
+  local can_handle = device:supports_capability(capabilities.statelessColorTemperatureStep)
+    or device:supports_capability(capabilities.statelessSwitchLevelStep)
+  if can_handle then
+    local subdriver = require("stateless_handlers")
+    return true, subdriver
+  end
+  return false
+end
--- a/drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
@@ -1,0 +1,56 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
+local st_utils = require "st.utils"
+local clusters = require "st.zigbee.zcl.clusters"
+local switch_utils = require "switch_utils"
+
+-- These values are the mired versions of the config bounds in the default profile (e.g. color-temp-bulb)
+local DEFAULT_MIRED_MAX_BOUND = 370 -- 2700 Kelvin (Mireds are the inverse of Kelvin)
+local DEFAULT_MIRED_MIN_BOUND = 154 -- 6500 Kelvin (Mireds are the inverse of Kelvin)
+
+-- Transition Time: The time that shall be taken to perform the step change, in units of 1/10ths of a second.
+local TRANSITION_TIME = 3 -- default: 0.3 seconds
+
+-- Options Mask & Override: Indicates which options are being overridden by the Level/ColorControl cluster commands
+local OPTIONS_MASK = 0x01 -- default: The `ExecuteIfOff` option is overriden
+local IGNORE_COMMAND_IF_OFF = 0x00 -- default: the command will not be executed if the device is off
+
+local function step_color_temperature_by_percent_handler(driver, device, cmd)
+  local step_percent_change = cmd.args and cmd.args.stepSize or 0
+  if step_percent_change == 0 then return end
+  -- Reminder, stepSize > 0 == Kelvin UP == Mireds DOWN. stepSize < 0 == Kelvin DOWN == Mireds UP
+  local step_mode = (step_percent_change > 0) and clusters.ColorControl.types.CcStepMode.DOWN or clusters.ColorControl.types.CcStepMode.UP
+  local min_mireds = device:get_field(switch_utils.MIRED_MIN_BOUND)
+  local max_mireds = device:get_field(switch_utils.MIRED_MAX_BOUND)
+  -- since colorTemperatureRange is only set after both custom bounds are, use defaults if any custom bound is missing
+  if not (min_mireds and max_mireds) then
+    min_mireds = DEFAULT_MIRED_MIN_BOUND
+    max_mireds = DEFAULT_MIRED_MAX_BOUND
+  end
+  local step_size_in_mireds = st_utils.round((max_mireds - min_mireds) * (math.abs(step_percent_change)/100.0))
+  device:send(clusters.ColorControl.server.commands.StepColorTemperature(device, step_mode, step_size_in_mireds, TRANSITION_TIME, min_mireds, max_mireds, OPTIONS_MASK, IGNORE_COMMAND_IF_OFF))
+end
+
+local function step_level_handler(driver, device, cmd)
+  local step_size = st_utils.round((cmd.args and cmd.args.stepSize or 0)/100.0 * 254)
+  if step_size == 0 then return end
+  local step_mode = (step_size > 0) and clusters.Level.types.MoveStepMode.UP or clusters.Level.types.MoveStepMode.DOWN
+  device:send(clusters.Level.server.commands.Step(device, step_mode, math.abs(step_size), TRANSITION_TIME, OPTIONS_MASK, IGNORE_COMMAND_IF_OFF))
+end
+
+local stateless_handlers = {
+  NAME = "Zigbee Stateless Step Handlers",
+  capability_handlers = {
+    [capabilities.statelessColorTemperatureStep.ID] = {
+      [capabilities.statelessColorTemperatureStep.commands.stepColorTemperatureByPercent.NAME] = step_color_temperature_by_percent_handler,
+    },
+    [capabilities.statelessSwitchLevelStep.ID] = {
+      [capabilities.statelessSwitchLevelStep.commands.stepLevel.NAME] = step_level_handler,
+    },
+  },
+  can_handle = require("stateless_handlers.can_handle")
+}
+
+return stateless_handlers
--- a/drivers/SmartThings/zigbee-switch/src/sub_drivers.lua
+++ b/drivers/SmartThings/zigbee-switch/src/sub_drivers.lua
@@ -35,5 +35,6 @@ return {
     lazy_load_if_possible("tuya-multi"),
     lazy_load_if_possible("frient"),
     lazy_load_if_possible("frient-IO"),
-    lazy_load_if_possible("color_temp_range_handlers")
+    lazy_load_if_possible("color_temp_range_handlers"),
+    lazy_load_if_possible("stateless_handlers")
 }
--- a/drivers/SmartThings/zigbee-switch/src/switch_utils.lua
+++ b/drivers/SmartThings/zigbee-switch/src/switch_utils.lua
@@ -1,7 +1,18 @@
 -- Copyright 2025 SmartThings, Inc.
 -- Licensed under the Apache License, Version 2.0
 
+local st_utils = require "st.utils"
+
 local switch_utils = {}
+
+switch_utils.MIRED_MAX_BOUND = "__max_mired_bound"
+switch_utils.MIRED_MIN_BOUND = "__min_mired_bound"
+
+switch_utils.MIREDS_CONVERSION_CONSTANT = 1000000
+
+switch_utils.convert_mired_to_kelvin = function(mired)
+  return st_utils.round(switch_utils.MIREDS_CONVERSION_CONSTANT / mired)
+end
 
 switch_utils.emit_event_if_latest_state_missing = function(device, component, capability, attribute_name, value)
   if device:get_latest_state(component, capability.ID, attribute_name) == nil then
--- a/drivers/SmartThings/zigbee-switch/src/test/test_sengled_color_temp_bulb.lua
+++ b/drivers/SmartThings/zigbee-switch/src/test/test_sengled_color_temp_bulb.lua
@@ -4,6 +4,7 @@
 local test = require "integration_test"
 local t_utils = require "integration_test.utils"
 local clusters = require "st.zigbee.zcl.clusters"
+local capabilities = require "st.capabilities"
 local zigbee_test_utils = require "integration_test.zigbee_test_utils"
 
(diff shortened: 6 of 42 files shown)


drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,45 @@
...
+-- Options Mask & Override: Indicates which options are being overriden by the Level/ColorControl cluster commands
+local OPTIONS_MASK = 0x01 -- default: The `ExecuteIfOff` option is overriden
+local IGNORE_COMMAND_IF_OFF = 0x00 -- default: the command will not be executed if the device is off
+
+local function step_color_temperature_by_percent_handler(driver, device, cmd)
+    local step_percent_change = cmd.args and cmd.args.stepSize or 0
+    if step_percent_change == 0 then return end
+    local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.CcStepMode and clusters.ColorControl.types.CcStepMode.DOWN or 3) or (clusters.ColorControl.types.CcStepMode and clusters.ColorControl.types.CcStepMode.UP or 1)
+    local min_mireds = device:get_field(constants.KELVIN_MIN) or constants.COLOR_TEMPERATURE_MIRED_MIN -- default min mireds
+    local max_mireds = device:get_field(constants.KELVIN_MAX) or constants.COLOR_TEMPERATURE_MIRED_MAX -- default max mireds
```
greens: I think you'd want to get the values from the `colorTemperature` `range` attribute instead. And those should be in Kelvin,

greens: Also `constants.COLOR_TEMPERATURE_MIRED_MIN` will only be defined in versions > 61 (same for `KELVIN_MAX`), so that's why you're getting errors.

You'll probably want to include some local values.

wkhenon: Would it make sense to move the shared values between the `stateless_handlers` and `color_temp_range_handlers` to `switch_utils.lua`?
```
local MIREDS_CONVERSION_CONSTANT = 1000000
local COLOR_TEMPERATURE_KELVIN_MAX = 15000
local COLOR_TEMPERATURE_KELVIN_MIN = 1000
local COLOR_TEMPERATURE_MIRED_MAX = utils.round(MIREDS_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN) -- 1000
local COLOR_TEMPERATURE_MIRED_MIN = utils.round(MIREDS_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX) -- 67
```

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,50 @@
...
+-- Options Mask & Override: Indicates which options are being overriden by the Level/ColorControl cluster commands
+local OPTIONS_MASK = 0x01 -- default: The `ExecuteIfOff` option is overriden
+local IGNORE_COMMAND_IF_OFF = 0x00 -- default: the command will not be executed if the device is off
+
+local function step_color_temperature_by_percent_handler(driver, device, cmd)
+    local step_percent_change = cmd.args and cmd.args.stepSize or 0
+    if step_percent_change == 0 then return end
+    local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.CcStepMode and clusters.ColorControl.types.CcStepMode.DOWN or 3) or (clusters.ColorControl.types.CcStepMode and clusters.ColorControl.types.CcStepMode.UP or 1)
+    local min_mireds = capabilities.colorTemperature.colorTemperatureRange.minimum or COLOR_TEMPERATURE_MIRED_MIN -- default min mireds
+    local max_mireds = capabilities.colorTemperature.colorTemperatureRange.maximum or COLOR_TEMPERATURE_MIRED_MAX -- default max mireds
```
hcarter-775: Isn't the `colorTemperatureRange` defined in kelvin? Since these are handled in mireds I believe these need to be converted. Also, the values of `colorTemperatureRange` are required in the capability definition, so doing it this way I do not think we'd need an "or" case

hcarter-775: Also (could be mistaken) but I think this implementation cannot be matched today in the native handlers, so if we want to map this, we'll have to use the previous get_field implementation. I don't really know what should be done with that but calling it out now.

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,55 @@
...
+-- Transition Time: The time that shall be taken to perform the step change, in units of 1/10ths of a second.
+local TRANSITION_TIME = 3 -- default: 0.3 seconds
+-- Options Mask & Override: Indicates which options are being overriden by the Level/ColorControl cluster commands
+local OPTIONS_MASK = 0x01 -- default: The `ExecuteIfOff` option is overriden
+local IGNORE_COMMAND_IF_OFF = 0x00 -- default: the command will not be executed if the device is off
+
+local function step_color_temperature_by_percent_handler(driver, device, cmd)
+    local step_percent_change = cmd.args and cmd.args.stepSize or 0
+    if step_percent_change == 0 then return end
+    local step_mode = step_percent_change > 0 and (clusters.ColorControl.types.CcStepMode and clusters.ColorControl.types.CcStepMode.DOWN or 3) or (clusters.ColorControl.types.CcStepMode and clusters.ColorControl.types.CcStepMode.UP or 1)
```
greens: is `clusters.ColorControl.types.CcStepMode` not always defined/available?

wkhenon: That's a good point, I originally added it to match the `handle_step_color_temperature_by_percent` Matter handling but I don't think there are any hub firmware versions where it wouldn't be defined; fixed with 4ad4ca3cad6c11d657b9ffa0b816d419d6355f06

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
hcarter-775: I think this is a good idea @wkhenon

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,54 @@
...
+local OPTIONS_MASK = 0x01 -- default: The `ExecuteIfOff` option is overriden
+local IGNORE_COMMAND_IF_OFF = 0x00 -- default: the command will not be executed if the device is off
+
+local function step_color_temperature_by_percent_handler(driver, device, cmd)
+    local step_percent_change = cmd.args and cmd.args.stepSize or 0
+    if step_percent_change == 0 then return end
+    local step_mode = step_percent_change > 0 and clusters.ColorControl.types.CcStepMode.DOWN or clusters.ColorControl.types.CcStepMode.UP
+    local color_temp_range = device:get_latest_state("main", capabilities.colorTemperature.ID, capabilities.colorTemperature.colorTemperatureRange.NAME);
+    local min_mireds = (color_temp_range and (device:get_field(KELVIN_MIN) or st_utils.round(MIREDS_CONVERSION_CONSTANT/color_temp_range.maximum))) or COLOR_TEMPERATURE_MIRED_MIN
+    local max_mireds = (color_temp_range and (device:get_field(KELVIN_MAX) or st_utils.round(MIREDS_CONVERSION_CONSTANT/color_temp_range.minimum))) or COLOR_TEMPERATURE_MIRED_MAX
```
hcarter-775: Just because `color_temp_range` is defined does not necessary mean that KELVIN_MAX/KELVIN_MIN are defined I don't think, since this attribute can be set in the profile config prior to driver setup.

Also, KELVIN_MAX and KELVIN_MIN are not defined in mireds.

wkhenon: Good call —  The pattern that I'm thinking to solve this is as followed:
- First, try to use the values from the device
- Second, use the `color_temp_range` if it's available
- Lastly, use the default range that we defined

It can be captured in 3 legs of an `if-else` statement:
```
    local color_temp_range = device:get_latest_state("main", capabilities.colorTemperature.ID, capabilities.colorTemperature.colorTemperatureRange.NAME);
    local kelvin_min = device:get_field(KELVIN_MIN);
    local kelvin_max = device:get_field(KELVIN_MAX);

    local min_mireds
    local max_mireds
    if kelvin_min and kelvin_max then
        -- First tier: use device values if available
        min_mireds = st_utils.round(MIREDS_CONVERSION_CONSTANT / kelvin_max)
        max_mireds = st_utils.round(MIREDS_CONVERSION_CONSTANT / kelvin_min)
    elseif color_temp_range then
        -- Second tier: use color_temp_range if available
        min_mireds = st_utils.round(MIREDS_CONVERSION_CONSTANT / color_temp_range.maximum)
        max_mireds = st_utils.round(MIREDS_CONVERSION_CONSTANT / color_temp_range.minimum)
    else
        -- Third tier: use defaults
        min_mireds = COLOR_TEMPERATURE_MIRED_MIN
        max_mireds = COLOR_TEMPERATURE_MIRED_MAX
    end
```
Or if we want to evaluate the `min_mireds` and `max_mireds` independently:
```
    local color_temp_range = device:get_latest_state("main", capabilities.colorTemperature.ID, capabilities.colorTemperature.colorTemperatureRange.NAME);
    local kelvin_min = device:get_field(KELVIN_MIN);
    local kelvin_max = device:get_field(KELVIN_MAX);

    local min_mireds = (kelvin_min and kelvin_max) and st_utils.round(MIREDS_CONVERSION_CONSTANT / kelvin_max) or (color_temp_range and st_utils.round(MIREDS_CONVERSION_CONSTANT / color_temp_range.maximum)) or COLOR_TEMPERATURE_MIRED_MIN
    local max_mireds = (kelvin_min and kelvin_max) and st_utils.round(MIREDS_CONVERSION_CONSTANT / kelvin_min) or (color_temp_range and st_utils.round(MIREDS_CONVERSION_CONSTANT / color_temp_range.minimum)) or COLOR_TEMPERATURE_MIRED_MAX

```

hcarter-775: Ok the 3-tier if-else makes sense to me :+1:

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,64 @@
...
+local IGNORE_COMMAND_IF_OFF = 0x00 -- default: the command will not be executed if the device is off
+
+local function step_color_temperature_by_percent_handler(driver, device, cmd)
+    local step_percent_change = cmd.args and cmd.args.stepSize or 0
+    if step_percent_change == 0 then return end
+    local step_mode = step_percent_change > 0 and clusters.ColorControl.types.CcStepMode.DOWN or clusters.ColorControl.types.CcStepMode.UP
+
+    local color_temp_range = device:get_latest_state("main", capabilities.colorTemperature.ID, capabilities.colorTemperature.colorTemperatureRange.NAME);
+    local kelvin_min = device:get_field(switch_utils.KELVIN_MIN);
+    local kelvin_max = device:get_field(switch_utils.KELVIN_MAX);
```
greens: these are only used to store one value until we get the second, the range attribute's value should be considered canonical over these.

drivers/SmartThings/zigbee-switch/src/stateless_handlers/can_handle.lua
```diff
@@ -0,0 +1,14 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
+
+return function(opts, driver, device)
+    local can_handle = device:supports_capability(capabilities.statelessColorTemperatureStep)
```
greens: nit: we use 2-space tabs

greens (approved):

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
hcarter-775: I think this speaks to a small miss in how these fields are stored. They should be persisted. Otherwise, we cannot guarantee their existence for native handlers.

Therefore, this PR (or the native handler PR) should change these setters to:
```
  device:set_field(KELVIN_MAX..endpoint, temp_in_kelvin, {persist = true})
```
and
```
  device:set_field(KELVIN_MAX..endpoint, temp_in_kelvin, {persist = true})
```

hcarter-775 (approved):

drivers/SmartThings/zigbee-switch/src/sub_drivers.lua
```diff
@@ -35,5 +35,6 @@ return {
     lazy_load_if_possible("tuya-multi"),
     lazy_load_if_possible("frient"),
     lazy_load_if_possible("frient-IO"),
-    lazy_load_if_possible("color_temp_range_handlers")
+    lazy_load_if_possible("color_temp_range_handlers"),
+    lazy_load_if_possible("stateless_handlers")
```
tpmanley, line 39: Do we need a version check here because older versions won't know about the new stateless capabilities?

drivers/SmartThings/zigbee-switch/src/test/test_rgbw_bulb.lua
```diff
@@ -149,6 +150,60 @@ test.register_message_test(
...
+    },
+    {
+      channel = "zigbee",
+      direction = "send",
+      message = {
+        mock_device.id,
+        Level.server.commands.Step(mock_device, Level.types.MoveStepMode.UP, 254, 3, 0x01, 0x00)
+      }
+    }
+  }
```
tpmanley, line 149: Should probably add the `min_api_version` to these tests like in the other tests

drivers/SmartThings/zigbee-switch/src/sub_drivers.lua
hcarter-775: Hub knowledge of capabilities is not defined by scripting engine / hub fw version

drivers/SmartThings/zigbee-switch/src/test/test_rgbw_bulb.lua
hcarter-775: @cjswedes @aleclorimer Can you remind me what these checks are for? These capabilities should not be gated by the api version 19 I don't think. Is that what `min_api_version` is saying?

cjswedes: Min api version specifies the minimum lua lib api version that we will run against. So I think this is needed, since the stateless step caps are newer, and are not present in old lua libs artifacts.

hcarter-775: added to all new tests

drivers/SmartThings/zigbee-switch/src/switch_utils.lua
```diff
@@ -1,8 +1,17 @@
 -- Copyright 2025 SmartThings, Inc.
 -- Licensed under the Apache License, Version 2.0
+local utils = require "st.utils"
 
 local switch_utils = {}
 
+switch_utils.KELVIN_MAX = "_max_kelvin"
+switch_utils.KELVIN_MIN = "_min_kelvin"
+switch_utils.MIREDS_CONVERSION_CONSTANT = 1000000
+switch_utils.COLOR_TEMPERATURE_KELVIN_MAX = 15000
```
tpmanley: I think we need to adjust these max/min's similar to what you did or were going to do for Matter

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,64 @@
...
+-- Transition Time: The time that shall be taken to perform the step change, in units of 1/10ths of a second.
+local TRANSITION_TIME = 3 -- default: 0.3 seconds
+-- Options Mask & Override: Indicates which options are being overriden by the Level/ColorControl cluster commands
+local OPTIONS_MASK = 0x01 -- default: The `ExecuteIfOff` option is overriden
+local IGNORE_COMMAND_IF_OFF = 0x00 -- default: the command will not be executed if the device is off
+
+local function step_color_temperature_by_percent_handler(driver, device, cmd)
+  local step_percent_change = cmd.args and cmd.args.stepSize or 0
+  if step_percent_change == 0 then return end
+  local step_mode = step_percent_change > 0 and clusters.ColorControl.types.CcStepMode.DOWN or clusters.ColorControl.types.CcStepMode.UP
```
tpmanley: It'd be helpful to add a comment here that a step up in Kelvin is a step down in mireds

drivers/SmartThings/zigbee-switch/src/switch_utils.lua
hcarter-775: The Matter change (seen [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/144ddea03e4500f445864f52c25a9e49bbd6d2b7)) does not actually alter the max/min values- only the defaults used as max and min for the step level.

One reason I haven't done that in this PR is because I'm not sure what the correct default should be. In matter there is a relatively clear default from my perspective, but not here.

tpmanley: I think the same range would be fine to use here. It's going to be closer to reality than 1000-15000

drivers/SmartThings/zigbee-switch/src/test/test_rgbw_bulb.lua
hcarter-775: Talked to @aleclorimer about this and I'm not sure we should be gating what tests can be run based on the capabilities in old lua libs artifacts, since that isn't an accurate representation of actual hub characteristics.

I am going to leave these for now since it matches the current behavior across drivers, but in a vacuum I don't think this test should be gated by api 19.

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,64 @@
...
+
+  local color_temp_range = device:get_latest_state("main", capabilities.colorTemperature.ID, capabilities.colorTemperature.colorTemperatureRange.NAME);
+  local kelvin_min = device:get_field(switch_utils.KELVIN_MIN);
+  local kelvin_max = device:get_field(switch_utils.KELVIN_MAX);
```
nickolas-deboom: These fields are set with the endpoint appended to the field name, but are being read here without the endpoint.

nickolas-deboom: Also noticed that the 3 lines above here have semicolons at the end that should be removed

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,64 @@
...
+    max_mireds = st_utils.round(switch_utils.MIREDS_CONVERSION_CONSTANT / color_temp_range.minimum)
+  elseif kelvin_min and kelvin_max then
+    -- Second tier: use device values if available
+    min_mireds = st_utils.round(switch_utils.MIREDS_CONVERSION_CONSTANT / kelvin_max)
+    max_mireds = st_utils.round(switch_utils.MIREDS_CONVERSION_CONSTANT / kelvin_min)
+  else
+    -- Third tier: use defaults
+    min_mireds = switch_utils.COLOR_TEMPERATURE_MIRED_MIN
+    max_mireds = switch_utils.COLOR_TEMPERATURE_MIRED_MAX
+  end
```
nickolas-deboom: It would probably be worth adding test cases to cover tiers 1 & 2

drivers/SmartThings/zigbee-switch/src/switch_utils.lua
hcarter-775: Done!

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
hcarter-775: done!

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
hcarter-775: done to both. This was a bit of a weird issue- in matter we have the component_to_endpoint logic to get endpoints from a component, but in zigbee we don't. For now, I removed the endpoint logic altogether, so we cannot support different bounds across parent/child zigbee bulbs. I think this is an ok trade-off for the work it might take to add new infrastructure to handle this.

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
hcarter-775: I removed the tier 1 case... since I now have us persist the max/min fields there is never a case where the tier 1 exists and tier 2 doesn't. 

That said, I added a test for the tier 2 (now tier 1) case

drivers/SmartThings/zigbee-switch/src/color_temp_range_handlers/init.lua
```diff
@@ -4,55 +4,63 @@
...
-local MIREDS_CONVERSION_CONSTANT = 1000000
-local COLOR_TEMPERATURE_KELVIN_MAX = 15000
-local COLOR_TEMPERATURE_KELVIN_MIN = 1000
-local COLOR_TEMPERATURE_MIRED_MAX = utils.round(MIREDS_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MIN) -- 1000
-local COLOR_TEMPERATURE_MIRED_MIN = utils.round(MIREDS_CONVERSION_CONSTANT/COLOR_TEMPERATURE_KELVIN_MAX) -- 67
+local switch_utils = require "switch_utils"
+
+-- These values are a "sanity check" to ensure that max/min values we are getting are reasonable
+local COLOR_TEMPERATURE_MIRED_MAX = 1000 -- 15000 Kelvin
+local COLOR_TEMPERATURE_MIRED_MIN = 67 -- 1000 Kelvin
```
hcarter-775: A note to reviewers- while looking at this all, I started noticing that while we are saving these fields in Kelvin, almost all functionality involving these occurs in Mireds. Therefore, I have altered the flow to more directly use mireds, only converting into kelvin at the specific `colorTemperatureRange` step.

I believe this should simplify the constant conversion back and forth.

drivers/SmartThings/zigbee-switch/src/stateless_handlers/can_handle.lua
```diff
@@ -0,0 +1,14 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
```
aleclorimer, line 4: nit: Move this inside the function call for slight memory savings

drivers/SmartThings/zigbee-switch/src/test/test_rgbw_bulb.lua
aleclorimer: There's certainly going to be a weird grey area/time where tests would run on earlier versions of lua-libs than what is set by the `min_api_version`. If you would like to test this against earlier versions of lua_libs then update the min_api_version to the versions that you tested against, you could do that. You could also run the `tools/update_min_tags.py --filter <driver>` with the driver and it will update those values automatically (with a bit of lua-libs modifications).

drivers/SmartThings/zigbee-switch/src/stateless_handlers/can_handle.lua
hcarter-775: I'm not sure that will cause savings in this case, since st.capabilities is already required [at the top](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-switch/src/init.lua#L5) of the main init.lua file.

drivers/SmartThings/zigbee-switch/src/test/test_rgbw_bulb.lua
hcarter-775: I'm not sure that should be the case, as this blocks all automatic pre-60 testing from matter switch, for example? And since new capabilities (with commands) are always defined in the top-level init file today, this means that anytime we create a new one and implement it, we can never automatically run tests against lower lua libs. This seems like a strange, artificial requirement to me.  

For example, if we had older fw tests for switch, the fix implemented [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2865) would have been caught in CI

drivers/SmartThings/zigbee-switch/src/color_temp_range_handlers/init.lua
```diff
@@ -3,56 +3,59 @@
...
```
nickolas-deboom:
```suggestion
local COLOR_TEMPERATURE_MIRED_MAX = 1000 -- 1000 Kelvin
local COLOR_TEMPERATURE_MIRED_MIN = 67 -- 15000 Kelvin
```

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,51 @@
...
+
+local function step_level_handler(driver, device, cmd)
+  local step_size = st_utils.round((cmd.args and cmd.args.stepSize or 0)/100.0 * 254)
+  if step_size == 0 then return end
+  local step_mode = (step_size > 0) and clusters.Level.types.MoveStepMode.UP or clusters.Level.types.MoveStepMode.DOWN
+  device:send(clusters.Level.server.commands.Step(device, step_mode, math.abs(step_size), TRANSITION_TIME, OPTIONS_MASK, IGNORE_COMMAND_IF_OFF))
+end
+
+local stateless_handlers = {
+  Name = "Zigbee Stateless Step Handlers",
```
nickolas-deboom:
```suggestion
  NAME = "Zigbee Stateless Step Handlers",
```
not 100% sure this is necessary but its in all caps in other subdrivers

drivers/SmartThings/zigbee-switch/src/test/test_sengled_color_temp_bulb.lua
```diff
@@ -151,4 +155,64 @@ test.register_coroutine_test(
...
+  end,
+  {
+    min_api_version = 19
+  }
+)
+
+test.register_coroutine_test(
+  "Step Level command test",
+  function()
+    test.timer.__create_and_queue_test_time_advance_timer(1, "oneshot")
```
nickolas-deboom: im not sure if the timers in some of the new test cases are needed since it looks like they dont get advanced during the test

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,51 @@
...
+local switch_utils = require "switch_utils"
+
+-- These values are the mired versions of the config bounds in the default profiles
+local DEFAULT_MIRED_MAX_BOUND = 455 -- 2200 Kelvin (Mireds are the inverse of Kelvin)
+local DEFAULT_MIRED_MIN_BOUND = 154 -- 6500 Kelvin (Mireds are the inverse of Kelvin)
+
+-- Transition Time: The time that shall be taken to perform the step change, in units of 1/10ths of a second.
+local TRANSITION_TIME = 3 -- default: 0.3 seconds
+
+-- Options Mask & Override: Indicates which options are being overriden by the Level/ColorControl cluster commands
```
nickolas-deboom:
```suggestion
-- Options Mask & Override: Indicates which options are being overridden by the Level/ColorControl cluster commands
```

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -0,0 +1,51 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
+local st_utils = require "st.utils"
+local clusters = require "st.zigbee.zcl.clusters"
+local switch_utils = require "switch_utils"
+
+-- These values are the mired versions of the config bounds in the default profiles
```
nickolas-deboom: Which default profiles is this comment referring to? `color-temp-bulb.yml` uses 2700K for example. I'm guessing this really just means a default range that would represenet the widest supported color temp range across all of the profiles?

nickolas-deboom (approved):
Left some minor comments but the rest looks good to me 👍

drivers/SmartThings/zigbee-switch/src/test/test_sengled_color_temp_bulb.lua
hcarter-775: removed a couple of these

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
hcarter-775: Just made some updates, I'd originally used the Matter default profiles (which use 2200) as the basis, but decided to use the default Zigbee bounds instead, like you'd pointed out.

aleclorimer (approved):

drivers/SmartThings/zigbee-switch/src/color_temp_range_handlers/init.lua
```diff
@@ -3,56 +3,59 @@
...
-  local temp_in_kelvin = utils.round(MIREDS_CONVERSION_CONSTANT / temp_in_mired)
-  device:set_field(KELVIN_MIN..endpoint, temp_in_kelvin)
-  local max = device:get_field(KELVIN_MAX..endpoint)
-  if max ~= nil then
-    if temp_in_kelvin < max then
-      device:emit_event_for_endpoint(endpoint, capabilities.colorTemperature.colorTemperatureRange({ value = {minimum = temp_in_kelvin, maximum = max}}))
-    else
-      device.log.warn_with({hub_logs = true}, string.format("Device reported a min color temperature %d K that is not lower than the reported max color temperature %d K", temp_in_kelvin, max))
-    end
+  device:set_field(switch_utils.MIRED_MAX_BOUND, max_mired_bound, {persist = true})
```
cjswedes, line 46: Whats the reason for removing the endpoint from these field names? Same here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/82cdfe213802a9b360bb18e2491525dcc105d5b5/drivers/SmartThings/zigbee-switch/src/color_temp_range_handlers/init.lua#L21

cjswedes (approved):

drivers/SmartThings/zigbee-switch/src/color_temp_range_handlers/init.lua
hcarter-775: oops, I resolved the conversation where this was discussed. See [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2818#discussion_r3024539282)

PR #2826, opened 2026-03-06
Matrix testing, don't merge it

aleclorimer:
DO NOT MERGE THIS

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -196,7 +196,7 @@ jobs:
   publish:
     uses: ./.github/workflows/publish-test-results.yml
     if: ${{ always() }}
-    needs: [ run-driver-tests]
+    needs: [ run-driver-tests ]
 
   event-file:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


.github/workflows/run-tests.yml
```diff
@@ -9,48 +9,75 @@ jobs:
...
+      fail-fast: false
+
     steps:
-      - name: Find the lua lib release version
+      - name: Find the lua lib release version for 0.${{ matrix.version}}
         id: lib-version
         run: |
-          curl "https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/releases/latest" -s -L -I -o /dev/null -w '%{url_effective}' > test.log
-          echo "url=$(cat test.log)" >> $GITHUB_OUTPUT
+          curl -s \
```
varzac, line 28: This looks like you are trying to get these from the Edge Drivers uploaded artifacts, and not from artifactory.  I think we'll want to get them from artifactory because 1) that is truly authoritative, 2) it will be easier than this pagination stuff since they are already organized by hubcore release version and 3) for backporting the test code to support tagged test version, we won't be uploading those to the edge drivers repo, instead we will just upload them to artifactory

aleclorimer: Is the artifactory route working? I updated my test for a bit to try to run the artifactory route and it would always fail with an authentication error. I don't see `get-dev-artifact` running for beta, production or regular drivers.

PR #2858, opened 2026-03-30
Adding support for Hager WAASYS devices

JanJakubiszyn:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [X] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests
Tests performed on the physical device.

### DIFF
--- a/drivers/SmartThings/matter-switch/profiles/motion-illuminance.yml
+++ b/drivers/SmartThings/matter-switch/profiles/motion-illuminance.yml
@@ -1,0 +1,14 @@
+name: motion-illuminance
+components:
+- id: main
+  capabilities:
+  - id: motionSensor
+    version: 1
+  - id: illuminanceMeasurement
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: MotionSensor
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -1,4 +1,4 @@
--- Copyright © 2025 SmartThings, Inc.
+-- Copyright © 2026 SmartThings, Inc.
 -- Licensed under the Apache License, Version 2.0
 
 local MatterDriver = require "st.matter.driver"
@@ -374,6 +374,7 @@ local matter_driver_template = {
     switch_utils.lazy_load_if_possible("sub_drivers.aqara_cube"),
     switch_utils.lazy_load("sub_drivers.camera"),
     switch_utils.lazy_load_if_possible("sub_drivers.eve_energy"),
+    switch_utils.lazy_load_if_possible("sub_drivers.hager"),
     switch_utils.lazy_load_if_possible("sub_drivers.ikea_scroll"),
     switch_utils.lazy_load_if_possible("sub_drivers.third_reality_mk1")
   },
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/hager/can_handle.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/hager/can_handle.lua
@@ -1,0 +1,13 @@
+-- Copyright © 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+return function(opts, driver, device, ...)
+    local device_lib = require "st.device"
+    local get_product_override_field = require "switch_utils.utils".get_product_override_field
+
+    local checked_device = device.network_type == device_lib.NETWORK_TYPE_MATTER and device or device:get_parent_device()
+    if get_product_override_field(checked_device, "needs_hager_subdriver") then
+        return true, require("sub_drivers.hager")
+    end
+    return false
+end
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
@@ -1,0 +1,475 @@
+-- Copyright © 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
+local clusters = require "st.matter.clusters"
+local cluster_base = require "st.matter.cluster_base"
+local device_lib = require "st.device"
+local buttonCfg = require "switch_utils.device_configuration".ButtonCfg
+local create_child = require "switch_utils.device_configuration".ChildCfg.create_or_update_child_devices
+local switch_utils = require "switch_utils.utils"
+local fields = require "switch_utils.fields"
+
+local ACTIVE_EPS = "__active_EPS"
+local MATTER_DEVICE_ID = "MATTER_DEVICE_ID"
+local PARENT_ID = "PARENT_ID"
+local CURRENT_LIFT = "__current_lift"
+local BUTTON_EPS = "__button_eps"
+local MAIN_WC_EP = "__main_wc_ep"
+
+local PRODUCT_ID = {
+  SWITCH_1G = 0x0005,
+  SWITCH_2G = 0x0006,
+  PIR_1_1M = 0x0007,
+  PIR_2_2M = 0x000A,
+}
+
+local function subscribe (device, endpoint_id, cluster_id, attr_id, event_id)
+  device:send(cluster_base.subscribe(device, endpoint_id, cluster_id, attr_id, event_id))
+end
+
+local function get_parent (driver, device)
+  return driver:get_device_info(device:get_field(PARENT_ID) or nil)
+end
+
+local function get_matter_device(device)
+  local matter_device_id = device:get_field(MATTER_DEVICE_ID)
+  if matter_device_id then
+    local driver = device.driver
+    if driver then
+      local matter_device = driver:get_device_info(matter_device_id)
+      if matter_device then
+        return matter_device
+      end
+    end
+  end
+  return nil
+end
+
+local function extract_reported_endpoints(ib)
+  local eps = {}
+  if ib.data and ib.data.elements then
+    for _, el in ipairs(ib.data.elements) do
+      local ep_id = el.value
+      if type(ep_id) == "number" and ep_id ~= 1 and ep_id ~= 2 then
+        table.insert(eps, ep_id)
+      end
+    end
+  end
+  return eps
+end
+
+local function detect_endpoint_changes (device, ib_elements)
+  local stored_eps = device:get_field(ACTIVE_EPS) or {}
+  ib_elements = ib_elements or {}
+
+  local old_set, new_set = {}, {}
+  for _, ep_id in ipairs(stored_eps) do
+    old_set[ep_id] = true
+  end
+  for _, ep_id in ipairs(ib_elements) do
+    new_set[ep_id] = true
+  end
+
+  local removed_eps, added_eps = {}, {}
+
+  for ep_id in pairs(old_set) do
+    if not new_set[ep_id] then
+      table.insert(removed_eps, ep_id)
+    end
+  end
+  for ep_id in pairs(new_set) do
+    if not old_set[ep_id] then
+      table.insert(added_eps, ep_id)
+    end
+  end
+  return removed_eps, added_eps
+end
+
+local function assign_profile_for_endpoint(device_type_id)
+  local profile = fields.device_type_profile_map[device_type_id] or "switch-binary"
+  return function(device, ep_id, is_child_device)
+    return profile, nil
+  end
+end
+
+local function link_matter_device_and_parent(matter_device)
+  local parent = matter_device:get_parent_device()
+  matter_device:set_field(PARENT_ID, parent.id, { persist = true })
+  matter_device:set_field(MATTER_DEVICE_ID, matter_device.id, { persist = true })
+  parent:set_field(PARENT_ID, parent.id, { persist = true })
+  parent:set_field(MATTER_DEVICE_ID, matter_device.id, { persist = true })
+end
+
+local function handle_set_preset(driver, device, cmd)
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local lift_value = device.preferences.presetPosition
+  local hundredths_lift_percent = (100 - tonumber(lift_value)) * 100
+  device:send(clusters.WindowCovering.server.commands.GoToLiftPercentage(device, endpoint_id, hundredths_lift_percent))
+end
+
+local function handle_close(driver, device, cmd)
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local req = clusters.WindowCovering.server.commands.DownOrClose(device, endpoint_id)
+  if device.preferences.reverse then
+    req = clusters.WindowCovering.server.commands.UpOrOpen(device, endpoint_id)
+  end
+  device:send(req)
+end
+
+local function handle_open(driver, device, cmd)
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local req = clusters.WindowCovering.server.commands.UpOrOpen(device, endpoint_id)
+  if device.preferences.reverse then
+    req = clusters.WindowCovering.server.commands.DownOrClose(device, endpoint_id)
+  end
+  device:send(req)
+end
+
+local function handle_pause(driver, device, cmd)
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local req = clusters.WindowCovering.server.commands.StopMotion(device, endpoint_id)
+  device:send(req)
+end
+
+local function handle_shade_level(driver, device, cmd)
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local lift_percentage_value = 100 - cmd.args.shadeLevel
+  local hundredths_lift_percentage = lift_percentage_value * 100
+  local req = clusters.WindowCovering.server.commands.GoToLiftPercentage(
+    device, endpoint_id, hundredths_lift_percentage
+  )
+  device:send(req)
+end
+
+local current_pos_handler = function(attribute)
+  return function(driver, device, ib, response)
+    if ib.data.value == nil then
+      return
+    end
+    local windowShade = capabilities.windowShade.windowShade
+    local position = 100 - math.floor(ib.data.value / 100)
+    local reverse = device.preferences.reverse
+    device:emit_event_for_endpoint(ib.endpoint_id, attribute(position))
+    if attribute == capabilities.windowShadeLevel.shadeLevel then
+      device:set_field(CURRENT_LIFT, position)
+    end
+    local lift_position = device:get_field(CURRENT_LIFT)
+    if lift_position == 100 then
+      device:emit_event_for_endpoint(ib.endpoint_id, reverse and windowShade.closed() or windowShade.open())
+    elseif lift_position > 0 then
+      device:emit_event_for_endpoint(ib.endpoint_id, windowShade.partially_open())
+    elseif lift_position == 0 then
+      device:emit_event_for_endpoint(ib.endpoint_id, reverse and windowShade.open() or windowShade.closed())
+    end
+  end
+end
+
+local function current_status_handler(driver, device, ib, response)
+  local windowShade = capabilities.windowShade.windowShade
+  local reverse = device.preferences.reverse
+  local state = ib.data.value & clusters.WindowCovering.types.OperationalStatus.GLOBAL
+  if state == 1 then
+    device:emit_event_for_endpoint(ib.endpoint_id, reverse and windowShade.closing() or windowShade.opening())
+  elseif state == 2 then
+    device:emit_event_for_endpoint(ib.endpoint_id, reverse and windowShade.opening() or windowShade.closing())
+  elseif state ~= 0 then
+    device:emit_event_for_endpoint(ib.endpoint_id, windowShade.unknown())
+  end
+end
+
+local function info_changed(driver, device, event, args)
+  if not device or not device.id or not device.profile then
+    return
+  end
+
+  if device.profile.id ~= args.old_st_store.profile.id or device.network_type == device_lib.NETWORK_TYPE_CHILD then
+    local parent = device:get_parent_device()
+    local matter_device = get_matter_device(parent)
+    local map = {}
+    device.thread:call_with_delay(2, function()
+      if device:supports_capability(capabilities.button) then
+        local button_eps = parent:get_field(BUTTON_EPS)
+        local clean_eps = {}
+        table.sort(button_eps)
+
+        for _, v in ipairs(button_eps or {}) do
+          table.insert(clean_eps, v)
+        end
+
+        buttonCfg.update_button_component_map(matter_device, clean_eps[1], clean_eps)
+        for _, ep_id in ipairs(clean_eps) do
+          subscribe(device, ep_id, clusters.Switch.ID, nil, clusters.Switch.events.MultiPressComplete.ID)
+          subscribe(device, ep_id, clusters.Switch.ID, nil, clusters.Switch.events.LongPress.ID)
+          device:emit_event_for_endpoint(ep_id, capabilities.button.supportedButtonValues({ "pushed", "double", "held" }))
+        end
+        return
+      elseif device:supports_capability(capabilities.switch) then
+        map = { main = 3 }
+        subscribe(device, nil, clusters.OnOff.ID, clusters.OnOff.attributes.OnOff.ID, nil)
+        if device:supports_capability(capabilities.switchLevel) then
+          map = { main = 4 }
+          subscribe(device, nil, clusters.LevelControl.ID, clusters.LevelControl.attributes.CurrentLevel.ID, nil)
+          subscribe(device, nil, clusters.LevelControl.ID, clusters.LevelControl.attributes.MaxLevel.ID)
+          subscribe(device, nil, clusters.LevelControl.ID, clusters.LevelControl.attributes.MinLevel.ID)
+        end
+      elseif device:supports_capability(capabilities.motionSensor) then
+        subscribe(device, nil, clusters.OccupancySensing.ID, clusters.OccupancySensing.attributes.Occupancy.ID, nil)
+        subscribe(device, nil, clusters.IlluminanceMeasurement.ID, clusters.IlluminanceMeasurement.attributes.MeasuredValue.ID, nil)
+      elseif device:supports_capability(capabilities.windowShadeLevel) then
+        map = { main = 5 }
+        subscribe(device, nil, clusters.WindowCovering.ID, clusters.WindowCovering.attributes.OperationalStatus.ID)
+        subscribe(device, nil, clusters.WindowCovering.ID, clusters.WindowCovering.attributes.CurrentPositionLiftPercent100ths.ID, nil)
+      end
+      if device.id == matter_device.id then
+        parent:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, map, { persist = true })
+        matter_device:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, map, { persist = true })
+      end
+    end)
+  end
+end
+
+local function device_init (driver, device)
+  if device.network_type ~= device_lib.NETWORK_TYPE_MATTER then
+    device.thread:call_with_delay(4, function()
+      info_changed(driver, device, nil, {
+        old_st_store = { profile = {  } }
+      })
+    end)
+    return
+  end
+
+  if device:get_parent_device() ~= nil then
+    link_matter_device_and_parent(device)
+    local parent = device:get_parent_device()
+    device:extend_device("send", function(self, message)
+      return parent:send(message)
+    end)
+  else
+    subscribe(device, 2, clusters.Descriptor.ID, clusters.Descriptor.attributes.PartsList.ID)
+    device.thread:call_with_delay(3, function()
+      local matter_device = get_matter_device(device)
+      device:extend_device("emit_event_for_endpoint", function(self, ep_info, event)
+        local endpoint_id = type(ep_info) == "number" and ep_info or ep_info.endpoint_id
+        local child = self:get_child_by_parent_assigned_key(string.format("%d", endpoint_id))
+        if child then
+          return child:emit_event(event)
+        end
+        if matter_device then
+          return matter_device:emit_event_for_endpoint(ep_info, event)
+        end
+      end)
+    end)
+  end
+
+  device:set_component_to_endpoint_fn(switch_utils.component_to_endpoint)
+  device:set_endpoint_to_component_fn(switch_utils.endpoint_to_component)
+end
+
+local function handle_descriptor_report(driver, device, ib, response)
+  local product_id = device.manufacturer_info and device.manufacturer_info.product_id
+  local parent = get_parent(driver, device)
+
+  if not parent or ib.endpoint_id ~= 2 then
+    return
+  end
+
+  local matter_device = get_matter_device(device)
+
+  local new_eps = extract_reported_endpoints(ib) or {}
+  table.sort(new_eps)
+  local removed_eps, added_eps = detect_endpoint_changes(device, new_eps)
+
+  device:set_field(ACTIVE_EPS, new_eps, { persist = true })
+
+  for _, ep_id in ipairs(added_eps or {}) do
+    subscribe(parent, ep_id, clusters.Descriptor.ID, clusters.Descriptor.attributes.DeviceTypeList.ID, nil)
+
+    if product_id == PRODUCT_ID.SWITCH_1G and ep_id == 3 then
+      matter_device:try_update_metadata({ profile = "light-binary" })
+    elseif product_id == PRODUCT_ID.SWITCH_2G then
+      if ep_id == 3 then
+        local profile = switch_utils.tbl_contains(new_eps, 4) and "light-binary" or "2-button"
+        matter_device:try_update_metadata({ profile = profile })
+      elseif ep_id == 4 then
+        local profile = switch_utils.tbl_contains(new_eps, 3) and "light-binary" or "2-button"
+        matter_device:try_update_metadata({ profile = profile })
+      end
+    end
+  end
+
+  for _, ep_id in ipairs(removed_eps) do
+    local button_eps = parent:get_field(BUTTON_EPS) or {}
+    local clean_eps = {}
+
+    for _, value in ipairs(button_eps) do
+      if value ~= ep_id then
+        table.insert(clean_eps, value)
+      end
+    end
+    table.sort(clean_eps)
+    parent:set_field(BUTTON_EPS, clean_eps, { persist = true })
+    matter_device:set_field(BUTTON_EPS, clean_eps, { persist = true })
+
+    local child = parent:get_child_by_parent_assigned_key(tostring(ep_id))
+    if child then
+      driver:try_delete_device(child.id)
+    end
+
+    if ep_id == 3 then
+      if device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_1G then
+        matter_device:try_update_metadata({ profile = "2-button" })
+      elseif device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_2G then
+        local has_ep4 = switch_utils.tbl_contains(new_eps, 4)
+        matter_device:try_update_metadata({ profile = has_ep4 and "2-button" or "4-button" })
+      end
+    elseif ep_id == 4 then
+      if device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_2G then
+        local button_comb = switch_utils.tbl_contains(new_eps, 3)
+        if button_comb then
+          matter_device:try_update_metadata({ profile = "2-button" })
+          create_child(driver, parent, { 3 }, 1, assign_profile_for_endpoint(fields.DEVICE_TYPE_ID.LIGHT.ON_OFF))
+        else
+          matter_device:try_update_metadata({ profile = "4-button" })
+        end
+      end
+    end
+  end
+end
+
+local function device_type_handler (driver, device, ib)
+  local parent = get_parent(driver, device)
+  if not parent then
+    return
+  end
+
+  local matter_device = get_matter_device(device)
+  local stored_btn_eps = parent:get_field(BUTTON_EPS) or {}
+  local new_btn_eps = {}
+  local ep_id = ib.endpoint_id
+  local value = ib.data.elements
+
+  for _, v in ipairs(stored_btn_eps) do
+    table.insert(new_btn_eps, v)
+  end
+  for _, element in ipairs(value) do
+    local device_type_field = element.elements.device_type
+    local device_type_id = device_type_field and device_type_field.value
+
+    if device_type_id == fields.DEVICE_TYPE_ID.GENERIC_SWITCH then
+      if not switch_utils.tbl_contains(new_btn_eps, ep_id) then
+        switch_utils.set_field_for_endpoint(parent, fields.SUPPORTS_MULTI_PRESS, ep_id)
+        switch_utils.set_field_for_endpoint(parent, fields.IGNORE_NEXT_MPC, ep_id)
+        table.insert(new_btn_eps, ep_id)
+        table.sort(new_btn_eps)
+        parent:set_field(BUTTON_EPS, new_btn_eps, { persist = true })
+      end
+    end
+
+    if device_type_id == fields.DEVICE_TYPE_ID.LIGHT.ON_OFF then
+      local active_eps = device:get_field(ACTIVE_EPS)
+
+      if ep_id == 3 and device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_1G then
+        return
+      elseif ep_id == 4 and device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_2G then
+        if switch_utils.tbl_contains(active_eps, 3) then
+          local ep3 = parent:get_child_by_parent_assigned_key("3")
+          if ep3 then
+            driver:try_delete_device(ep3.id)
+          end
+        else
+          create_child(driver, parent, { 4 }, 1, assign_profile_for_endpoint(device_type_id))
+          return
+        end
+      elseif ep_id == 3 and device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_2G then
+        if not switch_utils.tbl_contains(active_eps, 4) then
+          create_child(driver, parent, { 3 }, 1, assign_profile_for_endpoint(device_type_id))
+        else
+          driver:try_delete_device(ep_id)
+        end
+        return
+      end
+      create_child(driver, parent, { ep_id }, 1, assign_profile_for_endpoint(device_type_id))
+    elseif device_type_id == fields.DEVICE_TYPE_ID.LIGHT.DIMMABLE then
+      if ep_id == 4 and device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_1G then
+        return
+      end
+      create_child(driver, parent, { ep_id }, 1, assign_profile_for_endpoint(device_type_id))
+    elseif device_type_id == fields.DEVICE_TYPE_ID.WINDOW_COVERING then
+      if matter_device:get_field(MAIN_WC_EP) == nil then
+        create_child(driver, parent, { ep_id }, 1, assign_profile_for_endpoint(device_type_id))
+      end
+    end
+  end
+end
+
+local function do_configure (driver, device)
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+
+    local wc_eps = device:get_endpoints(clusters.WindowCovering.ID)
+    local oc_eps = device:get_endpoints(clusters.OccupancySensing.ID)
+    local bt_eps = device:get_endpoints(clusters.Switch.ID)
+    local lvl_eps = device:get_endpoints(clusters.LevelControl.ID)
+    local product_id = device.manufacturer_info.product_id
+
+    table.sort(wc_eps)
+    table.sort(oc_eps)
+    table.sort(lvl_eps)
+
+    if #oc_eps > 0 then
+      device:try_update_metadata({ profile = "motion-illuminance" })
+    elseif #wc_eps > 0 and product_id == PRODUCT_ID.SWITCH_1G then
+      device:try_update_metadata({ profile = "window-covering" })
+      device:set_field(MAIN_WC_EP, wc_eps[1], { persist = true })
+    elseif #bt_eps == 4 then
+      device:try_update_metadata({ profile = "4-button" })
+    elseif #bt_eps == 2 then
+      device:try_update_metadata({ profile = "2-button" })
+    elseif #lvl_eps > 0 and product_id == PRODUCT_ID.SWITCH_1G then
+      device:try_update_metadata({ profile = "light-level" })
+    end
+    device:set_field(BUTTON_EPS, bt_eps, { persist = true })
+  end
+end
+
+local function added (driver, device) end
+local function driver_switched(driver, device) end
+
+local hager_switch = {
+  NAME = "Hager Subdriver",
+  lifecycle_handlers = {
+    added = added,
+    init = device_init,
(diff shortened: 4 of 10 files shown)


hcarter-775:
One way in which I think we could quickly simplify some of the Hager subdriver would be by require-ing some fields and functions from the main driver, and changing their inputs as necessary in the overriden driver, rather than rewriting everything completely from scratch.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -373,6 +373,7 @@ local matter_driver_template = {
     switch_utils.lazy_load_if_possible("sub_drivers.aqara_cube"),
     switch_utils.lazy_load("sub_drivers.camera"),
     switch_utils.lazy_load_if_possible("sub_drivers.eve_energy"),
+    switch_utils.lazy_load("sub_drivers.Hager"),
```
hcarter-775: nit: let's keep the lowercase formatting
```suggestion
    switch_utils.lazy_load("sub_drivers.hager"),
```

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -115,6 +115,12 @@ SwitchFields.vendor_overrides = {
   [0x1189] = { -- LEDVANCE_MANUFACTURER_ID
     [0x0891] = { target_profile = "switch-binary", initial_profile = "light-binary" },
   },
+  [0x1285] = { -- HAGER_MANUFACTURER_ID
+    [0x0005] = {}, -- Hager WAASYS 1g switch
+    [0x0006] = {}, -- Hager WAASYS 2g switch
+    [0x0007] = {}, -- Hager WAASYS PIR 1.1M
+    [0x000A] = {}, -- Hager WAASYS PIR 2.2M
+  },
```
hcarter-775, line 128:
```suggestion
  [0x1285] = { -- HAGER_MANUFACTURER_ID
    [0x0005] = { needs_hager_subdriver = true }, -- WAASYS 1g switch
    [0x0006] = { needs_hager_subdriver = true }, -- WAASYS 2g switch
    [0x0007] = { needs_hager_subdriver = true }, -- WAASYS PIR 1.1M
    [0x000A] = { needs_hager_subdriver = true }, -- WAASYS PIR 2.2M
  },
```

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -161,6 +161,14 @@ function utils.find_default_endpoint(device)
...
   end
 
+    -- Hager vendor special handling
+    if device.manufacturer_info.vendor_id == 0x1285 then
+      if #momentary_switch_ep_ids > 0 then
+        log.info("Hager device with buttons detected, using default endpoint")
+        return device.MATTER_DEFAULT_ENDPOINT
+      end
+    end
+
```
hcarter-775: I don't think this is required. With the rebase, it appears `momentary_switch_ep_ids` is nil here in fact, so I'm not sure we need this.

drivers/SmartThings/matter-switch/src/sub_drivers/Hager/init.lua
```diff
@@ -0,0 +1,642 @@
+-- Copyright © 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
+local clusters = require "st.matter.clusters"
+local cluster_base = require "st.matter.cluster_base"
+local device_lib = require "st.device"
+local utils = require "st.utils"
+local ButtonCfg = require "switch_utils.device_configuration"
```
hcarter-775:
```suggestion
local button_cfg = require "switch_utils.device_configuration".ButtonCfg
```
and update the appropriate sections here.

drivers/SmartThings/matter-switch/src/sub_drivers/Hager/init.lua
```diff
@@ -0,0 +1,642 @@
...
+end
+
+local function contains_ep (list, ep)
+    for _, v in ipairs(list) do
+        if v == ep then
+            return true
+        end
+    end
+    return false
+end
```
hcarter-775: Use `switch_utils.tbl_contains` instead.

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -259,7 +268,24 @@ function DeviceConfiguration.match_profile(driver, device)
...
   local server_onoff_ep_ids = device:get_endpoints(clusters.OnOff.ID) -- get_endpoints defaults to return EPs supporting SERVER or BOTH
   if #server_onoff_ep_ids > 0 then
-    ChildConfiguration.create_or_update_child_devices(driver, device, server_onoff_ep_ids, default_endpoint_id, SwitchDeviceConfiguration.assign_profile_for_onoff_ep)
+    if device.manufacturer_info.vendor_id == 0x1285 then
+    else
+      ChildConfiguration.create_or_update_child_devices(driver, device, server_onoff_ep_ids, default_endpoint_id, SwitchDeviceConfiguration.assign_profile_for_onoff_ep)
+    end
+  end
+
+    -- Hager vendor override checks
```
hcarter-775: I'm having a hard time understanding what these are doing... If you are trying to ignore doConfigure's normal logic, I might suggest that you just override it with an empty function or whatever you need in the subdriver?

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
hcarter-775: However, I'm not 100% sure what these are actually for, especially given my suggestions in utils.lua of removing the one piece of code and of overriding doConfigure.

drivers/SmartThings/matter-switch/src/sub_drivers/Hager/init.lua
```diff
@@ -0,0 +1,642 @@
...
+        for _, v in ipairs(stored) do
+            table.insert(button_endpoints, v)
+        end
+    end
+
+    for _, element in ipairs(value) do
+        local device_type_field = element.elements.device_type
+        local device_type_id = device_type_field and device_type_field.value
+
+        if device_type_id == 15 then
```
hcarter-775: This is a general suggestion, so I won't make it for all the cases, but I would strongly recommend using the `fields.DEVICE_TYPE_ID` constants instead of magic numbers

drivers/SmartThings/matter-switch/src/sub_drivers/Hager/init.lua
```diff
@@ -0,0 +1,642 @@
...
+local HOST_ID = "HOST_ID"
+local SUBHUB_ID = "SUBHUB_ID"
+local CURRENT_LIFT = "__current_lift"
+local REVERSE_POLARITY = "__reverse_polarity"
+local PRESET_LEVEL_KEY = "__preset_level_key"
+local BUTTON_EPS = "__button_eps"
+
+local function subscribe (device, endpoint_id, cluster_id, attr_id, event_id)
+    device:send(cluster_base.subscribe(device, endpoint_id, cluster_id, attr_id, event_id))
+end
```
hcarter-775: Rather than sending a subscribe every time you need something, I might suggest using `device:add_subscribed_attribute()` instead of this particular override, and then just using `device:subscribe()` at the end of whatever function requires a subscription to be sent.

drivers/SmartThings/matter-switch/src/sub_drivers/Hager/init.lua
```diff
@@ -0,0 +1,642 @@
...
+
+local function get_subhub (driver, device)
```
hcarter-775:
```suggestion
local function get_child_device (driver, device)
```

I am not sure where the `subhub` terminology comes from (so please feel free to push back), but since this is related to child devices, I might recommend moving the subhub terminology to child_device terminology.

drivers/SmartThings/matter-switch/src/sub_drivers/Hager/init.lua
```diff
@@ -0,0 +1,642 @@
...
+    return driver:get_device_info(device:get_field(SUBHUB_ID) or nil)
+end
+
+local function get_host (driver, device)
```
hcarter-775: similar to my comment about subhub -> child_device, I might suggest we use parent device instead of host as terminology. This will help keep this device contextualized to other Matter device functionality.

hcarter-775 (commented):
Thank you for implementing more of the pre-existing functionality. At this point, my main concern is with the way this device intersects with the main driver, all of which I've commented on at this point. As far as I can tell, we may not really have to interact with the main driver at all, and if so that would be my preference, to not include as many vendor checks in otherwise generic-ish code.

drivers/SmartThings/matter-switch/src/sub_drivers/Hager/init.lua
JanJakubiszyn: In my driver I called a functional matter type child device by "host" terminology as before the approach was to tunnel communication through it. Currently whole communication is based on the "subhub" device and host is not used anymore. I changed following names: 
subhub -> parent
host -> matter_device

drivers/SmartThings/matter-switch/profiles/motion-illuminance.yml
```diff
@@ -0,0 +1,14 @@
...
+  - id: motionSensor
+    version: 1
+  - id: illuminanceMeasurement
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: MotionSensor
```
hcarter-775: nit: newline

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -7,6 +7,7 @@ local version = require "version"
 local fields = require "switch_utils.fields"
 local switch_utils = require "switch_utils.utils"
 local embedded_cluster_utils = require "switch_utils.embedded_cluster_utils"
+local log = require "log"
```
JanJakubiszyn: to be deleted

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
JanJakubiszyn: These fields are created purely for the purpose of the can_handle function. There are other Hager devices which are from a different line of products and shouldn't be handled by my subdriver.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/can_handle.lua
```diff
@@ -0,0 +1,24 @@
...
+        return false
+    end
+    if device.network_type == device_lib.NETWORK_TYPE_MATTER
+            and device.manufacturer_info.vendor_id == 0x1285
+            and vendor_overrides[0x1285][device.manufacturer_info.product_id]
+    then
+        return true, require("sub_drivers.hager")
+    end
+    return false
+end
```
hcarter-775:
```suggestion
end

```

nit: newline

drivers/SmartThings/matter-switch/src/sub_drivers/hager/can_handle.lua
```diff
@@ -0,0 +1,24 @@
+-- Copyright © 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+return function(opts, driver, device, ...)
```
hcarter-775, line 4:
```suggestion
-- Licensed under the Apache License, Version 2.0

return function(opts, driver, device, ...)
```
nit: required formatting, thank you

drivers/SmartThings/matter-switch/src/sub_drivers/hager/can_handle.lua
hcarter-775:
```suggestion
    local device_lib = require "st.device"
    local get_product_override_field = require "switch_utils.utis".get_product_override_field

    local checked_device = device.network_type == device_lib.NETWORK_TYPE_MATTER and device or device:get_parent_device()

    if get_product_override_field(checked_device, "needs_hager_subdriver") then
      return true, require("sub_drivers.hager")
    end
    return false
```
This `get_product_override_field` function hooks directly into the `vendor_overrides` table for easy use. So if you add a value into it like in my other suggestion we can do this.

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -1,4 +1,4 @@
--- Copyright © 2025 SmartThings, Inc.
```
hcarter-775, line 1: nit: do not alter

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -146,17 +146,16 @@ function ButtonDeviceConfiguration.update_button_component_map(device, default_e
...
     if ep ~= default_endpoint_id then
-      local button_component = "button"
-      if #button_eps > 1 then
-        button_component = button_component .. component_num
+        local button_component = "button"
+        if #button_eps > 1 then
+          button_component = button_component .. component_num
+        end
+        component_map[button_component] = ep
       end
```
hcarter-775: nit: I think you added an extra tab

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -258,7 +257,7 @@ function DeviceConfiguration.match_profile(driver, device)
 
   local server_onoff_ep_ids = device:get_endpoints(clusters.OnOff.ID) -- get_endpoints defaults to return EPs supporting SERVER or BOTH
   if #server_onoff_ep_ids > 0 then
-    ChildConfiguration.create_or_update_child_devices(driver, device, server_onoff_ep_ids, default_endpoint_id, SwitchDeviceConfiguration.assign_profile_for_onoff_ep)
+      ChildConfiguration.create_or_update_child_devices(driver, device, server_onoff_ep_ids, default_endpoint_id, SwitchDeviceConfiguration.assign_profile_for_onoff_ep)
```
hcarter-775: same here, extra tab

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -146,17 +146,16 @@ function ButtonDeviceConfiguration.update_button_component_map(device, default_e
...
+        end
+        component_map[button_component] = ep
       end
-      component_map[button_component] = ep
-    end
   end
   device:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, component_map, {persist = true})
 end
 
-
```
hcarter-775, line 159: Since I'm already making suggestions, let's remove the extra line?

drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
```diff
@@ -160,7 +160,6 @@ function utils.find_default_endpoint(device)
     end
     return nil
   end
-
```
hcarter-775, line 163: let's maybe keep this line? Then this file just stays unaltered

hcarter-775 (approved):
There are two more minor comments that should be addressed imo, otherwise LGTM!

hcarter-775:
And reminder, make sure to squash your commits before merging. Thanks!

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
hcarter-775: bump on this

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
hcarter-775: bump on this

JanJakubiszyn: I have already removed 159th line. I'm not sure If this is still relevant.

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
JanJakubiszyn: Fixed. Thank you.

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -34,7 +34,9 @@ SwitchFields.DEVICE_TYPE_ID = {
   ELECTRICAL_SENSOR = 0x0510,
   FAN = 0x002B,
   GENERIC_SWITCH = 0x000F,
+  ILLUMINATION_SENSOR = 0x0106,
   IRRIGATION_SYSTEM = 0x0040,
+  MOTION_SENSOR = 0x0107,
```
hcarter-775: OCCUPANCY_SENSOR is the Matter-defined device type

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -34,7 +34,9 @@ SwitchFields.DEVICE_TYPE_ID = {
   ELECTRICAL_SENSOR = 0x0510,
   FAN = 0x002B,
   GENERIC_SWITCH = 0x000F,
+  ILLUMINATION_SENSOR = 0x0106,
```
hcarter-775: LIGHT_SENSOR is the Matter-defined device type

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+local function handle_set_preset(driver, device, cmd)
+    local endpoint_id = device:component_to_endpoint(cmd.component)
+    local lift_value = device:get_field(PRESET_LEVEL_KEY) or 50
+    local hundredths_lift_percent = (100 - tonumber(lift_value)) * 100
+    device:send(clusters.WindowCovering.server.commands.GoToLiftPercentage(device, endpoint_id, hundredths_lift_percent))
+end
+
+local function handle_close(driver, device, cmd)
+    local endpoint_id = device:component_to_endpoint(cmd.component)
+    endpoint_id = tonumber(endpoint_id)
```
hcarter-775: these tonumber() calls are not needed

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+    local req = clusters.WindowCovering.server.commands.DownOrClose(device, endpoint_id)
+    if device:get_field(REVERSE_POLARITY) then
+        req = clusters.WindowCovering.server.commands.UpOrOpen(device, endpoint_id)
+    end
+    device:send(req)
+end
+
+local function handle_open(driver, device, cmd)
+    local endpoint_id = device:component_to_endpoint(cmd.component)
+    endpoint_id = tonumber(endpoint_id)
```
hcarter-775: this one as well

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+    local req = clusters.WindowCovering.server.commands.UpOrOpen(device, endpoint_id)
+    if device:get_field(REVERSE_POLARITY) then
+        req = clusters.WindowCovering.server.commands.DownOrClose(device, endpoint_id)
+    end
+    device:send(req)
+end
+
+local function handle_pause(driver, device, cmd)
+    local endpoint_id = device:component_to_endpoint(cmd.component)
+    endpoint_id = tonumber(endpoint_id)
```
hcarter-775: and this one

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+    local req = clusters.WindowCovering.server.commands.StopMotion(device, endpoint_id)
+    device:send(req)
+end
+
+local function handle_shade_level(driver, device, cmd)
+    local endpoint_id = device:component_to_endpoint(cmd.component)
+    endpoint_id = tonumber(endpoint_id)
```
hcarter-775: and this one

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+                end
+            elseif device:supports_capability(capabilities.motionSensor) then
+                parent:send(clusters.OccupancySensing.attributes.Occupancy:read(parent))
+                parent:send(clusters.IlluminanceMeasurement.attributes.MeasuredValue:read(parent))
+            elseif device:supports_capability(capabilities.windowShadeLevel) then
+                map = {main = 5}
+                parent:send(clusters.WindowCovering.attributes.CurrentPositionLiftPercent100ths:read(parent))
+            end
+            device:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, map)
+            parent:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, map)
```
hcarter-775:
```suggestion
            device:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, map, {persist = true})
            parent:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, map, {persist = true})
```
Otherwise a driver restart will delete these maps

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+
+    local stored_eps = device:get_field(ACTIVE_EPS)
+
+    for _, ep in ipairs(added or {}) do
+        parent:send(clusters.Descriptor.attributes.DeviceTypeList:read(parent, ep))
+        if device.network_type == device_lib.NETWORK_TYPE_MATTER then
+            local product_id = device.manufacturer_info and device.manufacturer_info.product_id
+            if product_id == 0x0005 and ep == 3 then
+                matter_device:try_update_metadata({ profile = "light-binary" })
+            elseif product_id == 0x0006 then
```
hcarter-775: we should add constants for these various product ids that are scattered around

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+                if ep == 3 then
```
hcarter-775: we should also add understandable variable names for these endpoints

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+    local parent = get_parent(driver, device)
+    local matter_device = get_matter_device(device)
+
+    if not parent then
+        return
+    end
+
+    local new_eps = extract(ib) or {}
+    table.sort(new_eps)
+    local removed, added = diff(device, new_eps)
```
hcarter-775: "removed, added" what? I think you mean endpoints, I would suggest adding some more clarity to this. Similarly for the function names "diff" and "extract". It is difficult to maintain and follow along these kinds of generic names in my opinion

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+    end
+end
+
+local function occupancy_measured_value_handler(driver, device, ib, response)
+    if ib.data.value ~= nil then
+        device:emit_event_for_endpoint(ib.endpoint_id, ib.data.value == 0x01 and
+                capabilities.motionSensor.motion.active() or
+                capabilities.motionSensor.motion.inactive())
+    end
+end
```
hcarter-775: we shouldn't need this unique handler, since it is already handled in the main driver?

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+            subscribe(parent, ep, clusters.IlluminanceMeasurement.ID, clusters.IlluminanceMeasurement.attributes.MeasuredValue.ID)
+        end
+    end
+end
+
+local function do_configure (driver, device) end
+local function added (driver, device) end
+local function driver_switched(driver, device)  end
+
+local Hager_switch = {
```
hcarter-775:
```suggestion
local hager = {
```

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+    NAME = "Hager matter switch handler",
```
hcarter-775:
```suggestion
    NAME = "Hager Subdriver",
```

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+    table.sort(wc_eps)
+    table.sort(oc_eps)
+    table.sort(lvl_eps)
+
+    if device:get_parent_device() ~= nil then
+        link_matter_device_and_parent(device)
+        local parent = device:get_parent_device()
+        device:extend_device("send", function(self, message)
+            return parent:send(message)
+        end)
```
hcarter-775: if a device is a child, send is already handled by the parent?

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+    local bt_eps = device:get_endpoints(clusters.Switch.ID)
+    local lvl_eps = device:get_endpoints(clusters.LevelControl.ID)
+    local product_id = device.manufacturer_info.product_id
+
+    table.sort(wc_eps)
+    table.sort(oc_eps)
+    table.sort(lvl_eps)
+
+    if device:get_parent_device() ~= nil then
+        link_matter_device_and_parent(device)
```
hcarter-775: I'm not sure I understand what this link logic is meant to do. It appears like it could be removed if I'm understanding it correctly

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+        return "window-covering"
+    end
+    return "switch-binary", nil
+end
+
+local function create_assign_profile_wrapper(device_type_id)
+    local profile_name = assign_profile_for_endpoint(device_type_id)
+    return function(device, ep_id, is_child_device)
+        return profile_name, nil
+    end
```
hcarter-775:
```suggestion
   return profile_name, nil
```
though at this point I might suggest just removing this wrapper, since no real translation is occurring

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,508 @@
...
+        end)
+
+        local main_onOff_at_join = device:get_field(MAIN_ONOFF_EP)
+        if main_onOff_at_join and (product_id == 0x0005 or product_id == 0x0006) then
+            device:set_field(MAIN_ONOFF_EP, 3, { persist = true })
+            device.thread:call_with_delay(6, function()
+                if device:supports_capability(capabilities.switchLevel) then
+                    device:set_field(MAIN_ONOFF_EP, 4, { persist = true })
+                end
+            end)
```
hcarter-775: Are you hoping to check this after the profile has been altered or something? Wondering due to the delay call.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: Yes, that's correct, but there is also a matter device which requires this override.

There are following devices handled by my driver:
- Parent (matter hub).
- Matter type device (this one is the reason for such override as I keep the communication through the parent device due to the limitation related to the endpoint activation and deactivation).
- Child devices.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: Its due  to the reason that the main driver utilizes ":emit_event" instead of "emit_event_for_endpoint" in the occupancy attr handler so it would emit on the "Parent" device instead of the matter type device which is bridged to it. 

I need to evaluate if implementing an `emit_event` override on the matter device would introduce additional complexity. Will take a look tomorrow.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: You are right, thanks! I've fixed it.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: I'm not sure if I can recall the reason right now. Tomorrow I will be able to confirm it with the device.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: The linking in my opinion is crucial for this driver. I utilize it for easier handling of the communication between the matter device and parent device (e.g. info_changed).

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,501 @@
...
+                    matter_device:try_update_metadata({ profile = "4-button" })
+                end
+            end
+
+        end
+    end
+end
+
+local function handle_set_preset(driver, device, cmd)
+    local endpoint_id = device:component_to_endpoint(cmd.component)
```
hcarter-775: Hi, I hadn't noticed before, but all of this is written with tab spacing. As you can see elsewhere, other Matter drivers use 2-space spacing instead.

JanJakubiszyn: Sure, I will fix it right away.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,501 @@
...
+
+    if not parent then
+        return
+    end
+
+    local new_eps = extract_reported_endpoints(ib) or {}
+    table.sort(new_eps)
+    local removed_eps, added_eps = detect_endpoint_changes(device, new_eps)
+
+    device:set_field(ACTIVE_EPS, new_eps, { persist = true })
```
hcarter-775: I'm wondering, should the endpoint change detection be comparing the device and the new endpoints every time? Because we store the new eps into ACTIVE_EPS, wouldn't this mean the ACTIVE_EPS should be compared against in the future, not the device itself?

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,498 @@
...
+      local child = self:get_child_by_parent_assigned_key(string.format("%d", endpoint_id))
+
+      if child then
+        return child:emit_event(event)
+      end
+      if matter_device then
+        return matter_device:emit_event_for_endpoint(ep_info, event)
+      end
+    end)
+  end
```
hcarter-775: why is this happening in init? Seems like it should be happening in doConfigure (which would automatically have this happen only once), and then subsequent updates would occur in the descriptor cluster handlers

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,498 @@
...
+
+  local new_eps = extract_reported_endpoints(ib) or {}
+  table.sort(new_eps)
+  local removed_eps, added_eps = detect_endpoint_changes(device, new_eps)
+
+  device:set_field(ACTIVE_EPS, new_eps, { persist = true })
+
+  for _, ep_id in ipairs(added_eps or {}) do
+    subscribe(parent, ep_id, clusters.Descriptor.ID, clusters.Descriptor.attributes.DeviceTypeList.ID, nil)
+    if device.network_type == device_lib.NETWORK_TYPE_MATTER then
```
hcarter-775: this will always be a matter device, since it is a device report

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,498 @@
...
+      capabilities.motionSensor.motion.active() or
+      capabilities.motionSensor.motion.inactive())
+  end
+end
+
+local function info_changed(driver, device, event, args)
+  local parent = get_parent(driver, device)
+  local map = {}
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and device.profile.id ~= args.old_st_store.profile.id then
+    device.thread:call_with_delay(5, function()
```
hcarter-775: why is this a delayed call? Does not appear like that would be necessary

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,498 @@
...
+    device:emit_event_for_endpoint(ib.endpoint_id, ib.data.value == 0x01 and
+      capabilities.motionSensor.motion.active() or
+      capabilities.motionSensor.motion.inactive())
+  end
+end
+
+local function info_changed(driver, device, event, args)
+  local parent = get_parent(driver, device)
+  local map = {}
+  if device.network_type == device_lib.NETWORK_TYPE_MATTER and device.profile.id ~= args.old_st_store.profile.id then
```
hcarter-775: I'm confused, why would we only update these things for the main device and not any children (EDGE_CHILD)? This appears to just be basic configuration logic, no? Is the idea that these would have been configured elsewhere?

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,498 @@
...
+        if device:supports_capability(capabilities.switchLevel) then
+          map = { main = 4 }
+          parent:send(clusters.LevelControl.attributes.CurrentLevel:read(parent))
+        end
+      elseif device:supports_capability(capabilities.motionSensor) then
+        parent:send(clusters.OccupancySensing.attributes.Occupancy:read(parent))
+        parent:send(clusters.IlluminanceMeasurement.attributes.MeasuredValue:read(parent))
+      elseif device:supports_capability(capabilities.windowShadeLevel) then
+        map = { main = 5 }
+        parent:send(clusters.WindowCovering.attributes.CurrentPositionLiftPercent100ths:read(parent))
```
hcarter-775: I see all these reads, I do not think that is necessary. We should just set up a subscription and the same thing will basically happen. Looking at it more, I also strongly suggest simply using the subscribe logic from the main driver here (where we extend subscribe to use the custom logic in Matter Switch). It is specifically built to ensure subscriptions are created as expected for parent/child devices of varying types, like this.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
hcarter-775: When this is true ` device.network_type == device_lib.NETWORK_TYPE_MATTER` you will never have a `parent`. The parent only comes via being an EDGE_CHILD device.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
hcarter-775: Oh sure, in that case then I'd say just change the main driver to use `emit_event_on_endpoint`. There's no real reason why it can't be that. Then we can remove this override

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,498 @@
...
+      end
+      device:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, map, { persist = true })
+      parent:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, map, { persist = true })
+    end)
+  elseif args.old_st_store.preferences.reverse ~= device.preferences.reverse then
+    if device.preferences.reverse then
+      device:set_field(REVERSE_POLARITY, true, { persist = true })
+    else
+      device:set_field(REVERSE_POLARITY, false, { persist = true })
+    end
```
hcarter-775: Do we really need this field, or this info_changed logic at all? I'm sure this is how its implemented in the window covering driver, but I feel like we could just directly use `device.preferences.reverse` in place of adding driver state to track preference state?

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,498 @@
...
+  elseif args.old_st_store.preferences.presetPosition ~= device.preferences.presetPosition then
+    local new_preset_value = device.preferences.presetPosition
+    device:set_field(PRESET_LEVEL_KEY, new_preset_value, { persist = true })
+  end
```
hcarter-775: similar thing here... it's not like `device.preferences.presetPosition` disappears, so why not just use that directly?

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,498 @@
...
+      local active_eps = device:get_field(ACTIVE_EPS)
+      device.thread:call_with_delay(6, function()
+        subscribe(parent, ep_id, clusters.OnOff.ID, clusters.OnOff.attributes.OnOff.ID)
+      end)
+
+      if ep_id == 3 and device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_1G then
+        return
+      elseif ep_id == 4 and device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_2G then
+        if switch_utils.tbl_contains(active_eps, 3) then
+          local ep3 = parent:get_child_by_parent_assigned_key("3") or nil
```
hcarter-775:
```suggestion
          local ep3 = parent:get_child_by_parent_assigned_key("3")
```
nit: not necessary

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: The reason for custom subscription handling in my driver is that after an endpoint is deactivated and/or activated, the device's JSON might not update accurately. It won't show all the available endpoints, making the `device.endpoints` call miss some of them. The subscription through `cluster_base` eliminates the possibility of such behaviour by ensuring that all of the dynamic endpoints receive the subscription regardless of the cached device JSON.

I will change those reads to subscriptions.

hcarter-775: Ah, that makes sense. I am still wondering though if we even need these at all? Don't you already subscribe to these attributes elsewhere?

JanJakubiszyn: Yes, these reads were likely added to get initial state faster before subscriptions arrive. Since I'm already subscribing to these clusters elsewhere, they're redundant. I'll verify with the devices and remove them.

hcarter-775: ahh, sure that also makes sense. In the main driver we also update the subscription upon an infoChanged event. In that case, I might say we should make these subscribes, and then remove the other place where the subscription is happening then (in the descriptor cluster handlers)

JanJakubiszyn:
Hello @hcarter-775 , I've implemented the changes to the code and tested it with the physical devices. Please let me know if this state is acceptable. After that, I will adjust the unit tests.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,461 @@
...
+  parent:set_field(MATTER_DEVICE_ID, matter_device.id, { persist = true })
+end
+
+local function device_init (driver, device)
+  if device.network_type ~= device_lib.NETWORK_TYPE_MATTER then
+    return
+  end
+
+  device:set_field(fields.profiling_data.POWER_TOPOLOGY, false, { persist = true })
+  device:set_field(fields.profiling_data.BATTERY_SUPPORT, fields.battery_support.NO_BATTERY, { persist = true })
```
hcarter-775: Since you're no longer using the main driver's logic, you shouldn't need these field sets.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,461 @@
...
+
+  if device:get_parent_device() ~= nil then
```
hcarter-775: Since you have:
```
  if device.network_type ~= device_lib.NETWORK_TYPE_MATTER then
    return
  end
```
there will never be a parent device at this point. If you're trying to focus in on child devices with this, I might suggest you add an "if device.network_type == device_lib.EDGE_CHILD" check at the top or bottom of init and do this logic.

hcarter-775: then you don't need the "else" either, as you will know you're interacting with the primary matter device.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,461 @@
...
+    device:emit_event_for_endpoint(ib.endpoint_id, reverse and windowShade.closing() or windowShade.opening())
+  elseif state == 2 then
+    device:emit_event_for_endpoint(ib.endpoint_id, reverse and windowShade.opening() or windowShade.closing())
+  elseif state ~= 0 then
+    device:emit_event_for_endpoint(ib.endpoint_id, windowShade.unknown())
+  end
+end
+
+local function info_changed(driver, device, event, args)
+  if device.profile.id ~= args.old_st_store.profile.id or device.network_type == device_lib.NETWORK_TYPE_CHILD then
```
hcarter-775, line 186:
```suggestion
  if device.profile.id ~= args.old_st_store.profile.id then
```
I don't think you should just blanket allow all this to happen if it's a child device. The initial "device.profile.id" should be good enough, since the old profile would be non-existent if this is a new device.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,461 @@
...
+    local parent = device:get_parent_device()
+    local map = {}
+    device.thread:call_with_delay(2, function()
```
hcarter-775, line 190: you don't need this delay

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,461 @@
...
+    matter_device:set_field(BUTTON_EPS, clean_eps, { persist = true })
+
+    local child = parent:get_child_by_parent_assigned_key(tostring(ep_id))
+    if child then
+      driver:try_delete_device(child.id)
+    end
+
+    if ep_id == 3 then
+      if device.manufacturer_info.product_id == PRODUCT_ID.SWITCH_1G then
+        matter_device:try_update_metadata({ profile = "2-button" })
```
hcarter-775, line 321: One thing I find strange about this logic is that you are able to call try_update_metadata more than once for the same device. I might try and re-work this logic to ensure that "try_update_metadata" is only ever called once in this handler, to limit calls.

Also, we don't necessarily want to try this call every time we get a descriptor cluster response (like if the device or hub ever go offline, for example).

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,461 @@
...
+      device:try_update_metadata({ profile = "window-covering" })
+      device:set_field(MAIN_WC_EP, wc_eps[1])
+    elseif #bt_eps == 4 then
+      device:try_update_metadata({ profile = "4-button" })
+    elseif #bt_eps == 2 then
+      device:try_update_metadata({ profile = "2-button" })
+    elseif #lvl_eps > 0 and product_id == PRODUCT_ID.SWITCH_1G then
+      device:try_update_metadata({ profile = "light-level" })
+    end
+    device:set_field(BUTTON_EPS, bt_eps)
```
hcarter-775:
```suggestion
    device:set_field(BUTTON_EPS, bt_eps, { persist = true })
```
I think you want this

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
hcarter-775: @JanJakubiszyn In your current subscription implementation, one concern I have is whether the subscription will be re-initialized correctly when the hub is restart. This will delete the current subscription cache, and so this will need to be re-populated somehow. For this reason, I think you may want to include the main driver's subscription method. The primary issue with this at the moment is related to the fact that this is a subdriver. For that reason, I might suggest you look at the work I did in [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2898/changes#diff-bd38bdeee2eadd5833df0b7df2af3de9353bca01fb410f5517f393df12c3dbb0) and implement that subscription method here in utils.lua. Then, you can add a `subscriptions.lua` file like in this PR. Then you can just call "device:subscribe()" the same way that it's done in the main driver. You can also use the example in the current Matter Camera driver as an example of what to do instead, though I'm partial to this method in PR.

But basically, the issue now is that when the hub restarts, your subscription will be removed and you currently have no way of re-creating it.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: During testing, I observed that removing this delay causes subscriptions to be sent before component mapping completes for 1G and PIR devices. This results in devices remaining in 'connected' state with no initial state reported until the first attribute change occurs.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: This guard handles the Matter child device in a parent-child architecture. There is always one parent device (`MATTER `type, no parent) and one Matter child device (`MATTER `type, with parent reference). which require a different initialization. 

Additional EDGE_CHILD devices are filtered out by the `NETWORK_TYPE_MATTER` check.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: I've noticed that child devices don't pass the profile change condition on creation. The component-endpoint mapping is primarily for the matter_device's command routing. Child devices route commands through the parent by default, so the mapping field has no impact on their behavior. Testing confirms this works correctly.

hcarter-775 (approved):

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -0,0 +1,475 @@
...
+    init = device_init,
+    infoChanged = info_changed,
+    doConfigure = do_configure,
+    driverSwitched = driver_switched
+  },
+  matter_handlers = {
+    attr = {
+      [clusters.Descriptor.ID] = {
+        [clusters.Descriptor.attributes.PartsList.ID] = handle_descriptor_report,
+        [clusters.Descriptor.attributes.DeviceTypeList.ID] = device_type_handler,
```
hcarter-775, line 451: Hi Jan, looking at this again, I realized that you'll need to add the "DeviceTypeList" attribute to the embedded cluster def for Matter Switch. Otherwise, devices on per-59 FW will crash if they try to run this.

PR #2751, opened 2026-02-03
Add support for additional Closure device types

nickolas-deboom:
This adds support for the Covering, Gate, Door, and GarageDoor closure subtypes.

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2714 is a prerequisite for this PR.

PR #2875, opened 2026-04-03
Support Closure device type

nickolas-deboom:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

Add support for the Closure device type.
[ ](https://smartthings.atlassian.net/browse/CHAD-17522)

Note that these changes were originally in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2751, but that PR was based on a branch that would migrate the matter-window-covering driver to matter-switch, which might be done at some point but is a much larger scope. In the interest of supporting this device type sooner, this PR implements this code in matter-window-covering.

# Summary of Completed Tests

Tested with VDA Closure and CHIP example app.


hcarter-775:
@nickolas-deboom would it make more sense to implement this closure logic as a subdriver? That way, we would overwrite the capability handlers rather than add a bunch of else's to the handlers

nickolas-deboom:
> @nickolas-deboom would it make more sense to implement this closure logic as a subdriver? That way, we would overwrite the capability handlers rather than add a bunch of else's to the handlers

I think that might be a good idea

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -4,11 +4,19 @@
...
 --Note: No support for setting device into calibration mode, it must be done manually
+
 local capabilities = require "st.capabilities"
 local im = require "st.matter.interaction_model"
 local log = require "log"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
+local version = require "version"
+
+-- TODO: change this to < 20 once the lua libs have been updated for hub-core 61
```
ctowns: Can we make a ticket to track this?

drivers/SmartThings/matter-window-covering/profiles/covering.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Blind
+- id: windowShade3
+  optional: true
+  capabilities:
+    - id: windowShadeLevel
+      version: 1
+      optional: true
+  categories:
+    - name: Blind
+- id: windowShade4
```
ctowns, line 46: Why are there multiple components in each of these profiles, and why are there 4 in each? Is 4 an arbitrarily selection?

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
```diff
@@ -0,0 +1,468 @@
...
+local t_utils = require "integration_test.utils"
+local uint32 = require "st.matter.data_types.Uint32"
+local version = require "version"
+
+if version.api < 21 then
+  clusters.ClosureControl = require "embedded_clusters.ClosureControl"
+  clusters.ClosureDimension = require "embedded_clusters.ClosureDimension"
+end
+
+local mock_device = test.mock_device.build_test_matter_device(
```
ctowns, line 20: These test cases only cover the covering type - could we add some coverage for the door/garage/gate types as well? I know that currently we would expect these to generally behave the same since the attribute handler just switches between `windowShade` and `doorControl` when checking what the device supports [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2875/changes#diff-4e580361496e6c9c49784e600fee7954e34cea65c669902917513f94a26336d3R585), but I think it'd be good to at least have a couple door-specific unit tests to make sure this parity is maintained.

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -34,9 +57,46 @@ local function find_default_endpoint(device, cluster)
...
 end
 
+local function get_closure_dimension_eps(device)
+  local eps = device:get_endpoints(clusters.ClosureDimension.ID) or {}
+  table.sort(eps)
+  local result = {}
+  for _, ep in ipairs(eps) do
+    if ep ~= 0 then
+      table.insert(result, ep)
+      if #result >= 4 then break end
```
ctowns: I'm not sure where this 4 value comes from - could you define it as a constant somewhere for documentation purposes?

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -59,8 +119,114 @@ local function match_profile(device, battery_supported)
...
+
+  -- ClosureDimension capabilities: windowShadeLevel (covering) or level (door types)
+  local dim_eps = get_closure_dimension_eps(device)
+  if #dim_eps > 0 then
+    local dim_cap = is_door_type and capabilities.level.ID or capabilities.windowShadeLevel.ID
+    if #dim_eps == 1 then
+      -- Single ClosureDimension: enable the capability on the main component.
+      local found_main = false
+      for _, entry in ipairs(optional_caps) do
+        if entry[1] == "main" then
```
ctowns: this might be cleaner to do something like this where you define a `main_component_capabilities` table earlier in the function and then insert it into the table at the end, that way you can just consistently add to the main capabilities without needing to find them within this optional caps struct. You can just define `main_component_capabilities` up on line 143 and then insert into it directly for the battery capability checking in line 144-148, and then insert directly in this handling as well.

Then at the end of the function, just push the `main_component_capabilities` into the `optional_caps` if it is not null.

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -283,18 +501,143 @@ local function battery_charge_level_attr_handler(driver, device, ib, response)
 end
 
 local function power_source_attribute_list_handler(driver, device, ib, response)
+  local is_closure = #device:get_endpoints(clusters.ClosureControl.ID) > 0
   for _, attr in ipairs(ib.data.elements) do
     -- Re-profile the device if BatPercentRemaining (Attribute ID 0x0C) is present.
     if attr.value == 0x0C then
-      match_profile(device, battery_support.BATTERY_PERCENTAGE)
+      if is_closure then
+        device:set_field(CLOSURE_BATTERY_SUPPORT, battery_support.BATTERY_PERCENTAGE, {persist = true})
```
ctowns: Could we just pass the battery support value in the `match_profile_for_closure` function, like what is done for the `match_profile` function?

Furthermore, could we have the `match_profile` function itself handle both closures and other types? You could just add a line at the top of the `match_profile` function that calls `match_profile_for_closure` if the device has a `ClosureControl` ep. It would essentially just be centralizing the logic in this function to the `match_profile` function itself, which I think would be a little cleaner.

drivers/SmartThings/matter-window-covering/profiles/covering.yml
nickolas-deboom: Right, 4 is arbitrary since i wasn't sure how many closure panels to typically expect (the spec just says 0+). The example devices I've seen so far have only had up to 2 so I think 4 should be good however I will define this as a constant as you suggested in another comment

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: This wasn't a permanent todo as this is already done as of 61.2 so I will update this in a commit within this PR.

drivers/SmartThings/matter-window-covering/profiles/covering.yml
nickolas-deboom: To clarify, a closure consists of 1 endpoint implementing the ClosureControl cluster and then any number of endpoints of the Closure Panel device type that each implement the ClosureDimension cluster. So the number of components in these profiles that are used would reflect the number of ClosureDimension endpoints.

nickolas-deboom:
@hcarter-775 I made an update to move the closure handling to a subdriver in 541b8969675c1b9de371582018a83f1012733b6f. Lemmeknow how that looks

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: Updated in 68d4763879d99b4789d629ef701534f169f843e4

drivers/SmartThings/matter-window-covering/src/sub_drivers/closure/closure_handlers/capability_handlers.lua
```diff
@@ -0,0 +1,71 @@
+-- Copyright 2025 SmartThings, Inc.
```
hcarter-775: quick nit, gotta update these to '26 😺

hcarter-775:
@nickolas-deboom I haven't read the specific closures logic too closely (even when making the initial suggestion), but from the subdriver perspective idea, this looks really great!

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
nickolas-deboom: Added in 1a2667abe3d6c87d8f598b0443ff3baedcbc53f0!

drivers/SmartThings/matter-window-covering/src/sub_drivers/closure/closure_handlers/capability_handlers.lua
nickolas-deboom: Done!

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: Done in latest commit and I left a [comment](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/0eb4b1fecb13f3c958989f2680434d05897273b1/drivers/SmartThings/matter-window-covering/src/sub_drivers/closure/closure_utils/fields.lua#L31) with an explanation

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: Nice idea, implemented in 9ebf898b

drivers/SmartThings/matter-window-covering/src/init.lua
nickolas-deboom: We need to save it in a field because we need to wait until we have both the battery support and closure tag info before profiling, and the ordering isn't guaranteed. This matches the system used in other drivers where it's required to wait for multiple attribute reports before running match_profile. 

> Furthermore, could we have the match_profile function itself handle both closures and other types?

Closure support was moved into a subdriver in 541b8969675c1b9de371582018a83f1012733b6f, so now it has it's own `match_profile` function that is specific to closures - is that sufficient in your eyes?

hcarter-775 (approved):
With the testing done and the gating we have behind the subdriver implementation, I think this is good to go for merging to main

PR #2729, opened 2026-01-26
WWSTCERT-10189 Ledvance zigbee meter plug

LQ107:
Check all that apply

# Type of Change

- [ ] WWST Certification Request
     - If this is your first time contributing code:
          - [ ] I have reviewed the README.md file
          - [ ] I have reviewed the CODE_OF_CONDUCT.md file
          - [ ] I have signed the CLA
     - [ ] I plan on entering a WWST Certification Request or have entered a request through the WWST Certification console at developer.smartthings.com
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have verified my changes by testing with a device or have communicated a plan for testing
- [ ] I am adding new behavior, such as adding a sub-driver, and have added and run new unit tests to cover the new behavior

# Description of Change

# Summary of Completed Tests

PR #2876, opened 2026-04-06
Delete guest user when SetYearDaySchedule is failed

HunsupJung:
# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests
Tested with Yale Smart Lock with Matter. Unit test added to explicitly test the flow.

### DIFF
--- a/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
+++ b/drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
@@ -1554,15 +1554,29 @@ local function clear_user_response_handler(driver, device, ib, response)
     device.log.warn(string.format("Failed to clear user: %s", status))
   end
 
-  -- Update commandResult
-  local command_result_info = {
-    commandName = cmdName,
-    userIndex = userIdx,
-    statusCode = status
-  }
-  device:emit_event(capabilities.lockUsers.commandResult(
-    command_result_info, {state_change = true, visibility = {displayed = false}}
-  ))
+  -- This occurs in the "defaultSchedule" command failure path, when a guest user's credentials are set but
+  -- the scheduling fails during default setup. In this case, those set credentials should be removed, and we
+  -- wait to log lock credentials (note: as a "failure", though it technically succeeded) until here.
+  if cmdName == "defaultSchedule" then
+    local command_result_info = {
+      commandName = "addCredential",
+      userIndex = userIdx,
+      statusCode = "failure"
+    }
+    device:emit_event(capabilities.lockCredentials.commandResult(
+      command_result_info, {state_change = true, visibility = {displayed = false}}
+    ))
+    device:set_field(lock_utils.BUSY_STATE, false, {persist = true})
+  else
+    local command_result_info = {
+      commandName = cmdName,
+      userIndex = userIdx,
+      statusCode = status
+    }
+    device:emit_event(capabilities.lockUsers.commandResult(
+      command_result_info, {state_change = true, visibility = {displayed = false}}
+    ))
+  end
   device:set_field(lock_utils.BUSY_STATE, false, {persist = true})
 end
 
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1550,6 +1550,19 @@ local function clear_user_response_handler(driver, device, ib, response)
...
+      }
+      device:emit_event(capabilities.lockCredentials.commandResult(
+        command_result_info, {state_change = true, visibility = {displayed = false}}
+      ))
+      device:set_field(lock_utils.BUSY_STATE, false, {persist = true})
+      return
+    end
   else
     device.log.warn(string.format("Failed to clear user: %s", status))
   end
```
hcarter-775, line 1555: In the case this fails, what is the expected result?

Given the current logic, a failure would mean that 
1. lockCredentials commandResult is never emitted (HIGH importance)
2. lockUsers commandResult for defaultSchedules will be emitted below, which will fail (MEDIUM importance)

Should we retry the `ClearUser` command again? If so, how many times?

Also, if it fails after some number of retries, I think we should emit a success result for lockCredentials. What do you think?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1550,6 +1550,19 @@ local function clear_user_response_handler(driver, device, ib, response)
     delete_aliro_from_table_as_user(device, userIdx)
     delete_week_schedule_from_table_as_user(device, userIdx)
     delete_year_schedule_from_table_as_user(device, userIdx)
+    if cmdName == "defaultSchedule" then
```
ctowns: I think this handling that is specific to the `defaultSchedule` command could benefit from a short comment that explains why this is needed to improve readability.

ctowns (commented):
@HunsupJung could you also fill out the `Summary of Completed Tests` section on the top PR comment?

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: > lockCredentials commandResult is never emitted (HIGH importance)
lockUsers commandResult for defaultSchedules will be emitted below, which will fail (MEDIUM importance)

I don't understand what you said. The logic up to here is as follows.
1. Plugin sends `setCredential` for guest
2. Hub sends `SetCredential` to Door Lock
3. Door Lock creates a user and set the credential
4. Door Lock returns a success result
5. Hub sends `SetYearDaySchedule` for guest
6. Door Lock returns a failure result
7. Hub sends `ClearUser` to Door Lock
8. Door Lock clear the user and return a success result
9. Hub set the `commandResult` of `lockCredentials` to failure

There is no additional action. I don't think it's the right way to continue follow-up on failures. And now we are putting the default schedule directly from the driver for guest, but I will discuss with Plugin and ask Plugin to put the default schedule on its own. (It is dangerous for the driver to put the schedule on device because the schedule operation may vary from device to device.)

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
HunsupJung: Okay, I will add the comment.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: Hi @HunsupJung, I am wondering about the following scenario:

Steps 1-6 are the same.

7. Hub sends ClearUser to Door Lock
8. ClearUser fails and returns a failure result
9. The `commandResult` of lockCredentials never occurs.

Even if we don't do retries, we should account for this possibility. Basically, I'm saying that we should emit a `commandResult` whether or not the `ClearUser` command is a success.

HunsupJung: @hcarter-775 
Do you mean that the `commandResult` of `lockCredentials` should be updated even if `ClearUser` fails? If that's what you mean, I agree.
However, it is not necessary to update the commandResult of other capabilities. the default schedule is set only when a credential is created for the guest, so we only need to update the `commandResult` of the `lockCredentials`.

hcarter-775: Yes, I agree with that. I am only talking about updating lockCredentials in the failure case

HunsupJung: Okay, I will update it right now.

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
```diff
@@ -1554,15 +1554,28 @@ local function clear_user_response_handler(driver, device, ib, response)
...
-  local command_result_info = {
-    commandName = cmdName,
-    userIndex = userIdx,
-    statusCode = status
-  }
-  device:emit_event(capabilities.lockUsers.commandResult(
-    command_result_info, {state_change = true, visibility = {displayed = false}}
-  ))
+  -- If the guest user's default schedule input fails, the created guest user is deleted.
+  -- Below is the logic executed when the deletion of the guest user information.
```
hcarter-775:
```suggestion
  -- If we attempt to create a guest user and the credentials are set correctly, but afterwards the default schedule command for that user fails on-device, the credentials we'd set for the guest user should be deleted. This all will occur in the "defaultSchedule" command failure path.
  -- Below is the logic executed in response to an attempt to remove that guest user's credentials.
```

Added a little more commentary here, since I do think this is a confusing code path

hcarter-775 (approved):

hcarter-775:
Hi Hunsup, before merging, please squash all commits. Thank you

drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua
hcarter-775: Updated manually, thanks!

hcarter-775:
For posterity as an FYI, I am merging this because the related test failures are due to [this unrelated PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2729) in an unrelated driver.

PR #2877, opened 2026-04-06
Matter Switch: Use defaults bounds if any custom bound is missing

hcarter-775:
# Description of Change
In the StatelessColorTemperatureStep command, we read the bounds or use defaults. However, since the colorTemperature capability's colorTemperatureRange attribute is only set after both custom bounds are, we should actually use the default bounds if any custom bound is missing to avoid "out of range" errors.

tag-along update: use hardcoded mired fields to avoid messy and unecessary rounding in the fields file

Follow-up discussion may be worthwhile to set the colorTemperatureRange if even one bound is present.

# Summary of Completed Tests
Tested with Zemismart device that gives max/min mireds that are out of "sane" range.


ctowns (approved):

ctowns:
Do these changes need to be made for zigbee as well?

hcarter-775:
@ctowns this has already been done for Zigbee in the initial Stateless PR- it was just a miss on my part for the Matter one a few months ago

PR #2886, opened 2026-04-09
Fixing zigbee-switch tests for ColorTempPhysMireds

aleclorimer:
- Fixing tests against dev for expected new messages for the `ColorTempPhysical(min/max)Mireds` attribute
- Adding `dev` to jenkins-driver-tests.yml workflow file matrix

### DIFF
--- a/.github/workflows/jenkins-driver-tests.yml
+++ b/.github/workflows/jenkins-driver-tests.yml
@@ -4,6 +4,10 @@ on:
     paths:
       - 'drivers/**'
 
+  pull_request_target:
+    paths:
+      - 'drivers/**'
+
 permissions:
   statuses: write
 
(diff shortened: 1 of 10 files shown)


aleclorimer:
These tests are failing because the lua libs being used for 60 and the "Run driver tests" GH workflow don't yet have the filtering on api version, so this is trying to run the tests meant for `api.version = 20`.

varzac:
> These tests are failing because the lua libs being used for 60 and the "Run driver tests" GH workflow don't yet have the filtering on api version, so this is trying to run the tests meant for `api.version = 20`.

I would have expected the Jenkins runs to pass though? I'll check the failure details tomorrow morning, but I'm not sure why they would be failing

cjswedes (approved):

PR #2887, opened 2026-04-09
Matter Camera: Fix bugs related to profile matching and re-init of capabilites

samadDotDev:
# Type of Change

- [x] Bug fix
- [x] Refactor

# Checklist

# Description of Change

I started with a fix to feature-map change, which if occurred on a SW update and didn't result in a profile change won't re-init or update any capability attributes (such as `supportedFeatures`) that should be updated on such changes.

In a specific example, I observed one of the cameras getting updated for its watermark/OSD bits in feature map of AVSM cluster, and despite the SW update (which triggered the `infoChanged` handler with SW version diff), the `videoStreamSettings::supportedFeatures` wasn't updated to reflect the changes in watermark/OSD support.

While addressing this issue I noticed that this was applicable to other capabilities and features as well, so I refactored that a bit. There were a few more issues that I addressed as part of this PR since they were tightly linked with the new changes to observe these diffs:

1. Added subscription to feature map attribute of 3 clusters we track features of (AVSM, AVSUL, ZoneManagement), just so if the feature map changes without a SW version bump (which can technically occur until the formal enforcement of fixed-quality attributes, and bump of `ConfigurationVersion` from spec) then we still reflect those changes in profile and/or relevant supported features of capabilities. 
 - Unlike `AttributesList`, I believe we aren't tracking FeatureMap global attribute for clusters in the Lua libs, so this is defined as a const until then. This was also surfaced in the recent interest of Door Lock TF in subscribing to feature map changes.
2. Refactored how to re-init capabilities based on profile changes and/or other changes (like of feature map and attributes list) that may not result in a profile update. Only selective capabilities are updated/re-initialized based on last state (only if there is a diff), and the "reset" of last state is accounted for cases of re-enablement of a capability (imagine populating `supportedFeature` of a capability, disabling it in profile, re-enabling it - we'd want to re-upsert the `supportedFeature` even if there is no diff in that case)
  - This also addresses a bug where an event could unintentionally get emitted (and potentially mark an offline device as online) for capabilities in a profile that we don't control (enable/disable) such as `vision.clipAnalysis` - it will result in an `infoChanged` with profile diff that'd previously call the re-init of all capabilities.
3. Status light and light brightness are now tracked based on the `AttributesList` handler. This was handled previously too, but wasn't tracked, so there was a bug where this could be reset (removing the status led components from profile) if you receive the `AttributesList` first and then another event occurred which would re-match the profile from various paths (as that re-match was defaulting to `false` for all spots outside of `AttributesList` handler). We can't rely (yet) on the attributes list within device's object (unlike feature map) due to this being potentially empty, hence it is tracked based on last report received for this attribute, and then used in any future re-matches from any path.

# Summary of Completed Tests

- Added unit tests covering some of the behavior additions/updates
- Performed functional testing with a camera, specifically looking out for following combinations:
  - Driver update on existing camera
  - Hub reboot / driver restart on offline camera
  - Hub reboot / driver restart on online camera
  - SW update of existing camera
  - Device object update without relevant diffs (unrelated changes in `infoChanged`) of existing camera
  - Onboarding a new camera

### DIFF
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
@@ -141,7 +141,11 @@ local function capabilities_needing_reinit(device)
   end
 
   if device:supports_capability(capabilities.audioRecording) then
-    local audio_enabled_state = device:get_latest_state(main, capabilities.audioRecording.ID, capabilities.audioRecording.audioRecording.NAME)
+    local audio_enabled_state = device:get_latest_state(
+      camera_fields.profile_components.main,
+      capabilities.audioRecording.ID,
+      capabilities.audioRecording.audioRecording.NAME
+    )
     if audio_enabled_state == nil then
       capabilities_to_reinit.audio_recording = true
     end
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
@@ -375,14 +375,6 @@ function CameraDeviceConfiguration.initialize_camera_capabilities(device)
   init_camera_privacy_mode(device)
 end
 
-function CameraDeviceConfiguration.initialize_camera_capabilities_and_subscriptions(device)
-  CameraDeviceConfiguration.initialize_camera_capabilities(device)
-  device:subscribe()
-  if #switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.DOORBELL) > 0 then
-    button_cfg.configure_buttons(device, device:get_endpoints(clusters.Switch.ID, {feature_bitmap=clusters.Switch.types.SwitchFeature.MOMENTARY_SWITCH}))
-  end
-end
-
 local function initialize_selected_camera_capabilities(device, capabilities_to_reinit)
   local reinit_targets = capabilities_to_reinit or {}
 
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
@@ -6,6 +6,7 @@ local camera_fields = require "sub_drivers.camera.camera_utils.fields"
 local camera_utils = require "sub_drivers.camera.camera_utils.utils"
 local capabilities = require "st.capabilities"
 local clusters = require "st.matter.clusters"
+local st_utils = require "st.utils"
 local device_cfg = require "switch_utils.device_configuration"
 local fields = require "switch_utils.fields"
 local switch_utils = require "switch_utils.utils"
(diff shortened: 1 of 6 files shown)


drivers/SmartThings/matter-switch/src/sub_drivers/camera/init.lua
```diff
@@ -42,24 +41,19 @@ end
...
   if #device:get_endpoints(clusters.CameraAvStreamManagement.ID) == 0 then
-    camera_cfg.match_profile(device, false, false)
+    camera_cfg.match_profile(device)
   end
 end
 
 function CameraLifecycleHandlers.info_changed(driver, device, event, args)
   local software_version_changed = device.matter_version ~= nil and args.old_st_store.matter_version ~= nil and
     device.matter_version.software ~= args.old_st_store.matter_version.software
+  local profile_changed = not switch_utils.deep_equals(device.profile, args.old_st_store.profile, { ignore_functions = true })
```
hcarter-775, line 51:
```suggestion
  local profile_changed = not switch_utils.deep_equals(device.profile, args.old_st_store.profile)
```
There are no functions needed to ignore/not ignore

samadDotDev: This was actually carried over from your change in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/a2914159e857907d73b72f27e3f14e303dce4df0#diff-7b6146aa3224eca03e6dad8843d29bb67487ef3e71bb2620114186a2d983b06cR50.

Was that done out of caution for deep comparison of profiles (incase component tables contained any function fields)?

hcarter-775: Well, not just caution, but actually profile structs do include an `emit_event` function written into them that is not a straight copy. though it should be treated as such for the purposes of the profiling case.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -275,6 +370,89 @@ function CameraDeviceConfiguration.initialize_camera_capabilities(device)
...
+  end
+  if reinit_targets.camera_privacy_mode then
+    init_camera_privacy_mode(device)
+  end
+end
+
+local function profile_capability_set(profile)
+  local capability_set = {}
+  for _, component in pairs((profile or {}).components or {}) do
+    for _, capability in ipairs(component.capabilities or {}) do
```
hcarter-775:
```suggestion
    for _, capability in pairs(component.capabilities or {}) do
```
I think `ipairs` may cause issues, depending on how the structure of capabilities changes. For example, running this test on FW 61 integration tests causes this to fail without using `pairs`

hcarter-775: Even if no issues are explictly found as a part of this, the structure of the profile table object being parsed is slightly different than what may be expected here. The capabilities table has explicit string keys rather than the ordered integer keys.

samadDotDev: Good catch, since order doesn't really matter here. I changed this as part of the https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2887/commits/d387762cd8fe7d88f00796f0dd910d89851bbb01

drivers/SmartThings/matter-switch/src/sub_drivers/camera/init.lua
samadDotDev: Pending question offline, I'm unclear if you're requesting change on the behavior here. That particular profile comparison is just a carry-over in this PR, so want to make sure that we're intentional if wanting to change the behavior of that comparison.

hcarter-775: Oh my bad, we should definitely keep this. Sorry for the confusion. I had meant to comment this on a different section, in the original `state_differs` fn where the ignore_functions tag was not needed, which you've since removed.

hcarter-775 (approved):

ctowns (approved):
LGTM! Excellent unit tests, and thank you for including all the real device test cases in the top comment!

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -275,6 +375,89 @@ function CameraDeviceConfiguration.initialize_camera_capabilities(device)
   init_camera_privacy_mode(device)
 end
 
+function CameraDeviceConfiguration.initialize_camera_capabilities_and_subscriptions(device)
```
nickolas-deboom: Where does this get called?

samadDotDev: Apologies, this is unused and was superseded by the other re-init function (`reinitialize_changed_camera_capabilities_and_subscriptions`). I just removed it in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2887/commits/ff3cdb042f37b18b9d4adf4d4ad6a8d19b5921f8

nickolas-deboom (approved):
LGTM!

drivers/SmartThings/matter-switch/src/sub_drivers/camera/init.lua
```diff
@@ -33,7 +32,7 @@ end
 function CameraLifecycleHandlers.do_configure(driver, device)
   camera_utils.update_camera_component_map(device)
   if #device:get_endpoints(clusters.CameraAvStreamManagement.ID) == 0 then
-    camera_cfg.match_profile(device, false, false)
+    camera_cfg.match_profile(device)
   end
   camera_cfg.create_child_devices(driver, device)
   camera_cfg.initialize_camera_capabilities(device)
```
hcarter-775, line 38: a related change, I think we can remove this call, and therefore this function. There will never be any capabilities added to the profile at this step.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -275,6 +375,81 @@ function CameraDeviceConfiguration.initialize_camera_capabilities(device)
...
+  local reinit_targets = capabilities_to_reinit or {}
+
+  if reinit_targets.webrtc then
+    init_webrtc(device)
+  end
+  if reinit_targets.ptz then
+    init_ptz(device)
+  end
+  if reinit_targets.zone_management then
+    init_zone_management(device)
```
hcarter-775, line 388: There is a device:supports_capability() check in all of these functions. Assuming the initial, raw call to these functions is removed, with the current ways that we call into there is no way that the capability will not be available at the point these init_<blank> functions are called.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/init.lua
samadDotDev: I think that's reasonable ask given the current profile structure, but I wouldn't depend on it (we could possibly add a new required capability to existing profile or add new ones that match to this subdriver as well).

Basically, I'd favor a no-op init here rather than missing an init in a future profile update.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
samadDotDev: Discussed offline for this and the above comment: since the original init isn't removed, this is still needed. Resolving to proceed with PR merge, and then we can follow up with further cleanup in the other in-flight PR.

cjswedes:
Please squash commits in the future. Now we have to cherry pick 5 commits instead of 1 to beta this week and prod next week.

samadDotDev:
> Please squash commits in the future. Now we have to cherry pick 5 commits instead of 1 to beta this week and prod next week.

My bad, I have been getting used to of leaving the logical changes with other contributors in separate commits in hub-core, but definitely something I'll keep in mind for this repo since we seem to be cherry-picking instead of pulling the branch here.

cjswedes:
No worries, typically I am heavily in favor more commits for independent changes, but this repo is an exception atm

PR #2339, opened 2025-08-14
Matter Switch: remove double init on device added

ctowns:
The device_added handler was calling device_init in order to account for situations where device_init was not called due to a race condition with the device cache. Since lifecycle events are now guaranteed and ordered with FW 58, we no longer need to depend on this extra init.

# Type of Change

- [ ] WWST Certification Request
     - If this is your first time contributing code:
          - [ ] I have reviewed the README.md file
          - [ ] I have reviewed the CODE_OF_CONDUCT.md file
          - [ ] I have signed the CLA
     - [ ] I plan on entering a WWST Certification Request or have entered a request through the WWST Certification console at developer.smartthings.com
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have verified my changes by testing with a device or have communicated a plan for testing
- [ ] I am adding new behavior, such as adding a sub-driver, and have added and run new unit tests to cover the new behavior

# Description of Change

# Summary of Completed Tests

PR #2898, opened 2026-04-13
Matter Switch/Camera: Remove Unique Camera Lifecycle Handlers

hcarter-775:
# Description of Change

Primarily, this PR removes the camera subdriver lifecycle events in order to create more cohesive device configuration steps within the Matter Switch driver. It does this by using the main driver util functions where appropriate and by integrating a new step into the Matter Switch-specific subscribe function, where the function reads a "subscriptions.lua" file for sub-driver specific information.
  - As a part of this, work from [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2339) was included to avoid a double init for post-58 FW versions. Almost all test files were edited to comply with this update.
  - Also, the "avoid bridge" logic in device profiling that was introduced [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/d1bf0df8375cfcae048ae0aaa955f01eb0ef7a4e#diff-7ab55ddbc6a9760ea5b6f928abee244c3142c74389743ce05b895b938ad81f10) has been re-worked. The original case described in the above commit has now been taken care of naturally, though with the introduction of new devices in Matter Switch like buttons and water valves, we still are returning early if the default endpoint found is of the aggregator type, after the child on/off devices are created.

As a secondary piece, it rewrites the Matter Camera integration tests. These tests were originally handled in a single file that took a very long time (at least for drivers) to run, and duplicated many parts of tests. I therefore broke it into two parts- one that handles device configuration logic, and the other that handles the attribute/event/capability handlers, which is not likely to change.

It also re-works some helper functions and helper systems to hopefully be more clear/concise.

# Summary of Completed Tests
Tested with a series of spot-checks for common device types:
- Extended color light did not re-profile, as expected
- Binary light re-profiled, as expected
- Matter Bridge did not re-profile, as expected


drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -3,57 +3,10 @@
...
-        cluster_id = clusters.CameraAvStreamManagement.ID,
-        attribute_ids = {
-          clusters.CameraAvStreamManagement.attributes.MicrophoneMuted.ID,
-          clusters.CameraAvStreamManagement.attributes.MicrophoneVolumeLevel.ID,
-          clusters.CameraAvStreamManagement.attributes.MicrophoneMaxLevel.ID,
-          clusters.CameraAvStreamManagement.attributes.MicrophoneMinLevel.ID,
-        },
-      }
-    end
-    if CameraUtils.feature_supported(device, clusters.CameraAvStreamManagement.ID, clusters.CameraAvStreamManagement.types.Feature.VIDEO) then
```
ldeora, line 40: Is it really VIDEO or is it SPEAKER?

`clusters.CameraAvStreamManagement.types.Feature.SPEAKER)`

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/subscriptions.lua
```diff
@@ -0,0 +1,106 @@
+-- Copyright © 2026 SmartThings, Inc.
```
ctowns, line 1: This feels like a missing features of our Lua Libraries that sub drivers cannot provide/update a subscription map. I think we could address this in the lua libs by keeping the same "subscribed_attributes" field that is used in the parent driver, but we could just include the subdriver's attribute table as a part of the considered attributes in the `populate_subscribed_attributes_from_driver_defaults` function . I think you should be able to repurpose the code you've used below for aggregating the sub driver subscriptions.

This can solution here can be a short term solution, but if the goal of this is to provide a long term solution for sub drivers to provide unique subscriptions, then I think we should make that change in the lua libs themselves in order to keep the structure of sub drivers similar to those of the parent drivers.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
hcarter-775: cc: @nickolas-deboom @samadDotDev Could I get confirmation one way or the other about this particular feature? It does seem like it should be SPEAKER.

PR #2902, opened 2026-04-13
Matter Switch: Update Orein fingerprint

hcarter-775:
# Description of Change
Update the device fingerprint to use the latest device handling, introduced [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2653). Add a comment to deprecate this profile, since it no longer is required.

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -155,7 +155,7 @@ matterManufacturer:
     deviceLabel: OREIN Bath Fan OLO5S
     vendorId: 0x1396
     productId: 0x1001
-    deviceProfileName: light-color-level-fan
+    deviceProfileName: fan-modular
   - id: "5014/4214"
     deviceLabel: Linkind Smart Light Bulb
     vendorId: 0x1396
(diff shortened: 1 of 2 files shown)


ctowns (approved):

drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -155,7 +155,7 @@ matterManufacturer:
     deviceLabel: OREIN Bath Fan OLO5S
     vendorId: 0x1396
     productId: 0x1001
-    deviceProfileName: light-color-level-fan
+    deviceProfileName: fan-modular
```
tpmanley, line 158: `fan-modular` doesn't have all the capabilities that `light-color-level-fan` has

hcarter-775: That's true- it has all the fan related capabilities though, and the light capabilities will be included as a child device. This work was completed in [this PR](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2653) after some discussion a few months ago.

PR #2914, opened 2026-04-21
Matter Switch: Add Stateless Native Handler Registration

hcarter-775:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

Add native handler registration for stateless capabilities

# Summary of Completed Tests

Tested with Zemismart Bulb

### DIFF
--- a/drivers/SmartThings/matter-switch/src/test/test_stateless_step.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_stateless_step.lua
@@ -63,6 +63,14 @@ test.register_message_test(
       }
     },
     {
+      channel = "devices",
+      direction = "send",
+      message = {
+        "register_native_capability_cmd_handler",
+        { device_uuid = mock_device_color_temp.id, capability_id = "statelessColorTemperatureStep", capability_cmd_id = "stepColorTemperatureByPercent" }
+      }
+    },
+    {
       channel = "matter",
       direction = "send",
       message = {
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/src/test/test_stateless_step.lua
cjswedes, line 1: If we were to pin a min_api_version for these tests, I think we would use version 11 since that is when the register native capability command api became available (FW 0.54). Not sure if were going back that far in our testing, but just wanted to call it out as what I think is the best min version for these tests.

cjswedes (approved):

nickolas-deboom (approved):

drivers/SmartThings/matter-switch/src/test/test_stateless_step.lua
hcarter-775: The stateless capabilities themselves are currently gated at v17 I think, also just a callout cause I don't disagree

mocelet:
Is it possible to pass or configure the transition time the native handler will use for the Step command?

While for some lights like KAJPLATS 0.3 seconds is fine, for others like WiZ the experience is better with 0 seconds (they apply their own fade and looks less "jagged") and for Nanoleaf 0.1 seconds seems to be better although that might change with their update to Matter 1.4 later this year.

An additional thought, while the number of devices using knob is low at the moment, there's the risk of a knob emitting events fast enough that the previous step hasn't finished yet and gets canceled (this actually happens in WiZ lights but for the opposite reason, they seem to take more than 0.3 seconds to perform the step and sometimes the last step is killed).

hcarter-775:
@mocelet, I plan to make this an available feature, since, as you mentioned, many devices respond to the transition time differently. I appreciate the callout / reminder!

PR #2918, opened 2026-04-22
Fix ledvance unit tests

cjswedes:
Device init messages must be disabled before adding the mock device to the test framework for these tests.

For release, this should go with the ledvance WWST commit in #2729

### DIFF
--- a/drivers/SmartThings/zigbee-switch/src/test/test_ledvance_metering_plug.lua
+++ b/drivers/SmartThings/zigbee-switch/src/test/test_ledvance_metering_plug.lua
@@ -23,6 +23,7 @@ local mock_device = test.mock_device.build_test_zigbee_device(
 zigbee_test_utils.prepare_zigbee_env_info()
 
 local function test_init()
+  test.disable_startup_messages()
   test.mock_device.add_test_device(mock_device)
 end
 
@@ -37,7 +38,10 @@ test.register_coroutine_test(
     test.wait_for_events()
     assert(mock_device:get_field(zigbee_constants.SIMPLE_METERING_MULTIPLIER_KEY) == 1)
     assert(mock_device:get_field(zigbee_constants.SIMPLE_METERING_DIVISOR_KEY) == 100)
-  end
+  end,
+  {
+    min_api_version = 15
+  }
 )
 
 test.register_coroutine_test(
@@ -49,7 +53,10 @@ test.register_coroutine_test(
     test.wait_for_events()
     assert(mock_device:get_field(zigbee_constants.SIMPLE_METERING_MULTIPLIER_KEY) == 5)
     assert(mock_device:get_field(zigbee_constants.SIMPLE_METERING_DIVISOR_KEY) == 1000)
-  end
+  end,
+  {
+    min_api_version = 15
+  }
 )
 
 test.run_registered_tests()


cjswedes:
@hcarter-775 ~~FYI, the test failing on 0.60 lua libs is: "matter camera FAILED on setStream with resolution change should trigger reallocation"~~ The base was out of date, nvm

aleclorimer (approved):

drivers/SmartThings/zigbee-switch/src/test/test_ledvance_metering_plug.lua
```diff
@@ -37,7 +38,10 @@ test.register_coroutine_test(
     test.wait_for_events()
     assert(mock_device:get_field(zigbee_constants.SIMPLE_METERING_MULTIPLIER_KEY) == 1)
     assert(mock_device:get_field(zigbee_constants.SIMPLE_METERING_DIVISOR_KEY) == 100)
-  end
+  end,
+  {
+    min_api_version = 15
```
aleclorimer, line 43: Was this tested against 15? HC 0.59 is running lua-libs v17.

cjswedes: I put it at 15 because that is when the disable_startup_messages was included in the test framework, so in theory it should work all the way back to version 15 if we ever include those earlier versions

cjswedes: Testing against 15 has unrelated failures due to stateless step capabilities in the driver. New capabilities in a driver are a separate issue that would need to be dealt with if we use those older api versions. Testing on 17 does pass since the capabilities are present

hcarter-775: I don't know if it makes sense to do backwards compat tests with the artificial limitation of capabilities not being generated, which is a UT only situation. To me, this should be worked around by the jenkins test framework rather than the devs, since new capabilities will continue to be made.

cjswedes: If we ever include tests for lua libs <17 we will need to address that. I agree that a work around would be best done in jenkins when running the tests. For now I am opting for setting the min_api_version to what it should be, but I can change to 17 if people want.

hcarter-775: I think I'm wondering what will happen when we add statelessWindowStep (or whatever) in the next release and roll that out to the drivers. Will that cause everything to fail?

aleclorimer: Ill look into updating jenkins to be using the latest capabilities. IDRC if we do 15 or 17 as we likely aren't going to be adding 58 to our testing matrix in the future. The only way we'd go below 17 is if we wanted more support into the Cell hubs back on 52.

hcarter-775: I think we should support 58, right? We still have some tv hubs on 58, no? Calling this out since HQ caught a 58 bug in one of my PRs about a month ago.

cjswedes: 60 is almost fully released to TVs, so there is not much need to include it. 60 will be the lowest FW version out there (besides the ST Wifi hubs on 52).

PR #2921, opened 2026-04-22
Matter/Zigbee Switch: Make transition time for stateless capabilities configurable

hcarter-775:
# Description of Change
Since one particular transition time has varied results across different drivers, there may be opportunity / need to alter this value in the future. In the drivers this is as easy as updating a value, but this change will allow the native handlers to also be updated by a driver restart, rather than requiring a full hub release to change. This also more easily allows different devices to have different transition times, as the need arises.

To avoid multiple writes of the same field, I placed the Matter logic for this in device init, so it only runs once on startup. However, because the lifecycle handlers for Zigbee are so often overwritten by devices that we'd like to include this functionality, I thought the most appropriate place for this would be in the handler itself. I could but did not add any special state logic to only set this once.

# Summary of Completed Tests
E2E test completed with a Matter Zemismart Bulb, showing that an overriden transition time was used in both a driver and native handler context.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -101,6 +101,7 @@ function SwitchLifecycleHandlers.device_init(driver, device)
     if #device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY}) == 0 then
       device:set_field(fields.profiling_data.BATTERY_SUPPORT, fields.battery_support.NO_BATTERY, {persist = true})
     end
+    switch_utils.set_transition_times_for_stateless_capabilities(device)
     switch_utils.handle_electrical_sensor_info(device)
     device:extend_device("subscribe", switch_utils.subscribe)
     device:subscribe()
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -101,7 +101,6 @@ function SwitchLifecycleHandlers.device_init(driver, device)
     if #device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY}) == 0 then
       device:set_field(fields.profiling_data.BATTERY_SUPPORT, fields.battery_support.NO_BATTERY, {persist = true})
     end
-    switch_utils.set_transition_times_for_stateless_capabilities(device)
     switch_utils.handle_electrical_sensor_info(device)
     device:extend_device("subscribe", switch_utils.subscribe)
     device:subscribe()
--- a/drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
@@ -194,8 +194,8 @@ SwitchFields.TEMP_BOUND_RECEIVED = "__temp_bound_received"
 SwitchFields.TEMP_MIN = "__temp_min"
 SwitchFields.TEMP_MAX = "__temp_max"
 
-SwitchFields.NULL_TRANSITION_TIME = 0 -- 0.0 seconds
-SwitchFields.TRANSITION_TIME_FAST = 3 -- 0.3 seconds, measured in tenths of a second as per the Matter spec
+SwitchFields.ZERO_TRANSITION_TIME = 0 -- 0.0 seconds
+SwitchFields.DEFAULT_STEP_TRANSITION_TIME = 3 -- 0.3 seconds, measured in tenths of a second as per the Matter spec
 
 SwitchFields.TRANSITION_TIME = {
   SWITCH_LEVEL_STEP = "__switch_level_step_transition_time",
--- a/drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
@@ -194,8 +194,13 @@ SwitchFields.TEMP_BOUND_RECEIVED = "__temp_bound_received"
 SwitchFields.TEMP_MIN = "__temp_min"
 SwitchFields.TEMP_MAX = "__temp_max"
 
-SwitchFields.TRANSITION_TIME = 0 -- number of 10ths of a second
-SwitchFields.TRANSITION_TIME_FAST = 3 -- 0.3 seconds
+SwitchFields.NULL_TRANSITION_TIME = 0 -- 0.0 seconds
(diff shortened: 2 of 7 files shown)


drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -194,8 +194,13 @@ SwitchFields.TEMP_BOUND_RECEIVED = "__temp_bound_received"
...
 
-SwitchFields.TRANSITION_TIME = 0 -- number of 10ths of a second
-SwitchFields.TRANSITION_TIME_FAST = 3 -- 0.3 seconds
+SwitchFields.NULL_TRANSITION_TIME = 0 -- 0.0 seconds
+SwitchFields.TRANSITION_TIME_FAST = 3 -- 0.3 seconds, measured in tenths of a second as per the Matter spec
+
+SwitchFields.TRANSITION_TIME = {
+  SWITCH_LEVEL_STEP = "__switch_level_step_transition_time",
+  COLOR_TEMP_STEP = "__color_temp_step_transition_time",
+}
```
gharveymn, line 203: IMO, inverting the struct (ie. `SwitchFields.COLOR_TEMP_STEP.TRANSITION_TIME` instead of `SwitchFields.TRANSITION_TIME.COLOR_TEMP_STEP`) might make sense since it would match the key names here, and also would be easier to organize if you want to add more dynamic fields in the future.

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
```diff
@@ -194,8 +194,13 @@ SwitchFields.TEMP_BOUND_RECEIVED = "__temp_bound_received"
 SwitchFields.TEMP_MIN = "__temp_min"
 SwitchFields.TEMP_MAX = "__temp_max"
 
-SwitchFields.TRANSITION_TIME = 0 -- number of 10ths of a second
-SwitchFields.TRANSITION_TIME_FAST = 3 -- 0.3 seconds
+SwitchFields.NULL_TRANSITION_TIME = 0 -- 0.0 seconds
```
gharveymn: I went looking in the spec because I thought you were setting the fields to Null rather than 0. Maybe `TRANSITION_TIME_ZERO`?

mocelet:
> but this change will allow the native handlers to also be updated by a driver restart

Can a custom driver with configurable transition time preferences update the native handlers without a driver restart? If just changing the field is not enough, would triggering a re-registration of the native handler work?

hcarter-775:
> Can a custom driver with configurable transition time preferences update the native handlers without a driver restart? If just changing the field is not enough, would triggering a re-registration of the native handler work?

Hi @mocelet, Yes, when the work here is complete, to update the transition time in the native handler, all that will be required is to update the associated field, whether via a custom preference or otherwise. Any change in that field will be automatically relayed to the already registered native handler.

mocelet:
> when the work here is complete, to update the transition time in the native handler, all that will be required is to update the associated field

Thanks @hcarter-775  , that's excellent!

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
```diff
@@ -20,6 +32,8 @@ local IGNORE_COMMAND_IF_OFF = 0x00 -- default: the command will not be executed
 local function step_color_temperature_by_percent_handler(driver, device, cmd)
   local step_percent_change = cmd.args and cmd.args.stepSize or 0
   if step_percent_change == 0 then return end
+  set_transition_time(device, TRANSITION_TIME.COLOR_TEMP_STEP)
+  local transition_time = DEFAULT_STATELESS_TRANSITION_TIME
```
cjswedes: I get that we want the transition time to be a field so that native handlers can key into any changes a subdriver wants to make, but if there ever was a device with a preference that changes the field, this would overwrite it. We can update the native handler to look for this field and fallback to the default if its not there, but I dont really see why we need this change for zigbee right now.

hcarter-775: That's a good point- I think we could work around that by doing an "if field nil check" before setting it so that other callers could update it. I do think this is a change we'd want for zigbee, to keep parity across supported protocols.

drivers/SmartThings/matter-switch/src/init.lua
```diff
@@ -101,6 +101,7 @@ function SwitchLifecycleHandlers.device_init(driver, device)
     if #device:get_endpoints(clusters.PowerSource.ID, {feature_bitmap = clusters.PowerSource.types.PowerSourceFeature.BATTERY}) == 0 then
       device:set_field(fields.profiling_data.BATTERY_SUPPORT, fields.battery_support.NO_BATTERY, {persist = true})
     end
+    switch_utils.set_transition_times_for_stateless_capabilities(device)
```
cjswedes: This is not applicable to all devices, but its getting set on them all. IMO this is not necessary; the handler can check for the fields presence and then fallback to the default there (which it already does), rather than setting the default for every device.

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
cjswedes: I think this handler should just be checking for the fields presence and then falling back to the default if the value is nil.

drivers/SmartThings/matter-switch/src/init.lua
hcarter-775: You're right, I think this helps answer the weird Zigbee behavior in its current state. Basically, rather than setting the field we should check if the field has been set and use it if so

hcarter-775: done in latest

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
hcarter-775: done in latest

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
hcarter-775: agree, done in latest- well actually did ZERO_TRANSITION_TIME but effectively the same

drivers/SmartThings/matter-switch/src/switch_utils/fields.lua
hcarter-775: Yeah, that makes sense to me, but having a higher order table with only one element seems a little odd, and there are already many "color temp" and "switch level" fields in use today.

So I was inspired by your suggestion and made the following, semi-dramatic but I think useful changeset, [in PR here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2922) to keep the work separate.

cjswedes (approved):

PR #2922, opened 2026-04-23
Organize the Matter Switch field naming conventions

hcarter-775:
# Description of Change
Done in response to [this comment](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2921#discussion_r3126989665), which made me think that if we were going to try to make a COLOR_TEMP_STEP table, what we really wanted was a COLOR_TEMP table that holds all the color temp-related fields. Then, to make this make sense with all the other assorted fields, I re-organized most of them into structures like this, which I think makes what field is associated with what make more sense from a top-down way.

However, since this is clearly a larger change than was initially suggested, I put this in its own PR for separate review.

As we can see, none of the tests were actually changed, as none of the driver functionality was changed in any way, just the naming conventions.
 
# Summary of Completed Tests


mocelet:
Would it be possible to somehow tag, differentiate, indicate with a comment or, less ideally, document elsewhere, which field names are not arbitrary but a contract since they are read by native handlers or other components outside the driver code?

Edit: Just realized inspecting the lua libs that they store the native handler context field names in a field so at least there is a programmatic way to know the names that are expected.

PR #2929, opened 2026-04-27
St protocol handlers single device thread

aleclorimer:
Enabling of `shared_device_thread` on the remaining ST protocol handlers. Split into a different commit for each driver.

### DIFF
--- a/drivers/SmartThings/matter-pump/src/init.lua
+++ b/drivers/SmartThings/matter-pump/src/init.lua
@@ -321,6 +321,7 @@ local matter_driver_template = {
     capabilities.pumpOperationMode,
     capabilities.pumpControlMode,
   },
+  shared_device_thread_enabled = true,
 }
 
 local matter_driver = MatterDriver("matter-pump", matter_driver_template)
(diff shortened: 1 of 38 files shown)


cjswedes (commented):
This field is only considered at the top level driver, subdriver init files do not need an update.

varzac:
# Possible Redundant  in Subdrivers

The following files set  and appear to be subdrivers (not top-level driver entry files):

- drivers/SinuxSoft/britzyhub/src/elevator/init.lua
- drivers/SinuxSoft/britzyhub/src/gas-valve/init.lua

Can the maintainers confirm if these fields are intentionally required in subdrivers? If not, it may be cleaner to remove for consistency, as only the primary driver should generally require this per ST convention.

**AIREVIEW** tag added for internal tracking.

---
This review was generated using the OpenAI/GPT automation for code review.

varzac:
It did not do a very good job of identifying the list of subdrivers.  And Carter had already mentioned the general comment, so not sure how much value the AIReview added to this PR.

drivers/SmartThings/matter-pump/src/init.lua
aleclorimer, line 1: Must be a CRLF vs LF line ending change that only has popped up from this file. Only shows a change on line 323->324 when you hide whitespace changes.

aleclorimer:
> It did not do a very good job of identifying the list of subdrivers. And Carter had already mentioned the general comment, so not sure how much value the AIReview added to this PR.

Just a general question, but any way I can kick this agent to run again or will it run automatically?

varzac (approved):

PR #2930, opened 2026-04-28
Matter Switch: use parent device for get_field calls in capability commands

hcarter-775:
# Description of Change
Explicitly use the parent device when getting the custom color temperature fields in the capability handlers. Without this, child devices will not be able to access their endpoint-specific fields, as these are only stored on the parent device.

# Summary of Completed Tests
Unit tests updated with tests to handle the capability commands from child devices.

Further, the entire parent/child test file was re-worked in order to more clearly describe in several places what is actually being tested, with more focused tests being created and several tests being merged together to better understand what is being tested.

Also, there are several "on off parent child"-like tests across the integration tests, and I have moved them all into one file to further add clarity. Doing this involved some re-work of the matter_switch_device_types unit tests, which has had several tests moved/consolidated, and has been renamed to matter_on_off_device_configuration for further clarity.


drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
```diff
@@ -108,14 +108,16 @@ end
 function CapabilityHandlers.handle_set_color_temperature(driver, device, cmd)
   local endpoint_id = device:component_to_endpoint(cmd.component)
   local temp_in_kelvin = cmd.args.temperature
-  local min_temp_kelvin = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_KELVIN..fields.COLOR_TEMP_MIN, endpoint_id)
-  local max_temp_kelvin = switch_utils.get_field_for_endpoint(device, fields.COLOR_TEMP_BOUND_RECEIVED_KELVIN..fields.COLOR_TEMP_MAX, endpoint_id)
+  -- note: the field containing the color temp bounds will be associated with a parent device
+  local field_device = device:get_parent_device() or device
+  local min_temp_kelvin = switch_utils.get_field_for_endpoint(field_device, fields.COLOR_TEMP_BOUND_RECEIVED_KELVIN..fields.COLOR_TEMP_MIN, endpoint_id)
```
ctowns, line 114: Could you update the `get_field_for_endpoint` util function itself to just get the parent if the current device is a child? That would prevent the need to add in these additional checks to get the "field device" in these handlers. I assume there is no case where we save fields on the child device?

Also, one thing I might be misunderstanding is -if the fields are not handled on child devices, how would they be used by the native handlers for those devices?

hcarter-775: To your first point, I think that's a cool idea, though this sort of logic is also implemented rn for a non "for_endpoint" field, see [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua#L224), and I don't love the idea of having 2 paths for this. What do you think? I do like the clarity of the current approach.

To your second, parent/child modelled devices are not handled today by native handlers, partially for reasons related to complexity like this.

drivers/SmartThings/matter-switch/src/test/test_matter_on_off_device_configuration.lua
```diff
@@ -0,0 +1,306 @@
...
+      },
+      device_types = {
+        {device_type_id = 0x0103, device_type_revision = 1} -- On/Off Light Switch
+      }
+    }
+  }
+})
+
+test.register_coroutine_test(
+  "Test init for onoff parent cluster as client",
```
cjswedes, line 77: Does this device remain as a matter thing?

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
cjswedes, line 1: Could you add a test for use of these fields for zigbee (something like the "Extended Color Child: colorTemperatureRange, setColorTemperature, stepColorTemperatureByPercent handled appropriately" you added for matter). I am not seeing one right now.

cjswedes (approved):

drivers/SmartThings/matter-switch/src/test/test_matter_on_off_device_configuration.lua
hcarter-775: Yes, since we don't want to re-profile devices with on/off clusters that are handled as "client" rather than "server".

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
hcarter-775: So here for example is a Zigbee test that uses these fields: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-switch/src/test/test_sengled_color_temp_bulb.lua#L283, but for a parent device. The issue I'm seeing in testing children (like the test you mentioned) is that there is kinda no way today for the Zigbee Switch driver to generically produce an EDGE_CHILD device with the stateless step capabilities. None of the manufacturer-defined subdrivers create children that utilize any of these capabilities.

drivers/SmartThings/matter-switch/src/switch_handlers/capability_handlers.lua
ctowns: Yeah I think the current approach is good, I didn't realize we had another use case for this as well so thanks for linking that!

Thanks for the clarification on parent/child devices with native handlers, I had forgotten about that :+1:

ctowns (approved):

drivers/SmartThings/zigbee-switch/src/stateless_handlers/init.lua
cjswedes: Sounds good, no need to add anything then.

PR #2931, opened 2026-04-29
[Aqara/Locks] Remove unnecessary sequence number increment check

haedo-doo:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [x] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/Aqara/aqara-lock/src/init.lua
+++ b/drivers/Aqara/aqara-lock/src/init.lua
@@ -253,16 +253,11 @@ local function lock_state_handler(driver, device, value, zb_rx)
     local payload = string.sub(text, 4, string.len(text))
     local func_id = toValue(payload, 1, 1) .. "." .. toValue(payload, 2, 1) .. "." .. toValue(payload, 3, 2)
     local serial_num = toValue(msg, 3, 2)
-    local last_serial_num = device:get_field(SERIAL_NUM_RX) or 0
 
-    if serial_num > last_serial_num then
-      device:set_field(SERIAL_NUM_RX, serial_num)
-      if resource_id[func_id] then
-        resource_id[func_id].event_handler(driver, device, resource_id[func_id].event_name,
-          toValue(payload, 6, string.byte(payload, 5)))
-      end
-    else
-      request_generate_shared_key(device)
+    device:set_field(SERIAL_NUM_RX, serial_num)
+    if resource_id[func_id] then
+      resource_id[func_id].event_handler(driver, device, resource_id[func_id].event_name,
+        toValue(payload, 6, string.byte(payload, 5)))
     end
   end
 end
--- a/drivers/Aqara/aqara-lock/src/init.lua
+++ b/drivers/Aqara/aqara-lock/src/init.lua
@@ -23,7 +23,6 @@ local CLOUD_PUBLIC_KEY = "__cloud_public_key"
 local SUPPORTED_ALARM_VALUES = { "damaged", "forcedOpeningAttempt", "unableToLockTheDoor", "notClosedForALongTime",
   "highTemperature", "attemptsExceeded" }
 local SERIAL_NUM_TX = "serial_num_tx"
-local SERIAL_NUM_RX = "serial_num_rx"
 local SEQ_NUM = "seq_num"
 local THRESHOLD_BATTERY = {
   ["aqara.lock.akr011"] = { low = 47, dryout = 28 },  -- K100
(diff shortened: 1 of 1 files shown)


haedo-doo:
Avoid redundant key exchanges triggered by out-of-order Zigbee messages under poor signal conditions.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -253,16 +253,11 @@ local function lock_state_handler(driver, device, value, zb_rx)
...
 
-    if serial_num > last_serial_num then
-      device:set_field(SERIAL_NUM_RX, serial_num)
-      if resource_id[func_id] then
-        resource_id[func_id].event_handler(driver, device, resource_id[func_id].event_name,
-          toValue(payload, 6, string.byte(payload, 5)))
-      end
-    else
-      request_generate_shared_key(device)
+    device:set_field(SERIAL_NUM_RX, serial_num)
```
tpmanley: Do you anticipate a future use of this or could be considered for removal? In the current code I don't see anything else reading the `SERIAL_NUM_RX` field.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -253,16 +253,11 @@ local function lock_state_handler(driver, device, value, zb_rx)
     local payload = string.sub(text, 4, string.len(text))
     local func_id = toValue(payload, 1, 1) .. "." .. toValue(payload, 2, 1) .. "." .. toValue(payload, 3, 2)
     local serial_num = toValue(msg, 3, 2)
-    local last_serial_num = device:get_field(SERIAL_NUM_RX) or 0
 
-    if serial_num > last_serial_num then
```
cjswedes, line 258: With this check removed, how will duplicate/stale messages that are sent by a device be detected? It seems like this could result in extra/stale lock events being emitted.

drivers/Aqara/aqara-lock/src/init.lua
```diff
@@ -253,16 +253,11 @@ local function lock_state_handler(driver, device, value, zb_rx)
...
-      device:set_field(SERIAL_NUM_RX, serial_num)
-      if resource_id[func_id] then
-        resource_id[func_id].event_handler(driver, device, resource_id[func_id].event_name,
-          toValue(payload, 6, string.byte(payload, 5)))
-      end
-    else
-      request_generate_shared_key(device)
```
cjswedes, line 265: What is the initial reason for having this here? are there cases where a key does need to be generated if the message serial numbers start getting out of order?

cjswedes (commented):
There currently are not any tests for how an old serial number in a message is handled. Please add a test to show what events the driver will emit if such a message is received.

drivers/Aqara/aqara-lock/src/init.lua
haedo-doo: Thanks for the feedback. I've removed the unnecessary fields.

drivers/Aqara/aqara-lock/src/init.lua
haedo-doo: Thanks for the feedback. The main reason for this change is to prevent an infinite loop of shared key regeneration caused by out-of-order messages in unstable Zigbee networks.

In the current logic, if a message (e.g., N+1) is lost but subsequent messages (N+2 to N+8) arrive first, the driver expects the next sequence to be higher than N+8. When the door lock later retries the lost N+1, the driver treats it as an invalid or stale packet and immediately triggers the shared key regeneration process. In poor network conditions, this cycle repeats indefinitely, leading to a deadlock where the device becomes unusable.

drivers/Aqara/aqara-lock/src/init.lua
haedo-doo: While this validation was initially implemented following the manufacturer's guide, it causes severe side effects in real-world environments. Therefore, I decided to remove it to ensure connectivity and stability.

Regarding the test cases, the SmartThings Edge Driver framework does not support simulating the encryption key generation process or assigning arbitrary keys. Since the door lock communicates all data with the hub exclusively through encrypted messages, it is currently impossible to implement meaningful test cases that incorporate this encryption logic. We ask for your understanding regarding these platform-level constraints.

cjswedes: Given the lack of ability to add unit tests, I assume this is tested manually.

cjswedes (approved):

hcarter-775 (approved):

kdbang:
As far as I know, this logic is for preventing key mismatch issue. so the checking sequence number logic was added. This PR is about removing key-regeneration. Is there any other way to address key mismatch issue?

PR #2938, opened 2026-04-30
Matter Switch: Version gate the extra init event on added

hcarter-775:
# Description of Change
As of Hub FW 58, the caching issues have been solved and so this extra event is not required any longer. 

# Summary of Completed Tests
This changeset is large primarily because I took this opportunity to restructure several test files that were checking the entire added -> init -> doConfigure cycle for every test, or that were testing pieces of code that were already tested elsewhere verbatim. I also just removed the added check in several places for simplicity, rather than re-versioning ~100 tests.

In particular, I really took the hammer to [test_matter_switch.lua‎](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2938/changes#diff-e3a5fd063cee499ccdbc8f49df165cd06bbc0b389e4bddacc8acfd407dab37f3) and [test_light_illuminance_motion.lua‎](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2938/changes#diff-c6d721e804fe1a90b244e4d823e7f396abd57142a36bf0cc8e98497073890f2e), since there was significant overlap between these 2 and they were the first ones I looked at. I think a similar treatment is likely due for many of our Generic Switch tests, but I selectively reworked pieces as time permitted.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/test/test_matter_button.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_matter_button.lua
@@ -6,9 +6,10 @@ local capabilities = require "st.capabilities"
 local t_utils = require "integration_test.utils"
 local clusters = require "st.matter.clusters"
 local button_attr = capabilities.button.button
-local utils = require "st.utils"
-local dkjson = require "dkjson"
 local uint32 = require "st.matter.data_types.Uint32"
+local fields = require "switch_utils.fields"
+local switch_utils = require "switch_utils.utils"
+local st_utils = require "st.utils"
 
 local mock_device = test.mock_device.build_test_matter_device({
   profile = t_utils.get_profile_definition("button.yml"),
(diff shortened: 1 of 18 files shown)


drivers/SmartThings/matter-switch/src/test/test_matter_button.lua
```diff
@@ -498,67 +412,205 @@ test.register_message_test(
...
+      mock_device_battery, 1, {uint32(clusters.PowerSource.attributes.BatPercentRemaining.ID)}
+    )})
+    expect_configure_button(mock_device_battery)
+    mock_device_battery:expect_metadata_update({ profile = "button-battery" })
+  end,
+  {
+    test_init = test_init_profile_change_with_battery,
+    min_api_version = 17
+  }
+)
```
cjswedes, line 542: This is the same test as the one above.

cjswedes (approved):

drivers/SmartThings/matter-switch/src/test/test_matter_button.lua
hcarter-775: very true, good callout. I'd initially only been planning to test the one path, but expanded it to the other cases and forgot to remove the initial one.

cjswedes:
Looks like there are several failing tests.

hcarter-775:
@cjswedes yeah I'm looking into that, we just merged the Irrigation System PR and I need to update those tests

hcarter-775:
New unit tests updated in a similar way.

PR #2937, opened 2026-04-30
CHAD-16364 & CHAD-16365: Add lockUsers and lockCredentials capability

cbaumler:
This PR replaces https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2416. I've re-arranged the commits into two: one covering Zigbee and one covering Z-Wave to make review easier. I've also rebased onto `main` and done some extensive conflict resolution (using AI). Here's the summary of the changes required to resolve conflicts:

- Kept the new split-driver architecture in zigbee-lock/src/init.lua and zwave-lock/src/init.lua because the rebase was introducing capability-mode routing, not reverting it.
- Resolved both top-level [sub_drivers.lua](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/feature/chad-16364/drivers/SmartThings/zigbee-lock/src/sub_drivers.lua) and [sub_drivers.lua](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/feature/chad-16364/drivers/SmartThings/zwave-lock/src/sub_drivers.lua) to load the old/new capability mode selectors directly, which is the correct shape for the new dispatcher layout.
- Flattened the mode wrappers in `zigbee-lock/src/using-old-capabilities/init.lua` and `zwave-lock/src/using-old-capabilities/init.lua` because wrapping require("...sub_drivers") in an extra table would have produced the wrong subdriver structure.
- Preserved `zwave-lock/src/apiv6_bugfix/init.lua` and added it back to the Z-Wave subdriver list because it is still a live compatibility workaround, not dead code.
- Kept the legacy fingerprint-based handlers in the moved using-old-capabilities tree, since the merge was a directory split and path rewrite, not a behavior change for those devices.
- Merged the Yale fingerprint-lock test in `zigbee-lock/src/test/test_zigbee_yale-fingerprint-lock.lua` so both the old-capability and migrated-lock paths remain covered after the capability migration.

# Type of Change

- [ ] WWST Certification Request
     - If this is your first time contributing code:
          - [ ] I have reviewed the README.md file
          - [ ] I have reviewed the CODE_OF_CONDUCT.md file
          - [ ] I have signed the CLA
     - [ ] I plan on entering a WWST Certification Request or have entered a request through the WWST Certification console at developer.smartthings.com
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

- [x] I have performed a self-review of my code
- [ x] I have commented my code in hard-to-understand areas
[...]

PR #2939, opened 2026-05-01
Zigbee Lock: Support SLGA Migration

hcarter-775:
# Description of Change

Note on structure: The legacy-handlers subdriver keeps some subdrivers that have specific non-conformant behavior that relates specifically to lockCodes. This includes lock-without-codes, which no longer needs to exist in the modern implementation, as well as the yale-fingerprint-lock subdriver, which sets the lockCodes maxCodes state. Similarly, the yale subdriver has lockCodes-specific behavior as well.

However, upon analysis it was clear that the sub-sub-driver yale-bad-battery-reporter subdriver was 1. not unique to yale, and 2. was distinct from any lockCodes behavior, so it was pulled out into the now default subdrivers. Also, the samsungds subdriver had minimal differences between the old and new handling, so I condensed them.

The other primary thing that was removed from the now-legacy-handlers subdriver is the old added handler, for the following 2 reasons. 1. Generally, devices should not be added to this driver any longer, and 2. We now use the added main driver implementation to initially set the migrated keyword to true.

I find the fact that the entire migrated state depends on a keyword set only once in added to be a little precarious, but nothing stands out to me initially at this not working.

# Summary of Completed Tests
Migration process, plus lockUsers/lockCredentials functionality, has been tested on-device with two different kwikset locks.

Unit tests added (I renamed some of the previous tests to be `_legacy`, to clarify that they are testing pre-migration functionality.


drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -96,228 +59,471 @@ local alarm_handler = function(driver, device, zb_mess)
...
-    if (lock_utils.get_lock_codes(device)[code_slot] ~= nil) then
-      -- Code has been deleted
-      lock_utils.lock_codes_event(device, lock_utils.code_deleted(device, code_slot))
-    else
-      -- Code is unset
-      event.value = code_slot .. " unset"
-      device:emit_event(event)
-    end
+local function device_added(driver, device)
+  if device:supports_capability_by_id(capabilities.lockCodes.ID) and device._provisioning_state == "TYPED" then
```
cjswedes, line 63: This provisioning state typed check should only not be true for devices that are driver switched, do we need to consider the driver switch case wrt migration? I assume this would rarely happen and it it did, it would be a user moving from a custom driver to ours, in which case IMO we can ask them to re-onboard if anything goes wrong.

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -96,228 +59,471 @@ local alarm_handler = function(driver, device, zb_mess)
...
     end
-    device:emit_event(LockCodes.maxCodes(value.value, { visibility = { displayed = false } }))
   end
 end
 
-local handle_max_code_length = function(driver, device, value)
-  device:emit_event(LockCodes.maxCodeLength(value.value, { visibility = { displayed = false } }))
+local add_user_handler = function(driver, device, command)
+  if lock_utils.busy_check_and_set(device, {name = lock_utils.ADD_USER, type = lock_utils.LOCK_USERS}) then
+    return
```
cjswedes, line 173: If this happens, whats the user experience? Does the app just time out?

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -96,228 +59,471 @@ local alarm_handler = function(driver, device, zb_mess)
...
+    device:set_field(lock_utils.CHECKING_CODE, 1)
+  end
+
+  device:send(LockCluster.server.commands.GetPINCode(device, device:get_field(lock_utils.CHECKING_CODE)))
+end
+
+local init = function(driver, device)
+  lock_utils.reload_tables(device)
+  device.thread:call_with_delay(15, function(d)
+    reload_all_codes(device)
```
tpmanley, line 138: Do we have to reload all codes every time or can we put a guard on this so it's only done in the cases where it's needed?

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -96,228 +59,471 @@ local alarm_handler = function(driver, device, zb_mess)
...
+    if lock_utils.get_credential(device, credential_index) ~= nil then
+      -- Credential has been deleted.
+      lock_utils.delete_credential(device, credential_index)
+      emit_event = true
+    end
+  end
+
+  if (credential_index == device:get_field(lock_utils.CHECKING_CODE)) then
+    -- the credential we're checking has arrived
+    local last_slot = device:get_latest_state("main", capabilities.lockCredentials.ID,
```
tpmanley, line 460: Are we guaranteed to have `pinUsersSupported` state at this point or do we need to handle the case where it's not set yet?

drivers/SmartThings/zigbee-lock/src/init.lua
hcarter-775: Seems that all of the actual sends in this function are gated on if the cache is actually empty, so I think that's effectively done.

tpmanley: It doesn't look like `SendPINOverTheAir` and `GetPINCode` are gated

hcarter-775: Ah yeah, my bad. Seems like this was 1-1 copied from the current [Zigbee Lock handling](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-lock/src/init.lua#L33) (which is now in legacy-handlers), and the capability cache checks have just been moved to the new capabilities.

drivers/SmartThings/zigbee-lock/src/legacy-handlers/init.lua
```diff
@@ -0,0 +1,374 @@
...
+  local max_codes    = device:get_latest_state("main", capabilities.lockCodes.ID, capabilities.lockCodes.maxCodes.NAME, 0)
+  if (code_length ~= nil) then
+    max_code_len = code_length
+    min_code_len = code_length
+  end
+
+  device:emit_event(LockCredentials.minPinCodeLen(min_code_len, { visibility = { displayed = false } }))
+  device:emit_event(LockCredentials.maxPinCodeLen(max_code_len, { visibility = { displayed = false } }))
+  device:emit_event(LockCredentials.pinUsersSupported(max_codes, { visibility = { displayed = false } }))
+  device:emit_event(LockCredentials.credentials(lock_credentials, { visibility = { displayed = false } }))
```
cjswedes, line 324: You could consider setting the credential and user tables on the device at this point. I think everything should work since the tables are reloaded when accessed, but its a bit fragile and potentially prone to race conditions to rely on the state cache to be immediately updated.

drivers/SmartThings/zigbee-lock/src/zigbee_lock_utils.lua
```diff
@@ -0,0 +1,309 @@
...
+  TABLES_LOADED = "tablesLoaded",
+  UPDATE_CREDENTIAL = "updateCredential",
+  UPDATE_USER = "updateUser",
+  USER_INDEX = "userIndex",
+  USER_NAME = "userName",
+  USER_TYPE = "userType",
+  SLGA_MIGRATED = "slgaMigrated"
+}
+
+lock_utils.DELETE_ALL_USERS = 0xFF
```
cjswedes, line 39: This overrides what was previously set (`"deleteAllUsers"`). I think everything will still work as intended (just a weird command status name during the delete operation, but it does look like a potential bug, and is misleading. I think you should a new constant for the protocol credential index special value that means all users.

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -96,228 +59,471 @@ local alarm_handler = function(driver, device, zb_mess)
...
+      device:set_field(lock_utils.CHECKING_CODE, checkingCode)
+      device:send(LockCluster.server.commands.GetPINCode(device, checkingCode))
+    end
+  end
+
+  if emit_event then
+    lock_utils.send_events(device)
+  end
+  -- ignore handling the busy state for these commands, they are handled within their own handlers
+  if command ~= nil and command ~= lock_utils.DELETE_ALL_CREDENTIALS and command ~= lock_utils.DELETE_ALL_USERS then
```
cjswedes, line 476: These `command ~= ...` checks will always be true since command is a table.

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -96,228 +59,471 @@ local alarm_handler = function(driver, device, zb_mess)
...
-  device:send(LockCluster.server.commands.GetPINCode(device, command.args.codeSlot))
-end
+local add_credential_handler = function(driver, device, command)
+  if lock_utils.busy_check_and_set(device, {name = lock_utils.ADD_CREDENTIAL, type = lock_utils.LOCK_CREDENTIALS}) then
+    return
+  end
+  local user_index = tonumber(command.args.userIndex)
+  local user_type = command.args.userType
+  local credential_type = command.args.credentialType
+  local credential_data = command.args.credentialData
```
cjswedes, line 281: I think we should also extract credentialName, and include that in the `ACTIVE_CREDENTIAL` field, so the data is not lost when we emit an event. Same goes for the other places we set ACTIVE_CREDENTIAL in update_credential_handler and delete_credential_handler, and then the lock_utils.add_credential function would need to consider it as well (I dont think I missed any...)

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -349,22 +555,22 @@ local lock_operation_event_handler = function(driver, device, zb_rx)
...
-          lock_codes[code_id] ~= nil) then
-        code_name = lock_codes[code_id]
+      local code_name = "Code " .. code_id
+      local user = lock_utils.get_user(device, code_id)
+      if user ~= nil then
+        code_name = user.userName
       end
-      event.data = {method = METHOD[0], codeId = code_id .. "", codeName = code_name}
+
+      event.data = { method = METHOD[0], codeId = code_id .. "", codeName = code_name }
```
cjswedes, line 573: I think this is the old event data, and should include the new lock capability fields. The full list of fields in the lock attribute are:
- codeId
- codeName
- userIndex
- userName
- userType
- timeout
- unlockDirection

I am not sure what the plugin expects for these events, but the requirements say that unlock events should include the user information, and the codeId/codeName are legacy fields.

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -349,22 +555,22 @@ local lock_operation_event_handler = function(driver, device, zb_rx)
...
-    if (source == 0 and device:supports_capability_by_id(capabilities.lockCodes.ID)) then --keypad
+    if (source == 0 and device:supports_capability_by_id(capabilities.lockUsers.ID)) then --keypad
       local code_id = zb_rx.body.zcl_body.user_id.value
-      local code_name = "Code "..code_id
-      local lock_codes = device:get_field("lockCodes")
-      if (lock_codes ~= nil and
-          lock_codes[code_id] ~= nil) then
-        code_name = lock_codes[code_id]
+      local code_name = "Code " .. code_id
+      local user = lock_utils.get_user(device, code_id)
```
cjswedes, line 568: The zigbee message's `user_id` maps to our credential capability not the users capability. 

```suggestion
      local credential = lock_utils.get_credential(device, code_id)
      local user = lock_utils.get_user(device, credential.userIndex)
```

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -96,228 +59,471 @@ local alarm_handler = function(driver, device, zb_mess)
...
+  local delay = 0
+  for _, credential in pairs(credentials) do
+    local credential_index = tonumber(credential.credentialIndex)
+    device.thread:call_with_delay(delay, function()
+      device:send(LockCluster.server.commands.ClearPINCode(device, credential_index))
+    end)
+    device.thread:call_with_delay(delay + 2, function(d)
+      device:send(LockCluster.server.commands.GetPINCode(device, credential_index))
+    end)
+    delay = delay + 2
```
cjswedes, line 384: If there are more than 5 credentials being deleted, the "busy state" will be cleared prior to the command being completed.

We could consider upping the busy state timeout to be 30 or 60 seconds, but it comes with the possibility that the user will see unexplained failures if they try to deal with any failure situations by issuing more commands within that window of the initial failed command.

cjswedes: A similar timing situation occurs with delete_all_users

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -96,228 +59,471 @@ local alarm_handler = function(driver, device, zb_mess)
...
+    device:send(LockCluster.attributes.MaxPINCodeLength:read(device))
   end
+  if (device:get_latest_state("main", capabilities.lockCredentials.ID, capabilities.lockCredentials.minPinCodeLen.NAME) == nil) then
+    device:send(LockCluster.attributes.MinPINCodeLength:read(device))
+  end
+  if (device:get_latest_state("main", capabilities.lockCredentials.ID, capabilities.lockCredentials.pinUsersSupported.NAME) == nil) then
+    device:send(LockCluster.attributes.NumberOfPINUsersSupported:read(device))
+  end
+  if (device:get_latest_state("main", capabilities.lockUsers.ID, capabilities.lockUsers.totalUsersSupported.NAME) == nil) then
+    device:send(LockCluster.attributes.NumberOfTotalUsersSupported:read(device))
```
cjswedes, line 126: I dont think this is needed. 1 there is no handler for the attribute, and 2 it has no bearing on our credentials/user capabilities. I think the total number of users supported is equal to the total number of pin users supported, although I am not entirely sure how the legacy driver supported RFID credentials on zigbee locks...the max users according to the spec is the larger of NumberOfPINUsersSupported or NumberOfRFIDUsersSupported

drivers/SmartThings/zigbee-lock/src/legacy-handlers/yale-fingerprint-lock/can_handle.lua
```diff
@@ -0,0 +1,13 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+return function(opts, driver, device)
+  local fingerprints = require("yale-fingerprint-lock.fingerprints")
```
nickolas-deboom, line 5: I think this should be
```suggestion
  local fingerprints = require("legacy-handlers.yale-fingerprint-lock.fingerprints")
```

drivers/SmartThings/zigbee-lock/src/legacy-handlers/legacy_lock_utils.lua
```diff
@@ -0,0 +1,89 @@
...
+  device:emit_event(event)
+  legacy_lock_utils.reset_code_state(device, code_slot)
+  return lock_codes
+end
+
+function legacy_lock_utils.populate_state_from_data(device)
+  if device.data.lockCodes ~= nil and device:get_field(legacy_lock_utils.MIGRATION_COMPLETE) ~= true then
+    -- build the lockCodes table
+    local lockCodes = {}
+    local lc_data = json.decode(device.data.lockCodes)
```
nickolas-deboom, line 74: Maybe add a nil check for device.data.lockCodes? in case migration is incomplete or something

drivers/SmartThings/zigbee-lock/src/init.lua
ctowns: FWIW in the matter lock driver, we just have a separate handler for the `driverSwitched` event that mirrors `do_configure`,  If we want to address the driver switch case, I  think we could add handling for the driver switch case by encapsulating the configuration logic in a helper function and just calling that same function from the `doConfigure` and the `driverSwitched` handler.

cjswedes: IMO driver switching locks should be treated like a LAN device, and basically not be supported officially. Should anyone try to do it and face issues, the solution is re-onboarding.

drivers/SmartThings/zigbee-lock/src/zigbee_lock_utils.lua
cjswedes: Actually I think we will emit an event for command status with 0xFF as the command name.

drivers/SmartThings/zigbee-lock/src/init.lua
hcarter-775: That was my thinking. Agreed that we could add a driverSwitched handler if we wanted to support that, but it should not match the normal user experience.

drivers/SmartThings/zigbee-lock/src/legacy-handlers/yale-fingerprint-lock/can_handle.lua
hcarter-775: Nah, it is correctly pointed. The same fingerprints accessing this legacy subdriver should access the updated subdriver.

drivers/SmartThings/zigbee-lock/src/init.lua
hcarter-775: Fyi I've decided to make some updates to all the command name stuff, so I'd hold off on reviewing this for the moment.

drivers/SmartThings/zigbee-lock/src/init.lua
hcarter-775: Also, fyi the `added` handler shouldn't really be related to the `SWITCHED` state.

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -96,228 +59,471 @@ local alarm_handler = function(driver, device, zb_mess)
...
-    local last_slot = device:get_latest_state("main", capabilities.lockCodes.ID, capabilities.lockCodes.maxCodes.NAME) - 1
-    if (code_slot >= last_slot) then
-      device:emit_event(LockCodes.scanCodes("Complete", { visibility = { displayed = false } }))
-      device:set_field(lock_utils.CHECKING_CODE, nil)
-    else
-      local checkingCode = device:get_field(lock_utils.CHECKING_CODE) + 1
-      device:set_field(lock_utils.CHECKING_CODE, checkingCode)
-      device:send(LockCluster.server.commands.GetPINCode(device, checkingCode))
-    end
+  driver:inject_capability_command(device, {
```
ctowns, line 72: Do you know why this is done via injecting the `refresh` command rather than just calling a separate function? If a delay is necessary, a timer could be used. This might just be something historical from the old driver code, I just didn't know the reason why it was done this way and was curious.

drivers/SmartThings/zigbee-lock/src/legacy-handlers/init.lua
hcarter-775: Not sure what you mean by this, we are just updating our own internal models here, so we shouldn't need to update the device itself.

cjswedes: We are emitting capability events for the credentials and users that the device has based on what was stored in the lock codes (persisted device field). After migration credentials and users are stored as transient device fields, and are essentially a cache of what the state store has for the capability attribute state. There is a potential race condition where the events issued during migration are not cached prior to the next get_latest_state being called to read the state of those attributes. I think the way the hub to driver messaging currently works makes this guarantee to work as expected, but its an implicit guarantee. 

In my mind there are 3 options:
1. Do nothing as implicitly we are guaranteed to get the correct state do to the way the driver -> hub messaging works.
2. Call get_latest_state for lockUsers and lockCredentials prior to emitting the events (the device object will cache the existing values, and the events emitted will update the cache before the driver even sends the event to the hub).
3. Do what is done in reload_tables and set the lock users/lock credentials fields on the device object.

I am fine with 1, just wanted to call it out that we are relying on multiple round trip communications to the hub by relying on the state cache to be updated immediately following event emission.

hcarter-775:
Closing in favor of: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2956

PR #2942, opened 2026-05-04
[Aqara] add Aqara Bath Heater T1

seojune79:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/Aqara/aqara-bath-heater/src/init.lua
+++ b/drivers/Aqara/aqara-bath-heater/src/init.lua
@@ -487,7 +487,11 @@ local aqara_bathroom_heater_driver_template = {
   supported_capabilities = {
     capabilities.switch,
     capabilities.switchLevel,
-    capabilities.colorTemperature
+    capabilities.colorTemperature,
+    capabilities.thermostatMode,
+    capabilities.thermostatHeatingSetpoint,
+    capabilities.fanOscillationMode,
+    capabilities.fanMode
   },
 
   capability_handlers = {
--- a/drivers/Aqara/aqara-bath-heater/src/init.lua
+++ b/drivers/Aqara/aqara-bath-heater/src/init.lua
@@ -324,6 +324,12 @@ local function handle_fan_mode(driver, device, cmd)
   send_ac_code(device, { fan = fan })
 end
 
+local function handle_refresh(driver, device, cmd)
+  device:send(OnOff.attributes.OnOff:read(device))
+  device:send(Level.attributes.CurrentLevel:read(device))
+  device:send(ColorControl.attributes.ColorTemperatureMireds:read(device))
+end
+
 -- Zigbee attribute handlers
 local function on_off_attr_handler(driver, device, value, zb_rx)
   device:emit_event(capabilities.switch.switch(value.value and "on" or "off"))
@@ -538,6 +544,7 @@ local aqara_bathroom_heater_driver = ZigbeeDriver("aqara-bathroom-heater-t1", {
     capabilities.thermostatHeatingSetpoint,
     capabilities.fanOscillationMode,
     capabilities.fanMode,
+    capabilities.refresh,
   },
 
   capability_handlers = {
@@ -562,6 +569,9 @@ local aqara_bathroom_heater_driver = ZigbeeDriver("aqara-bathroom-heater-t1", {
     },
     [capabilities.fanMode.ID] = {
       [capabilities.fanMode.commands.setFanMode.NAME] = handle_fan_mode,
+    },
+    [capabilities.refresh.ID] = {
+      [capabilities.refresh.commands.refresh.NAME] = handle_refresh,
     },
   },
 
--- a/drivers/Aqara/aqara-bath-heater/src/init.lua
+++ b/drivers/Aqara/aqara-bath-heater/src/init.lua
@@ -273,10 +273,6 @@ local function handle_heating_setpoint(driver, device, cmd)
   if cur == ST_MODE.HEAT then
     send_ac_code(device, { setpoint = temp_c })
   end
-
-  device:emit_event(capabilities.thermostatHeatingSetpoint.heatingSetpoint(
-    { value = temp_c, unit = "C" }
-  ))
 end
 
 local function handle_fan_oscillation_mode(driver, device, cmd)
@@ -285,7 +281,6 @@ local function handle_fan_oscillation_mode(driver, device, cmd)
 
   device:set_field("fan_mode", st_fan)
   send_ac_code(device, { swing = swing })
-  device:emit_event(capabilities.fanOscillationMode.fanOscillationMode(st_fan))
 end
 
 local function handle_fan_mode(driver, device, cmd)
@@ -293,7 +288,6 @@ local function handle_fan_mode(driver, device, cmd)
   local fan      = MODE_TO_FAN[fan_mode] or FAN_MID
   device:set_field("fan_mode_ac", fan)
   send_ac_code(device, { fan = fan })
-  device:emit_event(capabilities.fanMode.fanMode(fan_mode))
 end
 
 local function handle_refresh(driver, device)
--- a/drivers/Aqara/aqara-bath-heater/src/init.lua
+++ b/drivers/Aqara/aqara-bath-heater/src/init.lua
@@ -2,6 +2,7 @@
 -- Licensed under the Apache License, Version 2.0
 local capabilities    = require "st.capabilities"
 local ZigbeeDriver    = require "st.zigbee"
+local defaults        = require "st.zigbee.defaults"
 local cluster_base    = require "st.zigbee.cluster_base"
 local zcl_clusters    = require "st.zigbee.zcl.clusters"
 local data_types      = require "st.zigbee.data_types"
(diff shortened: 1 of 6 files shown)


drivers/Aqara/aqara-bath-heater/src/init.lua
```diff
@@ -0,0 +1,602 @@
...
+local SWING_OFF       = 0x1
+
+-- SmartThings fanMode capability values
+local SPEED           = {
+  LOW    = "low",
+  MEDIUM = "medium",
+  HIGH   = "high",
+}
+local MODE_TO_FAN     = { [SPEED.LOW] = FAN_LOW, [SPEED.MEDIUM] = FAN_MID, [SPEED.HIGH] = FAN_HIGH }
+local FAN_TO_MODE     = { [0] = SPEED.LOW, [1] = SPEED.MEDIUM, [2] = SPEED.HIGH }
```
aleclorimer:
```suggestion
local FAN_TO_MODE     = { [FAN_LOW] = SPEED.LOW, [FAN_MID] = SPEED.MEDIUM, [FAN_HIGH] = SPEED.HIGH }
```
nit: Keep them interlocked

drivers/Aqara/aqara-bath-heater/src/init.lua
```diff
@@ -0,0 +1,602 @@
...
+--   fan      : bits23-20  0=low    1=mid     2=high  3=auto
+--   swing    : bits17-16  0=swing  1=fixed
+--   setpoint : bits63-48  Celsius x 100
+local function send_ac_code(device, params)
+  local hi32 = 0xFFFFFFFF
+  local lo32 = 0xFFFFFFFF
+
+  if params.setpoint ~= nil then
+    local sp_raw = math.floor(clamp(params.setpoint, 16, 45) * 100) & 0xFFFF
+    hi32 = (sp_raw << 16) | (hi32 & 0xFFFF)
```
aleclorimer:
```suggestion
    hi32 = (sp_raw << 16) | 0xFFFF
```
Since `hi32` is unmodified from being set to `0xFFFFFFFF` and `(hi32 & 0xFFFF)` will always evaluate to `0xFFFF`

drivers/Aqara/aqara-bath-heater/src/init.lua
seojune79: I have made the requested changes.

drivers/Aqara/aqara-bath-heater/src/init.lua
seojune79: The unnecessary computations have been removed.

drivers/Aqara/aqara-bath-heater/src/init.lua
```diff
@@ -0,0 +1,602 @@
...
+  ))
+  device:emit_event(capabilities.fanMode.supportedFanModes(
+    SUPPORTED_SPEED_MODES, { visibility = { displayed = false } }
+  ))
+  device:emit_event(capabilities.thermostatHeatingSetpoint.heatingSetpointRange(
+    { value = { minimum = 16, maximum = 45, step = 1 }, unit = "C" }
+  ))
+  device:send(OnOff.attributes.OnOff:read(device))
+  device:send(Level.attributes.CurrentLevel:read(device))
+  device:send(ColorControl.attributes.ColorTemperatureMireds:read(device))
```
cjswedes: You should call `handle_refresh` or inject the refresh capability command to avoid duplicating this code.

drivers/Aqara/aqara-bath-heater/src/init.lua
```diff
@@ -0,0 +1,602 @@
...
+    [capabilities.switch.ID] = {
+      [capabilities.switch.commands.on.NAME]  = handle_switch_on,
+      [capabilities.switch.commands.off.NAME] = handle_switch_off,
+    },
+    [capabilities.switchLevel.ID] = {
+      [capabilities.switchLevel.commands.setLevel.NAME] = handle_switch_level,
+    },
+    [capabilities.colorTemperature.ID] = {
+      [capabilities.colorTemperature.commands.setColorTemperature.NAME] = handle_color_temperature,
+    },
```
cjswedes: These handlers are all essentially duplicating the default handlers, and should be removed. There are improvements in those handlers that include UI bugfixes for the color temperature capability, and native handler use for all attributes/commands.

drivers/Aqara/aqara-bath-heater/src/init.lua
```diff
@@ -0,0 +1,602 @@
...
+    attr = {
+      [OnOff.ID] = {
+        [OnOff.attributes.OnOff.ID] = on_off_attr_handler,
+      },
+      [Level.ID] = {
+        [Level.attributes.CurrentLevel.ID] = current_level_handler,
+      },
+      [ColorControl.ID] = {
+        [ColorControl.attributes.ColorTemperatureMireds.ID] = color_temp_handler,
+      },
```
cjswedes: These handlers are also covered by the defaults

drivers/Aqara/aqara-bath-heater/src/init.lua
```diff
@@ -0,0 +1,602 @@
...
+      [aqara.CLUSTER_ID] = {
+        [aqara.ATTR_AC_CODE] = ac_code_attr_handler,
```
cjswedes, line 518: I dont see any reporting configuration setup for this attribute. Does Aqara just automatically do the reporting?

drivers/Aqara/aqara-bath-heater/src/init.lua
```diff
@@ -0,0 +1,602 @@
...
+local function handle_fan_oscillation_mode(driver, device, cmd)
+  local st_fan = cmd.args.fanOscillationMode
+  local swing  = ST_FAN_TO_SWING[st_fan] or SWING_ON
+
+  device:set_field("fan_mode", st_fan)
+  send_ac_code(device, { swing = swing })
+  device:emit_event(capabilities.fanOscillationMode.fanOscillationMode(st_fan))
+end
+
+local function handle_fan_mode(driver, device, cmd)
```
cjswedes, line 286: Other handlers emit the event after the ac_code has been sent. Why not do that here? Should the other command handlers be following this pattern to ensure we are emitting events that match the reported device state (i.e. `handle_fan_oscillation_mode`, `handle_heating_setpoint`, `handler_thermostat_mode`) ?

cjswedes (commented):
I dont see a reason to not use the defaults, so I would prefer those are used.

drivers/Aqara/aqara-bath-heater/src/init.lua
seojune79: Replaced the duplicated code block by explicitly calling handle_refresh to adhere to the DRY principle.

drivers/Aqara/aqara-bath-heater/src/init.lua
seojune79: Added register_for_default_handlers to utilize the default configurations as suggested.

drivers/Aqara/aqara-bath-heater/src/init.lua
seojune79: Yes, the aqara.ATTR_AC_CODE attribute under the aqara.CLUSTER_ID cluster is automatically reported by the device whenever there is a change in the device status. No additional reporting configuration is required for this specific attribute.

drivers/Aqara/aqara-bath-heater/src/init.lua
seojune79: Good catch. I've adjusted the event emission timing to execute after the ac_code is sent, ensuring consistency across all command handlers as suggested.

nand-nor (approved):
This looks good to me with all the added fixes addressing others comments

drivers/Aqara/aqara-bath-heater/src/init.lua
cjswedes: What is the reason for emitting the even prior to receiving the updated state from the device? If there is a failure on the device, the platform will show the wrong state.

seojune79: That's a valid point, and I was already aware of this trade-off — I understood the risk of showing an inconsistent state if the device reports a failure.

The reason I added this optimistic emit was twofold:
1. Improved UX responsiveness — emitting early gives users immediate visual feedback, reducing perceived latency.
2. Consistency with thermostat mode changes — the same pattern was already being used there, so I followed the same approach for uniformity.

That said, your concern is well-founded, and I've updated the code to wait for the confirmed state from the device before emitting. One exception: I've kept the optimistic emit for thermostat mode changes, as that behavior is intentional to match the UX of the Aqara app. Please take note of that when reviewing.

drivers/Aqara/aqara-bath-heater/src/init.lua
```diff
@@ -0,0 +1,533 @@
...
+      aqara.MFG_CODE, data_types.Uint8, new[thermostatCtrl] and 1 or 0))
+  end
+end
+
+local aqara_bathroom_heater_driver_template = {
+  supported_capabilities = {
+    capabilities.switch,
+    capabilities.switchLevel,
+    capabilities.colorTemperature
+  },
```
cjswedes, line 495: These should be updated to include thermostatMode, thermostatHeatingSetpoint, fanOscilalationMode, and fanMode. Its not super critical since you are not relying on any default functionality, but it also affects the default driver switch behavior (which is a rarely used path)

cjswedes (approved):

drivers/Aqara/aqara-bath-heater/src/init.lua
seojune79: I removed them earlier since we weren't using them, but after reading your comment, I agree it's better to have them for the default behavior. I've updated the code to reflect this. Thanks!

PR #2944, opened 2026-05-04
Zwave Lock: Support SLGA Migration

hcarter-775:
# Description of Change

Note on structure: The `legacy-handlers` subdriver is nearly untouched from before- the only real difference is that the key-we logic was lifted out from there since it is so close in from to the "new", now default key-we overriden behavior.

The other primary thing that was removed from the now-`legacy-handlers` subdriver is the old `added` handler, for the following 2 reasons. 1. Generally, devices should not be added to this driver any longer, and 2. We now use the `added` main driver implementation to initially set the migrated keyword to `true`.

I find the fact that the entire migrated state depends on a keyword set only once in `added` to be a little precarious, but nothing stands out to me initially at this not working.

# Summary of Completed Tests


nickolas-deboom:
For some reason my comment is failing to save when i try to add an in line comment, but I was wondering if in addition to the lockCodes.migrated attribute should we also set a persisted flag to indicated the migrated state? Just since otherwise it would fallback to the legacy handling if the state cache is cleared

cjswedes:
> For some reason my comment is failing to save when i try to add an in line comment, but I was wondering if in addition to the lockCodes.migrated attribute should we also set a persisted flag to indicated the migrated state? Just since otherwise it would fallback to the legacy handling if the state cache is cleared

I agree with this: we should use the datastore. The state should always be available now that we are using the hubs state store to back the driver state_cache, but we have not always had this guarantee, and the pattern of expecting cached state to be present is not good IMO. Should we disable that for this driver, or make a change to how the hub handles state, it could have unintended consequences here.

hcarter-775:
@cjswedes @nickolas-deboom 
Agreed, makes sense to me.

If we're wondering about the state cache, I wonder how secure `lockUsers` and `lockCredentials` are as storage places for all this data. Should we be storing that information via the datastore as well?

drivers/SmartThings/zwave-lock/profiles/base-lock.yml
```diff
@@ -6,6 +6,10 @@ components:
     version: 1
   - id: lockCodes
     version: 1
+  - id: lockCredentials
```
nickolas-deboom, line 9: What if an existing device on one of these profiles that would not be migrated doesn't support these capabilities? I'm just wondering if new profiles should be created instead

drivers/SmartThings/zwave-lock/src/init.lua
```diff
@@ -159,22 +303,41 @@ local driver_template = {
...
+      [LockCredentials.commands.deleteAllCredentials.NAME] = delete_all_credentials_handler,
+    },
+    [capabilities.lockCodes.ID] = {
+      [capabilities.lockCodes.commands.migrate.NAME] = migrate,
+    },
   },
   zwave_handlers = {
     [cc.TIME] = {
-      [Time.GET] = time_get_handler -- used by DanaLock
+      [0x01] = time_get_handler -- used by DanaLock
```
nickolas-deboom, line 327: Just curious but why did Time.GET change to a hardcoded value?

drivers/SmartThings/zwave-lock/src/init.lua
```diff
@@ -2,140 +2,282 @@
...
-    device.thread:call_with_delay(
-      SCAN_CODES_CHECK_INTERVAL,
-      function(d)
-        periodic_codes_state_verification(driver, device)
-      end
-    )
-  end
-end
 
-local function populate_state_from_data(device)
```
cjswedes, line 45: This function was removed from the legacy init handler. I think it needs to be included in the legacy path so we keep functionality the same for devices that are not yet migrated.

drivers/SmartThings/zwave-lock/src/zwave_lock_utils.lua
```diff
@@ -0,0 +1,595 @@
...
+    if user_index ~= nil then
+      lock_utils.create_user(device, nil, user_type, user_index)
+    else
+      status = lock_utils.STATUS_RESOURCE_EXHAUSTED
+    end
+  end
+
+  if status == lock_utils.STATUS_SUCCESS then
+    device:set_field(lock_utils.ACTIVE_CREDENTIAL,
+      { userIndex = user_index, userType = user_type, credentialType = credential_type, credentialIndex = credential_index })
```
cjswedes, line 523: We should include the credentialName data here as well so it is not lost. A few other places would need to be updated as well (update credential, and add_credential for sure)

drivers/SmartThings/zwave-lock/src/init.lua
```diff
@@ -151,6 +293,8 @@ local driver_template = {
   supported_capabilities = {
     capabilities.lock,
```
cjswedes, line 294: The default lock handlers use codeId and codeName and do not include the userIndex, userName, or userType fields. I think for clarity, we should override these handlers in the base driver, and stop using the default handlers here; doing so will have implications on the legacy handlers too though, we would have to make sure that subdriver still includes the legacy event fields.

drivers/SmartThings/zwave-lock/src/zwave_lock_utils.lua
```diff
@@ -0,0 +1,595 @@
...
+lock_utils.get_available_user_index = function(device)
+  local max = device:get_latest_state("main", capabilities.lockUsers.ID,
+    capabilities.lockUsers.totalUsersSupported.NAME, DEFAULT_SUPPORTED_PIN_SLOTS)
+  local current_users = lock_utils.get_users(device)
+  local available_index = nil
+  local used_index = {}
+  for _, user in pairs(current_users) do
+    used_index[user.userIndex] = true
+  end
+  if current_users ~= {} then
```
cjswedes, line 151: This will always be true.

```suggestion
  if utils.table_size(current_users) == 0 then
```

drivers/SmartThings/zwave-lock/src/zwave_lock_utils.lua
```diff
@@ -0,0 +1,595 @@
...
+lock_utils.get_available_credential_index = function(device)
+  local max = device:get_latest_state("main", capabilities.lockCredentials.ID,
+    capabilities.lockCredentials.pinUsersSupported.NAME, DEFAULT_SUPPORTED_PIN_SLOTS)
+  local current_credentials = lock_utils.get_credentials(device)
+  local available_index = nil
+  local used_index = {}
+  for _, credential in pairs(current_credentials) do
+    used_index[credential.credentialIndex] = true
+  end
+  if current_credentials ~= {} then
```
cjswedes, line 201:
```suggestion
  if utils.table_size(current_credentials) == 0 then
```

drivers/SmartThings/zwave-lock/src/zwave_lock_utils.lua
```diff
@@ -0,0 +1,595 @@
...
+local INITIAL_INDEX = 1
+
+local lock_utils = {
+  -- Constants
+  ADD_CREDENTIAL = "addCredential",
+  ADD_USER = "addUser",
+  BUSY = "busy",
+  COMMAND_IN_PROGRESS = "commandInProgress",
+  CREDENTIAL_TYPE = "pin",
+  CHECKING_CODE = "checkingCode",
```
ctowns, line 15: Could you add a brief comment explaining what `checkingCode` is supposed to be used for? It can be a little confusing to understand without parsing through the code.

drivers/SmartThings/zwave-lock/src/zwave_lock_utils.lua
```diff
@@ -0,0 +1,595 @@
...
+  UPDATE_USER = "updateUser",
+  USER_INDEX = "userIndex",
+  USER_NAME = "userName",
+  USER_TYPE = "userType",
+  SLGA_MIGRATED = "slgaMigrated"
+}
+
+local DEFAULT_SUPPORTED_PIN_SLOTS = 8
+
+-- check if we are currently busy performing a task.
```
ctowns, line 41: Could we add more documentation on why the busy state is used/neccessary here? This is a source of complexity in both of these lock drivers and I think better documentation would help with reviews and future maintainability.

hcarter-775:
Closing once more. This has gone through a few iterations of PR, apologies.

Issue #2940, opened 2026-05-03
Matter Lock: modular reprofiling should run when the target profile changes, not only when optional capabilities change


ldeora:
## Summary

The `new-matter-lock` sub-driver can detect that a Matter lock supports the DoorLock `UNBOLT` feature and correctly emits:

```json
"supportedLockValues": ["locked", "unlocked", "unlatched", "not fully locked"],
"supportedLockCommands": ["lock", "unlock", "unlatch"]
```

However, the device may still remain on a profile that does not expose the `unlatch` UI, because `match_profile_modular()` only calls `try_update_metadata()` when the optional capability list changes.

This means a device can have the correct capability state, but the SmartThings app does not show the unlatch feature because the profile was not updated.

## Tested with a real device

This was tested with a real Nuki Smart Lock Ultra.

With the current behavior, the device reported `unlatch` support in component status, but the SmartThings app did not show the unlatch control because the device remained on the non-unlatch profile.

After applying the proposed reprofiling fix locally, the same real device successfully migrated to:

```text
lock-modular-embedded-unlatch
```

The SmartThings app then showed the unlatch feature as expected.

## Observed behavior

Tested with a Nuki Smart Lock Ultra using the Matter Lock driver.

The device initially used a non-unlatch fallback profile. The driver detected `UNBOLT` and emitted:

```json
"supportedLockCommands": ["lock", "unlock", "unlatch"]
```

but the app did not show the unlatch feature until the profile was actually changed to:

```text
lock-modular-embedded-unlatch
```

The key difference was that `try_update_metadata()` did not run when only the target profile name changed.

## Current code pattern

In `drivers/SmartThings/matter-lock/src/new-matter-lock/init.lua`, `match_profile_modular()` determines the correct modular profile name:

```lua
modular_profile_name = "lock-modular-embedded-unlatch"
```

when `DoorLock.types.Feature.UNBOLT` is supported.

But the metadata update is gated only by:

```lua
if lock_utils.optional_capabilities_list_changed(enabled_optional_component_capability_pairs, device.profile.components) then
  device:try_update_metadata({
    profile = modular_profile_name,
    optional_component_capabilities = enabled_optional_component_capability_pairs
  })
end
```

[...]

PR #2945, opened 2026-05-04
Matter Lock: Re-profile a Lock if it requires an unlatch embedded config

hcarter-775:
# Description of Change
Issue reported: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/issues/2940

In latest production release last week, a profiling bug related to a subset of Nuki devices supporting UNLATCH was introduced which led the embedded config for these devices to not be exposed, since the capabilities may not have explicitly changed, though their fingerprints did not initially include the unlatch embedded configs.

# Summary of Completed Tests
New unit test added to match the Nuki Ultra Lock profiling behavior.

### DIFF
--- a/drivers/SmartThings/matter-lock/src/test/test_matter_lock_modular.lua
+++ b/drivers/SmartThings/matter-lock/src/test/test_matter_lock_modular.lua
@@ -627,8 +627,8 @@ test.register_coroutine_test(
 local mock_nuki_smart_lock_ultra = test.mock_device.build_test_matter_device({
   profile = t_utils.get_profile_definition("lock-nocodes-notamper.yml"),
   manufacturer_info = {
-    vendor_id = 0x115f,
-    product_id = 0x2802,
+    vendor_id = 0x135D,
+    product_id = 0x00A1,
   },
   endpoints = {
     {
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-lock/src/test/test_matter_lock_modular.lua
```diff
@@ -623,4 +623,77 @@ test.register_coroutine_test(
   }
 )
 
+
+local mock_nuki_smart_lock_ultra = test.mock_device.build_test_matter_device({
+  profile = t_utils.get_profile_definition("lock-nocodes-notamper.yml"),
+  manufacturer_info = {
```
ldeora, line 629: This is for an Aqara lock.

```suggestion
  manufacturer_info = {
    vendor_id = 0x135D,
    product_id = 0x00A1,
  }
```

Because the mock device is called Nuki Smart Lock Ultra.
Yes, I'm pedantic.

ldeora:
Tested it with a real device (while not being at home! No risk, no fun!) and it works.

drivers/SmartThings/matter-lock/src/test/test_matter_lock_modular.lua
hcarter-775: Ah, good catch! Did a bad copy-paste job on my part (:

ctowns (approved):
Thank you @hcarter-775 and @ldeora!

PR #2950, opened 2026-05-05
Matter Lock: Add clear user failure test during defaultSchedule

hcarter-775:
# Description of Change
Add test for ClearUser case where it fails to remove the credentials during the defaultSchedule creation.

# Summary of Completed Tests
Test added


HunsupJung (approved):
The conclusion is somewhat absurd, but it seems unavoidable for now. it would be better to separate setCredential and setYearDaySchedule as soon as possible.

nickolas-deboom (approved):

PR #2956, opened 2026-05-11
Zigbee Lock: Support SLGA Migration

hcarter-775:
# Description of Change

Note on structure: The legacy-handlers subdriver keeps some subdrivers that have specific non-conformant behavior that relates specifically to lockCodes. This includes lock-without-codes, which no longer needs to exist in the modern implementation, as well as the yale-fingerprint-lock subdriver, which sets the lockCodes maxCodes state. Similarly, the yale subdriver has lockCodes-specific behavior as well.

However, upon analysis it was clear that the sub-sub-driver yale-bad-battery-reporter subdriver was 1. not unique to yale, and 2. was distinct from any lockCodes behavior, so it was pulled out into the now default subdrivers. Also, the samsungds subdriver had minimal differences between the old and new handling, so I condensed them.

The other primary thing that was removed from the now-legacy-handlers subdriver is the old added handler, for the following 2 reasons. 1. Generally, devices should not be added to this driver any longer, and 2. We now use the added main driver implementation to initially set the migrated keyword to true.

I find the fact that the entire migrated state depends on a keyword set only once in added to be a little precarious, but nothing stands out to me initially at this not working.

# Summary of Completed Tests
Migration process, plus lockUsers/lockCredentials functionality, has been tested on-device with a kwikset lock.

Various Unit tests added.


drivers/SmartThings/zigbee-lock/src/lock_utils/tables.lua
```diff
@@ -0,0 +1,251 @@
...
+  persist_table(device, def, {})
+  return COMMAND_RESULT.SUCCESS
+end
+
+-- Restore capability state from the persistent field store.
+-- Called during init to re-emit table events if the capability state cache
+-- has been wiped (e.g. after a hub reboot). Only emits for tables that have
+-- persisted data, are in a nil state, and whose capability is supported by the device.
+function table_utils.restore_from_persistent_store(device)
+  for _, internal in ipairs(DEFS) do
```
tpmanley: I think this is supposed to be `pairs` since `DEFS` is a dictionary table

drivers/SmartThings/zigbee-lock/src/lock_utils/tables.lua
```diff
@@ -0,0 +1,251 @@
...
+  local predicate = function(entry) return entry[def.match_key] == matcher end
+
+  local removed = nil
+  for i, entry in ipairs(t) do
+    if predicate(entry) then
+      removed = table.remove(t, i)
+      break
+    end
+  end
+  device:emit_event(def.attribute(t, {visibility = {displayed = false}}))
```
tpmanley: Perhaps this should be gated on whether or not an entry was removed from the table?

drivers/SmartThings/zigbee-lock/src/lock_utils/utils.lua
```diff
@@ -0,0 +1,75 @@
...
+
+  -- If we are missing the cached values for these attributes, read them so we can properly manage them locally
+  if (device:get_latest_state("main", capabilities.lockCredentials.ID, capabilities.lockCredentials.maxPinCodeLen.NAME) == nil) then
+    device:send(clusters.DoorLock.attributes.MaxPINCodeLength:read(device))
+  end
+  if (device:get_latest_state("main", capabilities.lockCredentials.ID, capabilities.lockCredentials.minPinCodeLen.NAME) == nil) then
+    device:send(clusters.DoorLock.attributes.MinPINCodeLength:read(device))
+  end
+  if (device:get_latest_state("main", capabilities.lockCredentials.ID, capabilities.lockCredentials.pinUsersSupported.NAME) == nil) or
+     (device:get_latest_state("main", capabilities.lockUsers.ID, capabilities.lockUsers.totalUsersSupported.NAME) == 0) then
```
tpmanley: Should this be checking `== nil` instead of `== 0`?

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
```diff
@@ -0,0 +1,209 @@
...
+    device:send(LockCluster.server.commands.ClearAllPINCodes(device))
+  end
+end
+
+
+-- [[ REFRESH CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.refresh(driver, device, cmd)
+  device:refresh()
+  device:send(LockCluster.attributes.LockState:read(device))
```
tpmanley, line 225: Is the explicit read of LockState needed since I think that'd be covered by refresh?

drivers/SmartThings/zigbee-lock/src/legacy-handlers/can_handle.lua
```diff
@@ -0,0 +1,12 @@
+-- Copyright 2026 SmartThings
```
tpmanley:
```suggestion
-- Copyright 2026 SmartThings, Inc.
```

drivers/SmartThings/zigbee-lock/src/legacy-handlers/lock-without-codes/can_handle.lua
```diff
@@ -0,0 +1,13 @@
...
+
+return function(opts, driver, device, cmd)
+    local fingerprints = require("legacy-handlers.lock-without-codes.fingerprints")
+    for _, fingerprint in ipairs(fingerprints) do
+        if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
+            local subdriver = require("legacy-handlers.lock-without-codes")
+            return true, subdriver
+        end
+    end
+    return false
```
tpmanley:
```suggestion
  local fingerprints = require("legacy-handlers.lock-without-codes.fingerprints")
  for _, fingerprint in ipairs(fingerprints) do
      if device:get_manufacturer() == fingerprint.mfr and device:get_model() == fingerprint.model then
          local subdriver = require("legacy-handlers.lock-without-codes")
          return true, subdriver
      end
  end
  return false
```

drivers/SmartThings/zigbee-lock/src/lock_handlers/zigbee_responses.lua
```diff
@@ -0,0 +1,395 @@
...
+  local OP_EVENT_SOURCE_CAPABILITY_MAP = {
+    [OpEventSource.KEYPAD] = "keypad",
+    [OpEventSource.RF]     = "command",
+    [OpEventSource.MANUAL] = "manual",
+    [OpEventSource.RFID]   = "rfid",
+    -- These last two sources are not found in the spec, but they were in the legacy driver
+    [4] = "fingerprint",
+    [5] = "bluetooth",
+  }
+  if (op_event_source ~= OpEventSource.KEYPAD and (
```
tpmanley: The legacy handler doesn't have the parenthesis after the `and` so the logic is slightly changed. Looking at table 7-29 in ZCL8 it looks like AutoLock, ScheduleLock and ScheduleUnlock should only come from source Manual so it seems like if we were going to do a source check it would be for Manual instead of not Keypad. It's also possible the legacy code was written that way for a reason (like to handle a specific lock behavior). I'm not sure if there would be any harm in just ignoring the source altogether and treating any AutoLock, ScheduleLock and ScheduleUnlock events as "auto". That's slightly more permissive than the legacy code.

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
hcarter-775: Not sure, I'd have to look into it. This explicit read is in the current refresh function

drivers/SmartThings/zigbee-lock/src/lock_utils/tables.lua
hcarter-775: good catch, accidentally screwed this up making some changes last night!

drivers/SmartThings/zigbee-lock/src/lock_utils/tables.lua
hcarter-775: agreed

drivers/SmartThings/zigbee-lock/src/lock_handlers/zigbee_responses.lua
hcarter-775: makes sense, works for me.

hcarter-775: looks like a samsungsds test failed when I made that alteration.

hcarter-775: It appears the fingerprint/bluetooth handling is also specific to the Samsung SDS, so I'm going to move the handler as-is to that subdriver, and will make the main driver implementation more generic/permissive.

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
hcarter-775: I think this is because the LockState handling was added at a later date, so it wasn't in the default immediately.

tpmanley: So we can remove it now?

drivers/SmartThings/zigbee-lock/src/lock_handlers/zigbee_responses.lua
```diff
@@ -0,0 +1,391 @@
...
+function ZigbeeHandlers.programming_event_notification(driver, device, zb_rx)
+  -- zb response values
+  local user_id = tonumber(zb_rx.body.zcl_body.user_id.value)
+  local event_code = tonumber(zb_rx.body.zcl_body.program_event_code.value)
+
+  if device:get_manufacturer() == "ASSA ABLOY iRevo" or device:get_manufacturer() == "Yale" then
+    if user_id >= 256 then -- Index is incorrectly written on these devices. Attempt to shift it to get an actual value
+      user_id = user_id >> 8
+    end
+  end
```
cjswedes, line 205: I dont see where this was being done in the legacy behavior. Usually it would be a subdriver too, but I mainly just want to understand why its being done since I do not see where it was being done previously

drivers/SmartThings/zigbee-lock/src/lock_handlers/zigbee_responses.lua
```diff
@@ -0,0 +1,391 @@
...
+        credentialType  = credential_args_in_use.credentialType,
+        credentialName  = credential_args_in_use.credentialName, -- optional
+      })
+    elseif event_code == ProgramEventCode.PIN_CODE_CHANGED and command_in_progress == consts.LOCK_CREDENTIALS.UPDATE then
+      result_status = consts.COMMAND_RESULT.SUCCESS
+    elseif event_code == ProgramEventCode.PIN_CODE_DELETED and command_in_progress == consts.LOCK_CREDENTIALS.DELETE then
+      result_status = tables.delete_entry(device, "credentials", user_id)
+    end
+    if command_in_progress then
+      lock_utils.emit_command_result(device,
```
cjswedes, line 259: This is only emitting a lockCredentials command result, when this could happen as a part of lockUsers command (i.e. you remove a lock user, which results in the pin code associated with that users credential being removed, and then this event is sent from the device; the user deleted the user, but receives a credential command result).

The lockUsers table will not have its entry deleted. I think this failsafe path in general needs to consider the lockUsers commands as well as the lockCredentials.

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
```diff
@@ -0,0 +1,209 @@
...
+-- [[ LOCK CREDENTIALS CAPABILITY COMMANDS ]] --
+
+function CapabilityHandlers.add_credential(driver, device, command)
+  if lock_utils.is_device_busy(device) then
+    lock_utils.emit_command_result(device, capabilities.lockCredentials, consts.LOCK_CREDENTIALS.ADD, consts.COMMAND_RESULT.BUSY)
+    return
+  end
+
+  -- A userIndex of 0 means "auto-assign the next available slot"
+  local user_index = command.args.userIndex == 0 and tables.next_index(device, "users") or command.args.userIndex
```
cjswedes, line 131:
```suggestion
  local user_index = command.args.userIndex == 0 and tables.next_index(device, "credentials") or command.args.userIndex
```
This is a defensive suggestion in case the credential and user tables ever get out of sync.

drivers/SmartThings/zigbee-lock/src/init.lua
```diff
@@ -1,450 +1,149 @@
...
-    device:emit_event_for_endpoint(zb_rx.address_header.src_endpoint.value, event)
-  end
-end
+  device:send(device_management.build_bind_request(device, clusters.Alarms.ID, self.environment_info.hub_zigbee_eui))
+  device:send(clusters.Alarms.attributes.AlarmCount:configure_reporting(device, 0, 21600, 0))
 
-local function lock(driver, device, command)
-  device:send_to_component(command.component, LockCluster.server.commands.LockDoor(device))
+  if device:supports_capability(capabilities.lockCredentials) then
+    device.thread:call_with_delay(2, function(d) lock_utils.sync_device_state(device) end)
```
cjswedes: Since the device init is also doing a delayed (15s) sync_device_state, this seems unnecessary.

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
cjswedes, line 1: Is it safe to rely on receiving an `addCredential`/`addUser` or a `deleteCredential`/`deleteUser` in pairs from the app in order to maintain the 1:1 credential to user mapping? If we receive a `deleteUser` command, we will delete the associated credential on the device and emit the corresponding `commandResult`, but the opposite is not true; is the split of SoT for the credentials being in the driver and the users in the app intentional?

drivers/SmartThings/zigbee-lock/src/legacy-handlers/init.lua
```diff
@@ -0,0 +1,490 @@
...
+    if user_id then
+      table.insert(users, { userIndex = user_id, userType = "guest", userName = code_name })
+      table.insert(credentials, { userIndex = user_id, credentialIndex = user_id, credentialType = post_migration_consts.CRED_TYPE_PIN })
+    end
+  end
+  -- manually ensure user/cred state is persisted, then emit events to populate the new capabilities with the migrated data.
+  device:set_field("persistedUsers", users, { persist = true })
+  device:set_field("persistedCredentials", credentials, { persist = true })
+  device:emit_event(LockUsers.users(users, { visibility = { displayed = false } }))
+  device:emit_event(LockCredentials.credentials(credentials, { visibility = { displayed = false } }))
```
cjswedes, line 427: We only ever emit a `lockCredentials.credentials`/`lockUsers.users` attribute event during migration, and rely on the `commandResult` attribute events to make sure the platform has the correct credentials capability state. It seems a bit brittle to rely on the device responses and `commandResult` event emissions being handled properly to maintain the `lockUser` and `lockCredential` capability state; perhaps on refresh/init we should be emitting the `lockCredential.credentials` and `lockUser.users` attributes based on the state we have in the driver.

cjswedes (commented):
I am a still a little bit confused on how credentials and users will work in the capabilities for zigbee locks because it seems fragile. The zigbee user_index maps directly to the smart things `credentialIndex`, and there should be a 1:1 mapping of `credentialIndex` to `userIndex` in the `lockCredentials` and `lockUsers` capabilities. Is that right? It seems like an artificial constraint, and somewhat unexpected: why even have the lockUsers capability on zigbee locks. It makes sense to me for the app to allow 1 user to have multiple credentials, even for zigbee locks; the users are just something that the ST platform maintains, and the lock is only concerned with credentials.

I really need to play with this in the app to see what the desired user experience is in the new world, but I am not sure I fully understand what the intended user experience is, and I worry we are implementing something that is not what stakeholders want.

drivers/SmartThings/zigbee-lock/src/lock_handlers/zigbee_responses.lua
hcarter-775: It was in the yale subdriver. Initally that subdriver had a lot of overrides that were 1-1 with this handling, except for this, so I just added this here. I could (and have thought about) breaking this function into a subdriver.

drivers/SmartThings/zigbee-lock/src/lock_handlers/zigbee_responses.lua
hcarter-775: Agreed, I updated this since you last reviewed I think.

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
hcarter-775: I think the system should be resilient to these getting out of sync

drivers/SmartThings/zigbee-lock/src/init.lua
hcarter-775: Also updated since you last reviewed haha, I no longer have init do this since it is unnecessary.

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
hcarter-775: A successful addCredential leads to an updateUser command, and yes, we can rely on the app sending deleteUser in response to deleteCredential.

drivers/SmartThings/zigbee-lock/src/legacy-handlers/init.lua
hcarter-775: That is what happens currently, see the `table_utils.restore_from_persistent_store(device)` function.

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
hcarter-775: One reason there is a distinction between these two is that a credential can be associated with a user we do not want to remove.

drivers/SmartThings/zigbee-lock/src/lock_handlers/zigbee_responses.lua
hcarter-775: Sorry, let me clarify. It was in the "new" yale subdriver [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/2612d80660ee88523b837cb7fd1e0af9b8c9aa03/drivers/SmartThings/zigbee-lock/src/using-new-capabilities/yale/init.lua#L111). I do not know why it was added, but I initially aimed to re-create the handling more or less 1-1

hcarter-775:
Hey @aleclorimer @cjswedes, looks like one of the tests passing now is failing in dev, and as far as I can tell it really shouldn't be. Would one of you mind taking a look?

Specifically, the check `device:supports_capability(capabilities.lockCodes)` is being evaluated as true after a generate_info_changed event to a profile without that capability.

It's not really an important test, the AI generated it and it's not a path that can actually happen today, but like it technically could

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
hcarter-775: Well, also I actually made an update here since you last reviewed I think, as well. I know I keep saying that, so clearly we were looking at the same weird spots (:

Let me know what you think of this system, which I think tracks state a little better

hcarter-775:
> I am a still a little bit confused on how credentials and users will work in the capabilities for zigbee locks because it seems fragile. The zigbee user_index maps directly to the smart things credentialIndex, and there should be a 1:1 mapping of credentialIndex to userIndex in the lockCredentials and lockUsers capabilities. Is that right? It seems like an artificial constraint, and somewhat unexpected: why even have the lockUsers capability on zigbee locks. It makes sense to me for the app to allow 1 user to have multiple credentials, even for zigbee locks; the users are just something that the ST platform maintains, and the lock is only concerned with credentials.

@cjswedes I agree that much of this Users/Credentials state seems a little fragile. I am not 100% sure how to make it less fragile than it already was/is (in Matter Lock for example and in the initial versions of this work), so have mostly endeavored to make the system as clear as possible.

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
hcarter-775: I'm really not sure. I might lean towards just leaving it as-is

cjswedes:
> Hey @aleclorimer @cjswedes, looks like one of the tests passing now is failing in dev, and as far as I can tell it really shouldn't be. Would one of you mind taking a look?
> 
> Specifically, the check `device:supports_capability(capabilities.lockCodes)` is being evaluated as true after a generate_info_changed event to a profile without that capability.
> 
> It's not really an important test, the AI generated it and it's not a path that can actually happen today, but like it technically could

The integration test framework, does not actually mock out profile changes on the mock device objects, it just validates that the event was there. I think maybe delete the test since it is not a path that can really happen

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
cjswedes: I do not see a reporting configuration for the LockState attribute in the default handler for lock, so we should leave this read in because the default refresh reads the configured attributes.

drivers/SmartThings/zigbee-lock/src/lock_handlers/capabilities.lua
cjswedes: Looks good to me.

cjswedes (approved):
I have given this a light test on my home, and an in depth review.

For the test, I started with the old lock driver created some codes, and then did the migration. The only unique thing i did was have non-consecutive codes stored on the device prior to migration, and that was handled well. Subsequent adds re-filled the unused slot.

tpmanley (approved):

ctowns (approved):

PR #2971, opened 2026-05-14
add firstled-io_M4S4BAC

thinkaName:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests

### DIFF
--- a/drivers/SmartThings/zigbee-switch/profiles/switch-button-light-restore-wireless.yml
+++ b/drivers/SmartThings/zigbee-switch/profiles/switch-button-light-restore-wireless.yml
@@ -1,0 +1,39 @@
+name: switch-button-light-restore-wireless
+components:
+- id: main
+  capabilities:
+  - id: switch
+    version: 1
+  - id: button
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+    - name: RemoteController
+preferences:
+  - title: "背光灯(backlight/백라이트)"
+    name: backlight
+    description: "背光灯(backlight/백라이트)"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        0: "关闭(off/닫다)"
+        1: "打开(on/열다)"
+        2: "人体接近感应(proximity detection/근접 감지)"
+      default: 2
+  - title: "开关上电状态(relay powerOn state)"
+    name: powerOnStatus
+    description: "开关上电状态(relay powerOn state/릴레이 초기 동작 상태)"
+    required: false
+    preferenceType: enumeration
+    definition:
+      options:
+        0: "关闭(off/닫다)"
+        1: "打开(on/열다)"
+        2: "恢复记忆状态(restore/복원)"
+      default: 2
+  - preferenceId: stse.changeToWirelessSwitch
+    explicit: true
--- a/drivers/SmartThings/zigbee-switch/profiles/switch-button-wireless.yml
+++ b/drivers/SmartThings/zigbee-switch/profiles/switch-button-wireless.yml
@@ -1,0 +1,17 @@
+name: switch-button-wireless
+components:
+- id: main
+  capabilities:
+  - id: switch
+    version: 1
+  - id: button
+    version: 1
+  - id: firmwareUpdate
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+    - name: RemoteController
+preferences:
+  - preferenceId: stse.changeToWirelessSwitch
+    explicit: true
--- a/drivers/SmartThings/zigbee-switch/src/firstled-io/fingerprints.lua
+++ b/drivers/SmartThings/zigbee-switch/src/firstled-io/fingerprints.lua
@@ -1,0 +1,9 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+--The number of children determines the number of sub-devices to be created. Each sub-device has the capability to switch between a switch and a button.
+--The number of buttons determines how many buttons devices will be created.
+--The driver supports a series of device combinations, such as 4+4, 3+3, 2+2, 4+0, etc., of switch and button type products.
+return {
+  { mfr = "FIRSTLED", model = "M4S4BAC", children = 4, buttons = 4, child_profile = "switch-wireless" }
+}
--- a/drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
+++ b/drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
@@ -1,0 +1,209 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local device_lib = require "st.device"
+local capabilities = require "st.capabilities"
+local cluster_base = require "st.zigbee.cluster_base"
+local data_types = require "st.zigbee.data_types"
+local zcl_clusters = require "st.zigbee.zcl.clusters"
+
+local Scenes = zcl_clusters.Scenes
+local PRIVATE_CLUSTER_ID = 0xFCCA
+local MFG_CODE = 0x1235
+local FINGERPRINTS = require("firstled-io.fingerprints")
+
+--stse.changeToWirelessSwitch
+--switch mode:The local switch and the app can control the relay.The button capability is not working.
+--wirelessSwitch mode:The local switch does not control the relay. Once triggered, it will report to the system as "RecallScene",emit_event button.pushed. The relay can be controlled via the app.
+local preference_map = {
+  ["backlight"] = {
+    cluster_id = PRIVATE_CLUSTER_ID,
+    attribute_id = 0x0000,
+    mfg_code = MFG_CODE,
+    data_type = data_types.Uint8,
+  },
+  ["powerOnStatus"] = {
+    cluster_id = PRIVATE_CLUSTER_ID,
+    attribute_id = 0x0001,
+    mfg_code = MFG_CODE,
+    data_type = data_types.Uint8,
+  },
+  ["stse.changeToWirelessSwitch"] = {
+    cluster_id = PRIVATE_CLUSTER_ID,
+    attribute_id = 0x0002,
+    mfg_code = MFG_CODE,
+    data_type = data_types.Boolean
+  }
+}
+
+local function is_parent_device(device)
+    local parent = device:get_parent_device()
+    return parent == nil
+end
+
+--If it is in the switch mode, only the switch will be displayed. If it is in the wirelessSwitch mode, both the switch and the button will be displayed.
+local function toggle_button_visibility(device, show)
+  local is_parent = is_parent_device(device)
+  if is_parent then
+    if show then
+      device:try_update_metadata({profile = "switch-button-light-restore-wireless"})
+    else
+      device:try_update_metadata({profile = "switch-light-restore-wireless"})
+    end
+  else
+    if show then
+      device:try_update_metadata({profile = "switch-button-wireless"})
+    else
+      device:try_update_metadata({profile = "switch-wireless"})
+    end
+  end
+end
+
+--When stse.changeToWirelessSwitch switching to the wirelessSwitch mode to listen for profile changes
+local function listen_profile_button_transition(device, args)
+  local current_has = device:supports_capability(capabilities.button, "main")
+
+  local old_has = false
+  if args and args.old_st_store and args.old_st_store.profile then
+    local old_main = args.old_st_store.profile.components.main
+    if old_main and old_main.capabilities then
+      old_has = old_main.capabilities["button"] ~= nil
+    end
+  end
+  --Capabilities button from non-existent to existing
+  if not old_has and current_has then
+    device:emit_event(capabilities.button.numberOfButtons({value = 1}, {visibility = {displayed = false}}))
+    device:emit_event(capabilities.button.supportedButtonValues({"pushed"}, {visibility = {displayed = false}}))
+  end
+end
+
+local function device_info_changed(driver, device, event, args)
+  listen_profile_button_transition(device, args)
+
+  local preferences = device.preferences
+  local old_preferences = args.old_st_store.preferences
+  if preferences ~= nil then
+    for id, attr in pairs(preference_map) do
+      local old_value = old_preferences[id]
+      local value = preferences[id]
+      if value ~= nil and value ~= old_value then
+        if attr.data_type == data_types.Uint8 then
+          value = tonumber(value)
+        end
+        device:send(cluster_base.write_manufacturer_specific_attribute(device, attr.cluster_id, attr.attribute_id,
+          attr.mfg_code, attr.data_type, value))
+        --Switch to the corresponding profile based on stse.changeToWirelessSwitch
+        if id == "stse.changeToWirelessSwitch" then
+          toggle_button_visibility(device, value)
+        end
+      end
+    end
+  end
+end
+
+local function get_children_amount(device)
+  for _, fingerprint in ipairs(FINGERPRINTS) do
+    if device:get_model() == fingerprint.model then
+      return fingerprint.children
+    end
+  end
+end
+
+local function get_button_amount(device)
+  for _, fingerprint in ipairs(FINGERPRINTS) do
+    if device:get_model() == fingerprint.model then
+      return fingerprint.buttons
+    end
+  end
+end
+
+local function get_child_profile_name(device)
+  for _, fingerprint in ipairs(FINGERPRINTS) do
+    if device:get_model() == fingerprint.model then
+      return fingerprint.child_profile
+    end
+  end
+end
+
+local function find_child(parent, ep_id)
+  return parent:get_child_by_parent_assigned_key(string.format("%02X", ep_id))
+end
+
+--Create composite switches such as 4+4, 1+1, 4+0, 3+0 children+button
+local function device_added(driver, device)
+  -- Create the corresponding number of child devices based on the value of fingerprint.children
+  if device.network_type == device_lib.NETWORK_TYPE_ZIGBEE then
+    local children_amount = get_children_amount(device)
+    if children_amount >= 2 then
+      for i = 2, children_amount, 1 do
+        if find_child(device, i) == nil then
+          local name = string.format("%s%d", string.sub(device.label, 0, -2), i)
+          local child_profile = get_child_profile_name(device)
+          local metadata = {
+            type = "EDGE_CHILD",
+            label = name,
+            profile = child_profile,
+            parent_device_id = device.id,
+            parent_assigned_child_key = string.format("%02X", i),
+            vendor_provided_label = name
+          }
+          driver:try_create_device(metadata)
+        end
+      end
+    end
+  -- Create the corresponding number of button devices based on the value of fingerprint.buttons
+    local button_amount = get_button_amount(device)
+    if button_amount >= 1 then
+      for i = children_amount + 1, children_amount + button_amount, 1 do
+        if find_child(device, i) == nil then
+          local name = string.format("%s%d", string.sub(device.label, 0, -2), i)
+          local metadata = {
+            type = "EDGE_CHILD",
+            label = name,
+            profile = "button",
+            parent_device_id = device.id,
+            parent_assigned_child_key = string.format("%02X", i),
+            vendor_provided_label = name,
+          }
+          driver:try_create_device(metadata)
+        end
+      end
+    end
+  elseif device.network_type == "DEVICE_EDGE_CHILD" then
+    device:emit_event(capabilities.button.numberOfButtons({ value = 1 },
+      { visibility = { displayed = false } }))
+    device:emit_event(capabilities.button.supportedButtonValues({ "pushed" },
+      { visibility = { displayed = false } }))
+  end
+end
+
+local function scenes_cluster_handler(driver, device, zb_rx)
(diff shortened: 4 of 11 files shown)


drivers/SmartThings/zigbee-switch/src/test/test_firstled_switch.lua
```diff
@@ -0,0 +1,729 @@
...
+
+local mock_base_device = test.mock_device.build_test_zigbee_device(
+  {
+    profile = parent_switch_profile_def,
+    fingerprinted_endpoint_id = 0x01,
+    zigbee_endpoints = {
+      [1] = {
+        id = 1,
+        manufacturer = "FIRSTLED",
+        model = "MU4S4BAC",
```
cjswedes:
```suggestion
        model = "M4S4BAC",
```
This is the model the fingerprint and can_handle use. Which is correct?

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,164 @@
...
+local data_types = require "st.zigbee.data_types"
+local zcl_clusters = require "st.zigbee.zcl.clusters"
+
+local Scenes = zcl_clusters.Scenes
+local PRIVATE_CLUSTER_ID = 0xFCCA
+local PRIVATE_ATTRIBUTE_ID = 0x0000
+local MFG_CODE = 0x1235
+local FINGERPRINTS = require("firstled-io.fingerprints")
+
+local preference_map = {
```
cjswedes, line 18: The `switch-button-light-restore-wireless` profile also defines `backlight` and `powerOnStatus`, what is the purpose of those preferences if the driver doesnt do anything with them?

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,164 @@
...
+      for i = children_amount+1,children_amount + button_amount, 1 do
+        if find_child(device, i) == nil then
+          local name = string.format("%s%d", string.sub(device.label, 0, -2), i)
+          local metadata = {
+            type = "EDGE_CHILD",
+            label = name,
+            profile = "button",
+            parent_device_id = device.id,
+            parent_assigned_child_key = string.format("%02X", i),
+            vendor_provided_label = name .. i,
```
cjswedes:
```suggestion
            vendor_provided_label = name,
```

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,164 @@
...
+        end
+      end
+    end
+
+    -- for wireless button
+    device:emit_event(capabilities.button.numberOfButtons({ value = children_amount },
+      { visibility = { displayed = false } }))
+
+    device:send(cluster_base.write_manufacturer_specific_attribute(device,
+      PRIVATE_CLUSTER_ID, PRIVATE_ATTRIBUTE_ID, MFG_CODE, data_types.Uint8, 0x01)) -- private
```
cjswedes: I dont really know what this attribute write does, but is it something that should be done in device init? The difference being it would be written anytime the hub restarts or the driver is updated

drivers/SmartThings/zigbee-switch/src/test/test_firstled_switch.lua
cjswedes: This means the test isnt actually testing the subdriver since can_handle will return false.

cjswedes (commented):
Test failures will need to be resolved prior to merging.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
cjswedes: Looks like this was moved to infoChanged to handle preference changes. It seems like it might still be needed in the added handler so that the initial preference default value is written.

thinkaName: When the device is entering the pairing process, it will restore the data to the default values, and both ends will have the same default values.

nand-nor:
Can code coverage be improved in `firstled-io/init.lua`? Device init tests, preference changes, scenes cluster handling, profile switching, that kind of thing

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,164 @@
...
+          }
+          driver:try_create_device(metadata)
+        end
+      end
+    end
+
+  -- Create Button if necessary
+    local button_amount = get_button_amount(device)
+    if button_amount >= 1 then
+      for i = children_amount+1,children_amount + button_amount, 1 do
```
nand-nor:
```suggestion
      for i = children_amount + 1, children_amount + button_amount, 1 do
```

thinkaName:
> Can code coverage be improved in `firstled-io/init.lua`? Device init tests, preference changes, scenes cluster handling, profile switching, that kind of thing

The maximum improvement can only reach 88%. What suggestions do you have?

drivers/SmartThings/zigbee-switch/src/test/test_wallhero_switch.lua
cjswedes, line 1: Why is this test file modified? It seems like this device is unrelated to the changes.

cjswedes:
> The maximum improvement can only reach 88%. What suggestions do you have?

You  can run coverage locally and get the report yourself. When I do it I see that none of the child button behavior is tested.

To run it locally first run your test file with coveraged enabled (note you need luacov [installed](https://lunarmodules.github.io/luacov/))
```bash
## run test with coverage enabled from zigbee-switch/src
lua -lluacov test/test_firstled_switch.lua
## create the report from the stats output
luacov -c ./luacov.stats.out --config ~/SmartThingsEdgeDrivers/tools/config.luacov
```
This will create a `luacov.report.out` with counts next to each line in each file that was required during the lua script. Here is what I was seeing relating to the `firstled-io/init.lua` file:
```
       local function get_child_profile_name(device)
*****0   for _, fingerprint in ipairs(FINGERPRINTS) do
*****0     if device:get_model() == fingerprint.model then
*****0       return fingerprint.child_profile
           end
         end
       end

       local function find_child(parent, ep_id)
    23   return parent:get_child_by_parent_assigned_key(string.format("%02X", ep_id))
       end

         -- Only create children for the actual Zigbee device and not the children
     8   if device.network_type == device_lib.NETWORK_TYPE_ZIGBEE then
     1     local children_amount = get_children_amount(device)
     1     if children_amount >= 2 then
     4       for i = 2, children_amount, 1 do
     3         if find_child(device, i) == nil then
*****0           local name = string.format("%s%d", string.sub(device.label, 0, -2), i)
*****0           local child_profile = get_child_profile_name(device)
*****0           local metadata = {
                   type = "EDGE_CHILD",
                   label = name,
                   profile = child_profile,
                   parent_device_id = device.id,
                   parent_assigned_child_key = string.format("%02X", i),
*****0             vendor_provided_label = name
                 }
*****0           driver:try_create_device(metadata)
               end
             end
           end

         -- Create Button if necessary
     1     local button_amount = get_button_amount(device)
     1     if button_amount >= 1 then
     5       for i = children_amount + 1, children_amount + button_amount, 1 do
     4         if find_child(device, i) == nil then
*****0           local name = string.format("%s%d", string.sub(device.label, 0, -2), i)
*****0           local metadata = {
                   type = "EDGE_CHILD",
                   label = name,
                   profile = "button",
                   parent_device_id = device.id,
                   parent_assigned_child_key = string.format("%02X", i),
                   vendor_provided_label = name,
                 }
*****0           driver:try_create_device(metadata)
               end
             end
           end
```

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,164 @@
...
+          local name = string.format("%s%d", string.sub(device.label, 0, -2), i)
+          local metadata = {
+            type = "EDGE_CHILD",
+            label = name,
+            profile = "button",
+            parent_device_id = device.id,
+            parent_assigned_child_key = string.format("%02X", i),
+            vendor_provided_label = name,
+          }
+          driver:try_create_device(metadata)
```
cjswedes, line 168: please add tests that test this part of device added.

cbaumler:
Hi @thinkaName , it looks like this PR branch is out of date with main. Please update it with the latest main and push the result. Then we can re-review.

drivers/SmartThings/zigbee-switch/src/firstled-io/fingerprints.lua
```diff
@@ -0,0 +1,6 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+return {
+  { mfr = "FIRSTLED", model = "M4S4BAC",  buttons = 4 ,  children = 4, child_profile = "switch-button-wireless" }
```
cjswedes: How are the number of buttons and children used? Please describe the way the device is intended to be integrated, since it is not clear from the code in its current state. My assumption is that there is a single zigbee device with 4 buttons, and you are trying to have a child device for each button.

What is the goal of using child devices to represent individual buttons? The approach we take on the platform is to use a multi component profile that has a component for each button.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,164 @@
...
+      return fingerprint.child_profile
+    end
+  end
+end
+
+local function find_child(parent, ep_id)
+  return parent:get_child_by_parent_assigned_key(string.format("%02X", ep_id))
+end
+
+local function device_added(driver, device)
```
cjswedes, line 133: When a new device is joined what is the expectation for children devices?

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,164 @@
...
+            vendor_provided_label = name
+          }
+          driver:try_create_device(metadata)
+        end
+      end
+    end
+
+  -- Create Button if necessary
+    local button_amount = get_button_amount(device)
+    if button_amount >= 1 then
```
cjswedes, line 156: This is always true, or is a nil error.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,164 @@
...
+
+local function find_child(parent, ep_id)
+  return parent:get_child_by_parent_assigned_key(string.format("%02X", ep_id))
+end
+
+local function device_added(driver, device)
+  -- Only create children for the actual Zigbee device and not the children
+  if device.network_type == device_lib.NETWORK_TYPE_ZIGBEE then
+    local children_amount = get_children_amount(device)
+    if children_amount >= 2 then
```
cjswedes, line 137: This is always true or is a nil error.

cjswedes (commented):
It seems like there was a lot of stuff added since the last reviews wrt child devices, but we cant tell because you are obfuscating the commit history by always force pushing; this makes it hard to review this PR and others you have submitted. Please keep incremental changes in their own commits until we have approved the PR.

drivers/SmartThings/zigbee-switch/src/firstled-io/fingerprints.lua
thinkaName: Since this is the first product submitted, a series of devices will be added to reuse this driver. Currently, this is the 4+4 model, consisting of 4 switches that can be set to button mode, plus 4 pure buttons. The number of child devices to be created is determined by "fingerprints.children", and the number of buttons to be created is determined by "fingerprints.buttons". Subsequently, models such as 3+3, 2+2, and 1+1 will be added. Additionally, there will be models without the button, and the setting for fingerprints will be fingerprints.buttons=0.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
thinkaName: Based on the number of fingerprints.buttons and fingerprints.children of the corresponding model, create the corresponding sub-devices and buttons.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
thinkaName: ScreenShot_2026-06-12_101256_549
This is an additional judgment made for the subsequent model products.

thinkaName:
> It seems like there was a lot of stuff added since the last reviews wrt child devices, but we cant tell because you are obfuscating the commit history by always force pushing; this makes it hard to review this PR and others you have submitted. Please keep incremental changes in their own commits until we have approved the PR.

Because it was previously informed that a device's PR could only have one commit, so the forced override push was carried out.

cjswedes:
> > It seems like there was a lot of stuff added since the last reviews wrt child devices, but we cant tell because you are obfuscating the commit history by always force pushing; this makes it hard to review this PR and others you have submitted. Please keep incremental changes in their own commits until we have approved the PR.
> 
> Because it was previously informed that a device's PR could only have one commit, so the forced override push was carried out.

I apologize for the confusion. We do want there to only be a single commit when we merge, but it is useful to see how feedback is addressed by keeping commits separate during the review period.

drivers/SmartThings/zigbee-switch/src/firstled-io/fingerprints.lua
cjswedes: Thank you, this clears up a lot of my confusion with what is going on with the child devices. You may want to add a comment at the top of this subdriver explaining how this device works.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,164 @@
...
+  local old_preferences = args.old_st_store.preferences
+  if preferences ~= nil then
+    for id, attr in pairs(preference_map) do
+      local old_value = old_preferences[id]
+      local value = preferences[id]
+      if value ~= nil and value ~= old_value then
+	    if attr.data_type == data_types.Uint8 then
+	      value = tonumber(value)
+		end
+        device:send(cluster_base.write_manufacturer_specific_attribute(device, attr.cluster_id, attr.attribute_id,
```
cjswedes, line 93: This is correctly changing the mode of the device based on the preference change, but the app will still be showing both the switch capability and the button capability to the user. What happens when a user has the device in "button mode" and they issue an on or off command to the device from the app? 

My guess is that the app will time out on the command (it depends on how the device responds). Which is not a good user experience.

drivers/SmartThings/zigbee-switch/profiles/switch-button-wireless.yml
```diff
@@ -0,0 +1,17 @@
+name: switch-button-wireless
+components:
+- id: main
+  capabilities:
+  - id: switch
```
cjswedes, line 5: This is potentially a good time to use the modular profiles, which make capabilities optional and allow for changing them on the fly to make sure the only the switch or button capability is presented to the user based on how the preference is set.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
thinkaName: The product design is as follows: When in the switch mode, both the APP and the local button can control the relay. The button trigger changes the relay status and reports "onoff". In the button mode, when the local button is triggered, it reports the "RecallScenes" command. In this mode, the local button trigger does not control the relay. Users can still control the relay status through the APP, and the switch command is triggered normally.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,202 @@
...
+      device:emit_event(capabilities.button.numberOfButtons({ value = 1 },
+        { visibility = { displayed = false } }))
+      device:emit_event(capabilities.button.supportedButtonValues({ "pushed" },
+        { visibility = { displayed = false } }))
+    else
+      device:try_update_metadata({profile = "switch-light-restore-wireless"})
+    end
+  else
+    if show then
+      device:try_update_metadata({profile = "switch-button-wireless"})
```
cjswedes, line 55: Rather than have 4 profiles to account for the button capability not being used we have a feature called modular profiles which allows for a profile to specify a capability as optional, and then you can update which optional capabilities are included using try update metadata.


```suggestion
      device:try_update_metadata({profile = "switch-button-wireless", optional_component_capabilities = {{"main", {"button"}}})
```

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,202 @@
...
+      device.thread:call_with_delay(5, function()
+        device:emit_event(capabilities.button.numberOfButtons({value = 1}, {visibility = {displayed = false}}))
+        device:emit_event(capabilities.button.supportedButtonValues({"pushed"}, {visibility = {displayed = false}}))
+      end)
+    else
+      device:try_update_metadata({profile = "switch-wireless"})
```
cjswedes, line 57:
```suggestion
      device:try_update_metadata({profile = "switch-wireless", optional_component_capabilities = {{"main", {}}})
```

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
```diff
@@ -0,0 +1,202 @@
...
+        { visibility = { displayed = false } }))
+      device:emit_event(capabilities.button.supportedButtonValues({ "pushed" },
+        { visibility = { displayed = false } }))
+    else
+      device:try_update_metadata({profile = "switch-light-restore-wireless"})
+    end
+  else
+    if show then
+      device:try_update_metadata({profile = "switch-button-wireless"})
+      device.thread:call_with_delay(5, function()
```
cjswedes: This is okay, but it will fail if there any situations of higher latency which do sometimes occur. Once the profile/optional capabilities are updated, another infoChanged event will be triggered based on the changed capabilities, so the better place to emit these button events would be from the infoChanged handler based on the capabilities of the device.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
cjswedes: @ctowns or @hcarter-775 this will work to remove all optional capabilities, correct?

drivers/SmartThings/zigbee-switch/profiles/switch-button-light-restore-wireless.yml
```diff
@@ -0,0 +1,39 @@
+name: switch-button-light-restore-wireless
+components:
+- id: main
+  capabilities:
+  - id: switch
+    version: 1
+  - id: button
+    version: 1
```
cjswedes, line 8:
```suggestion
    version: 1
    optional: true
```

cjswedes (commented):
The removal of the button capability based on the preference makes sense, but I think its a good use case for modular profiles.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
cjswedes: I got confirmation that this does work.

thinkaName: ScreenShot_2026-06-16_163441_977
The actual testing shows that this method has a very low success rate. There are always some issues that cannot be resolved, and the UI fails to switch successfully.

cbaumler:
Hi @thinkaName, it looks like this PR branch is out of date with main. Please update it with the latest main and push the result. Then let us know when you're ready for us to re-review.

drivers/SmartThings/zigbee-switch/src/firstled-io/init.lua
cjswedes: Thank you for trying this approach, since you are getting good results with switching profiles, it makes sense to me to go with that. With the `listen_profile_button_transition` in place, this functionality is looking pretty solid.

drivers/SmartThings/zigbee-switch/src/test/test_firstled_switch.lua
```diff
@@ -0,0 +1,521 @@
...
+    cluster_base.write_manufacturer_specific_attribute(mock_parent, PRIVATE_CLUSTER_ID, 0x0002, MFG_CODE, data_types.Boolean, false)
+  })
+  mock_parent:expect_metadata_update({ profile = "switch-light-restore-wireless" })
+  end,
+  {
+    min_api_version = 19
+  }
+)
+
+local function test_clild_changeToWirelessSwitch_true(ep, name)
```
cjswedes:
```suggestion
local function test_child_changeToWirelessSwitch_true(ep, name)
```

cjswedes (approved):

thinkaName:
> Hi @thinkaName, it looks like this PR branch is out of date with main. Please update it with the latest main and push the result. Then let us know when you're ready for us to re-review.

It ready to re-review.

cbaumler:
Hi @thinkaName, it looks like there are some failing tests now. Could you address those before we re-review? Thanks!

thinkaName:
> Hi @thinkaName, it looks like there are some failing tests now. Could you address those before we re-review? Thanks!

It ready to re-review.

aleclorimer (approved):

PR #2977, opened 2026-05-18
WWSTCERT-11699 Z-Wave Scene Control Switch

cbaumler:
Relevant WWST tickets:
- WWSTCERT-11699


aleclorimer (approved):
I see we have some other ids that are hex strings of the manufacturer id, product type, and product id, but this `id` field is in base 10. Not sure if that matters.

cjswedes (approved):

cbaumler:
> I see we have some other ids that are hex strings of the manufacturer id, product type, and product id, but this `id` field is in base 10. Not sure if that matters.

I don't think it should matter as long as the `id` is unique. @cjswedes what do you think?

cjswedes:
> > I see we have some other ids that are hex strings of the manufacturer id, product type, and product id, but this `id` field is in base 10. Not sure if that matters.
> 
> I don't think it should matter as long as the `id` is unique. @cjswedes what do you think?

I dont think it matters technically, but we should probably change it to match our existing convention of using hex.

nand-nor:
> > > I see we have some other ids that are hex strings of the manufacturer id, product type, and product id, but this `id` field is in base 10. Not sure if that matters.
> > 
> > 
> > I don't think it should matter as long as the `id` is unique. @cjswedes what do you think?
> 
> I dont think it matters technically, but we should probably change it to match our existing convention of using hex.

Agree, especially to avoid confusion since for example 11 in hex and 11 in decimal are different values so it should all be consistently either base 10 or base 16. Especially if these values ever need to be parsed

nand-nor (approved):

PR #2978, opened 2026-05-18
Matter Switch: Remove ghost events, re-provision after driver switch

hcarter-775:
# Description of Change
Remove `pushed` event during configuration. This is not a required operation, and generates events not pre-conditioned by a device event.

Set device provisioning state back to PROVISIONED after a `driverSwitched` event to block repeated events.  

# Summary of Completed Tests
Tests updated. Driver switch tested with button device switched. Fresh button onboarding tested.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/init.lua
+++ b/drivers/SmartThings/matter-switch/src/init.lua
@@ -61,6 +61,7 @@ function SwitchLifecycleHandlers.driver_switched(driver, device)
   if device.network_type == device_lib.NETWORK_TYPE_MATTER and not switch_utils.detect_bridge(device) then
     device_cfg.match_profile(driver, device)
   end
+  device:try_update_metadata({provisioning_state = "PROVISIONED"})
 end
 
 function SwitchLifecycleHandlers.info_changed(driver, device, event, args)
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/aqara_cube/init.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/aqara_cube/init.lua
@@ -184,7 +184,9 @@ end
 local function do_configure(driver, device) end
 
 -- override driver_switched to prevent it running in the main driver
-local function driver_switched(driver, device) end
+local function driver_switched(driver, device)
+  device:try_update_metadata({provisioning_state = "PROVISIONED"})
+end
 
 local function initial_press_event_handler(driver, device, ib, response)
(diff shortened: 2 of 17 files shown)


cjswedes:
> Remove pushed event during configuration. This is not a required operation, and generates events not pre-conditioned by a device event.

I am not 100% sure about this, but I thought we did this to make sure there was an event for the capability to allow the app to see the device as online (not greyed out in the app). Did you test onboarding a button without it?

mocelet:
> Set device provisioning state back to PROVISIONED after a `driverSwitched` event to block repeated events

May I ask what does it block exactly?

Could not find much documentation about the provisioning states and, in fact, just realized many of my Matter devices have a provisioning state of NONFUNCTIONAL (probably due to driver changes) but they are working perfectly fine, no warnings or anything. An old [post at the Community](https://community.smartthings.com/t/provisioning-state-provisioned-vs-nonfunctional/256177/8?u=mocelet) stated that it had no effect.

hcarter-775:
> I am not 100% sure about this, but I thought we did this to make sure there was an event for the capability to allow the app to see the device as online (not greyed out in the app). Did you test onboarding a button without it?

@cjswedes Yes, I have tested onboarding a button without it and there is no difference in the UI elements with or without the extra event. I agree that there might have been some client reason for this event in the past, but it is no longer required.

hcarter-775:
> May I ask what does it block exactly?
> 
> Could not find much documentation about the provisioning states and, in fact, just realized many of my Matter devices have a provisioning state of NONFUNCTIONAL (probably due to driver changes) but they are working perfectly fine, no warnings or anything. An old [post at the Community](https://community.smartthings.com/t/provisioning-state-provisioned-vs-nonfunctional/256177/8?u=mocelet) stated that it had no effect.

@mocelet Without this change, devices switched into this driver will get the `DRIVER_SWITCH` provisioning state and will not lose it, meaning that the `driverSwitched` event will be queued on each startup. The default handling of `driverSwitched` will switch a device into either `PROVISIONED` or `NONFUNCTIONAL` and neither really has any significant difference today besides developer use. However, for reference if desired, a `TYPED` state will queue doConfigure and `DRIVER_SWITCH` will queue driverSwitched ad nauseam on startup.

mocelet:
Perfect, thanks as always @hcarter-775 for the insights!

cjswedes (approved):

tpmanley (approved):

PR #2979, opened 2026-05-18
Provision driver switch in Matter Lock

hcarter-775:
See description [here](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2978). Copying this work in other drivers.


cjswedes (approved):
You could add a test for this.

PR #2991, opened 2026-05-21
Add Sombra Shades window shade sub-driver

Woofser:
## Summary
Adds a sub-driver to the zigbee-window-treatment driver for Sombra Shades
Zigbee window shade motors (manufacturer "Sombra Shades").

## Details
- Position is reported and commanded through the standard Window Covering
  cluster (0% = open, 100% = closed); no inversion.
- Opening/Closing status is inferred from the direction of position change,
  so it works across Sombra motor variants (battery and DC) regardless of
  any manufacturer-specific motion cluster. A short settle timer emits the
  resting state once movement stops.
- Battery level via the Power Configuration cluster; power source via the
  Basic cluster.
- Registered through the existing lazy-load mechanism in sub_drivers.lua.

## Testing
- Unit tests in test_zigbee_window_treatment_sombra.lua.
- Verified on physical Sombra hardware (battery and DC motors): open, close,
  pause, set level, preset, position reporting, battery, and power source.

### DIFF
--- a/drivers/SmartThings/zigbee-window-treatment/fingerprints.yml
+++ b/drivers/SmartThings/zigbee-window-treatment/fingerprints.yml
@@ -138,6 +138,16 @@ zigbeeManufacturer:
     manufacturer: HOPOsmart
     model: A2230011
     deviceProfileName: window-shade-only
+  - id: "Sombra Shades/WM25/L-Z"
+    deviceLabel: Sombra Shades Window Treatment
+    manufacturer: Sombra Shades
+    model: WM25/L-Z
+    deviceProfileName: window-treatment-battery
+  - id: "Sombra Shades/SS25/L-Z"
+    deviceLabel: Sombra Automated Shades and Blinds
+    manufacturer: Sombra Shades
+    model: SS25/L-Z
+    deviceProfileName: window-treatment-battery
   - id: "Sombra Shades/SOMBRA/Z-M"
     deviceLabel: Sombra Automated Shades
     manufacturer: Sombra Shades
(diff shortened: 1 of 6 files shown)


nand-nor:
Is the change to the fingerprints file intentional? Im not sure this will be backwards compatible, depending on if there are devices in the field that already use the exiting fingerprint

Woofser:
Good catch — that removal wasn't intentional. This branch predated the `Sombra Shades/WM25/L-Z` and `Sombra Shades/SS25/L-Z` fingerprints being added to `main`, so the diff was dropping them.

I've pushed a commit restoring both entries (unchanged, still mapped to `window-treatment-battery`), so the change is now purely additive — it only adds the `Sombra Shades/SOMBRA/Z-M` fingerprint (mapped to `window-treatment-powerSource` and handled by the new `sombra` sub-driver). No existing fingerprints are removed, so any field devices on `WM25/L-Z` or `SS25/L-Z` keep matching. Thanks for catching it!

cbaumler:
Hi @Woofser, could you also update this branch? Looks like it's out-of-date with main.

Woofser:
Thanks @cbaumler — done. I've updated the branch, so it's now up to date with `main` and merges cleanly. Let me know if anything else is needed.

cbaumler:
Hi @Woofser, it looks like the branch is still out-of-date with `main`.

Woofser:
Hi @cbaumler, That's odd. It looked like it went through. It should be good now

nand-nor (approved):

cjswedes (approved):

aleclorimer:
Looks like two tests are failing.
> A falling position reports opening, then settles to partially open
> A rising position reports closing, then settles to partially open

Woofser:
`d6633f6` should fix the two failures. The settle tests were advancing the mock clock without registering the timer first, so `partially_open` never fired — added the `__create_and_queue_test_time_advance_timer` call to match the `doConfigure`/`added` tests. Test-only change. Mind re-running the workflow when you get a sec? Thanks!

aleclorimer (approved):

PR #2993, opened 2026-05-21
Matter Switch: Use Total Wh for PCR if delta < 0.0

hcarter-775:
# Description of Change
Use total Wh instead of a negative value for the delta of the powerConsumptionReport. This should only ever be negative in the case that the device has had its energy meter reset, so we should avoid spiking the energy report for ST energy in a case like this. 

Since this would occur only on a rare and one-time case, I've elected not to include any extra state to account for the energy delta that would have also accumulated prior to the reset. I think the tradeoff of adding the state vs. what is actually lost favors just losing that bit of state.

# Summary of Completed Tests
Unit tests updated.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/utils.lua
@@ -402,12 +402,13 @@ function utils.report_power_consumption_to_st_energy(device, endpoint_id, total_
 
   local previous_imported_report = utils.get_latest_state_for_endpoint(device, endpoint_id, capabilities.powerConsumptionReport.ID,
     capabilities.powerConsumptionReport.powerConsumption.NAME, { energy = total_imported_energy_wh }) -- default value if nil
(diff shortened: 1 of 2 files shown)


Kwang-Hui (approved):
Looks good to me

mocelet:
> avoid spiking the energy report for ST energy

I thought the purpose was to avoid losing ST energy data for the interval (e.g. 15 minutes) when a reset happens.

Emitting the total Wh when assuming a reset happened prevents losing the part of the energy post-reset. However, the part before the reset in that interval would be lost.

Maybe the pre-reset part can be restored so the energy reportings for ST Energy are completely reset-proof. Would need to store the maximum energy received in a field and, when the reset is detected, calculate the delta before the reset and add it to the total Wh so the delta of that interval is accurate.

Matter does not have standard commands to reset the meter anyway so probably not worth it complicating it at the moment.

nickolas-deboom (approved):

PR #3019, opened 2026-06-08
Isolate the eve-energy subdriver to private cluster usage

samadDotDev:
# Type of Change

- [x] Bug fix

# Checklist

# Description of Change

The eve-energy sub-driver was being loaded even for other Eve devices that don't contain the private cluster (which eve energy subdriver uses for energy/power measurements pre-1.3). This change limits the scope of that subdriver to just the devices that contains such private cluster (id of which has the encoding for Eve's vendor id already) AND when device doesn't have the standard device type that was later introduced in Matter.

# Summary of Completed Tests

- Added unit tests (these are AI-generated)
- Tested with Eve Energy (containing standard + private clusters)
- Tested with another Eve device that uses matter-switch driver but doesn't contain electrical sensor or that private cluster

### DIFF
--- a/drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
+++ b/drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
@@ -15,28 +15,6 @@ local PRIVATE_ATTR_ID_WATT = 0x130A000A
 local PRIVATE_ATTR_ID_WATT_ACCUMULATED = 0x130A000B
 local PRIVATE_ATTR_ID_ACCUMULATED_CONTROL_POINT = 0x130A000E
 
--- Helper function to add get_endpoints method to mock devices for can_handle testing
-local function add_get_endpoints_to_mock(device)
-  device.get_endpoints = function(self, cluster_id, opts)
-    opts = opts or {}
-    local eps = {}
-    for _, ep in ipairs(self.endpoints) do
-      for _, cluster in ipairs(ep.clusters or {}) do
-        if cluster.cluster_id == cluster_id then
-          -- Check feature_bitmap if specified
-          if opts.feature_bitmap == nil or
-             (cluster.feature_map and (cluster.feature_map & opts.feature_bitmap) == opts.feature_bitmap) then
-            table.insert(eps, ep.endpoint_id)
-            break
-          end
-        end
-      end
-    end
-    return eps
-  end
-  return device
-end
-
 local mock_device = test.mock_device.build_test_matter_device({
   profile = t_utils.get_profile_definition("power-energy-powerConsumption.yml"),
   manufacturer_info = {
@@ -75,7 +53,6 @@ local mock_device = test.mock_device.build_test_matter_device({
     }
   }
 })
-add_get_endpoints_to_mock(mock_device)
 
 local mock_eve_device_using_electrical_sensor = test.mock_device.build_test_matter_device({
   profile = t_utils.get_profile_definition("plug-energy-powerConsumption.yml"),
@@ -135,7 +112,6 @@ local mock_eve_device_using_electrical_sensor = test.mock_device.build_test_matt
     }
   }
 })
-add_get_endpoints_to_mock(mock_eve_device_using_electrical_sensor)
 
 -- Mock device without Eve Private Cluster (should not match eve_energy sub-driver)
 local mock_device_without_private_cluster = test.mock_device.build_test_matter_device({
@@ -170,7 +146,6 @@ local mock_device_without_private_cluster = test.mock_device.build_test_matter_d
     }
   }
 })
-add_get_endpoints_to_mock(mock_device_without_private_cluster)
 
 local function test_init()
   local cluster_subscribe_list = {
@@ -188,7 +163,6 @@ local function test_init()
 end
 test.set_test_init_function(test_init)
 
--- Test can_handle logic
 test.register_coroutine_test(
   "Eve Energy sub-driver can_handle should return true for devices with Eve Private Cluster and no Electrical Sensor",
   function()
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
```diff
@@ -129,6 +188,44 @@ local function test_init()
 end
 test.set_test_init_function(test_init)
 
+-- Test can_handle logic
```
hcarter-775: nit: line not needed

drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
```diff
@@ -129,6 +188,44 @@ local function test_init()
...
+  "Eve Energy sub-driver can_handle should return false for devices with Electrical Sensor device type",
+  function()
+    local eve_energy_can_handle = require("sub_drivers.eve_energy.can_handle")
+    local result = eve_energy_can_handle(nil, nil, mock_eve_device_using_electrical_sensor)
+    assert(result == false, "can_handle should return false for Eve device with electrical sensor")
+  end,
+  {
+    min_api_version = 17
+  }
+)
```
hcarter-775: this is already tested in this file

drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
```diff
@@ -15,6 +15,28 @@ local PRIVATE_ATTR_ID_WATT = 0x130A000A
...
+            break
+          end
+        end
+      end
+    end
+    return eps
+  end
+  return device
+end
+
```
hcarter-775: this feels like an overcomplicated solution. I might suggest we just add one more mock device with both clusters present.

samadDotDev: I actually started with that i.e. by adding this util function to `can_handle` and that'd just work best for mock device in tests, but didn't want to influence the functional code just because of test infra not providing `device::get_endpoints`:

```diff
diff --git a/drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua b/drivers/SmartThings/matter-switch/s
rc/sub_drivers/eve_energy/can_handle.lua
index 02bc9e06..b5b536a4 100644
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/eve_energy/can_handle.lua
@@ -5,13 +5,24 @@ local device_lib = require "st.device"
 local fields = require "switch_utils.fields"
 local switch_utils = require "switch_utils.utils"

+local function has_cluster(device, cluster_id)
+  for _, ep in ipairs(device.endpoints) do
+    for _, cluster in ipairs(ep.clusters or {}) do
+      if cluster.cluster_id == cluster_id then
+        return true
+      end
+    end
+  end
+  return false
+end
+
 return function(opts, driver, device)
   local EVE_PRIVATE_CLUSTER_ID = 0x130AFC01
   -- this sub driver loads for devices that:
   -- 1. Contain the Eve Private Cluster (0x130AFC01)
   -- 2. Do NOT have the Standard Electrical Sensor device type
   if device.network_type == device_lib.NETWORK_TYPE_MATTER and
-    #device:get_endpoints(EVE_PRIVATE_CLUSTER_ID) > 0 and
+    has_cluster(device, EVE_PRIVATE_CLUSTER_ID) and
     #switch_utils.get_endpoints_by_device_type(device, fields.DEVICE_TYPE_ID.ELECTRICAL_SENSOR) == 0 then
     return true, require("sub_drivers.eve_energy")
   end
```

Let me know what you prefer, and curious about thoughts from @ctowns as well here!

drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
samadDotDev: Yeah this is indirectly tested by a larger test but I wanted to add specific tests for `can_handle` in all 3 combinations (with private cluster, electrical sensor device type, and both). I can remove this one though.

drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
samadDotDev: I made that change in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3019/commits/5d2a4d2c430262e1c301360b95d242d22a5e4eaf but LMK if you were thinking something different.

drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
samadDotDev: Removed in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3019/commits/5d2a4d2c430262e1c301360b95d242d22a5e4eaf

nickolas-deboom (approved):

drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
hcarter-775: Ah, I see the issue you're mentioning. Not to spend too much more time on this, but perhaps it would be better to implicitly test can_handle by just adding an init test instead of an explicit can_handle test? It is a little annoying that the current infrastructure doesn't well support direct can_handle tests, so this is more or less the way that other drivers do it.

hcarter-775: If you have issues with that, I do think it would be better to revert to the first version you made than to affect the actual can_handle code.

samadDotDev: Yeah at that point we're exercising what existing tests are already doing implicitly and seems odd to cover all cases of specifically can_handle. I have reverted the change in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3019/commits/d141a472e31185025671b66a12fe148283b88d6c

hcarter-775 (approved):

drivers/SmartThings/matter-switch/src/test/test_eve_energy.lua
```diff
@@ -15,6 +15,28 @@ local PRIVATE_ATTR_ID_WATT = 0x130A000A
 local PRIVATE_ATTR_ID_WATT_ACCUMULATED = 0x130A000B
 local PRIVATE_ATTR_ID_ACCUMULATED_CONTROL_POINT = 0x130A000E
 
+-- Helper function to add get_endpoints method to mock devices
+local function add_get_endpoints_to_mock(device)
```
ctowns, line 19: Just a note: technically this does not match the lua library definition of the `get_endpoints` function, but I think it covers the test appropriately. The alternative would be to manually register each of the new mock devices using their own test init function so that the device function metatable is populated with functions like `get_endpoints` from the lua libs, but I think that is too much overhead for just doing a quick `can_handle` test for each of these mock devices, and this works well for the purposes of this test.

ctowns (approved):

PR #3039, opened 2026-06-23
Matter Ikea Scroll: Add global scroll state

hcarter-775:
# Description of Change
Add global state to check for upper and lower bounds within a time frame for the ikea scroll. This should halt extra rotateAmount events from being processed after the threshold for "100%" has been reached.

# Summary of Completed Tests
Tested on-device

### DIFF
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/ikea_scroll/scroll_handlers/event_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/ikea_scroll/scroll_handlers/event_handlers.lua
@@ -17,7 +17,9 @@ local function rotate_amount_event_helper(device, endpoint_id, num_presses_to_ha
   local scroll_direction = switch_utils.tbl_contains(scroll_fields.ENDPOINTS_UP_SCROLL, endpoint_id) and 1 or -1
   local scroll_amount = st_utils.clamp_value(scroll_direction * scroll_fields.PER_SCROLL_EVENT_ROTATION * num_presses_to_handle, -100, 100)
 
-  device:emit_event_for_endpoint(endpoint_id, capabilities.knob.rotateAmount(scroll_amount, {state_change = true}))
+  if event_utils.is_valid_scroll_amount(device, scroll_amount) then
+    device:emit_event_for_endpoint(endpoint_id, capabilities.knob.rotateAmount(scroll_amount, {state_change = true}))
+  end
 end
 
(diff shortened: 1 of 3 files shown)


gharveymn:
Is the idea of the timer rather than just always tracking the state that we get protected for a certain amount of time, but still allow the device to send us new out of bounds events just in case, if if we were to get the state tracking wrong? Is it possible for us to ignore any valid events?

Also, is the behavior like the following?

1. User scrolls to 150% -> Driver clamps to 100%
2. User scrolls down 100% -> Driver sees this as scrolling to 50%

Does that feel intuitive? I'm not arguing for or against, just wondering if you've tested to see how it feels.

hcarter-775:
> Is the idea of the timer rather than just always tracking the state that we get protected for a certain amount of time, but still allow the device to send us new out of bounds events just in case, if if we were to get the state tracking wrong? Is it possible for us to ignore any valid events?

Basically, yeah. The timer basically protects us within a context window that we can be fairly confident about for this device. But as to why we need a timer, imagine this is scrolling a dimmable bulb. In the moment we can cap between -100 and 100, but what if later the user moves the level to 30? Then scrolling "up" should definitely do something. I think this strangeness speaks to the fact that this work is kind of a "stopgap" while the logic is moved to the bulb step logic more directly. So yeah, technically a >100 scroll is totally valid and so in a way, this would be ignoring "valid" events, but those >100 scrolls are not used by our current production logic.

> Also, is the behavior like the following?
User scrolls to 150% -> Driver clamps to 100%
User scrolls down 100% -> Driver sees this as scrolling to 50%
Does that feel intuitive? I'm not arguing for or against, just wondering if you've tested to see how it feels.

Yeah I've tested all this and as far as the current use cases for the scroll, this is totally fine today, since past 100% there is no way the dimmable bulb would have anywhere more to go up, vice versa for down. However, `User scrolls down 100% -> Driver sees this as scrolling to 50%` is kinda wrong in the current implementation, it would actually go down to 0% if they did that, since the global state is clamped at 100/-100. Because we don't care about the scroll state past 100%, so we don't track it as different than 100% here.

gharveymn:
Discussed offline for a bit. Clarified a typo in my original comment which should read "User scrolls down **to** 100%",  and confirmed the original postulation for the behavior was correct. Also confirmed that the behavior makes more sense than having an arbitrary gate at 100%, where there might not be any tactile feedback on where that 100% is.

gharveymn (approved):

tpmanley (approved):

ctowns (approved):

PR #3052, opened 2026-06-26
Update matter window covering edge driver

hongye-samsung:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests
Co-Authored-By: Cline SR (50%)
CGAI & AI Review

### DIFF
--- a/drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-battery.yml
+++ b/drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-battery.yml
@@ -8,6 +8,8 @@ components:
     version: 1
   - id: windowShadeLevel
     version: 1
+  - id: statelessWindowShadeLevelStep
+    version: 1
   - id: windowShadeTiltLevel
     version: 1
   - id: battery
--- a/drivers/SmartThings/matter-window-covering/src/init.lua
+++ b/drivers/SmartThings/matter-window-covering/src/init.lua
@@ -9,6 +9,7 @@ local im = require "st.matter.interaction_model"
 local log = require "log"
 local clusters = require "st.matter.clusters"
 local MatterDriver = require "st.matter.driver"
+local utils = require "st.utils"
 
 local CURRENT_LIFT = "__current_lift"
 local CURRENT_TILT = "__current_tilt"
@@ -20,6 +21,11 @@ local battery_support = {
 local REVERSE_POLARITY = "__reverse_polarity"
 local PRESET_LEVEL_KEY = "__preset_level_key"
 local DEFAULT_PRESET_LEVEL = 50
+local LATEST_TARGET_LEVEL = "_latest_target_level"
+local TARGET_LEVEL_TIME_OUT = "_target_level_timeout"
+local TARGET_LEVEL_TIME_OUT_SECONDS = 15
+
+local TARGET_REACH_TOLERANCE = 1
 
 local function find_default_endpoint(device, cluster)
   local res = device.MATTER_DEFAULT_ENDPOINT
@@ -175,15 +181,44 @@ local function handle_shade_level(driver, device, cmd)
   device:send(req)
 end
 
+local function handle_step_shade_level(driver, device, cmd)
+  local step = cmd.args.stepSize
+
+  local current_level = device:get_field(LATEST_TARGET_LEVEL) or
+    device:get_latest_state("main", capabilities.windowShadeLevel.ID,
+      capabilities.windowShadeLevel.shadeLevel.NAME) or 0
+
+  local target_level = utils.clamp_value(current_level + step, 0, 100)
+
+  device:set_field(LATEST_TARGET_LEVEL, target_level)
+
+  local old_timer = device:get_field(TARGET_LEVEL_TIME_OUT)
+  if old_timer ~= nil then
+    device.thread:cancel_timer(old_timer)
+  end
+
+  -- Set 15 second timeout timer to ensure target_level is cleared
+  local timer = device.thread:call_with_delay(TARGET_LEVEL_TIME_OUT_SECONDS, function(d)
+    device:set_field(LATEST_TARGET_LEVEL, nil)
+  end)
+  device:set_field(TARGET_LEVEL_TIME_OUT, timer)
+
+  driver:inject_capability_command(device, {
+    capability = capabilities.windowShadeLevel.ID,
+    component = cmd.component,
+    command = capabilities.windowShadeLevel.commands.setShadeLevel.NAME,
+    named_args = { shadeLevel = target_level }
+  })
+end
+
 -- move to shade tilt level between 0-100
 local function handle_shade_tilt_level(driver, device, cmd)
   local endpoint_id = device:component_to_endpoint(cmd.component)
   local tilt_percentage_value = 100 - cmd.args.level
   local hundredths_tilt_percentage = tilt_percentage_value * 100
-  local req = clusters.WindowCovering.server.commands.GoToTiltPercentage(
+  device:send(clusters.WindowCovering.server.commands.GoToTiltPercentage(
     device, endpoint_id, hundredths_tilt_percentage
-  )
-  device:send(req)
+  ))
 end
 
 -- current lift/tilt percentage, changed to 100ths percent
@@ -194,6 +229,22 @@ local current_pos_handler = function(attribute)
     end
     local windowShade = capabilities.windowShade.windowShade
     local position = 100 - math.floor(ib.data.value / 100)
+
+    -- Step control logic
+    local target_level_field = device:get_field(LATEST_TARGET_LEVEL)
+    if target_level_field and attribute == capabilities.windowShadeLevel.shadeLevel then
+      -- Allow ±1 degree tolerance for reaching target
+      if math.abs(position - target_level_field) <= TARGET_REACH_TOLERANCE then
+        -- Device reached target position, clear target marker and timeout timer
+        device:set_field(LATEST_TARGET_LEVEL, nil)
+        local timer = device:get_field(TARGET_LEVEL_TIME_OUT)
+        if timer ~= nil then
+          device.thread:cancel_timer(timer)
+          device:set_field(TARGET_LEVEL_TIME_OUT, nil)
+        end
+      end
+    end
+
     local reverse = device:get_field(REVERSE_POLARITY)
     device:emit_event_for_endpoint(ib.endpoint_id, attribute(position))
 
@@ -253,6 +304,16 @@ local function current_status_handler(driver, device, ib, response)
     device:emit_event_for_endpoint(ib.endpoint_id, reverse and windowShade.closing() or windowShade.opening())
   elseif state == 2 then -- closing
     device:emit_event_for_endpoint(ib.endpoint_id, reverse and windowShade.opening() or windowShade.closing())
+  elseif state == 0 then -- stopped (idle)
+    -- Clear target_level when device stops moving
+    local target_level_field = device:get_field(LATEST_TARGET_LEVEL)
+    if target_level_field then
+      device:set_field(LATEST_TARGET_LEVEL, nil)
+      local timer = device:get_field(TARGET_LEVEL_TIME_OUT)
+      if timer ~= nil then
+        device.thread:cancel_timer(timer)
+      end
+    end
   elseif state ~= 0 then -- unknown
     device:emit_event_for_endpoint(ib.endpoint_id, windowShade.unknown())
   end
@@ -353,6 +414,9 @@ local matter_driver_template = {
     },
     [capabilities.windowShadeLevel.ID] = {
       [capabilities.windowShadeLevel.commands.setShadeLevel.NAME] = handle_shade_level,
+    },
+    [capabilities.statelessWindowShadeLevelStep.ID] = {
+      [capabilities.statelessWindowShadeLevelStep.commands.stepShadeLevel.NAME] = handle_step_shade_level
     },
     [capabilities.windowShadeTiltLevel.ID] = {
       [capabilities.windowShadeTiltLevel.commands.setShadeTiltLevel.NAME] = handle_shade_tilt_level,
--- a/drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
+++ b/drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
@@ -1222,4 +1222,232 @@ test.register_coroutine_test(
   }
 )
 
+-- stepShadeLevel tests
+
+test.register_coroutine_test(
+  "WindowShade stepShadeLevel cmd handler - step up", function()
+    test.socket.capability:__queue_receive(
+      {
+        mock_device.id,
+        {capability = "statelessWindowShadeLevelStep", component = "main", command = "stepShadeLevel", args = { 10 }},
+      }
+    )
+    test.socket.matter:__expect_send(
+      {mock_device.id, WindowCovering.server.commands.GoToLiftPercentage(mock_device, 10, 9000)}
+    )
+  end,
+  {
+     min_api_version = 19
+  }
+)
+
+test.register_coroutine_test(
+  "WindowShade stepShadeLevel cmd handler - step down", function()
+    test.socket.matter:__queue_receive(
+      {
+        mock_device.id,
+        WindowCovering.attributes.CurrentPositionLiftPercent100ths:build_test_report_data(
+          mock_device, 10, 5000
+        ),
+      }
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShadeLevel.shadeLevel(50)
+      )
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShade.windowShade.partially_open()
+      )
+    )
+    test.wait_for_events()
+    test.socket.capability:__queue_receive(
+      {
+        mock_device.id,
+        {capability = "statelessWindowShadeLevelStep", component = "main", command = "stepShadeLevel", args = { -20 }},
+      }
+    )
+    test.socket.matter:__expect_send(
+      {mock_device.id, WindowCovering.server.commands.GoToLiftPercentage(mock_device, 10, 7000)}
+    )
+  end,
+  {
+     min_api_version = 19
+  }
+)
+
+test.register_coroutine_test(
+  "WindowShade stepShadeLevel cmd handler - continuous step with target tracking", function()
+    test.socket.matter:__queue_receive(
+      {
+        mock_device.id,
+        WindowCovering.attributes.CurrentPositionLiftPercent100ths:build_test_report_data(
+          mock_device, 10, 7000
+        ),
+      }
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShadeLevel.shadeLevel(30)
+      )
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShade.windowShade.partially_open()
+      )
+    )
+    test.wait_for_events()
+    test.socket.capability:__queue_receive(
+      {
+        mock_device.id,
+        {capability = "statelessWindowShadeLevelStep", component = "main", command = "stepShadeLevel", args = { 10 }},
+      }
+    )
+    test.socket.matter:__expect_send(
+      {mock_device.id, WindowCovering.server.commands.GoToLiftPercentage(mock_device, 10, 6000)}
+    )
+    test.wait_for_events()
+    test.socket.capability:__queue_receive(
+      {
+        mock_device.id,
+        {capability = "statelessWindowShadeLevelStep", component = "main", command = "stepShadeLevel", args = { 10 }},
+      }
+    )
+    test.socket.matter:__expect_send(
+      {mock_device.id, WindowCovering.server.commands.GoToLiftPercentage(mock_device, 10, 5000)}
+    )
+  end,
+  {
+     min_api_version = 19
+  }
+)
+
+test.register_coroutine_test(
+  "WindowShade stepShadeLevel - target reached clears target marker", function()
+    test.socket.matter:__queue_receive(
+      {
+        mock_device.id,
+        WindowCovering.attributes.CurrentPositionLiftPercent100ths:build_test_report_data(
+          mock_device, 10, 5000
+        ),
+      }
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShadeLevel.shadeLevel(50)
+      )
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShade.windowShade.partially_open()
+      )
+    )
+    test.wait_for_events()
+    test.socket.capability:__queue_receive(
+      {
+        mock_device.id,
+        {capability = "statelessWindowShadeLevelStep", component = "main", command = "stepShadeLevel", args = { 10 }},
+      }
+    )
+    test.socket.matter:__expect_send(
+      {mock_device.id, WindowCovering.server.commands.GoToLiftPercentage(mock_device, 10, 4000)}
+    )
+    test.wait_for_events()
+    test.socket.matter:__queue_receive(
+      {
+        mock_device.id,
+        WindowCovering.attributes.CurrentPositionLiftPercent100ths:build_test_report_data(
+          mock_device, 10, 4000
+        ),
+      }
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShadeLevel.shadeLevel(60)
+      )
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShade.windowShade.partially_open()
+      )
+    )
+  end,
+  {
+     min_api_version = 19
+  }
+)
+
+test.register_coroutine_test(
+  "WindowShade stepShadeLevel - step up to maximum (100)", function()
+    test.socket.matter:__queue_receive(
+      {
+        mock_device.id,
+        WindowCovering.attributes.CurrentPositionLiftPercent100ths:build_test_report_data(
+          mock_device, 10, 500
+        ),
+      }
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShadeLevel.shadeLevel(95)
+      )
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShade.windowShade.partially_open()
+      )
+    )
+    test.wait_for_events()
+    test.socket.capability:__queue_receive(
+      {
+        mock_device.id,
+        {capability = "statelessWindowShadeLevelStep", component = "main", command = "stepShadeLevel", args = { 10 }},
+      }
+    )
+    test.socket.matter:__expect_send(
+      {mock_device.id, WindowCovering.server.commands.GoToLiftPercentage(mock_device, 10, 0)}
+    )
+  end,
+  {
+     min_api_version = 19
+  }
+)
+
+test.register_coroutine_test(
+  "WindowShade stepShadeLevel - step down to minimum (0)", function()
+    test.socket.matter:__queue_receive(
+      {
+        mock_device.id,
+        WindowCovering.attributes.CurrentPositionLiftPercent100ths:build_test_report_data(
+          mock_device, 10, 9500
+        ),
+      }
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShadeLevel.shadeLevel(5)
+      )
+    )
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message(
+        "main", capabilities.windowShade.windowShade.partially_open()
+      )
+    )
+    test.wait_for_events()
+    test.socket.capability:__queue_receive(
+      {
+        mock_device.id,
+        {capability = "statelessWindowShadeLevelStep", component = "main", command = "stepShadeLevel", args = { -10 }},
+      }
+    )
+    test.socket.matter:__expect_send(
+      {mock_device.id, WindowCovering.server.commands.GoToLiftPercentage(mock_device, 10, 10000)}
+    )
+  end,
+  {
+     min_api_version = 19
+  }
+)
+
 test.run_registered_tests()
--- a/drivers/SmartThings/matter-window-covering/profiles/window-covering-battery.yml
+++ b/drivers/SmartThings/matter-window-covering/profiles/window-covering-battery.yml
@@ -8,6 +8,8 @@ components:
     version: 1
   - id: windowShadeLevel
     version: 1
+  - id: statelessWindowShadeLevelStep
+    version: 1
   - id: battery
     version: 1
   - id: firmwareUpdate
--- a/drivers/SmartThings/matter-window-covering/profiles/window-covering-batteryLevel.yml
+++ b/drivers/SmartThings/matter-window-covering/profiles/window-covering-batteryLevel.yml
@@ -8,6 +8,8 @@ components:
     version: 1
   - id: windowShadeLevel
     version: 1
+  - id: statelessWindowShadeLevelStep
+    version: 1
   - id: batteryLevel
     version: 1
   - id: firmwareUpdate
--- a/drivers/SmartThings/matter-window-covering/profiles/window-covering-profile.yml
+++ b/drivers/SmartThings/matter-window-covering/profiles/window-covering-profile.yml
@@ -9,6 +9,8 @@ components:
     version: 1
   - id: windowShadeLevel
     version: 1
+  - id: statelessWindowShadeLevelStep
+    version: 1
   - id: battery
     version: 1
   - id: firmwareUpdate
--- a/drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt.yml
+++ b/drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt.yml
@@ -8,6 +8,8 @@ components:
     version: 1
   - id: windowShadeLevel
     version: 1
+  - id: statelessWindowShadeLevelStep
+    version: 1
   - id: windowShadeTiltLevel
     version: 1
   - id: firmwareUpdate
--- a/drivers/SmartThings/matter-window-covering/profiles/window-covering.yml
+++ b/drivers/SmartThings/matter-window-covering/profiles/window-covering.yml
@@ -8,6 +8,8 @@ components:
     version: 1
   - id: windowShadeLevel
     version: 1
+  - id: statelessWindowShadeLevelStep
+    version: 1
   - id: firmwareUpdate
     version: 1
   - id: refresh


vangoran:
Edge driver development for R2 new feature (https://smartthings.atlassian.net/browse/CHAD-18661)

hongye-samsung:
@HunsupJung @hcarter-775 @z-michel

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -354,6 +436,9 @@ local matter_driver_template = {
     [capabilities.windowShadeLevel.ID] = {
       [capabilities.windowShadeLevel.commands.setShadeLevel.NAME] = handle_shade_level,
     },
+    [capabilities.statelessWindowShadeLevelStep.ID] = {
+      [capabilities.statelessWindowShadeLevelStep.commands.stepShadeLevel.NAME] = window_shade_step_level_cmd
```
hcarter-775:
```suggestion
      [capabilities.statelessWindowShadeLevelStep.commands.stepShadeLevel.NAME] = handle_step_shade_level
```
let's keep the same language consistency

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -373,4 +458,4 @@ local matter_driver_template = {
 }
 
 local matter_driver = MatterDriver("matter-window-covering", matter_driver_template)
-matter_driver:run()
+matter_driver:run()
```
hcarter-775: we shouldn't remove the newline

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -121,6 +126,7 @@ local function device_removed(driver, device) log.info("device removed") end
 
 -- capability handlers
 local function handle_preset(driver, device, cmd)
+  device:set_field(LATEST_TARGET_LEVEL, nil)
```
hcarter-775: why are we setting this to nil in every handler?

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +185,51 @@ local function handle_shade_level(driver, device, cmd)
...
+  local target_level_field = device:get_field(LATEST_TARGET_LEVEL)
+  local current_level = target_level_field or
+    device:get_latest_state("main", capabilities.windowShadeLevel.ID,
+      capabilities.windowShadeLevel.shadeLevel.NAME) or 0
+
+  -- Calculate new target (user level: 0-100, 0=closed, 100=open)
+  local target_level = current_level + step
+  if target_level > 100 then target_level = 100
+  elseif target_level < 0 then target_level = 0
+  end
```
hcarter-775:
```suggestion
  local target_level = st_utils.clamp_value(current_level + step, 0, 100)
```
and then you can do `local st_utils = require "st.utils"` at the top of the file.

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +185,51 @@ local function handle_shade_level(driver, device, cmd)
...
+  -- Set 30 second timeout timer to ensure target_level is cleared
+  local timer = device.thread:call_with_delay(TARGET_LEVEL_TIME_OUT_SECONDS, function(d)
+    device:set_field(LATEST_TARGET_LEVEL, nil)
+    device:set_field(TARGET_LEVEL_TIME_OUT, nil)
+  end)
+  device:set_field(TARGET_LEVEL_TIME_OUT, timer)
+
+  -- Matter uses inverted logic (like IKEA)
+  -- User level: 0=closed, 100=open
+  -- Matter level: 10000=open, 0=closed (in percent100ths)
```
hcarter-775: remove this comment

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +185,51 @@ local function handle_shade_level(driver, device, cmd)
...
+  local lift_percentage_value = 100 - target_level
+  local hundredths_lift_percentage = lift_percentage_value * 100
+
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local req = clusters.WindowCovering.server.commands.GoToLiftPercentage(
+    device, endpoint_id, hundredths_lift_percentage
+  )
+  device:send(req)
```
hcarter-775: we need to handle the case where the device is in the reverse mode

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +185,51 @@ local function handle_shade_level(driver, device, cmd)
...
+      capabilities.windowShadeLevel.shadeLevel.NAME) or 0
+
+  -- Calculate new target (user level: 0-100, 0=closed, 100=open)
+  local target_level = current_level + step
+  if target_level > 100 then target_level = 100
+  elseif target_level < 0 then target_level = 0
+  end
+
+  -- Update tracking state
+  device:set_field(LATEST_TARGET_LEVEL, target_level)
```
hcarter-775, line 193: why do we even need this field? What purpose is it really serving here?

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -194,6 +249,22 @@ local current_pos_handler = function(attribute)
...
+      if math.abs(position - target_level_field) <= TARGET_REACH_TOLERANCE then
+        -- Device reached target position, clear target marker and timeout timer
+        device:set_field(LATEST_TARGET_LEVEL, nil)
+        local timer = device:get_field(TARGET_LEVEL_TIME_OUT)
+        if timer ~= nil then
+          device.thread:cancel_timer(timer)
+          device:set_field(TARGET_LEVEL_TIME_OUT, nil)
+        end
+      end
+    end
```
hcarter-775, line 246: I don't think I understand why this is useful. The command will send the device here anyway? This just removes the field without changing anything as far as I can see

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -253,6 +324,17 @@ local function current_status_handler(driver, device, ib, response)
...
+    -- Clear target_level when device stops moving
+    local target_level_field = device:get_field(LATEST_TARGET_LEVEL)
+    if target_level_field then
+      device:set_field(LATEST_TARGET_LEVEL, nil)
+      local timer = device:get_field(TARGET_LEVEL_TIME_OUT)
+      if timer ~= nil then
+        device.thread:cancel_timer(timer)
+        device:set_field(TARGET_LEVEL_TIME_OUT, nil)
+      end
+    end
```
hcarter-775, line 316: I don't see why this is needed. kinda repeating myself here, but I'm not really sure what this logic is gaining for us.

hcarter-775 (commented):
Seems like the new fields may be unnecessary, and may be adding some unnecessary complexity

drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-only-battery.yml
```diff
@@ -6,6 +6,8 @@ components:
     version: 1
   - id: windowShadeTiltLevel
     version: 1
+  - id: statelessWindowShadeLevelStep
```
tpmanley: My initial thought is we'd only add the new capability to profiles that have the `windowShadeLevel` capability, and so it shouldn't be added to all these profiles that have `windowShadeTiltLevel` and not `windowShadeLevel`. In other words, we wouldn't support the stateless step commands on shades that only support tilt. If we want to do that in the future we should add a `statelessWindowShadeTiltStep` capability, but that isn't in scope for the current project.

drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-battery.yml
```diff
@@ -10,6 +10,8 @@ components:
     version: 1
   - id: windowShadeTiltLevel
     version: 1
+  - id: statelessWindowShadeLevelStep
+    version: 1
```
hcarter-775, line 12: I might recommend moving the `statelessWindowShadeLevelStep` capability just below the `windowShadeLevel` capability in all cases, for consistency. A few of these profile updates have the new capability in different spots.

hcarter-775:
Hi, I just had the workflows run and there are some warnings showing up. Can you fix these warnings as well? Thanks!

drivers/SmartThings/matter-window-covering/src/init.lua
partner-hub-chenxuelong: The purpose of `LATEST_TARGET_LEVEL` is to record the target value of the latest step command, supporting continuous stepping operations.

Example Scenario:

1. Initial position of the curtain: level = 30

2. Received +10 command → Target value = 30 + 10 = 40, set LATEST_TARGET_LEVEL = 40

3. The curtain starts moving, currently at level = 35 (still moving)

4. Received +10 command again

- Using LATEST_TARGET_LEVEL(40) as the baseline → New target value = 40 + 10 = 50 ✓

- If there is no LATEST_TARGET_LEVEL, the current actual position (35) will be used → New target value = 35 + 10 = 45 ✗

This ensures that during continuous steps, each adjustment is based on the previous target,

rather than based on the intermediate position during the movement, achieving the cumulative stepping effect expected by statelessWindowShadeLevelStep.

drivers/SmartThings/matter-window-covering/src/init.lua
partner-hub-chenxuelong: If the clearing operation is not performed, if the user manually pulls the curtain or sends the setShadeLevel command to change the curtain position within 30 seconds before the timer triggers, the baseline value (in line 194) will be an expired target value when the step command is sent subsequently, instead of the actual position of the curtain, causing the curtain to move to the wrong position.

drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-only-battery.yml
hcarter-775: this has been done. thanks!

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +186,50 @@ local function handle_shade_level(driver, device, cmd)
   device:send(req)
 end
 
+-- window shade step control handler
```
hcarter-775: appreciate the comments, though I think we can remove this one

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +186,50 @@ local function handle_shade_level(driver, device, cmd)
...
+local function handle_step_shade_level(driver, device, cmd)
+  local step = cmd.args.stepSize
+
+  -- Priority: use target_level if exists, otherwise use latest state
+  local target_level_field = device:get_field(LATEST_TARGET_LEVEL)
+  local current_level = target_level_field or
+    device:get_latest_state("main", capabilities.windowShadeLevel.ID,
+      capabilities.windowShadeLevel.shadeLevel.NAME) or 0
+
+  -- Calculate new target (user level: 0-100, 0=closed, 100=open)
```
hcarter-775: remove this one as well

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +186,50 @@ local function handle_shade_level(driver, device, cmd)
...
+  local target_level = utils.clamp_value(current_level + step, 0, 100)
+
+  -- Update tracking state
```
hcarter-775: remove this one as well

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +186,50 @@ local function handle_shade_level(driver, device, cmd)
...
+  device:set_field(LATEST_TARGET_LEVEL, target_level)
+
+  -- Cancel previous timeout timer if exists
```
hcarter-775: remove the timer comments, or bring them together into 1

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +186,50 @@ local function handle_shade_level(driver, device, cmd)
...
+  local old_timer = device:get_field(TARGET_LEVEL_TIME_OUT)
+  if old_timer ~= nil then
+    device.thread:cancel_timer(old_timer)
+  end
```
hcarter-775, line 198:
```suggestion
device.thread:cancel_timer(old_timer)
```
nit: we don't really need the check here

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +186,50 @@ local function handle_shade_level(driver, device, cmd)
...
+
+  -- Set 30 second timeout timer to ensure target_level is cleared
+  local timer = device.thread:call_with_delay(TARGET_LEVEL_TIME_OUT_SECONDS, function(d)
+    device:set_field(LATEST_TARGET_LEVEL, nil)
+    device:set_field(TARGET_LEVEL_TIME_OUT, nil)
```
hcarter-775: I don't think you actually need to explicitly make this nil. It will happen on its own once the timer completes

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +186,50 @@ local function handle_shade_level(driver, device, cmd)
...
+  else
+    lift_percentage_value = 100 - target_level
+  end
+  local hundredths_lift_percentage = lift_percentage_value * 100
+
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local req = clusters.WindowCovering.server.commands.GoToLiftPercentage(
+    device, endpoint_id, hundredths_lift_percentage
+  )
+  device:send(req)
```
hcarter-775:
```suggestion
  device:send(clusters.WindowCovering.server.commands.GoToLiftPercentage(
    device, endpoint_id, hundredths_lift_percentage
  ))
```

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,6 +186,50 @@ local function handle_shade_level(driver, device, cmd)
...
 
+-- window shade step control handler
+local function handle_step_shade_level(driver, device, cmd)
+  local step = cmd.args.stepSize
+
+  -- Priority: use target_level if exists, otherwise use latest state
+  local target_level_field = device:get_field(LATEST_TARGET_LEVEL)
+  local current_level = target_level_field or
+    device:get_latest_state("main", capabilities.windowShadeLevel.ID,
+      capabilities.windowShadeLevel.shadeLevel.NAME) or 0
```
hcarter-775:
```suggestion
  local current_target_level = device:get_field(LATEST_TARGET_LEVEL) or
    device:get_latest_state("main", capabilities.windowShadeLevel.ID,
      capabilities.windowShadeLevel.shadeLevel.NAME) or 0
```
I would remove the comment, as this is fairly clear from the logic

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -139,6 +146,7 @@ end
 
 -- close covering
 local function handle_close(driver, device, cmd)
+  device:set_field(LATEST_TARGET_LEVEL, nil)
```
hcarter-775: I don't think we actually have to do this, and I think adding this to the top of every command overcomplicates things.

Imagine that a close command comes in while we are still doing the stateless step logic. Either we will send a command in the opposite direction that overrides the last commands, or we will send a command in the same direction that will simply pass the "target value" and keep going.

Effectively, the target value field is not a "busy" state. Other things can happen without getting in the way.

---

The one rare case I can think of would be, if the user were scrolling "up" to 50, but then the device received a command to close (whether via ST or any other method), but then we started scrolling again before the 30 seconds were up, we might send the wrong command, since the scrolling would assume we were still trying to get to 50.

However, this would only be an issue if we were going the opposite way (scrolling up and then get a down command or vice versa). If we got a command to go up after scrolling up this would not be an issue. Therefore, to catch this case, I would suggest we try to nil this field whenever we are switching directions. We already do this if the device stops (returns 0), which I think makes sense and should catch most cases.

hcarter-775 (dismissed):

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -253,6 +324,17 @@ local function current_status_handler(driver, device, ib, response)
...
     device:emit_event_for_endpoint(ib.endpoint_id, reverse and windowShade.opening() or windowShade.closing())
+  elseif state == 0 then -- stopped (idle)
+    -- Clear target_level when device stops moving
+    local target_level_field = device:get_field(LATEST_TARGET_LEVEL)
+    if target_level_field then
+      device:set_field(LATEST_TARGET_LEVEL, nil)
+      local timer = device:get_field(TARGET_LEVEL_TIME_OUT)
+      if timer ~= nil then
+        device.thread:cancel_timer(timer)
+        device:set_field(TARGET_LEVEL_TIME_OUT, nil)
```
hcarter-775:
```suggestion
      local timer = device:get_field(TARGET_LEVEL_TIME_OUT)
      if timer ~= nil then
        device.thread:cancel_timer(timer)
```
nit: simplify

drivers/SmartThings/matter-window-covering/src/init.lua
hcarter-775: oops, this is not true. Not sure what I was thinking!

drivers/SmartThings/matter-window-covering/src/init.lua
hcarter-775: I think the issue here is rather that the timeout is too long. We should probably just make the timeout shorter, since all that the timer is doing is affecting the step-level functionality.

I propose we make the timer ~5 seconds or so, and then remove this logic altogether

drivers/SmartThings/matter-window-covering/src/init.lua
partner-hub-chenxuelong: We have completed the modifications.

drivers/SmartThings/matter-window-covering/src/init.lua
zhangshen001: We believe it's better to retain this part, as it makes the logic more rigorous and avoids some extreme cases.

drivers/SmartThings/matter-window-covering/profiles/window-covering-tilt-battery.yml
zhangshen001: this has been done.

drivers/SmartThings/matter-window-covering/src/init.lua
zhangshen001: this has been done.please review

drivers/SmartThings/matter-window-covering/src/init.lua
zhangshen001: this has been done.please review

drivers/SmartThings/matter-window-covering/src/init.lua
zhangshen001: this has been done.please review

drivers/SmartThings/matter-window-covering/src/init.lua
zhangshen001: this has been done.please review

drivers/SmartThings/matter-window-covering/src/init.lua
zhangshen001: This avoids some extreme situations.

drivers/SmartThings/matter-window-covering/src/init.lua
zhangshen001: this has been done.please review

drivers/SmartThings/matter-window-covering/src/init.lua
zhangshen001: Determining whether it's a reverse operation makes the logic more complex.

drivers/SmartThings/matter-window-covering/src/init.lua
zhangshen001: this has been done.please review

zhangshen001:
The statelessWindowShadeLevelStep capability is newly introduced in version 62 (dev), so the driver tests on versions 59 to 61 have failed.

drivers/SmartThings/matter-window-covering/src/init.lua
hcarter-775: We should not add this. It is unnecessary and adds significantly more state to track across multiple access points. If there is a real issue that you have seen that exists without this change, we should prioritize the work to handle this generically, as this does not (and cannot) handle any manual operations.

drivers/SmartThings/matter-window-covering/src/init.lua
hcarter-775: I propose we make the timeout shorter, to more like 10 or 15 seconds.

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -175,15 +186,51 @@ local function handle_shade_level(driver, device, cmd)
...
+  else
+    lift_percentage_value = 100 - target_level
+  end
+  local hundredths_lift_percentage = lift_percentage_value * 100
+
+  local endpoint_id = device:component_to_endpoint(cmd.component)
+  local req = clusters.WindowCovering.server.commands.GoToLiftPercentage(
+    device, endpoint_id, hundredths_lift_percentage
+  )
+  device:send(req)
```
hcarter-775:
```suggestion
    driver:inject_capability_command(device, {
      capability = capabilities.windowShadeLevel.ID,
      component = command.component,
      command = capabilities.windowShadeLevel.commands.setShadeLevel.NAME,
      named_args = { shadeLevel = target_level }
    })
```
This will allow the handler to support the closure device type as well. This also makes me realize that the current `setShadeLevel` handlers (here and in the closure subdriver) do not support "reverse". We must add this with this implementation. Thanks

drivers/SmartThings/matter-window-covering/src/init.lua
partner-hub-chenxuelong: Okay, we'll change it to 15 seconds.

drivers/SmartThings/matter-window-covering/src/init.lua
partner-hub-chenxuelong: We have removed the complex logic that sets LATEST_TARGET_LEVEL at the top of every command

LATEST_TARGET_LEVEL is now cleared only in three cases:

1. When the device reaches the target position (within ±1% tolerance)
2. When the curtain stops moving (OperationalStatus = stopped)
3. After the 15-second timeout expires

drivers/SmartThings/matter-window-covering/src/init.lua
partner-hub-chenxuelong: We've added reverse polarity support to `handle_shade_level`.

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
```diff
@@ -787,4 +782,339 @@ test.register_coroutine_test(
   {test_init = test_init_door}
 )
 
+-- Test: handle_level command for door type
```
hcarter-775: Hi, I think a lot of unrelated unit tests have been added to this file, related to various utilities of the closures subdriver. I think these are outside the scope of this PR and should be removed for now.

I also do not see any unit tests directly related to the stateless step handling, which would be the tests within scope that we'd need.

partner-hub-chenxuelong: Okay, tests irrelevant to this PR change have been removed.

partner-hub-chenxuelong:
TC has been updated. Please recalculate the coverage

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
hcarter-775: Hi @partner-hub-chenxuelong , perhaps I'm not seeing something, but that does not appear to be true? Would you mind checking this again?

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -1228,3 +1218,1334 @@ test.register_coroutine_test(
...
+      }
+    )
+    mock_device:expect_metadata_update({ profile = "window-covering-tilt-battery" })
+  end,
+  {
+     min_api_version = 17
+  }
+)
+
+-- Test power_source_attribute_list_handler - BatChargeLevel present
```
hcarter-775: we have way too many tests added in this file as well. Please prune these to only the relevant files to this PR

drivers/SmartThings/matter-window-covering/src/init.lua
partner-hub-chenxuelong: This PR also includes reverse polarity support for handling_shade_level, so some tests will focus on this aspect.

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
partner-hub-chenxuelong: This PR also includes reverse polarity support for handling_shade_level, so some tests will focus on this aspect.

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
partner-hub-chenxuelong: We re-examined these two files (test_matter_closure.lua, test_matter_window_covering.lua), removing unnecessary tests and keeping only those related to stepShadeLevel and reverse polarity.

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
partner-hub-chenxuelong: We re-examined these two files (test_matter_closure.lua, test_matter_window_covering.lua), removing unnecessary tests and keeping only those related to stepShadeLevel and reverse polarity.

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -900,16 +894,6 @@ test.register_coroutine_test(
   }
 )
 
-test.register_coroutine_test(
-  "Test mains powered device does not switch to battery profile",
-  function()
-  end,
-  {
```
hcarter-775, line 907: there is no need to remove this test

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -1171,7 +1158,10 @@ test.register_coroutine_test(
 
 test.register_coroutine_test(
   "Check that preference updates to reverse polarity after being set to true and that the shade tilt operates as expected when opening and closing", function()
-    test.socket.device_lifecycle():__queue_receive(mock_device:generate_info_changed({ preferences = { reverse = "true" } }))
+    test.socket.device_lifecycle():__queue_receive(mock_device:generate_info_changed({
+      old_st_store = { profile = { name = "window-covering-tilt-battery" }, preferences = { reverse = false } },
+      preferences = { reverse = "true" }
+    }))
```
hcarter-775: I am not sure why a profile update is required to occur here?

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -1113,7 +1097,10 @@ test.register_coroutine_test(
 
 test.register_coroutine_test(
   "Check that preference updates to reverse polarity after being set to true and that the shade lift operates as expected when opening and closing", function()
-    test.socket.device_lifecycle():__queue_receive(mock_device:generate_info_changed({ preferences = { reverse = "true" } }))
+    test.socket.device_lifecycle():__queue_receive(mock_device:generate_info_changed({
+      old_st_store = { profile = { name = "window-covering-tilt-battery" }, preferences = { reverse = false } },
+      preferences = { reverse = "true" }
+    }))
```
hcarter-775: I am not sure why a profile update needs to happen here?

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -1228,3 +1218,413 @@ test.register_coroutine_test(
...
+      {mock_device.id, WindowCovering.server.commands.GoToLiftPercentage(mock_device, 10, 10000)}
+    )
+  end,
+  {
+     min_api_version = 19
+  }
+)
+
+test.register_coroutine_test(
+  "WindowShade stepShadeLevel - other commands clear target marker", function()
```
hcarter-775: remove this test, since this no longer occurs.

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -1228,3 +1218,413 @@ test.register_coroutine_test(
...
+    )
+  end,
+  {
+     min_api_version = 19
+  }
+)
+
+-- Test handle_close with reverse polarity
+test.register_coroutine_test(
+  "WindowShade close cmd with reverse polarity true",
```
hcarter-775: remove this test, it's not in the scope of this PR and is tested elsewhere

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -1228,3 +1218,413 @@ test.register_coroutine_test(
...
+    )
+  end,
+  {
+     min_api_version = 17
+  }
+)
+
+-- Test handle_open with reverse polarity
+test.register_coroutine_test(
+  "WindowShade open cmd with reverse polarity true",
```
hcarter-775: remove this test, it's not in the scope of this PR and is tested elsewhere

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -1228,3 +1218,413 @@ test.register_coroutine_test(
...
+      {mock_device.id, WindowCovering.server.commands.GoToLiftPercentage(mock_device, 10, 5000)}
+    )
+  end,
+  {
+     min_api_version = 19
+  }
+)
+
+test.register_coroutine_test(
+  "WindowShade stepShadeLevel - target reached clears target marker", function()
```
hcarter-775, line 1327: remove this test, since this no longer occurs.

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -1228,3 +1218,413 @@ test.register_coroutine_test(
...
+    )
+  end,
+  {
+     min_api_version = 17
+  }
+)
+
+-- Test current_pos_handler - reverse polarity state handling
+test.register_coroutine_test(
+  "CurrentPosition - reverse polarity affects state interpretation",
```
hcarter-775: remove this test, it's not in the scope of this PR and is tested elsewhere

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
```diff
@@ -1228,3 +1218,413 @@ test.register_coroutine_test(
...
+    )
+  end,
+  {
+     min_api_version = 17
+  }
+)
+
+-- Test current_status_handler - stopped clears target level
+test.register_coroutine_test(
+  "OperationalStatus stopped clears target level field",
```
hcarter-775: remove this test, since this no longer occurs.

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
```diff
@@ -787,4 +782,69 @@ test.register_coroutine_test(
...
+    })
+    -- With reverse=true, open should send MOVE_TO_FULLY_CLOSED
+    test.socket.matter:__expect_send({
+      mock_device.id,
+      clusters.ClosureControl.server.commands.MoveTo(
+        mock_device, 10, clusters.ClosureControl.types.TargetPositionEnum.MOVE_TO_FULLY_CLOSED
+      )
+    })
+  end
+)
```
hcarter-775: Remove these tests, they are outside the scope of this PR

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
```diff
@@ -787,4 +782,69 @@ test.register_coroutine_test(
...
+      mock_device.id,
+      {capability = "windowShadeLevel", component = "windowShade1", command = "setShadeLevel", args = {75}},
+    })
+    -- With reverse=true, 75 should be inverted to 25
+    test.socket.matter:__expect_send({
+      mock_device.id,
+      clusters.ClosureDimension.server.commands.SetTarget(mock_device, 11, 25 * 100)
+    })
+  end
+)
```
hcarter-775: Let's add a test for this _without_ reverse polarity.

drivers/SmartThings/matter-window-covering/src/init.lua
```diff
@@ -167,23 +173,57 @@ end
 -- move to shade level between 0-100
 local function handle_shade_level(driver, device, cmd)
   local endpoint_id = device:component_to_endpoint(cmd.component)
-  local lift_percentage_value = 100 - cmd.args.shadeLevel
+  local lift_percentage_value
+  if device:get_field(REVERSE_POLARITY) then
+    lift_percentage_value = cmd.args.shadeLevel
+  else
+    lift_percentage_value = 100 - cmd.args.shadeLevel
+  end
```
hcarter-775: Hi all, I know I suggested this originally, but I have looked into it further and I am no longer comfortable making this change to the "reverse" handling here. I believe we should leave this as it was before for the time being, and if we decide that the current behavior is not good handling, we should add this stuff back later.

My primary issue is that I do not want this PR to possibly screw up the current reverse handling without understanding deeply why it was done like this in the first place. I suggest we remove the reverse handling in the closures subdriver as well. Thanks for understanding

hcarter-775 (changes requested):

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
partner-hub-chenxuelong: ok, we will restore this

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
partner-hub-chenxuelong: Okay, we've removed this tset.

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
partner-hub-chenxuelong: Okay, we've removed the reverse handling and its testing.

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
partner-hub-chenxuelong: Okay, we've removed the reverse handling and its testing.

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
partner-hub-chenxuelong: Okay, we've removed the reverse handling and its testing.

drivers/SmartThings/matter-window-covering/src/init.lua
partner-hub-chenxuelong: Okay, we've removed the reverse handling and its testing.

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
partner-hub-chenxuelong: Okay, we've removed this tset.

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
partner-hub-chenxuelong: It was originally added to test reserve polarity, but has now been removed.

drivers/SmartThings/matter-window-covering/src/test/test_matter_window_covering.lua
partner-hub-chenxuelong: It was originally added to test reserve polarity, but has now been removed.

hongye-samsung:
We've fixed the issue and re-pushed the code. Please approve the check workflow.

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
hcarter-775: I see that this has not been taken care of. I think a test or two should be added in a follow-up PR for closures @hongye-samsung. Thanks

hcarter-775 (approved):
LGTM.

hcarter-775:
There are 2 failing tests here, un-related to this PR. I cannot rebase this PR over the outside fix, so I will merge this as-is, as the fix will be applied in the main branch.

drivers/SmartThings/matter-window-covering/src/test/test_matter_closure.lua
partner-hub-chenxuelong: The test for setShadeLevel without reverse polarity already exists (line 485-498 in test_matter_closure.lua). If you feel that test is not sufficient, we can add more in a follow-up PR.

hcarter-775: Ah, I see. Apologies, I think there was a miscommunication. I think we need some unit tests for `stepShadeLevel` for closures.

partner-hub-chenxuelong: we will add it in a follow-up PR.

PR #3072, opened 2026-06-26
Hue: Fix conditional in hue command error logging

hcarter-775:
# Description of Change
Fix Hue conditional for logging command response errors.

### DIFF
--- a/drivers/SmartThings/philips-hue/src/handlers/commands.lua
+++ b/drivers/SmartThings/philips-hue/src/handlers/commands.lua
@@ -54,13 +54,14 @@ end
 ---@param err string? Error message returned from the Hue API call, if any
 ---@param action_desc string Description of the action being performed, for logging purposes
 local function log_command_response_errors(response, err, action_desc)
-  if not response or (response.errors and #response.errors == 0) then
-    if err ~= nil then
-      log.error_with({ hub_logs = true }, "Error performing " .. action_desc .. ": " .. err)
-    elseif response and #response.errors > 0 then
-      for _, error in ipairs(response.errors) do
-        log.error_with({ hub_logs = true }, "Error returned in Hue response for " .. action_desc .. ": " .. error.description)
-      end
+  if err ~= nil then
+    log.error_with({ hub_logs = true }, "Error performing " .. action_desc .. ": " .. err)
+  end
+
+  if response and response.errors and #response.errors > 0 then
+    for _, error in ipairs(response.errors) do
+      log.error_with({ hub_logs = true },
+        "Error returned in Hue response for " .. action_desc .. ": " .. error.description)
     end
   end
 end


drivers/SmartThings/philips-hue/src/handlers/commands.lua
```diff
@@ -54,7 +54,7 @@ end
 ---@param err string? Error message returned from the Hue API call, if any
 ---@param action_desc string Description of the action being performed, for logging purposes
 local function log_command_response_errors(response, err, action_desc)
-  if not response or (response.errors and #response.errors == 0) then
+  if not response or (response.errors and #response.errors > 0) then
```
ctowns: What about something like this? This way the err from the Hue API call is logged separately regardless of the errors from the response. That way it is still logged even if we don't have a response, which could be due to an error.
```
 if err ~= nil then
    log.error_with({ hub_logs = true }, "Error performing " .. action_desc .. ": " .. err)
  end

  if response and response.errors and #response.errors > 0 then
    for _, error in ipairs(response.errors) do
      log.error_with({ hub_logs = true },
        "Error returned in Hue response for " .. action_desc .. ": " .. error.description)
    end
  end
```

hcarter-775: works for me! can you just merge that on top of this branch? I don't mind

ctowns: just amended the last commit with these changes and force pushed :+1:

hcarter-775: well I will say that if we dont have a response, we'll trigger `not response` as-is, right?

ctowns: Yeah, but that conditional is just a combination of the two conditionals within it, right? It seems unnecessary to me.

hcarter-775: ah sure, yeah that's true. ok yeah I'm back to being like this is good

ctowns (approved):
don't mind me just approving my own code

drivers/SmartThings/philips-hue/src/handlers/commands.lua
```diff
@@ -54,13 +54,14 @@ end
...
-    elseif response and #response.errors > 0 then
-      for _, error in ipairs(response.errors) do
-        log.error_with({ hub_logs = true }, "Error returned in Hue response for " .. action_desc .. ": " .. error.description)
-      end
+  if err ~= nil then
+    log.error_with({ hub_logs = true }, "Error performing " .. action_desc .. ": " .. err)
+  end
+
+  if response and response.errors and #response.errors > 0 then
+    for _, error in ipairs(response.errors) do
```
cjswedes, line 62: Is ipairs correct here? Another option would be to use utils.stringify_table on the errors and log them all in a blob. Its fine as is as long as ipairs is correct.

cjswedes (approved):

drivers/SmartThings/philips-hue/src/handlers/commands.lua
hcarter-775: Yes, ipairs is correct.

PR #3088, opened 2026-07-02
Matter Window Covering: Add embedded clusters for Descriptor, Global clusters

hcarter-775:
# Description of Change
Add embedded cluster definitions for Descriptor and Global clusters. Update usages of these in tests and in get_endpoints calls. Use correct can_handle path for all subdrivers in case the v1 lazy loading is invoked.

# Summary of Completed Tests
Tests continue to pass as expected in lower firmwares as well as the latest.


cjswedes (approved):

drivers/SmartThings/matter-window-covering/src/sub_drivers/closure/closure_handlers/attribute_handlers.lua
```diff
@@ -63,6 +64,10 @@ end
 function ClosureAttrHandlers.tag_list_handler(driver, device, ib, response)
   if not ib.data.elements then return end
   local tag_value
+  if version.api < 16 then
+    clusters.Global.types.SemanticTagStruct:augment_type(ib.data)
+  end
+  print("TagList attribute value: ", ib.data)
```
ctowns: nit: remove

drivers/SmartThings/matter-window-covering/src/sub_drivers/closure/closure_handlers/attribute_handlers.lua
```diff
@@ -8,6 +8,7 @@ local version = require "version"
 if version.api < 20 then
   clusters.ClosureControl = require "embedded_clusters.ClosureControl"
   clusters.ClosureDimension = require "embedded_clusters.ClosureDimension"
+  clusters.Global = require "embedded_clusters.Global"
```
ctowns, line 11: Shouldn't Global be included if API is <16?

hcarter-775: maybe... the Closure clusters depend on Global too so it seemed cleaner to go this way.

NoahCornell (approved):

aleclorimer (approved):

drivers/SmartThings/matter-window-covering/src/sub_drivers/closure/closure_handlers/attribute_handlers.lua
hcarter-775: I actually think we have to do it this way. Since if we're under v16 the global type won't exist, we'll need the embedded closure clusters to reference the embedded globals. Then, if we're just under v20, those references will still point to an embedded global type. So unless we added version gating in the embedded defs (which we shouldn't do) I think we'll need to have it like this.

ctowns (approved):

PR #3101, opened 2026-07-07
Default handler missing log require patch

aleclorimer:
Quick patch of a missing require in the lua libs requiring some zigbee-switch drivers to crash. Coppied over the fixed/updated default handler for powerMeter instantaneous demand.


varzac (approved):

cjswedes (approved):
When we do a monkey patch like this to fix a driver with something a future lua libs will have fixed, I think we should include a comment about why it is included, and when it is okay to remove it from the driver so that it is easier to maintain in the future if were looking at it.

aleclorimer:
> we should include a comment about why it is included

If we choose to go this route ill update with a comment with this.

PR #3103, opened 2026-07-08
zigbee-window-treatment: Don't halve battery percentage for IKEA devices

### LINKED ISSUE #2679: zigbee-window-treatment: IKEA blinds report BatteryPercentageRemaining as 0–100, but driver halves it (battery shows ~50%)
## Summary
When using the stock **SmartThings zigbee-window-treatment** Edge driver, IKEA blinds show a battery percentage that is *exactly cut in half* (e.g., a freshly charged battery shows ~50% instead of ~100%). Switching to Mariano’s community driver shows the correct battery percentage.

This makes the built-in driver feel “partially supported” for IKEA blinds, even though they are now matched by the stock driver.

## Affected devices
- Manufacturer: `IKEA of Sweden`
- Device: IKEA smart blinds / window treatments (e.g., FYRTUR/KADRILJ/TREDANSEN)

## Steps to reproduce
1. Pair an IKEA blind to a SmartThings hub.
2. Assign it to the stock driver: `SmartThings / zigbee-window-treatment` (default/official driver).
3. Fully charge the IKEA battery pack and reinsert it (or otherwise ensure the blind reports a “full” battery).
4. Check the battery percentage in the SmartThings app.

## Actual behavior
Battery is displayed at ~50% when it should be ~100% (i.e., it appears halved).

## Expected behavior
Battery is displayed correctly (e.g., ~100% after a full charge).

## Analysis / likely root cause
Zigbee `PowerConfiguration -> BatteryPercentageRemaining (0x0001 / 0x0021)` is commonly interpreted as **0.5% units** (0–200), so many handlers divide by 2 to convert to 0–100%.

However, IKEA blinds appear to report `BatteryPercentageRemaining` already in **1% units** (0–100). If the handler divides by 2 anyway, the displayed value becomes half of the real percentage.

The stock driver currently registers default handlers (including Battery) without an IKEA-specific exception:
- Stock driver init.lua: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/tree/main/drivers/SmartThings/zigbee-window-treatment/src

Mariano’s driver fixes this by **skipping the divide-by-2 for IKEA**:
```lua
local function battery_perc_attr_handler(driver, device, value, zb_rx)
  if device:get_manufacturer() ~= "IKEA of Sweden" then
    value.value = math.floor(value.value / 2.0 + 0.5)
  end
  device:emit_event_for_endpoint(zb_rx.address_header.src_endpoint.value,
    capabilities.battery.battery(value.value))
end
```
## Source

Mariano’s IKEA subdriver (reference implementation):

https://github.com/Mariano-Github/Edge-Drivers-Beta/blob/7e27b19016354ecb640e73cc69dc70ea6f71117f/zigbee-window-treatment/src/IKEA/init.lua#L82-L87

## Proposed fix

Add IKEA-specific battery handling in the stock `zigbee-window-treatment` driver, for example:

- Add a small **IKEA subdriver** that overrides the attribute handler for:
  - `PowerConfiguration.attributes.BatteryPercentageRemaining`
- For `device:get_manufacturer() == "IKEA of Sweden"`, treat `value.value` as already **0–100** and **do not halve it** (still clamp to 0–100).
- Keep the existing/default behavior for other manufacturers/devices.
- Alternative approach: implement a manufacturer/model exception directly in the driver’s battery handling, but a subdriver keeps the logic isolated and consistent with other vendor quirks.

## Related discussion

SmartThings Community thread:
- https://community.smartthings.com/t/smartthings-shades-window-treatment-driver-bug/307114

pachuho:
This addresses #2679.

IKEA blinds (FYRTUR/KADRILJ) report `BatteryPercentageRemaining` in 1% units (0–100) rather than the 0.5% units defined by the ZCL spec, so the default handler halves the value and a fully charged battery shows ~50% in the app.

This adds an IKEA subdriver that emits the reported value as-is. The repo already handles IKEA battery reports this way in two other drivers — `zigbee-button` (`zigbee-multi-button/ikea/init.lua`) and `zigbee-dimmer-remote` (`IKEAofSweden/init.lua`) — so this brings `zigbee-window-treatment` in line with that existing convention.

The subdriver matches on manufacturer only, mirroring the `zigbee-button` precedent; both IKEA entries in this driver's `fingerprints.yml` report `IKEA of Sweden`.

Added battery test cases to `test_zigbee_window_shade_battery_ikea.lua` (which previously only covered window shade states). All zigbee-window-treatment tests pass locally against `lua_libs-api_v20`.

Fixes #2679

### DIFF
--- a/drivers/SmartThings/zigbee-window-treatment/src/ikea/can_handle.lua
+++ b/drivers/SmartThings/zigbee-window-treatment/src/ikea/can_handle.lua
@@ -1,0 +1,11 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local can_handle_ikea = function(opts, driver, device)
+  if device:get_manufacturer() == "IKEA of Sweden" then
+    return true, require("ikea")
+  end
+  return false
+end
+
+return can_handle_ikea
--- a/drivers/SmartThings/zigbee-window-treatment/src/ikea/init.lua
+++ b/drivers/SmartThings/zigbee-window-treatment/src/ikea/init.lua
@@ -1,0 +1,30 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
+local clusters = require "st.zigbee.zcl.clusters"
+local utils = require "st.utils"
+local PowerConfiguration = clusters.PowerConfiguration
+
+-- IKEA devices report BatteryPercentageRemaining in 1% units (0-100)
+-- rather than the 0.5% units (0-200) defined by the ZCL spec, so the
+-- default handler would halve the reported percentage. Emit the raw
+-- value instead, matching the IKEA handling in the zigbee-button and
+-- zigbee-dimmer-remote drivers.
+local function battery_perc_attr_handler(driver, device, value, zb_rx)
+  device:emit_event(capabilities.battery.battery(utils.clamp_value(value.value, 0, 100)))
+end
+
+local ikea_window_treatment = {
+  NAME = "ikea window treatment",
+  zigbee_handlers = {
+    attr = {
+      [PowerConfiguration.ID] = {
+        [PowerConfiguration.attributes.BatteryPercentageRemaining.ID] = battery_perc_attr_handler
+      }
+    }
+  },
+  can_handle = require("ikea.can_handle"),
+}
+
+return ikea_window_treatment
--- a/drivers/SmartThings/zigbee-window-treatment/src/sub_drivers.lua
+++ b/drivers/SmartThings/zigbee-window-treatment/src/sub_drivers.lua
@@ -16,5 +16,6 @@ local sub_drivers = {
    lazy_load_if_possible("screen-innovations"),
    lazy_load_if_possible("VIVIDSTORM"),
    lazy_load_if_possible("HOPOsmart"),
+   lazy_load_if_possible("ikea"),
 }
 return sub_drivers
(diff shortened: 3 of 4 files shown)


tpmanley:
Thanks for submitting the PR. As noted on the Github issue, the battery reporting behavior does seem to vary by FW version, so we'll want to incorporate that into the fix so that devices that follow the ZCL spec continue to have their battery level reported correctly. I'm trying to figure out exactly what versions have the issue, and then can either provide that info to you if you want to add to your PR or we can put up a fix for it.

ldeora:
>Starting in firmware 24.x it is fixed and the battery reading is correct.

https://community.smartthings.com/t/smartthings-shades-window-treatment-driver-bug/307114/8

tpmanley:
> > Starting in firmware 24.x it is fixed and the battery reading is correct.
> 
> https://community.smartthings.com/t/smartthings-shades-window-treatment-driver-bug/307114/8

I have seen some versions before 24.x that seem to report correctly as well ,so I'm checking with the manufacturer to determine exactly what firmware versions have the issue.

mocelet:
> behavior does seem to vary by FW version, so we'll want to incorporate that into the fix

@tpmanley Is it now possible to check the FW version in a Zigbee driver? 

It used to be possible, I used it for my custom drivers for IKEA buttons that had the same issue with old/new firmwares but eventually the feature was removed ([see this thread at the Community](https://community.smartthings.com/t/best-way-to-check-zigbee-device-firmware-version-in-do-configure/295173/4?u=mocelet)) and had to resort to a manual setting by the user which is a bad user experience when it used to be automatic.

PR #3105, opened 2026-07-08
Fix error resubscribe

KyuminAhn:
- Fix re-subscription timeout after hub replacement
- fingerprints.yml: add matterManufacturer entries (AIHomeKit 0x1241/0x804E, BritzyHub 0x1436/0x800C) and tag generic device types with vendorId

Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests


drivers/SinuxSoft/britzyhub/src/init.lua
```diff
@@ -4,7 +4,14 @@
 local MatterDriver = require "st.matter.driver"
 local log = require "log"
 
+local function bridge_init(driver, device)
```
ctowns, line 7: I wonder if the issue is that there was no subscription being sent my the bridge device if this `bridge_init` function was not previously populated?

@DongHoon-Ryu could you check if just removing this function causes the issue to occur again during hub replace? And then if adding it back fixes the issue?

The reason I am more suspicious of this is because the other subscribe logic above is very similar to what already exists in the subscribe function, so I'm not sure yet how that is providing a change in functionality.

KyuminAhn: I ran the test you asked for.                                                                                                                                                                                                                                                                                  
With bridge_init removed and the rest of the changes left in place, after a hub replace:                                                                                                                                                                                                                       
- Toggling on/off from the device list screen in the SmartThings app still works.                                                                                                                                                                                                                              
- But opening an individual device's detail view and toggling on/off fails with a connection error 
- ("There was a problem with the connection" - you can see the pic attached 
<img width="1080" height="2280" alt="Screenshot_20260717_211910_SmartThings" src="https://github.com/user-attachments/assets/aa21617b-c41c-4afd-a47a-429fb5fc1875" />
).

hcarter-775: HI @KyuminAhn, would you mind providing hub logs around the time of the error you are seeing in the app? Thank you

DongHoon-Ryu: @hcarter-775 Please check MTR-1083. If additional logs are needed, please leave a comment on the ticket.

ctowns: Thank you for testing this!

Based on this, I believe this shows that the part of these changes that were fixing this issue are isolated to this file. Therefore, to move forward, I propose that we keep the changes in these files:
* drivers/SinuxSoft/britzyhub/src/init.lua
* drivers/SinuxSoft/britzyhub/fingerprints.yml
* drivers/SinuxSoft/britzyhub/profiles/matter-bridge.yml

But I do not believe the changes in the sub drivers are neccessary, so we could remove the changes in the remaining files:
* drivers/SinuxSoft/britzyhub/src/elevator/init.lua
* drivers/SinuxSoft/britzyhub/src/gas-valve/init.lua
* drivers/SinuxSoft/britzyhub/src/vent/init.lua
* drivers/SinuxSoft/britzyhub/src/utils.lua

My comment on the ticket has more detail. I believe if we adjust the changes as I have mentioned above and re-test, we should see that the subscription is now handled properly without the need to add additional sub driver changes. Please let me know if you have any questions!

DongHoon-Ryu: @ctowns Problems continue to occur when only the files you suggested are modified. Please refer to the comments in MTR-1083.

drivers/SinuxSoft/britzyhub/src/utils.lua
```diff
@@ -0,0 +1,33 @@
...
+-- Licensed under the Apache License, Version 2.0
+
+local im = require "st.matter.interaction_model"
+local log = require "log"
+
+local utils = {}
+
+-- Sends subscription request directly via device:send(), following the matter-switch pattern.
+-- The default device:subscribe() reuses cached sessions, which causes timeout after hub
+-- replacement because the hub Node ID changes and all sessions expire.
```
ctowns, line 11: I am surprised that we have not seen this issue in other devices during hub replace, but overall the changes make sense and I think we can move forward. I am still not confident in the root cause, but these changes are well understood and follow along with other examples (see our default lua libs handling, or see how subscriptions are handled the matter-switch driver for a comparison). So, I think we can move forward with these changes since they will resolve the issue now, but we should try to better understand the root cause and investigate as we move on.

ctowns (approved):

PR #3111, opened 2026-07-09
Aligned handlers and fields in value magnitude and added tests

aleclorimer:
Fixed a bug where `(min|max)(Heat|Cool)SetPoint` was being set in centidegrees and then being used to clamp a value in degrees causing the clamped value to have an Int16 overflow error when then tried to scale to centidegrees.

Mutant testing: Failing without the `/ 100.0` fix/update.

### DIFF
--- a/drivers/SmartThings/zigbee-thermostat/src/test/test_zenwithin_thermostat.lua
+++ b/drivers/SmartThings/zigbee-thermostat/src/test/test_zenwithin_thermostat.lua
@@ -523,57 +523,61 @@ test.register_coroutine_test(
     }
 )
 
-test.register_message_test(
+test.register_coroutine_test(
     "MinHeatSetpointLimit report stores device field",
-    {
-      {
-        channel = "zigbee",
-        direction = "receive",
-        message = { mock_device.id, Thermostat.attributes.MinHeatSetpointLimit:build_test_attr_report(mock_device, 500) }
-      }
-    },
+    function ()
+      test.socket.zigbee:__queue_receive({
+        mock_device.id,
+        Thermostat.attributes.MinHeatSetpointLimit:build_test_attr_report(mock_device, 500)
+      })
+      test.wait_for_events()
+      assert(mock_device:get_field("minHeatSetpoint") == 500, "minHeatSetpoint field not stored correctly")
+    end,
     {
        min_api_version = 17
     }
 )
 
-test.register_message_test(
+test.register_coroutine_test(
     "MaxHeatSetpointLimit report stores device field",
-    {
-      {
-        channel = "zigbee",
-        direction = "receive",
-        message = { mock_device.id, Thermostat.attributes.MaxHeatSetpointLimit:build_test_attr_report(mock_device, 3500) }
-      }
-    },
+    function ()
+      test.socket.zigbee:__queue_receive({
+        mock_device.id,
+        Thermostat.attributes.MaxHeatSetpointLimit:build_test_attr_report(mock_device, 3500)
+      })
+      test.wait_for_events()
+      assert(mock_device:get_field("maxHeatSetpoint") == 3500, "maxHeatSetpoint field not stored correctly")
+    end,
     {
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/zigbee-thermostat/src/zenwithin/init.lua
```diff
@@ -70,7 +70,7 @@ end
 
 local setpoint_limit_handler = function(limit_type)
   return function(driver, device, limit)
-    device:set_field(limit_type, limit.value, {persist = true})
+    device:set_field(limit_type, limit.value / 100.0, {persist = true})
```
varzac: While this will fix it for future devices, this value is already persisted on existing devices, so even though I think this change makes more sense in a vacuum, for this fix, I think we need to keep persisting in centidegrees (and comment that we are doing such) and then do this division on the checks.  We'll have to update the DEFAULT values as well

drivers/SmartThings/zigbee-thermostat/src/zenwithin/init.lua
```diff
@@ -108,22 +108,24 @@ local update_device_setpoint = function(device)
...
+  if value >= FAHRENHEIT_THRESHOLD then -- we got a command in fahrenheit
+    -- have to scale to degrees first because f_to_c expects a degree value, not centidegrees, then back to centidegrees
+    value = utils.f_to_c(value/100.0) * 100.0
   end
   value = utils.clamp_value(value,
-    device:get_field(MIN_HEAT_LIMIT) or DEFAULT_MIN_SETPOINT,
-    device:get_field(MAX_HEAT_LIMIT) or DEFAULT_MAX_SETPOINT)
+    device:get_field(MIN_COOL_LIMIT) or DEFAULT_MIN_SETPOINT,
+    device:get_field(MAX_COOL_LIMIT) or DEFAULT_MAX_SETPOINT)
   device:set_field(COOLING_SETPOINT, value)
```
varzac, line 129: Isn't this going to set the value in centidegrees, and then correspondingly the update_device_setpoint call will emit an invalid event?

drivers/SmartThings/zigbee-thermostat/src/zenwithin/init.lua
varzac, line 103: Won't this be multiplied by 100 now?

drivers/SmartThings/zigbee-thermostat/src/test/test_zenwithin_thermostat.lua
```diff
@@ -523,4 +523,365 @@ test.register_coroutine_test(
     }
 )
 
+test.register_message_test(
+    "MinHeatSetpointLimit report stores device field",
```
cjswedes, line 527: What is actually being tested here? Maybe needs to be a coroutine test to check device fields with `get_field`

cjswedes (approved):

drivers/SmartThings/zigbee-thermostat/src/test/test_zenwithin_thermostat.lua
aleclorimer: That the fields were actually being handled. I verified that it breaks the tests if the handlers are not set, but you're right. I should just check those values for more sanity while im at it. Updated

drivers/SmartThings/zigbee-thermostat/src/zenwithin/init.lua
aleclorimer: Yes, that was an issue upon further testing. Fixed

drivers/SmartThings/zigbee-thermostat/src/zenwithin/init.lua
aleclorimer: Correct. Reverted and then fixed the scaling or no-scaling elsewhere so we won't encounter that error

varzac (approved):

cjswedes (approved):

PR #3116, opened 2026-07-13
Matter Sensor/Thermostat: Handle invalid unit conversions 

hcarter-775:
# Description of Change
Handle invalid unit conversions by checking for valid molecular weights and valid input values. For Matter Sensor, update handling to include a greater number of conversions and more invalid data handling, akin to what is in Matter Thermostat.

Earlier today, a massive number of coroutine errors were generated related to improper number to string arithmetic operations. This adds greater checks to ensure that does not happen. Also, Matter Sensor was fairly behind and so was updated more generally to match the logic of Matter Thermostat.

### DIFF
--- a/drivers/SmartThings/matter-thermostat/src/test/test_unsupported_unit_conversion.lua
+++ b/drivers/SmartThings/matter-thermostat/src/test/test_unsupported_unit_conversion.lua
@@ -1,0 +1,63 @@
+-- Copyright © 2025 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local test = require "integration_test"
+local capabilities = require "st.capabilities"
+
+-- Unit conversion utilities for thermostat
+local thermostat_utils = require "thermostat_utils.utils"
+local thermostat_fields = require "thermostat_utils.fields"
+
+local units = thermostat_fields.units
+
+test.register_coroutine_test(
+  "Unsupported unit conversion from PPT to MGM3 should return nil",
+  function()
+    local result = thermostat_utils.convert_value_to_unit(100, units.PPT, units.MGM3, capabilities.ozoneMeasurement.NAME)
+    assert(result == nil, "Expected nil for unsupported PPT to MGM3 conversion, got: " .. tostring(result))
+  end
+)
+
+test.register_coroutine_test(
+  "Unsupported unit conversion from PPB to UGM3 should return nil",
+  function()
+    local result = thermostat_utils.convert_value_to_unit(100, units.PPB, units.UGM3, capabilities.carbonMonoxideMeasurement.NAME)
+    assert(result == nil, "Expected nil for unsupported PPB to UGM3 conversion, got: " .. tostring(result))
+  end
+)
+
+test.register_coroutine_test(
+  "Unsupported unit conversion from NGM3 to PPM should return nil",
+  function()
+    local result = thermostat_utils.convert_value_to_unit(100, units.NGM3, units.PPM, capabilities.carbonMonoxideMeasurement.NAME)
+    assert(result == nil, "Expected nil for unsupported NGM3 to PPM conversion, got: " .. tostring(result))
+  end
+)
+
+test.register_coroutine_test(
+  "Supported unit conversion from PPM to PPB should return correct value",
+  function()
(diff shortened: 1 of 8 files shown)


drivers/SmartThings/matter-thermostat/src/thermostat_handlers/attribute_handlers.lua
```diff
@@ -494,23 +494,15 @@ end
...
 function AttributeHandlers.concentration_measured_value_factory(capability_name, attribute, target_unit)
   return function(driver, device, ib, response)
-    local reporting_unit = device:get_field(capability_name.."_unit")
+    if not ib.data.value then return end
+    local reporting_unit = device:get_field(capability_name.."_unit") or fields.unit_default[capability_name]
+    local converted_value = thermostat_utils.convert_value_to_unit(ib.data.value, reporting_unit, target_unit, capability_name)
 
-    if not reporting_unit then
-      reporting_unit = fields.unit_default[capability_name]
-      device:set_field(capability_name.."_unit", reporting_unit, {persist = true})
```
tpmanley, line 501: Why is this removed but there is still code that reads the field?

tpmanley:
Can some tests be added to test the invalid inputs/conversions?

drivers/SmartThings/matter-thermostat/src/thermostat_handlers/attribute_handlers.lua
hcarter-775: we set the field elsewhere, in the measurement unit read handler. There is no need to set a field to a default. If the unit is unavailable in the field, we just use the default.

hcarter-775:
> Can some tests be added to test the invalid inputs/conversions?

Added a few!

NoahCornell (approved):

drivers/SmartThings/matter-thermostat/src/test/test_unsupported_unit_conversion.lua
```diff
@@ -0,0 +1,65 @@
...
+-- Licensed under the Apache License, Version 2.0
+
+local test = require "integration_test"
+local capabilities = require "st.capabilities"
+
+-- Unit conversion utilities for thermostat
+local thermostat_utils = require "thermostat_utils.utils"
+local thermostat_fields = require "thermostat_utils.fields"
+
+test.set_rpc_version(0)
```
cjswedes: Why is this needed?

cjswedes (approved):

drivers/SmartThings/matter-thermostat/src/thermostat_utils/utils.lua
```diff
@@ -180,18 +180,15 @@ function ThermostatUtils.get_device_type(device)
...
+function ThermostatUtils.convert_value_to_unit(value, from_unit, to_unit, capability_name)
+  local conversion_function = fields.unit_conversion[from_unit] and fields.unit_conversion[from_unit][to_unit]
   if not conversion_function then
     log.info_with( {hub_logs = true} , string.format("Unsupported unit conversion from %s to %s", fields.unit_strings[from_unit], fields.unit_strings[to_unit]))
     return
-  end
-
-  if not value then
-    log.info_with( {hub_logs = true} , "unit conversion value is nil")
+  elseif type(value) ~= "number" then
```
ctowns, line 188: I might be missing some context - what non-number values were being passed into this function before? You mentioned some number to string arithmetic operations in the PR comment, what were those and are they a result of device-side issues?

ctowns (approved):

drivers/SmartThings/matter-thermostat/src/thermostat_utils/utils.lua
hcarter-775: our tables for dust measurement molecular weight are "N/A", which is where that would be an issue. In this context, the non-number at hand would be nil

drivers/SmartThings/matter-thermostat/src/test/test_unsupported_unit_conversion.lua
hcarter-775: Claude Haiku doing a low-credit job is why. Removed

PR #3118, opened 2026-07-14
Aqara FP400: Support MCD and Re-profiling

hcarter-775:
# Description of Change
Support MCD profile and endpoint-comparison re-profiling for the FP400.

The endpoint data model re-interview logic will be included in 0.62+. For hubs below this version, the infoChanged trigger will never occur, unless perhaps a software version update occurs.

# Summary of Completed Tests
On-device testing completed.

Adding/removing endpoints via the Aqara app updates the FP400's endpoint model, which in turn updates the device profile. All endpoint attribute updates are properly mapped to the appropriate component.

### DIFF
--- a/drivers/SmartThings/matter-sensor/profiles/aqara-fp400.yml
+++ b/drivers/SmartThings/matter-sensor/profiles/aqara-fp400.yml
@@ -54,6 +54,11 @@ components:
     optional: true
   categories:
   - name: PresenceSensor
+- id: sensor7
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
   categories:
   - name: PresenceSensor
 - id: sensor8
--- a/drivers/SmartThings/matter-sensor/profiles/aqara-fp400.yml
+++ b/drivers/SmartThings/matter-sensor/profiles/aqara-fp400.yml
@@ -12,3 +12,54 @@ components:
     version: 1
   categories:
   - name: PresenceSensor
+- id: sensor1
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor2
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor3
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor4
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor5
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor6
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+  categories:
+  - name: PresenceSensor
+- id: sensor8
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
--- a/drivers/SmartThings/matter-sensor/profiles/aqara-fp400.yml
+++ b/drivers/SmartThings/matter-sensor/profiles/aqara-fp400.yml
@@ -12,3 +12,59 @@ components:
     version: 1
   categories:
   - name: PresenceSensor
+- id: sensor1
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor2
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor3
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor4
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor5
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor6
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor7
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
+- id: sensor8
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
+  categories:
+  - name: PresenceSensor
--- a/drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
+++ b/drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
@@ -1,0 +1,45 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local UpdateMetadataRequest = {}
+UpdateMetadataRequest.__index = UpdateMetadataRequest
+
+function UpdateMetadataRequest.init()
+  local update_metadata_request = setmetatable({}, UpdateMetadataRequest)
+  update_metadata_request.profile = nil
+  update_metadata_request.enabled_components = {}
+  return update_metadata_request
+end
+
+function UpdateMetadataRequest:add_capabilities_to_component(component_id, capability_ids)
+  if not self.enabled_components[component_id] then
+    self.enabled_components[component_id] = {}
+  end
+  for _, capability_id in ipairs(capability_ids) do
+    table.insert(self.enabled_components[component_id], capability_id)
+  end
+  return self
+end
+
+function UpdateMetadataRequest:add_profile(profile)
+  self.profile = profile
+  return self
+end
+
+function UpdateMetadataRequest:formatted_enabled_components()
+  local formatted_enabled_components = {}
+  for component_id, capability_ids in pairs(self.enabled_components) do
+    table.insert(formatted_enabled_components, { component_id, capability_ids })
+  end
+  return formatted_enabled_components
+end
+
+function UpdateMetadataRequest:format_request()
+  local formatted_request = {
+    profile = self.profile,
+    optional_component_capabilities = self:formatted_enabled_components()
+  }
+  return formatted_request
+end
+
+return UpdateMetadataRequest
--- a/drivers/SmartThings/matter-sensor/src/sub_drivers/air_quality_sensor/air_quality_sensor_utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-sensor/src/sub_drivers/air_quality_sensor/air_quality_sensor_utils/device_configuration.lua
@@ -30,53 +30,37 @@ function DeviceConfiguration.supported_level_measurements(device)
   return measurement_caps, level_caps
 end
 
--- Match Modular Profile
 function DeviceConfiguration.match_profile(device)
+  local update_metadata_request = require "sensor_utils.update_metadata_request"
+
   local temp_eps = embedded_cluster_utils.get_endpoints(device, clusters.TemperatureMeasurement.ID)
   local humidity_eps = embedded_cluster_utils.get_endpoints(device, clusters.RelativeHumidityMeasurement.ID)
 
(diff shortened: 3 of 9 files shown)


drivers/SmartThings/matter-sensor/src/sub_drivers/air_quality_sensor/air_quality_sensor_utils/utils.lua
```diff
@@ -77,48 +77,4 @@ function AirQualitySensorUtils.set_supported_health_concern_values(device)
   end
 end
 
---- Deeply compare two values.
```
hcarter-775, line 93: since this is now being used in 2 subdrivers, I just moved this to the main utils file

Kwang-Hui:
@hdlee27 Could you share your test result?

drivers/SmartThings/matter-sensor/src/test/test_matter_aqara_fp400.lua
```diff
@@ -138,4 +239,68 @@ test.register_message_test(
   }
 )
 
+test.register_coroutine_test(
+  "Test profile change on endpoints increment in infoChanged for FP400",
+  function()
+    local current_profile_id = mock_device.profile.id
+    local incremented_matter_endpoints = matter_endpoints
```
cjswedes, line 247:
```suggestion
    local incremented_matter_endpoints = utils.deep_copy(matter_endpoints)
```
This is modifying the table that is used in the mock_device table itself. It is also modifying it for subsequent tests.

hcarter-775: haha, I actually did that semi on purpose for the subsequent tests. You're right that that's pretty contrived though

drivers/SmartThings/matter-sensor/profiles/aqara-fp400.yml
```diff
@@ -12,3 +12,59 @@ components:
     version: 1
   categories:
   - name: PresenceSensor
+- id: sensor1
+  capabilities:
+  - id: presenceSensor
+    version: 1
+    optional: true
```
ctowns, line 19: If we a "modularizing" this profile, could we also explicitly add the `optional` values for the capabilities in `main` as well? Even if they are just defaulting to false anyway, I think its consistent with other modular profiles

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
```diff
@@ -0,0 +1,45 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local UpdateMetadataRequest = {}
+UpdateMetadataRequest.__index = UpdateMetadataRequest
+
+function UpdateMetadataRequest.init()
+  local self = setmetatable({}, UpdateMetadataRequest)
+  self.profile = nil
```
ctowns: If there is some device-specific information, should we pass in the device and save it on the device table? It feels like this could be a potential source of confusion or bugs to share device-specific state in one of these shared modules. IMO, device-specific data should be stored in the device table, even if temporary.

I would not expect us to run into issues in the real world because this should all happen within the same driver on a single thread so theoretically we shouldn't seen instances where multiple devices get mixed up in this during the length of a profile configuration step, but it seems like a precedent we wouldn't want to set. What do you think?

drivers/SmartThings/matter-sensor/src/sub_drivers/air_quality_sensor/air_quality_sensor_utils/device_configuration.lua
```diff
@@ -30,53 +30,37 @@ function DeviceConfiguration.supported_level_measurements(device)
...
-  elseif #temp_eps > 0 then
-    profile_name = "aqs-modular-temp"
-  elseif #humidity_eps > 0 then
-    profile_name = "aqs-modular-humidity"
-  else
-    profile_name = "aqs-modular"
-  end
-
-  device:try_update_metadata({profile = profile_name, optional_component_capabilities = optional_supported_component_capabilities})
+  device:try_update_metadata(updated_metadata:add_profile("aqs-modular" .. preference_tags):format())
```
ctowns: What if we just had a function that was something like "build_update_metadata_request" that took the profile as a parameter and then just returned the formatted metadata request instead of needing to daisy chain these methods together? It would remove the need to call `format()` on these requests, and would simplify it to a single function.

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
```diff
@@ -0,0 +1,45 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local UpdateMetadataRequest = {}
```
ctowns, line 4: Would we plan to gradually update the rest of the drivers to utilize a new module like this? If so, this seems generic enough that we could even consider adding these to the lua libs if we want to proceed with this

drivers/SmartThings/matter-sensor/profiles/aqara-fp400.yml
hcarter-775: Hmm, what are you imagining? To be clear, the capabilities in main are not optional for this device, and this is pretty deeply a device-specific profile

drivers/SmartThings/matter-sensor/src/sub_drivers/air_quality_sensor/air_quality_sensor_utils/device_configuration.lua
hcarter-775: I don't love the idea of adding a profile parameter to this call, since it kinda takes away the concept that the internals of the type will handle everything for you, given the appropriate data is stored in it. I also don't think this:
```
updated_metadata:build_update_metadata_request("aqs-modular" .. preference_tags)
```
is significantly more clear than this:
```
updated_metadata:add_profile("aqs-modular" .. preference_tags):format()
```
Also, the current solution can do something like this:
```
updated_metadata:add_profile("aqs-modular" .. preference_tags)
... later, somewhere down the line ...
updated_metadata:format()
```

I do kinda like the wording `format_request()` over the generic `format()`, but idk about adding a parameter. Like obviously now the system only has 2 pieces, but this makes it a bit more extensible, and there's no real "reason" why the profile name should be a parameter but the optional table shouldn't be, except that it's less complex. So to me, this is the more generic way of dealing with a new type/structure like this. What do you think?

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
hcarter-775: As needed, I'd think so, why not! And I generally agree, I can make a ticket for that.

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
hcarter-775: Hmm, what are you imagining would happen? Since this table is locally instantiated within a function call, all its values would be local to the function, I think.

drivers/SmartThings/matter-sensor/profiles/aqara-fp400.yml
ctowns: I was just suggesting that `optional: false` be added for required capabilities, but I looked at other modular profiles and it seems that in cases where it was mandatory the `optional` value is just left off, I had misremembered. Ignore this comment

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
ctowns: Yeah, it shouldn't be an issue, I was more just calling out that it seemed strange to me storing device-specific variables on a type that is used globally. For example, if for some reason one of the functions on this type is called by another device before `init` has been called to clear the previous data, then the call would return stale data not relevant to the new device. It won't be an issue as long as `init` is called each time this type needs to be used for a new device, which is what is being done currently. So there are set of implicit requirements when using this type:
1. `init` must be called before using it
2. the lifetime of the data can only be guaranteed in a single function call

An alternative to avoid these implicit requirements is just to pass in a table as a parameter. I personally think that is cleaner, but also at that point I can see that it hardly requires this abstraction since that is essentially what is being done in the `match_profile` function already. I don't feel too strongly one way or the other, so I think this is good, but I'd just add a comment somewhere indicating that `init` must be called to clear out any stale data before use.

drivers/SmartThings/matter-sensor/src/sub_drivers/air_quality_sensor/air_quality_sensor_utils/device_configuration.lua
ctowns: Yeah I think `format_request` might be a little more self-documenting, but otherwise I think with how this helper type is set up now I agree that passing in the profile as a parameter is inconsistent since the table itself is not passed in and is state that is saved on the type. I left some comments related to this on the discussion above.

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
hcarter-775: I think I'm still a little confused. If we leave the function where this is called (match_profile in this case), this metatable instantiation will be garbage collected, no? Since "self" is the only thing that points to it, and that's instantiated in init, which is just local to match_profile.

drivers/SmartThings/matter-sensor/src/sub_drivers/air_quality_sensor/air_quality_sensor_utils/device_configuration.lua
hcarter-775: Sure, I'll make that function name update then!

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
ctowns: tbh I think I'm just a little slow on the uptake here, the code works but it just wasn't clicking with me so that's my bad

Some of the confusion for me was that self is used in the other functions throughout this file and it is not declared as a local in those function, so I had assumed it was a global on this type, but I also see that those functions are defined with `:` rather than `.`, so there is an implicit `self` passed in when they are called on an object of this type, which was a miss on my part since this differs from how `init` is set up. I find this module somewhat confusing to understand, but that appears to be more my own lack of understanding rather than anything wrong with the actual code. But it does feel like bad practice to overwrite the `self` keyword in the init function, but maybe that is standard practice in lua.

TLDR; my understanding bad, code good

ctowns (approved):

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
hcarter-775: Ah sure, I see your point. I'll make a change to not override self in init, I think that's a good point.

drivers/SmartThings/matter-sensor/src/test/test_matter_aqara_fp400.lua
hcarter-775: I looked into altering this- it gets a little messy due to how the current update() function in our integration tests works within `generate_info_changed`. Effectively, the update fn is unable to actually remove endpoints right now. Per AI (double checked by me): "The utils.update  function merges tables recursively by key. Since Lua arrays use numeric keys (1, 2, 3, ...), when you pass a shorter array (4 elements) to update a longer one (5 elements), it updates keys 1–4 but leaves the 5th element from the original array intact. So endpoint 5 from the original device remains, plus your new endpoint 5 is now at position 4 — resulting in a duplicate."

So I actually think that the current version of this is the cleanest method of testing at the moment. Would you want to try and update this really quick before merging the PR? I kinda think that might be adding some tension where there is none.

Edit: I do not think we should change the current update implementation in the integration test logic, so this is the best way to do it, I think.

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
hcarter-775: var name updated

drivers/SmartThings/matter-sensor/src/sub_drivers/air_quality_sensor/air_quality_sensor_utils/device_configuration.lua
hcarter-775: fn name updated

drivers/SmartThings/matter-sensor/src/sensor_utils/update_metadata_request.lua
hcarter-775: done

PR #3127, opened 2026-07-16
Zwave Lock: Remove Migration on Added

hcarter-775:
Tested on-device. 
After onboarding, device functions with lockCodes capability as expected.
Migration functions as expected.
After migration, new capabilities function as expected


drivers/SmartThings/zwave-lock/src/init.lua
```diff
@@ -40,8 +32,6 @@ function LockLifecycle.device_added(driver, device)
     command = capabilities.refresh.commands.refresh.NAME,
     args = {}
   })
-  -- ensure our user/credential state is accurate to the current device state
-  device.thread:call_with_delay(5, function() lock_utils.sync_device_state(device) end)
```
hcarter-775, line 44: This was newly added to match zigbee functionality. As the migration is being halted here, I am removing this for now.

varzac (approved):

NoahCornell (approved):

drivers/SmartThings/zwave-lock/src/init.lua
cjswedes: I think some form of this is needed for the old way of doing things (I think it was `reload_all_codes`).

drivers/SmartThings/zwave-lock/src/test/test_init_lifecycle_handlers.lua
```diff
@@ -62,31 +55,6 @@ end
 -- added (device_added)
 -- ============================================================================
 
-test.register_coroutine_test(
```
cjswedes, line 65: I think should we comment out tests/expectations that are needed, so we can keep the tests for when we re-enable this functionality.

hcarter-775: I would likely just point us back to this PR when that time comes. Since some files are being removed altogether (and therefore that'll have to happen anyway), this seemed like a cleaner solution. what do you think?

drivers/SmartThings/zwave-lock/src/init.lua
cjswedes: I understand the comment now. There was no reload codes behavior on init for zwave. This was an improvement, and is fine to remove as we move back to new devices being added using the old functionality.

cjswedes (approved):

drivers/SmartThings/zwave-lock/src/test/test_init_lifecycle_handlers.lua
cjswedes: Documentation in code is better imo; I will forget about these tests. Your call though, since you will probably do the impl anyway

ctowns (approved):

PR #3130, opened 2026-07-17
Fix/coroutine errors

cjswedes:
These are fixes for a handful of coroutine errors that we are seeing in production.

The JBL and Aqara fixes were the highest impact.

### DIFF
--- a/drivers/SmartThings/zigbee-range-extender/src/init.lua
+++ b/drivers/SmartThings/zigbee-range-extender/src/init.lua
@@ -3,6 +3,7 @@
 
 
 local capabilities = require "st.capabilities"
+local log = require "log"
 local defaults = require "st.zigbee.defaults"
 local Basic = (require "st.zigbee.zcl.clusters").Basic
 local ZigbeeDriver = require "st.zigbee"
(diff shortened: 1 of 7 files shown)


drivers/SmartThings/zigbee-range-extender/src/init.lua
```diff
@@ -33,8 +34,13 @@ local zigbee_range_extender_driver = ZigbeeDriver("zigbee-range-extender", zigbe
 function zigbee_range_extender_driver:device_health_check()
   local device_list = self.device_api.get_device_list()
   for _, device_id in ipairs(device_list) do
-    local device = self:get_device_info(device_id, false)
-    device:send(Basic.attributes.ZCLVersion:read(device))
+    local device, err = self:get_device_info(device_id, false)
+    if device == nil then
```
varzac, line 38: Given that we request the device list immediately before this, I'm not really sure how this is possible.  This one might be worth grabbing hub logs for some hubs to investigate because it seems like it shouldn't be possible to me.

varzac (approved):

drivers/SmartThings/zigbee-range-extender/src/init.lua
cjswedes: This can happen if there is an incomplete inventory record for the device. Worth looking into though, like you mention

ctowns (approved):

PR #3131, opened 2026-07-17
Zwave Lock: Force Migration Revert on Init

hcarter-775:
Force Migration revert on init, given lock codes is accessible and a migration has already occurred.

Test added to prove out the migration reversion logic, and that code setting works as expected afterwards. Testing done on-device showing good results as well.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,6 +4,7 @@ def getEnvName() {
   if (branch == "origin/main") {return "ALPHA"}
   else if (branch == "origin/beta") {return "BETA"}
   else if (branch == "origin/production") {return "PROD"}
+  else if (branch == "origin/un-migrate-zwave-locks") {return "TEST"}
 }
 
 def getChangedDrivers() {
--- a/drivers/SmartThings/zwave-lock/src/init.lua
+++ b/drivers/SmartThings/zwave-lock/src/init.lua
@@ -34,6 +34,28 @@ function LockLifecycle.device_added(driver, device)
   })
 end
 
+local function revert_migration(driver, device)
+  local json = require "st.json"
+
+  local latest_users = table_utils.get_state(device, "users") or {}
+  print("Latest users: ", #latest_users)
+  local lock_codes = {}
+  for _, user in ipairs(latest_users) do
+    lock_codes[string.format("%s", user.userIndex)] = user.userName or ("Code " .. user.userIndex)
+  end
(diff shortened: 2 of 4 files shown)


drivers/SmartThings/zwave-lock/src/test/test_lock_users_commands.lua
```diff
@@ -702,4 +703,92 @@ test.register_coroutine_test(
...
+      )
+    )
+
+    test.socket.capability:__expect_send(
+      mock_device:generate_test_message("main",
+        capabilities.lockCodes.migrated(false, { visibility = { displayed = false } })
+      )
+    )
+    test.wait_for_events()
+    assert(mock_device:get_field(constants.SLGA_MIGRATED) == nil, "Device should not be marked as migrated")
```
cjswedes:
```suggestion
    assert(mock_device:get_field(constants.DRIVER_STATE.SLGA_MIGRATED) == nil, "Device should not be marked as migrated")
```

hcarter-775: fixed, thanks!

NoahCornell (approved):

drivers/SmartThings/zwave-lock/src/init.lua
```diff
@@ -34,16 +34,36 @@ function LockLifecycle.device_added(driver, device)
   })
 end
 
+local function revert_migration(driver, device)
+  local json = require "st.json"
+
+  local latest_users = table_utils.get_state(device, "users") or {}
+  print("Latest users: ", #latest_users)
```
ctowns: nit: remove

Jenkinsfile
```diff
@@ -4,6 +4,7 @@ def getEnvName() {
   if (branch == "origin/main") {return "ALPHA"}
   else if (branch == "origin/beta") {return "BETA"}
   else if (branch == "origin/production") {return "PROD"}
+  else if (branch == "origin/un-migrate-zwave-locks") {return "TEST"}
```
ctowns, line 7: nit: idk if the name really matters or where exactly it shows up, but would something like `LOCK-UNMIGRATE` be more descriptive than `TEST` in this case?

hcarter-775: the channel is called TEST

ctowns (approved):

PR #3140, opened 2026-07-22
GHA: Validate Fingerprints

hcarter-775:
# Description of Change
Validate altered fingerprints via GHA. All associated fingerprints changes are an output of this script. 

# Summary of Completed Tests
Tested locally. See file alterations for output results

### DIFF
--- a/.github/scripts/validate_fingerprints.py
+++ b/.github/scripts/validate_fingerprints.py
@@ -1,0 +1,348 @@
+#!/usr/bin/env python3
+"""
+Validate fingerprints.yml files changed in a PR.
+
+Required fields per manufacturer section:
+  matterManufacturer  → id, deviceLabel, vendorId, productId, deviceProfileName
+  zigbeeManufacturer  → id, deviceLabel, manufacturer, model, deviceProfileName
+  zwaveManufacturer   → id, deviceLabel, manufacturerId, deviceProfileName
+                        + at least one of: productId, productType
+
+Hex fields (vendorId, productId, productType, manufacturerId) must use 0xNNNN notation.
+String fields must not be empty or have leading/trailing whitespace.
+id values that contain YAML-special characters must be quoted.
+
+Indentation rules (spaces only, no tabs):
+  Section key:               col 0   e.g. "matterManufacturer:"
+  Entry opening (- id: ...): 2-space e.g. "  - id: ..."
+  All other entry fields:    4-space e.g. "    vendorId: 0x115F"
+
+No trailing whitespace on any line.
+No duplicate id values within a file.
+
+Generic sections (zigbeeGeneric, zwaveGeneric, matterGeneric, etc.) are skipped.
+
+Usage:
+  python3 tools/validate_fingerprints.py                      # auto-detect via git diff
+  python3 tools/validate_fingerprints.py path/fingerprints.yml ...
+"""
+
+import os
+import re
+import sys
+import subprocess
+from pathlib import Path
+
+try:
+    import yaml
+except ImportError:
+    print("Error: pyyaml is required.  pip install pyyaml", file=sys.stderr)
+    sys.exit(2)
+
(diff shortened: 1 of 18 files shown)


.github/scripts/validate_fingerprints.py
```diff
@@ -0,0 +1,348 @@
...
+    'matterManufacturer': ['id', 'deviceLabel', 'vendorId', 'productId', 'deviceProfileName'],
+    'zigbeeManufacturer': ['id', 'deviceLabel', 'manufacturer', 'model', 'deviceProfileName'],
+    'zwaveManufacturer':  ['id', 'deviceLabel', 'manufacturerId', 'deviceProfileName'],
+}
+
+# These fields must be formatted as hex literals (0xNNNN).
+HEX_FIELDS = {'vendorId', 'productId', 'productType', 'manufacturerId'}
+
+# YAML characters that force quoting when present in an unquoted scalar.
+_YAML_SPECIAL_RE = re.compile(r'[:{}\[\],&#*?|<>=!%@`]')
```
cjswedes, line 59: The `_*` prefix is sort of weird here; usually means something internal, but being a script it is all internal...

drivers/SmartThings/zigbee-humidity-sensor/fingerprints.yml
```diff
@@ -6,6 +6,7 @@ zigbeeManufacturer:
     deviceProfileName: humidity-temp-battery-aqara
   - id: "eZEX/E282-KR0B0Z1-HA"
     deviceLabel: eZEX Multipurpose Sensor
+    manufacturer: eZEX
```
cjswedes: Only one of the manufacturer or model must be present...so I think this was intentional, and this change is now limiting the devices that will match the fingerprint.

cjswedes: Sometimes either a manufacturer or a model alone is enough for the matching, and sometimes it is desired to have that be a wildcard.

tpmanley: I think this is intentional as well, both here and in the other fingerprints that were updated. Some devices do not populate the `ManufacturerName` attribute and in those cases we did not include the `manufacturer` in the fingerprint.

hcarter-775: Hmmm, ok good to know. I was not sure, so I can just remove this check. In Matter we need both vid and pid, so I incorrectly assumed there was a similar requirement elsewhere.

hcarter-775: It seems like, as far as all examples we actually have, the only field ever unpopulated is Manufacturer, never model. Anyone know anything about that? Should model always be expected? Seems like yes from an initial thought

cjswedes: At least one must be present.

tpmanley: The requirement enforced by driver API is that either or both manufacturer or model need to be specified so I think that's what we should require here as well

hcarter-775: Thank guys. This alteration to the requirements has been applied

drivers/SmartThings/matter-switch/fingerprints.yml
```diff
@@ -2315,11 +2315,6 @@ matterManufacturer:
     vendorId: 0x100b
     productId: 0x21B3
     deviceProfileName: light-color-level-2200K-6500K
-  - id: "4107/8627"
```
ctowns, line 2383: Just noting that this and the other fingerprints that are removed by this change are deleted because they are duplicates

ctowns (approved):

cjswedes (approved):

PR #3145, opened 2026-07-23
Matter Window Covering: Ignore unchanged handling, stop motion on inverted scroll

hcarter-775:
# Description of Change
Do not send new commands to a device if no change is introduced, and if inversion is attempted while a device is moving, stop motion to revert state.

# Summary of Completed Tests
Not tested on-device, I am not sure if this will need any tweaking to be a smooth experience.


tpmanley (approved):
Looks good but can you add a unit test too?

partner-hub-chenxuelong:
The design looks good, but we encountered some problems during testing. We found that the curtain motor did not respond when it rotated in the opposite direction multiple times.
We created a page to track this issue: https://smartthings.atlassian.net/wiki/spaces/SHPP/pages/5526487368/20260727+-+test?atlOrigin=eyJpIjoiZDdjNDM4NTMzZjdhNGY2ZGFhMjJkMTk4NDNlNmZmZDMiLCJwIjoiYyJ9

PR #3146, opened 2026-07-24
[Aqara] Aqara Dual Relay Module T2

seojune79:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

# Summary of Completed Tests


drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
```diff
@@ -37,14 +43,16 @@ local function find_child(parent, ep_id)
...
 local function device_added(driver, device)
   -- Only create children for the actual Zigbee device and not the children
   if device.network_type == device_lib.NETWORK_TYPE_ZIGBEE then
     local children_amount = get_children_amount(device)
     if children_amount >= 2 then
       for i = 2, children_amount, 1 do
         if find_child(device, i) == nil then
-          local name = string.format("%s%d", string.sub(device.label, 0, -2), i)
+          -- child shares the parent's label (endpoint 1 is the parent, 2..n are children)
+          local name = string.format("%s", device.label)
```
cjswedes: Whats the reason for this change? I think it will be confusing for the user to have the parent and children all be labeled the same.

drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
```diff
@@ -60,22 +68,36 @@ local function device_added(driver, device)
     end
 
     -- for wireless button
-    device:emit_event(capabilities.button.numberOfButtons({ value = children_amount },
-      { visibility = { displayed = false } }))
-    device:emit_event(capabilities.powerMeter.power({ value = 0.0, unit = "W" }))
-    device:emit_event(capabilities.energyMeter.energy({ value = 0.0, unit = "Wh" }))
+    device:emit_event(capabilities.button.numberOfButtons({ value = children_amount }, { visibility = { displayed = false } }))
```
cjswedes, line 70: There is a re-ordering of the events that are emitted on added that is breaking some tests. I dont see a problem with the re-ordering, but the tests need to be fixed:
```
Unit test failures inSmartThingsEdgeDrivers/drivers/SmartThings/zigbee-switch/src/test/test_aqara_switch_power.lua:
    Lifecycle - added test [line 72]
Unit test failures inSmartThingsEdgeDrivers/drivers/SmartThings/zigbee-switch/src/test/test_aqara_switch_no_power.lua:
    Lifecycle - added test [line 78]
```

drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/fingerprints.lua
```diff
@@ -2,13 +2,14 @@
...
-  { mfr = "LUMI", model = "lumi.switch.b2nacn01", children = 2, child_profile = "aqara-switch-child" },
-  { mfr = "LUMI", model = "lumi.switch.b3n01",    children = 3, child_profile = "aqara-switch-child" }
+  { mfr = "LUMI",  model = "lumi.switch.n1acn1",   children = 1, child_profile = "" },
+  { mfr = "LUMI",  model = "lumi.switch.n2acn1",   children = 2, child_profile = "aqara-switch-child" },
+  { mfr = "LUMI",  model = "lumi.switch.n3acn1",   children = 3, child_profile = "aqara-switch-child" },
+  { mfr = "LUMI",  model = "lumi.switch.b1laus01", children = 1, child_profile = "" },
+  { mfr = "LUMI",  model = "lumi.switch.b2laus01", children = 2, child_profile = "aqara-switch-child" },
+  { mfr = "LUMI",  model = "lumi.switch.l2aeu1",   children = 2, child_profile = "aqara-switch-child" },
+  { mfr = "LUMI",  model = "lumi.switch.n2aeu1",   children = 2, child_profile = "aqara-switch-child" },
+  { mfr = "LUMI",  model = "lumi.switch.b4nacn01", children = 2, child_profile = "aqara-switch-child" },
```
cjswedes: What is the `lumi.switch.b4nacn01` model? It is added here, but not in the `aqara/fingerprints.lua` file or in the `fingerprints.yml` file.

drivers/SmartThings/zigbee-switch/profiles/aqara-dual-relay-module-unified.yml
cjswedes, line 1: This should include the firmwareUpdate capability like the other device profiles do. This allows the device FW version to be displayed to the user in the app.

drivers/SmartThings/zigbee-switch/profiles/aqara-dual-relay-module-unified.yml
```diff
@@ -0,0 +1,40 @@
...
+    explicit: true
+  - preferenceId: stse.pulseIntervalSetting
+    explicit: true
+  - preferenceId: stse.switchType
+    explicit: true
+  - preferenceId: stse.changeToWirelessSwitch
+    explicit: true
+metadata:
+  mnmn: SolutionsEngineering
+  vid: SmartThings-smartthings-Aqara_Dual_Relay_Module_T2
```
cjswedes: nit: trailing whitespace
```suggestion
  vid: SmartThings-smartthings-Aqara_Dual_Relay_Module_T2
```

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -112,26 +133,71 @@ local preference_map = {
     data_type = data_types.Uint8,
     value_map = { rocker = 0x01, rebound = 0x02 },
   },
+  -- External switch wiring type (same attribute as stse.electricSwitchType, with an extra "nc"
```
cjswedes, line 137: Why is this needed? Can we add the nc variant t othe existing electricSwitchType preference?

Since they map to the same attribute, this could get confusing in the future.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -112,26 +133,71 @@ local preference_map = {
...
+    mfg_code = MFG_CODE,
+    data_type = data_types.Uint8,
+    value_map = { restore = 0x01, on = 0x00, off = 0x02, reverse = 0x03 },
+  },
+  ["stse.pulseIntervalSetting"] = {
+    cluster_id = PRIVATE_CLUSTER_ID,
+    attribute_id = PULSE_INTERVAL_ATTRIBUTE_ID,
+    mfg_code = MFG_CODE,
+    data_type = data_types.Uint16,
+    value_type = { "number" }, -- presence flag: coerce the preference value with tonumber() before writing
```
cjswedes: Since this is used as a flag to coerce the preference value a single element table is a weird way to do it. IMO a boolean flag makes more sense. 

```suggestion
    coerce_number = true
```

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
```diff
@@ -23,9 +23,30 @@ local CHANGE_TO_WIRELESS_SWITCH_ATTRIBUTE_ID = 0x0200
...
+local PULSE_INTERVAL_ATTRIBUTE_ID = 0x00EB   -- pulse width in ms when running in pulse mode, Uint16
 local LAST_REPORT_TIME = "LAST_REPORT_TIME"
 local PRIVATE_MODE = "PRIVATE_MODE"
+-- "interlock" / "devicemode" are extra profile components on aqara-dual-relay-module-unified.yml.
+-- The order of the SUPPORTED_* lists matches the raw device values (0-based), see the handlers below.
+local COMPONENT_INTERLOCK = "interlock"
+local SUPPORTED_INTERLOCK = { "normal", "interlock" }
+local COMPONENT_DEVICE_MODE = "devicemode"
+-- Relay working modes mapped to their raw device values. dry_contact_open_pulse_mode (raw 2) is
+-- intentionally not exposed; the remaining modes keep their original device values (on_off stays 3).
```
cjswedes, line 39: What is the reason for not exposing this? The mode capability can take arbitrary strings, so it could be represented.

If a device reports this mode, the driver silently drops it, and then the device state is no longer correct; this seems misleading to the user. Although I am not sure I fully understand these modes, so perhaps I am wrong.

drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
seojune79: Thank you for catching this detail.
I will update the code to revert to the previous naming convention as suggested. Once the final adjustments are made, I will commit and push the changes to this PR.

drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/init.lua
seojune79: I have fixed the issue causing the Test Case (TC) failure.

drivers/SmartThings/zigbee-switch/src/aqara/multi-switch/fingerprints.lua
seojune79: I have updated the model information as an error was identified in the previous configuration.

drivers/SmartThings/zigbee-switch/profiles/aqara-dual-relay-module-unified.yml
seojune79: Since this device does not support firmware updates in the SmartThings environment, I have removed the firmwareUpdate capability from the device profile as suggested by Samsung.

drivers/SmartThings/zigbee-switch/profiles/aqara-dual-relay-module-unified.yml
seojune79: Done.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
seojune79: If we add the 'nc' variant to the existing electricSwitchType preference, I am wondering if there will be any impact or side effects on devices that previously did not support the 'nc' variant.
I would appreciate it if you could share your thoughts on whether this might cause any issues regarding backward compatibility.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
seojune79: Updated.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
seojune79: Mode support varies depending on the device using this driver, which is why I separated them

cjswedes: Will some devices report dry_contact_open_pulse_mode?

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
cjswedes: Your concern is valid; adding the variant to the preference will expose it for all devices that have the preference even if they do not support it. IMO this is probably okay, but would depend on how existing devices that do not support it handle the nc option being set in the attribute write. If the attribute write is rejected and it is essentially a no-op (what I would expect), then we can just add text in the preference definition that the option may not be supported. 

IMO, we should combine them, but I am curious if @hcarter-775 or @tpmanley have any thoughts

drivers/SmartThings/zigbee-switch/profiles/aqara-dual-relay-module-unified.yml
cjswedes: If the device supports reporting its firmware version, I think we should still have it (which pretty much all zigbee devices do support).

seojune79: We will add this section once Samsung's guidelines are updated.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
seojune79: Currently, only the Dual Relay Module T2 reports this value.

cjswedes: Please explain the intention for not handling this value? The driver silently dropping it seems like it would be misleading to users when the device is in a different mode than what we report..

seojune79: I am not sure I fully understand your question. Could you please provide a bit more detail on the specific scenario or concern?

For context, the status reflection logic in the current driver operates as follows:
1. Mode Changes: When a user changes the mode, the device reports its updated state. The driver then updates the plugin UI accordingly via the interlock_switch_handler.
2. Onboarding: During onboarding, the device reports its current state, allowing the plugin UI to display correctly from the start.
3. Refresh: During a refresh, the driver reads the current state from the device and updates the display with the latest values.

If there is a specific edge case or condition where you expect an issue to occur, please let me know so I can investigate further.

Thank you.

cjswedes: The Dual Relay Module T2 can report `dry_contact_open_pulse` for the device mode. The handler will not emit an event for that mode because [it is not in the map](https://github.com/seojune79/SmartThingsEdgeDrivers/blob/6d0eb03fd0202f2a29e24d2e3176a92c92ae49ce/drivers/SmartThings/zigbee-switch/src/aqara/init.lua#L38-L49); there is a comment saying it is intentionally ignored but does not explain why it is ignored,

The consequence of ignoring the mode is that when a device reports that mode, previous mode value will remain displayed to the user; this is a bad user experience. I am trying to understand why that mode is not handled when it seems like a valid state the device can be in.

seojune79: Thank you for the detailed explanation.

During the initial development, all modes were supported. However, to align with the features supported by the Aqara app, that specific mode was removed during the PR process, leaving behind a comment generated by AI. Functionally, since there is no way to enter the removed mode, this should not cause any issue.

If my understanding of the situation is correct, would it be sufficient to simply delete that comment?

cjswedes: Thanks for the explanation.

> If my understanding of the situation is correct, would it be sufficient to simply delete that comment?

I would suggest that you add to the comment explaining how it is a mode that is not supported by the Aqara app, and that it is not possible to get the device into that mode which is why it is omitted.

seojune79: Thank you for the review.
I have updated the code to include the requested comment.
The added comment clearly explains that this mode is not supported by the Aqara app and that the device cannot enter this mode, which is why it was omitted from the implementation.
Please let me know if you have any further feedback.

drivers/SmartThings/zigbee-switch/src/aqara/init.lua
cjswedes: @tpmanley do you have any thoughts here? I dont love the duplicate named preference, but it may be warranted here.

cjswedes:
This is looking good to me. It needs to be updated to resolve conflicts and have tests passing.

aleclorimer (approved):

cjswedes (approved):

PR #3148, opened 2026-07-27
Branching registryCredentialsId based off region and nodeLabel

aleclorimer:
This change looks to solve a jenkins workflow error in which the CN prod runner fails due to a credential error. The targeted logic of 'artifactory-credentials' when on the CN prod jenkins runner have been tested to work in a test environment. This change should  have the affect of changing `registryCredentialsId` only and the registryURL for CN Prod.
| Region | Environment | Credentials | RegistryURL |
|--------|--------|--------|--------|
| Global | Staging | artifactory | <empty/default> |
| Global | Prod | artifactory | <empty/default> |
| China | Staging | artifactory | <empty/default> |
| China | Prod | artifactory-credentials | registry.artifactoryedge.streleng.cn |

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,11 +24,42 @@ def getChangedDrivers() {
   return drivers
 }
 
+def get_region() {
+  def uri = new URI(env.JENKINS_URL)
+  def region = uri.host.endsWith('.cn') ? 'cn' : 'global'
+  return region
+}
+
+// Gate artifactory-credentials for prod nodes in cn due to a docker authentication error arising in that environment
+def getDockerCredentialId() {
+    def nodeLabel = params.NODE_LABEL ?: 'production'
+    def region = get_region()
+    if (nodeLabel == 'production' && region == 'cn') {
+      return 'artifactory-credentials'
+    }
+    else {
+      return 'artifactory'
+    }
+}
+// Gate RegistryUrl for prod nodes in cn due to a docker authentication error arising in that environment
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -24,11 +24,30 @@ def getChangedDrivers() {
...
+      return 'artifactory'
+    }
+}
+
 pipeline {
   agent {
     docker {
       image 'python:3.10'
       label  "${params.NODE_LABEL ?: 'production'}"
+      registryUrl 'https://registry.artifactoryedge.streleng.cn'
```
cjswedes: why the `.cn` registry URL?

aleclorimer: Whoops, this should probably be gated on CN and prod. Otherwise kept normal

hcarter-775 (approved):
Since this is fairly specific code that doesn't really have codified reasons to exist, I would suggest adding some commentary in the file itself (tbh, effectively what's in the PR description). The table you have helps clarify what we're expecting here, so I just don't want us to forget, or need to dig up the PR

elbe0046 (approved):

cjswedes (approved):

cjswedes:
To add some further context to this. The defaults in jenkins CN are correct; we should not need to specify `registryCredentialId` at all (and we dont need to specify `registryUrl` at all). The reason for this change is to avoid  a caching issue by forcing the credential to be evaluated from the source.

PR #3150, opened 2026-07-28
Matter Switch: Fix Electrical profiling coroutine errors and improve fallback handling

hcarter-775:
# Description of Change
This attempts to improve some open issues in Electrical Sensor handling:
1.  Avoid coroutine errors when handling nil Electrical Energy struct inputs for all hubs using a lua libs version below 21, above which this fix is already in place
2. Avoid nil-case errors when no default profile for a device type is found for the onOff endpoint- in this case, likely if the device type is Electrical Sensor
3. Add profiling attributes to refresh as needed, in case errors that can be fixed via retry occur in the future, so that a refresh can trigger re-profiling as needed. All associated handlers block re-profiling if the device has already marked itself as profiled

This solves the following issue: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/issues/3147

# Summary of Completed Tests
Unit tests continue to pass. Tested on-device: https://community.smartthings.com/t/driver-for-tp-link-tapo-p316m/310364.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/embedded_clusters/ElectricalEnergyMeasurement/types/EnergyMeasurementStruct.lua
+++ b/drivers/SmartThings/matter-switch/src/embedded_clusters/ElectricalEnergyMeasurement/types/EnergyMeasurementStruct.lua
@@ -42,6 +42,20 @@ EnergyMeasurementStruct.field_defs = {
     is_optional = true,
     data_type = require "st.matter.data_types.Uint64",
   },
+  {
+    name = "apparent_energy",
+    field_id = 5,
+    is_nullable = false,
+    is_optional = true,
+    data_type = require "st.matter.data_types.Int64",
+  },
+  {
+    name = "reactive_energy",
+    field_id = 6,
+    is_nullable = false,
+    is_optional = true,
+    data_type = require "st.matter.data_types.Int64",
+  },
 }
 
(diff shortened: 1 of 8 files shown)


cjswedes (approved):

drivers/SmartThings/matter-switch/src/embedded_clusters/ElectricalEnergyMeasurement/types/EnergyMeasurementStruct.lua
```diff
@@ -42,17 +42,31 @@ EnergyMeasurementStruct.field_defs = {
     is_optional = true,
     data_type = require "st.matter.data_types.Uint64",
   },
+  {
```
ctowns, line 45: Do we also need to make these updates in `matter-thermostat` and `matter-energy` where this is also defined via an embedded cluster definition?

ctowns (approved):

drivers/SmartThings/matter-switch/src/embedded_clusters/ElectricalEnergyMeasurement/types/EnergyMeasurementStruct.lua
hcarter-775: I don't think it matters in those drivers. 1. Nobody is using them 2. Even if they were, there'd be no real side effects of the coroutine error for them

PR #3158, opened 2026-07-31
Remove auto-migration for Zigbee locks on added

hcarter-775:


cjswedes (approved):
Is the plan to do the migration all at once for all devices with a driver change at some point this fall?

PR #3159, opened 2026-07-31
Zigbee Lock: Force Migration Revert on Init

hcarter-775:
Force Migration revert on init, given lock codes is accessible and a migration has already occurred.

Test added to prove out the migration reversion logic, and that code setting works as expected afterwards.

Testing completed: Spot checked on a Kwikset Zigbee lock supporting codes, that when added to the production driver and then updated to this logic, without triggering a driver ID switch, the device un-migrated as expected

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,6 +4,7 @@ def getEnvName() {
   if (branch == "origin/main") {return "ALPHA"}
   else if (branch == "origin/beta") {return "BETA"}
   else if (branch == "origin/production") {return "PROD"}
+  else if (branch == "origin/un-migrate-zigbee-locks") {return "TEST"}
 }
 
 def getChangedDrivers() {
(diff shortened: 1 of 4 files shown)


NoahCornell (approved):

ctowns (approved):
Cross referenced with https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3131 and it looks good.

There seem to be a lot more unit test failures than expected. I notice this branch is 24 commits behind main, is that intentional? It could explain the unit test failures

hcarter-775:
@ctowns it really doesn't matter whether it's up to date with main or not, nor if any tests pass other than the ones in Zigbee Lock. So I wouldn't say intentional per se, but rather just unimportant in this case, since the only driver that will be deployed to the TEST channel is Zigbee Lock

PR #3166, opened 2026-08-05
Matter Switch: Add logging on create/update of EDGE_CHILD devices

hcarter-775:
# Description of Change
Add some new logging.

### DIFF
--- a/drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
+++ b/drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
@@ -1,7 +1,6 @@
 -- Copyright © 2025 SmartThings, Inc.
 -- Licensed under the Apache License, Version 2.0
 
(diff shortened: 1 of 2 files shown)


drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
```diff
@@ -1,6 +1,7 @@
 -- Copyright © 2025 SmartThings, Inc.
 -- Licensed under the Apache License, Version 2.0
 
+local st_utils = require "st.utils"
```
cjswedes: unused require

cjswedes (approved):

drivers/SmartThings/matter-switch/src/switch_utils/device_configuration.lua
hcarter-775: whoops, removed the thing I'd added it for

PR #3167, opened 2026-08-05
Test backporting 57 58

aleclorimer:
- Updated min/max_api_version for tests passing and failing on HCv57 and 58.
- Fixed bad test manually setting `version.api = 9` and set min/max_api_version so that test only runs on v9

### DIFF
--- a/drivers/SmartThings/matter-appliance/src/test/test_cook_top.lua
+++ b/drivers/SmartThings/matter-appliance/src/test/test_cook_top.lua
@@ -91,7 +91,7 @@ test.register_coroutine_test(
     assert(component_to_endpoint_map["cookSurfaceTwo"] == COOK_SURFACE_TWO_ENDPOINT, "Cook Surface Two Endpoint must be 3")
   end,
   {
-     min_api_version = 17
+     min_api_version = 15
   }
 )
 
@@ -117,7 +117,7 @@ test.register_message_test(
     }
   },
   {
-     min_api_version = 17
+     min_api_version = 15
   }
 )
 
@@ -157,7 +157,7 @@ test.register_message_test(
     },
   },
   {
-     min_api_version = 17
+     min_api_version = 15
   }
 )
 
@@ -192,7 +192,7 @@ test.register_message_test(
     }
   },
   {
-     min_api_version = 17
+     min_api_version = 15
   }
 )
 
--- a/drivers/SmartThings/matter-appliance/src/test/test_dishwasher.lua
+++ b/drivers/SmartThings/matter-appliance/src/test/test_dishwasher.lua
@@ -103,7 +103,7 @@ test.register_message_test(
     }
   },
   {
-     min_api_version = 17
+     min_api_version = 15
   }
 )
 
@@ -171,7 +171,7 @@ test.register_message_test(
     }, -- on receiving NO ERROR we don't do anything.
   },
   {
-     min_api_version = 17
+     min_api_version = 15
   }
 )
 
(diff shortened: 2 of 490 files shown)


cjswedes (approved):
Needs to pass Luacheck.

I dont see the 57 and 58 test runs in the github actions either. Will those work once we merge? I assume they were run manually and pass.

aleclorimer:
> Needs to pass Luacheck.

Good transition into a question I had. The luacheck failures came from a big find and replace with `min_api_version = 14` => `<empty>` and it left the empty lines there. I can fix this, but I won't need to if we keep the min version option set to 14 rather than empty on all passing tests.
I personally would like to do this (and drop the commit that deletes all the version 14 to empty changes). Just because the formatting of having the `min_api_version` already there makes it a lot easier to come back around and modify this value to a different version. I'm only pushing this because it sounds like we haven't fully ruled out back-porting to 52 as well.

aleclorimer:
> I dont see the 57 and 58 test runs in the github actions either. Will those work once we merge? I assume they were run manually and pass.

Yeah the updated 57 and 58 won't run until this is merged and the matrix is updated. That part of the actions only runs off `origin/main`. I did run them locally.

cjswedes:
> > Needs to pass Luacheck.
> 
> Good transition into a question I had. The luacheck failures came from a big find and replace with `min_api_version = 14` => `<empty>` and it left the empty lines there. I can fix this, but I won't need to if we keep the min version option set to 14 rather than empty on all passing tests. I personally would like to do this (and drop the commit that deletes all the version 14 to empty changes). Just because the formatting of having the `min_api_version` already there makes it a lot easier to come back around and modify this value to a different version. I'm only pushing this because it sounds like we haven't fully ruled out back-porting to 52 as well.

I am onboard with leaving min_api_version = 14 in place

aleclorimer:
The one matter-camera test I think is the one that randomly fails from time to time. I can keep running the tests, but this is not a failure i've seen locally, nor was anything else changed.

varzac:
I'm not seeing the 57/58 tests called out as separate Jenkins results like the other versions, but it does look to be running something (with a test error still present, so not sure if it's an intermittently failing tests, or if a test wasn't tagged right)

aleclorimer:
> I'm not seeing the 57/58 tests called out as separate Jenkins results like the other versions, but it does look to be running something (with a test error still present, so not sure if it's an intermittently failing tests, or if a test wasn't tagged right)

This comment got lost in the sea of bot spam
> The one matter-camera test I think is the one that randomly fails from time to time. I can keep running the tests, but this is not a failure i've seen locally, nor was anything else changed.

For the 57/58 tests, they aren't running because the Jenkins workflow route is run off whatever is in `origin/main`, so those tests won't start triggering until this `.github/workflows/jenkins-driver-tests.yml` is merged.

aleclorimer:
The spam should be done. All tests 57-dev look to be passing.

aleclorimer:
## Things to note:
Excluding entire files failing to run for versions < 17
- `drivers/SmartThings/matter-switch/src/test/test_matter_camera.lua`
- `drivers/SmartThings/matter-switch/src/test/test_matter_irrigation_system.lua`

Exits at the start of the file to stop running the rest of the file. Tests need to be run to avoid throwing errors
```
if version.api < 17 then
  test.run_registered_tests()
  os.exit(0)
end
```

Kept tests passing at v14 to have `min_api_version=14` in-case we need to move to an earlier revision, this framing will make it easier to take that step.

varzac (approved):

cjswedes (approved):

PR #3170, opened 2026-08-10
Driver test crashing visibility

aleclorimer:
# Description of Change
- Added handling for error messages coming from tests files and reporting the errors to the XML output. This is then picked up by the GHA driver_test status output

# Summary of Completed Tests
- Made a test fail from a bad require and having the error show up on GH.

Screenshot of error being picked up by GH.
image

### DIFF
--- a/tools/run_driver_tests_p.py
+++ b/tools/run_driver_tests_p.py
@@ -63,6 +63,12 @@ def run_test(test_file):
     else:
       successes += 1
     test_cases.append(test_case)
+  if error and error != "" and len(test_cases) == 0:
+    failure_output += "\t{} ERROR: test file failed to run\n".format(test_suite_name)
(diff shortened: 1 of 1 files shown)


varzac (approved):

cjswedes (approved):
Just to make sure I understand. This is about reporting errors that cause the test runner to crash due to a syntax error or something in the test file itself, right? All the lua errors from the driver code itself are caught by the integration test framework and already reported by the test.

aleclorimer:
> Just to make sure I understand. This is about reporting errors that cause the test runner to crash due to a syntax error or something in the test file itself, right? All the lua errors from the driver code itself are caught by the integration test framework and already reported by the test.

Correct. This idea came from some tests in backporting where some zap cluster files were missing so a whole test file was not being registered but the suite would still say `n out of n test passed`

Issue #3184, opened 2026-08-17
Matter camera: single-axis MPTZSetPosition commands should not depend on cached values for the other axes


ldeora:
## Summary

The current `mechanicalPanTiltZoom` handler implements SmartThings `setPan`, `setTilt` and `setZoom` by loading the cached state of all three axes and then sending all three values in `MPTZSetPosition`.

This is unsafe because a camera can support an axis while its current SmartThings state is temporarily unknown.

A real G350 state demonstrated exactly this:

```text
pan  = -36
tilt = 15
zoom = null
```

while mechanical zoom itself was supported and a valid zoom range was reported.

For single-axis commands, the driver should send only the axis requested by the SmartThings command and leave the other optional MPTZSetPosition fields omitted.

## Environment / verification

Verified against `SmartThingsCommunity/SmartThingsEdgeDrivers` `main` at commit
`60bbf2716412fdf04fd59f967bdd37be6995d664` (2026-08-12).

Real-device validation used:

- Aqara Camera Hub G350
- VID/PID: `0x115F / 0x3013`
- firmware: `4.5.70` / `4005070`
- reported Matter SpecificationVersion: `1.5.1`
- Camera device type: `0x0142`
- Camera endpoint in this device: endpoint 2

The proposed changes are generic Matter-camera changes. No Aqara VID/PID fingerprint or proprietary Aqara cluster handling is required.

## Current behavior

The current handler:

1. obtains cached pan/tilt/zoom;
2. changes only the requested value;
3. clamps **all** cached values;
4. sends all three values.

See:

- [`camera_handlers/capability_handlers.lua` lines 151-172](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/60bbf2716412fdf04fd59f967bdd37be6995d664/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/capability_handlers.lua#L151-L172)

Conceptually:

```lua
setPan:
  current.pan = requested_pan

for every axis:
  current = clamp(current)

MPTZSetPosition(current.pan, current.tilt, current.zoom)
```

If an unrelated cached axis is `nil`, a valid single-axis command unnecessarily depends on unavailable state.

## Reproduction

1. Use a mechanical PTZ camera.
2. Have valid pan and tilt state but no current zoom state.
3. Execute SmartThings `setPan`.
4. The current code still attempts to clamp and send the zoom value.

[...]

PR #3192, opened 2026-08-22
Fix PTZ bugs with unsupported & unchanged fields

samadDotDev:
# Description of Change

### Issue # 1
The attribute handler and device config used `MPAN/MTILT/MZOOM/MPRESETS` (based on "feature code" from [spec](https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/CameraAVSettingsUserLevelManagement.adoc#4-features)), which don't exist in the generated files in lua libs (`CameraAvSettingsUserLevelManagement/types/Feature.lua` only defines `MECHANICAL_PAN/TILT/ZOOM/PRESETS` based on "feature name" in [SDK XML](https://github.com/project-chip/connectedhomeip/blob/b791201722360efc16b5a4fe152123f67f2acc2b/src/app/zap-templates/zcl/data-model/chip/camera-av-settings-user-level-management-cluster.xml#L193-L194)). 

In Lua, indexing an undefined table field returns nil rather than erroring, so every feature-support check silently evaluated to "supported," and `MatterDevice:get_endpoints`'s `opts.feature_bitmap == nil` short-circuit treated any device with the cluster as having all PTZ features. On a device whose FeatureMap excludes zoom, this caused the driver to read a zoom field that the device never sends, crashing with `attempt to index a nil value (field 'zoom')`. 

Fixed by using the correct `MECHANICAL_*` names everywhere.

### Issue # 2
Separately, `ptz_relative_move_factory` always sent all three deltas to `MPTZRelativeMove`, defaulting unused axes to 0 instead of omitting them. Per [spec](https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/CameraAVSettingsUserLevelManagement.adoc#72-mptzrelativemove-command), each delta's conformance is tied to its own feature bit, so an explicit 0 isn't the same as an omitted field and essentially a zoom-less device rejected the command with `INVALID_COMMAND`. 

Fixed by only setting the delta for the axis being moved and leaving the others nil.

# Summary of Completed Tests

- Tested with the same camera that doesn't support zoom in PTZ (has Pan-Tilt-only feature)
- Tested with a different camera that supports PTZ
- Added a unit test covering the first fix with a new mock device fixture (that doesn't support zoom), and updated existing tests for the second fix (omitting unchanged/unsupported fields instead of defaulting to 0)

### DIFF
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
@@ -300,13 +300,13 @@ function CameraAttributeHandlers.ptz_position_handler(driver, device, ib, respon
       ))
     end
   end
-  if camera_utils.feature_supported(device, clusters.CameraAvSettingsUserLevelManagement.ID, clusters.CameraAvSettingsUserLevelManagement.types.Feature.MPAN) then
+  if camera_utils.feature_supported(device, clusters.CameraAvSettingsUserLevelManagement.ID, clusters.CameraAvSettingsUserLevelManagement.types.Feature.MECHANICAL_PAN) then
     emit_event(camera_fields.PAN_IDX, ib.data.elements.pan.value)
   end
-  if camera_utils.feature_supported(device, clusters.CameraAvSettingsUserLevelManagement.ID, clusters.CameraAvSettingsUserLevelManagement.types.Feature.MTILT) then
+  if camera_utils.feature_supported(device, clusters.CameraAvSettingsUserLevelManagement.ID, clusters.CameraAvSettingsUserLevelManagement.types.Feature.MECHANICAL_TILT) then
     emit_event(camera_fields.TILT_IDX, ib.data.elements.tilt.value)
   end
-  if camera_utils.feature_supported(device, clusters.CameraAvSettingsUserLevelManagement.ID, clusters.CameraAvSettingsUserLevelManagement.types.Feature.MZOOM) then
+  if camera_utils.feature_supported(device, clusters.CameraAvSettingsUserLevelManagement.ID, clusters.CameraAvSettingsUserLevelManagement.types.Feature.MECHANICAL_ZOOM) then
     emit_event(camera_fields.ZOOM_IDX, ib.data.elements.zoom.value)
   end
 end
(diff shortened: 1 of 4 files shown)


ldeora:
One related PTZ issue remains in `ptz_set_position_factory()` and is tracked in #3184.

Single-axis commands still depend on the cached values of the other axes:

```text
setPan(20)
    → MPTZSetPosition(20, cachedTilt, cachedZoom)
```

For Matter's optional PTZ fields, this should instead be:

```text
setPan         → MPTZSetPosition(pan, nil, nil)
setTilt        → MPTZSetPosition(nil, tilt, nil)
setZoom        → MPTZSetPosition(nil, nil, zoom)
setPanTiltZoom → MPTZSetPosition(pan, tilt, zoom)
```

This is not only relevant for cameras without one of the mechanical features. On the G350, all three axes are supported, but we observed a valid state where `pan` and `tilt` were populated while the current `zoom` value was temporarily `null`.

So a `setPan` or `setTilt` command should not require unrelated cached axis state to be available.

samadDotDev:
@Ideora thanks for reporting that bug. The latest commit should address that as well and LMK if it doesn't: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3192/commits/a8fb177b98cccb38db80db6a7047af0a6150ab3c

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
```diff
@@ -300,13 +300,13 @@ function CameraAttributeHandlers.ptz_position_handler(driver, device, ib, respon
       ))
     end
   end
-  if camera_utils.feature_supported(device, clusters.CameraAvSettingsUserLevelManagement.ID, clusters.CameraAvSettingsUserLevelManagement.types.Feature.MPAN) then
+  if ib.data.elements.pan and camera_utils.feature_supported(device, clusters.CameraAvSettingsUserLevelManagement.ID, clusters.CameraAvSettingsUserLevelManagement.types.Feature.MECHANICAL_PAN) then
```
hcarter-775, line 304: Before any of this, I'd add a check like: `if not ib.data.elements then return end`, since this is not a guaranteed table to exist in all cases and would cause a fatal (I think) error here if not caught.

ldeora:
> @Ideora thanks for reporting that bug. The latest commit should address that as well and LMK if it doesn't: [a8fb177](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/a8fb177b98cccb38db80db6a7047af0a6150ab3c)

The `MPTZSetPosition` change looks good.

The implementation matches the one I've been testing: single-axis commands only set and clamp the requested coordinate and leave the other two fields `nil`, while `setPanTiltZoom` supplies all three coordinates.

My test driver additionally gates the MPTZ commands against `AcceptedCommandList`, but that's part of a separate metadata/optional-command change and unrelated to this fix.

I also like the additional presence checks for the individual fields in incoming `MPTZPosition` reports - that's a useful robustness improvement beyond my current implementation.

So from my side, the fix looks correct.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
samadDotDev: I can add that as a defensive check, but just noting here that handler is only used for reports of `MPTZPosition` attribute, conformance of which is `MPAN | MTILT | MZOOM`, so one of these elements should always be present: https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/CameraAVSettingsUserLevelManagement.adoc#6-attributes

hcarter-775: For sure, it would only be defensive. Technically this should never really be nil, I don't think. But it is sometimes, generically

hcarter-775 (approved):

tpmanley (approved):

PR #3200, opened 2026-08-28
Fix profile name in fingerprint for Govee COB Strip Light

samadDotDev:
This was just merged and is causing driver packaging failure. 
We should probably add some validation in CI/GHA for this.

### DIFF
--- a/drivers/SmartThings/matter-switch/fingerprints.yml
+++ b/drivers/SmartThings/matter-switch/fingerprints.yml
@@ -927,6 +927,7 @@ matterManufacturer:
     deviceLabel: Govee COB Strip Light 2 Pro 16.4ft/5m
     vendorId: 0x1387
     productId: 0x1AA5
+    deviceProfileName: light-color-level
   - id: "4999/6723"
     deviceLabel: Govee Strip Light 2
     vendorId: 0x1387


cbaumler (approved):

ctowns (approved):

ctowns:
> We should probably add some validation in CI/GHA for this.


Packaging and uploading the driver to a test channel is a part of the current GHA flow, so I am surprised this passed CI. Here is the packaging step for the PR that added this: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/actions/runs/33095683934/job/98599931727

I'm not sure what's different between how that driver is packaged in CI vs. otherwise, but I agree that it would be worth fixing so issues like this don't sneak through. I can investigate further

hcarter-775:
this is what this is for: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3140

hcarter-775:
@ctowns The driver packaging step does not actually block CI. We can see the channel failure here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/actions/runs/33104808024

hcarter-775 (approved):

PR #3201, opened 2026-08-28
Fix hard privacy mode presence in `supportedAttributes`

samadDotDev:
## Description of Change

The `cameraPrivacyMode.hardPrivacyMode` was getting dropped from `supportedAttributes`, even on devices that report it. `HardPrivacyModeOn` is of optional conformance in the Matter spec on top of the PRIV feature bit that gates the soft-privacy attributes - so it can only be detected from the Camera AVSM cluster's `AttributeList`, not from the cluster feature map.

The `build_camera_privacy_supported_attributes()` was hardcoded to always return just `{"softRecordingPrivacyMode", "softLivestreamPrivacyMode"}`. A separate reactive path (`enabled_state_factory`) tried to append `"hardPrivacyMode"` when a live `HardPrivacyModeOn` report came in, but any subsequent profile/capability reconciliation (triggered by any of several unrelated feature-map attribute reports) compared current state against that hardcoded baseline, saw a mismatch, and reset `supportedAttributes` back down to the 2-item list, silently clobbering `hardPrivacyMode` again.

### Fix

- Extended `camera_av_stream_management_attribute_list_handler` to also detect `HardPrivacyModeOn` presence in the reported `AttributeList` (same pattern already used there for `StatusLightEnabled`/`StatusLightBrightness`), and persist it via a new `update_hard_privacy_mode_attribute_presence`.
- `build_camera_privacy_supported_attributes` now takes device and conditionally includes `"hardPrivacyMode"` based on that persisted presence, instead of a static 2-item list.
- Since the baseline itself now reflects the device's real capabilities, reconciliation no longer clobbers a dynamically-added `hardPrivacyMode` entry.

## Summary of Completed Tests

- Tested on a Matter Camera that had `hardPrivacyMode` support (privacy feature bit + presence of `HardPrivacyModeOn` in attributes list) but wasn't previously reflecting it in the `cameraPrivacyMode:: supportedAttributes`.
- Tested on a Matter Camera that supports privacy feature but doesn't support `hardPrivacyMode`
- Tested on a Matter camera that doesn't support privacy feature at all (so no `cameraPrivacyMode` in the profile).
- Added a unit test that validates the capability based on the presence of relevant attributes in `Attributelist`

### DIFF
--- a/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
+++ b/drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
@@ -450,7 +450,7 @@ end
 
 function CameraAttributeHandlers.camera_av_stream_management_attribute_list_handler(driver, device, ib, response)
   if not ib.data.elements then return end
-  local status_light_enabled_present, status_light_brightness_present = false, false
+  local status_light_enabled_present, status_light_brightness_present, hard_privacy_mode_present = false, false, false
   local attribute_ids = {}
   for _, attr in ipairs(ib.data.elements) do
     if attr.value == clusters.CameraAvStreamManagement.attributes.StatusLightEnabled.ID then
@@ -459,6 +459,8 @@ function CameraAttributeHandlers.camera_av_stream_management_attribute_list_hand
     elseif attr.value == clusters.CameraAvStreamManagement.attributes.StatusLightBrightness.ID then
       status_light_brightness_present = true
       table.insert(attribute_ids, clusters.CameraAvStreamManagement.attributes.StatusLightBrightness.ID)
+    elseif attr.value == clusters.CameraAvStreamManagement.attributes.HardPrivacyModeOn.ID then
+      hard_privacy_mode_present = true
     end
   end
   local component_map = device:get_field(fields.COMPONENT_TO_ENDPOINT_MAP) or {}
@@ -469,6 +471,7 @@ function CameraAttributeHandlers.camera_av_stream_management_attribute_list_hand
   }
   device:set_field(fields.COMPONENT_TO_ENDPOINT_MAP, component_map, {persist=true})
   camera_cfg.update_status_light_attribute_presence(device, status_light_enabled_present, status_light_brightness_present)
+  camera_cfg.update_hard_privacy_mode_attribute_presence(device, hard_privacy_mode_present)
   camera_cfg.reconcile_profile_and_capabilities(device)
 end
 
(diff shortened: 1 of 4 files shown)


ldeora:
The fix looks good for the `PRIV + HardPrivacyModeOn` case, and using `AttributeList` as the source of truth for `HardPrivacyModeOn` presence is definitely the right direction.

One standards-related edge case still seems uncovered, though:

`HardPrivacyModeOn` does **not** depend on the `PRIV` feature. In the [Matter 1.5 Camera AV Stream Management data model](https://github.com/project-chip/connectedhomeip/blob/v1.5.0.0/src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml), the two soft privacy attributes are conditionally mandatory when `PRIV` is supported, while `HardPrivacyModeOn` has independent optional conformance.

So this is a valid combination:

```text
PRIV feature:                 absent
SoftRecordingPrivacyMode:    absent
SoftLivestreamPrivacyMode:   absent
HardPrivacyModeOn:           present
```

The PR correctly detects `HardPrivacyModeOn` from `AttributeList`, but `match_profile()` still adds `cameraPrivacyMode` only when `Feature.PRIVACY` is present.

I think the profile decision should therefore be based on either condition:

```text
PRIV present
OR
HardPrivacyModeOn present in AttributeList
```

and the capability metadata should then be built independently:

```text
PRIV
  -> softRecordingPrivacyMode
  -> softLivestreamPrivacyMode
  -> corresponding setter commands

HardPrivacyModeOn present
  -> hardPrivacyMode
```

For a hard-privacy-only camera, `supportedAttributes` should therefore contain only:

```text
["hardPrivacyMode"]
```

and the soft-privacy setter commands should not be advertised.

One smaller cleanup: now that `AttributeList` is authoritative for hard-privacy capability presence, I think the old `enabled_state_factory()` path that dynamically appends `"hardPrivacyMode"` to `supportedAttributes` can be removed. Otherwise there are still two separate paths modifying the same capability metadata. The attribute handler can simply report the enabled/disabled state, while supported capability metadata comes from `AttributeList`.

Other than that, the approach looks good.

samadDotDev:
@ldeora good callouts! The commit https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3201/commits/9fe802234aca929963540bb15e8c2b2e99dda50b (https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3201/commits/c66b601d3cfea1089417114af946189d60803643 after rebase) should address your feedback.

ldeora:
> @ldeora good callouts! The commit [9fe8022](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/9fe802234aca929963540bb15e8c2b2e99dda50b) ([c66b601](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/commit/c66b601d3cfea1089417114af946189d60803643) after rebase) should address your feedback.

Thanks! I reviewed it and this addresses both points I raised.

The added regression coverage for the no-"PRIV" / hard-privacy-only case looks good as well.

Looks good from my side.

hcarter-775 (approved):

ctowns (approved):

PR #3204, opened 2026-08-29
feat(zigbee): add SONOFF TRV-ZBT thermostat

QuAnnan2019:
## Summary
- Add SONOFF TRV-ZBT fingerprint and standard-capability thermostat profile.
- Add SONOFF-specific handling for thermostat state, battery percentage, work mode, and Bluetooth pairing.
- Add integration coverage for reports, commands, configuration, and refresh.

## Validation
- Verified with a physical SONOFF TRV-ZBT device.
- `test_sonoff_trv_zbt.lua`: 7/7 passed.
- Lua syntax checks passed for the SONOFF driver and test files.

### DIFF
--- a/drivers/SmartThings/zigbee-thermostat/profiles/thermostat-sonoff-trv-zbt.yml
+++ b/drivers/SmartThings/zigbee-thermostat/profiles/thermostat-sonoff-trv-zbt.yml
@@ -1,0 +1,27 @@
+name: thermostat-sonoff-trv-zbt
+components:
+- id: main
+  capabilities:
+  - id: temperatureMeasurement
+    version: 1
+  - id: thermostatHeatingSetpoint
+    version: 1
+    config:
+      values:
+      - key: "heatingSetpoint.value"
+        range: [5, 30]
+  - id: thermostatMode
+    version: 1
+  - id: battery
+    version: 1
+  - id: refresh
+    version: 1
+  categories:
+  - name: Thermostat
+- id: bluetooth
+  label: Bluetooth Pairing
+  capabilities:
+  - id: momentary
+    version: 1
+  categories:
+  - name: Button
--- a/drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
+++ b/drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
@@ -1,0 +1,240 @@
+-- Copyright 2026 SmartThings, Inc.
+-- Licensed under the Apache License, Version 2.0
+
+local capabilities = require "st.capabilities"
+local clusters = require "st.zigbee.zcl.clusters"
+local data_types = require "st.zigbee.data_types"
+local device_management = require "st.zigbee.device_management"
+local generic_body = require "st.zigbee.generic_body"
+local messages = require "st.zigbee.messages"
+local zcl_messages = require "st.zigbee.zcl"
+local zb_const = require "st.zigbee.constants"
+local utils = require "st.utils"
+
+local PowerConfiguration = clusters.PowerConfiguration
+local Thermostat = clusters.Thermostat
+
+local Battery = capabilities.battery
+local Momentary = capabilities.momentary
+local Refresh = capabilities.refresh
+local TemperatureMeasurement = capabilities.temperatureMeasurement
+local ThermostatHeatingSetpoint = capabilities.thermostatHeatingSetpoint
+local ThermostatMode = capabilities.thermostatMode
+
+local SONOFF_CLUSTER = 0xFC11
+local WORK_MODE_ATTR = 0x0018
+local BUTTON_COMMAND = 0x10
+local BLUETOOTH_PAIRING_START = 0x01
+
+local WORK_MODE_OFF = 0x01
+local WORK_MODE_MANUAL = 0x03
+local WORK_MODE_SCHEDULE = 0x04
+local WORK_MODE_TEMP_MANUAL = 0x05
+
+local MIN_SETPOINT = 5
+local MAX_SETPOINT = 30
+local SUPPORTED_MODES = {
+  ThermostatMode.thermostatMode.off.NAME,
+  ThermostatMode.thermostatMode.heat.NAME,
+  ThermostatMode.thermostatMode.auto.NAME,
+}
+
+local function emit_setpoint_range(device)
+  device:emit_event(ThermostatHeatingSetpoint.heatingSetpointRange({
+    value = { minimum = MIN_SETPOINT, maximum = MAX_SETPOINT },
+    unit = "C",
+  }, { visibility = { displayed = false } }))
+end
+
+local function emit_supported_modes(device)
+  device:emit_event(ThermostatMode.supportedThermostatModes(SUPPORTED_MODES, {
+    visibility = { displayed = false },
+  }))
+end
+
+local function emit_mode(device, mode)
+  device:emit_event(ThermostatMode.thermostatMode[mode]())
+end
+
+local function endpoint_for(device, cluster)
+  return device:get_endpoint(cluster) or 1
+end
+
+local function send_bluetooth_pairing_start(device)
+  local zcl_header = zcl_messages.ZclHeader({
+    cmd = data_types.ZCLCommandId(BUTTON_COMMAND),
+  })
+  zcl_header.frame_ctrl:set_cluster_specific()
+
+  device:send(messages.ZigbeeMessageTx({
+    address_header = messages.AddressHeader(
+      zb_const.HUB.ADDR,
+      zb_const.HUB.ENDPOINT,
+      device:get_short_address(),
+      endpoint_for(device, SONOFF_CLUSTER),
+      zb_const.HA_PROFILE_ID,
+      SONOFF_CLUSTER
+    ),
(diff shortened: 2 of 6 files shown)


drivers/SmartThings/zigbee-thermostat/profiles/thermostat-sonoff-trv-zbt.yml
```diff
@@ -0,0 +1,27 @@
...
+    version: 1
+  categories:
+  - name: Thermostat
+- id: bluetooth
+  label: Bluetooth Pairing
+  capabilities:
+  - id: momentary
+    version: 1
+  categories:
+  - name: Button
```
KKlimczukS, line 27: Please elaborate why Bluetooth Pairing is needed here for zigbee device ?
Is related to any peripheral devices ?

drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
```diff
@@ -0,0 +1,252 @@
...
+  ThermostatMode.thermostatMode.off.NAME,
+  ThermostatMode.thermostatMode.heat.NAME,
+  ThermostatMode.thermostatMode.auto.NAME,
+}
+
+local function emit_setpoint_range(device)
+  device:emit_event(ThermostatHeatingSetpoint.heatingSetpointRange({
+    value = { minimum = MIN_SETPOINT, maximum = MAX_SETPOINT, step = SETPOINT_STEP },
+    unit = "C",
+  }, { visibility = { displayed = false } }))
```
KKlimczukS, line 87: Please reformat this one

drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
```diff
@@ -0,0 +1,252 @@
...
+  }))
+end
+
+local function emit_mode(device, mode)
+  device:emit_event(ThermostatMode.thermostatMode[mode]())
+end
+
+local function endpoint_for(device, cluster)
+  return device:get_endpoint(cluster) or 1
+end
```
KKlimczukS: Do You need a separate function that is called just once in this subdriver ?

drivers/SmartThings/zigbee-thermostat/profiles/thermostat-sonoff-trv-zbt.yml
```diff
@@ -0,0 +1,27 @@
+name: thermostat-sonoff-trv-zbt
```
KKlimczukS, line 1: Since this device supports child lock function. 
Do You consider exposing it in the device profile / device preferences and subdriver?

drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
```diff
@@ -0,0 +1,252 @@
...
+  local hub_eui = driver.environment_info.hub_zigbee_eui
+  device:send(device_management.build_bind_request(device, Thermostat.ID, hub_eui))
+  device:send(device_management.build_bind_request(device, PowerConfiguration.ID, hub_eui))
+  device:send(device_management.build_bind_request(device, SONOFF_CLUSTER, hub_eui))
+  device:send(Thermostat.attributes.LocalTemperature:configure_reporting(device, 10, 300, 50))
+  device:send(Thermostat.attributes.OccupiedHeatingSetpoint:configure_reporting(device, 1, 600, 50))
+  device:send(Thermostat.attributes.SystemMode:configure_reporting(device, 1, 600, 1))
+  device:send(PowerConfiguration.attributes.BatteryPercentageRemaining:configure_reporting(device, 30, 21600, 5))
+  refresh(driver, device)
+end
```
KKlimczukS, line 148: You may define all configuration for each attribute, example here: 
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/4a77e6fab7a3eecdc5a47807ef911a56a2ef2f4b/drivers/SmartThings/zigbee-button/src/frient/init.lua#L37

and then use it in init handler, example: 
https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/4a77e6fab7a3eecdc5a47807ef911a56a2ef2f4b/drivers/SmartThings/zigbee-button/src/frient/init.lua#L147

then the framework will build bindings and configure reporting for Your attributes.

drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
```diff
@@ -0,0 +1,252 @@
...
+  local system_mode = mode_to_system_mode[mode]
+  if system_mode == nil then
+    return
+  end
+
+  device:send(Thermostat.attributes.SystemMode:write(device, system_mode))
+  emit_mode(device, mode)
+  device.thread:call_with_delay(1, function()
+    device:send(Thermostat.attributes.SystemMode:read(device))
+  end)
```
KKlimczukS, line 246: Does Your device need to be polled this way ? Once You configure reporting (see my previous comment) then it should report changes without being polled.

cjswedes: I believe this is how we manage the ThermostatMode in zigbee. I think we are relying on the Thermostat cluster binding to send a report if the mode changes out of band (no periodic reports), and then we issue reads after a command. I am not really sure why we dont have a ThermostatMode default and an attr reporting configuration for this.

drivers/SmartThings/zigbee-thermostat/profiles/thermostat-sonoff-trv-zbt.yml
QuAnnan2019: The TRV-ZBT is a Zigbee device for SmartThings onboarding and normal control,
but the product also has a separate Bluetooth radio and supports a
Bluetooth-pairing workflow required by SONOFF.

Zigbee remains the only SmartThings transport for this device:
- The TRV joins the SmartThings Hub over Zigbee.
- Temperature, setpoint, mode, battery, and all normal device control use
  Zigbee clusters.
- This component does not add a Bluetooth device to SmartThings and does not
  change the Zigbee onboarding path.

The Bluetooth Pairing button is an explicit user action requested by SONOFF.
When a user presses it in the SmartThings app, the Edge driver sends SONOFF's
documented proprietary Zigbee command to the already-joined TRV
(cluster 0xFC11, command 0x10, payload 03 01 01). The TRV then enables its
own Bluetooth pairing mode.

In other words, SmartThings uses the existing Zigbee connection only as the
control path to request that the dual-radio TRV enter Bluetooth pairing mode.
The Bluetooth pairing process itself is performed by the TRV and is separate
from SmartThings Zigbee onboarding.

The momentary child component is used because this is a one-shot action:
there is no persistent SmartThings state to expose, only a request to enter
Bluetooth pairing mode.

drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
QuAnnan2019: Thank you for the clarification.  
The delayed SystemMode read is intentionally a one-time post-command reconciliation, not periodic polling. After writing a requested mode, it confirms the mode that the TRV actually accepted and keeps the app state aligned if the device rejects or normalizes the request.  
Changes made outside SmartThings, such as through physical device interaction, are handled through Thermostat cluster binding and attribute reporting. In the latest update, SystemMode is also registered in the subdriver CONFIGURATIONS, so the framework configures the corresponding binding and reporting automatically.  
We will retain the single confirmation read after a command while relying on reporting for normal state changes.

drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
QuAnnan2019: Thank you for the suggestion. 
We have updated the subdriver to define the Thermostat and Power Configuration reporting entries in a CONFIGURATIONS table and register them during init with device:add_configured_attribute. The doConfigure handler now uses device:configure(), allowing the framework to create the required bindings and configure reporting automatically.

drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
QuAnnan2019: Good point. 
The helper was only used once, so it has been removed and the endpoint lookup is now inlined at the call site.

drivers/SmartThings/zigbee-thermostat/src/sonoff/init.lua
QuAnnan2019: OK. Reformatted in the latest update for readability.

drivers/SmartThings/zigbee-thermostat/profiles/thermostat-sonoff-trv-zbt.yml
QuAnnan2019: Thank you for the suggestion. 
Child lock has been added as a standard device preference in the profile.  
The subdriver handles preference changes through infoChanged and writes SONOFF proprietary cluster 0xFC11, attribute 0x0000 as a Boolean: true (0x01) locks the physical input and false (0x00) unlocks it. Integration coverage has been added for both values.

PR #3205, opened 2026-08-31
Zigbee Switch: Fix Climax power meter scaling

### LINKED ISSUE #236: ClimaxTechnology Power Meter | power unit divisor
Hi,
the power switch with meter mfr = "ClimaxTechnology", model = "PSM_00.00.00.35TC" have a wrong divisor key.
 
It display 800 mW (milliwatts) instead of 800 W (watts), so its incorrect by a factor of 1000.

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/29ccb18212bc1d84168d17722f9d437b2b5ed2e8/drivers/SmartThings/zigbee-switch/src/zigbee-switch-power/init.lua#L30

Thanks,
Pablo

sungmpar:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [x] Bug fix
- [ ] New feature
- [ ] Refactor

# Checklist

# Description of Change

## Summary

- Add an exact Climax-specific nested power-meter subdriver.
- Delegate Electrical Measurement `ActivePower` to the SDK-standard handler.
- Delegate Simple Metering `InstantaneousDemand` to the SDK-standard handler.
- Add regression tests covering multiplier, divisor, and kW-to-W conversion.

## Root cause

The shared `zigbee-switch-power` compatibility handlers divide reported values by the configured divisor but do not apply the corresponding multiplier. The shared `InstantaneousDemand` path also emits a kW-scaled value as watts without multiplying by 1000.

Changing these shared handlers globally could alter established compatibility behavior for other fingerprinted devices.

## Implementation

The nested subdriver matches only manufacturer `ClimaxTechnology` and model `PSM_00.00.00.35TC`. It overrides only Electrical Measurement `ActivePower` and Simple Metering `InstantaneousDemand`, delegating both attributes to the existing SDK-standard handlers.

The shared compatibility handlers and all non-Climax device behavior remain unchanged.

## Hardware validation

No SmartThings hub or physical Climax device was used. Automated integration tests verify driver and subdriver routing, multiplier/divisor field use, numerical scaling, kW-to-W conversion, and emitted `powerMeter` events.

The physical-device test plan is to pair a `PSM_00.00.00.35TC` with a SmartThings hub and verify report ordering, endpoint behavior, and emitted watt values against a known load when hardware is available. That hardware validation was not performed for this change, so firmware report ordering, endpoint quirks, and deployed-device behavior remain unverified.

Fixes #236

# Summary of Completed Tests

- `test_switch_power`: 7/7 passed
- `test_aurora_relay`: 4/4 passed
- `test_ledvance_metering_plug`: 4/4 passed
- Complete Zigbee-switch suite: 63 files, 575/575 tests passed
- Luacheck: 0 warnings and 0 errors in the four changed files
- Whitespace checks passed


sungmpar:
Hi @cjswedes, I hope you're doing well.

Whenever you have a chance, would you or someone on the team be able to take a look at this PR?

I noticed that the `Run driver tests`, `Luacheck`, and `Upload driver packages` workflows are still awaiting maintainer approval.

The change is intentionally scoped to `ClimaxTechnology / PSM_00.00.00.35TC` so that the existing shared compatibility behavior for other devices remains unchanged. It addresses #236 by routing `ActivePower` and `InstantaneousDemand` through the SDK-standard handlers, with regression tests covering the scaling behavior.

If a different approach would be preferable, or if any additional validation is needed, I would be very happy to make any necessary changes.

Thank you very much for your time and for reviewing contributions to the project.

cjswedes:
Thank you for putting up this change to address this issue, I think there is a simpler way to address the issue rather than adding another nested subdriver. 

First, some context. The zigbee-switch-power subdriver itself is a bit useless IMO. Basically all these subdrivers do is set divisor/multipliers to be used in the `ActivePower` and `InstantaneousDemand` report handlers. It then proceeds to override those handlers with ones that do not consider the multiplier, as you have mentioned. I am not sure if changing those handlers will cause an issue with other devices. The subdriver exists because when we switched from groovy DTHs to Lua edge drivers, we tried to do a 1 to 1 port of functionality; likely the defaults are sufficient for all these devices, like the ClimaxTechnology device you are fixing.

Since the fix here is to use the defaults, rather than adding nested subdriver which then overrides the already overridden defaults with the defaults again, we should just have this device stop using the zigbee-switch-power subdriver entirely. Remove it from the fingerprints.lua [file](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/19bb6f9b75a4a7590dfb5c5f9aed3bbf3308c77c/drivers/SmartThings/zigbee-switch/src/zigbee-switch-power/fingerprints.lua#L13) and then your tests pass.  You then do not need the nested subdriver.

cjswedes:
@sungmpar please update this to include the latest from `main`. Hopefully all tests will be passing and we can get it merged.

sungmpar:
Thanks @cjswedes! I've updated the branch with the latest main.

The Jenkins driver test matrix is now passing on the updated head. I also reran test_switch_power locally and it passes 7/7, with Luacheck reporting 0 warnings / 0 errors for the two changed files.

The remaining GitHub Actions workflows are currently awaiting maintainer approval.

PR #3215, opened 2026-09-07
Hager rotary dimmer implementation

JanJakubiszyn:
Check all that apply

# Type of Change

          - [ ] I have signed the CLA
- [ ] Bug fix
- [ ] New feature
- [X] Refactor

# Checklist

# Description of Change
Adding rotary dimmer device support for Hager.

# Summary of Completed Tests


drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -33,6 +34,12 @@ local function subscribe (device, endpoint_id, cluster_id, attr_id, event_id)
   device:send(cluster_base.subscribe(device, endpoint_id, cluster_id, attr_id, event_id))
 end
 
+local function is_standalone_device (device)
```
hcarter-775, line 37: what does this function imply?

JanJakubiszyn: This function is created to separate standalone devices which are from WAASYS series but do not utilize the matter bridge device type. They differ in the endpoint architecture and have to be handled in a different way but they also support manually set functions like adding a child device responsible for remote light control. 

This function allows to determine this type of device and eliminate/include them the driver behavior.

for example, standalone devices do not require mapping done in the info_changed function.

hcarter-775: So you are saying that these are just devices from Hager, rather than bridged devices from the Hager Bridge? I am curious, but I also think this would be a worthwhile thing to explain in a comment, since `is_standalone` does not express a lot (in my opinion) about what is being checked, especially when the internals are extremely device specific.

I also suggest that you add some commentary in the function itself to explain what is being checked, since the endpoint id checks are not understandable without Hager-specific context.

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
```diff
@@ -356,10 +380,19 @@ local function device_type_handler (driver, device, ib)
   local new_btn_eps = {}
   local ep_id = ib.endpoint_id
   local value = ib.data.elements
+  local reports_dimmable = false
+
+  for _, element in ipairs(value) do
```
hcarter-775, line 385:
```suggestion
  for _, element in ipairs(value or {}) do
```

drivers/SmartThings/matter-switch/src/sub_drivers/hager/init.lua
JanJakubiszyn: Yes, they are just simple Hager devices without the bridge functionality. 

I will change the function name to "is_non_bridge_type" and add comments for better clarity 
and understanding of the architectural differences between them.

hcarter-775: Yeah, I'd maybe just check then if the device is 1. an Aggregator device type and 2. is a Matter (not EDGE child) device? Or something like that. If it isn't, then we'd know it is a non-bridge Matter device. I am not sure we need this extremely specific endpoint checking to know whether it is a regular device or not

JanJakubiszyn: I've renamed the function and added comments for clarity.

I couldn't implement the device type check quite as suggested. The Aggregator type only shows up on the parent right after pairing, and once child devices are created (still as Matter type, not EDGE children) it's no longer there, so that check alone doesn't reliably tell bridge children apart from an actual non-bridge device like the Rotary Dimmer. It seemed a bit too timing-dependent, so I went with a static per-product flag in `fields.lua` instead.

PR #3227, opened 2026-09-10
Add tool script for testing individual drivers

hcarter-775:
# Description of Change
Add a `test_driver.sh` script, along with a one-time setup script, to the tools directory, so that developers and agents have access to a slightly more ergonomic driver unit testing framework.

Some context: I have been using basically this script for a couple years now, and am trying to bring this to the more generic experience (like for AI tools). I had AI mostly generate the AGENTS documentation, and the QOL script for setting up an alias. The actual `test_driver.sh` script was pre-existing for me

### DIFF
--- a/tools/test_driver.sh
+++ b/tools/test_driver.sh
@@ -1,0 +1,340 @@
+#!/bin/bash
+#
+# test_driver.sh
+#
+# Run the Lua unit tests for a single SmartThings Edge Driver.
+#
+# Run this from the driver's `src` directory, e.g.:
+#   cd drivers/SmartThings/matter-switch/src
+#   $(git rev-parse --show-toplevel)/tools/test_driver.sh [options]
+#
+# The $(git rev-parse --show-toplevel) form finds the repo root regardless of
+# how deep you are, so it works the same from any driver without figuring out
+# a relative path back to tools/.
+#
+# (If you run it from the driver's root directory instead, it will `cd src`
+# for you automatically.)
+#
+# Run with -h for the full list of options.
+
+usage() {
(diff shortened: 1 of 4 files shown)


tools/test_driver.sh
```diff
@@ -0,0 +1,340 @@
...
+        export LUA_PATH="$LUA_PATH${rocks_lua_path:+;$rocks_lua_path}"
+    else
+        export LUA_PATH="./?.lua;./?/init.lua${rocks_lua_path:+;$rocks_lua_path}"
+    fi
+    echo "LUA_PATH set to: $LUA_PATH"
+fi
+
+# ---------------------------------------------------------------------------
+# -r: luacheck (mirrors the repo's .github/workflows/.luacheckrc — keep the
+# two in sync if that file changes)
```
cjswedes, line 151: We should not have to update 2 places to use this

tools/test_driver.sh
cjswedes, line 1: IMO this is a somewhat complex bash script and python is more maintainable. That aside, it is another barrier for developers on windows, which is the main reason why we should not add this script.

cjswedes (commented):
I see why you have been using this. However, rather than having two different test scripts, I think I would prefer to see new functionality integrated into the existing script (luacheck, installable to PATH, output file option, etc.).

If there is a different format for output that you think works better for agents (to save tokens or whatever the reason), we could add a new option that changes output to the existing script.

I dont love jamming everything in one script, but it would be useful for development to have luacheck/coverage and test running possible with one command. But I dont think we want to maintain this script along the existing one.

PR #3239, opened 2026-09-17
WWST: add Zemismart KES606 four-gang relay/scene switch

zemismart-dev:
# Type of Change

  - [x] I have signed the CLA
- [ ] Bug fix
- [x] New feature
- [ ] Refactor

# Checklist

# Description of Change

Add support for the Zemismart KES-606US-L4 composite wall switch, matching only `_TZ3000_hurauima / TS0726`, to the Zigbee Switch driver. The parent is a display/configuration container and creates four Edge children (`gang1` through `gang4`). Each child's Mode preference switches between relay control (`switch` + `refresh`) and a single-press scene button (`button`).

The sub-driver maps the four relay endpoints, real scene commands, mode reports, indicator settings, countdown, and global/per-gang power-on settings. Relay attribute reports and initialization never generate a synthetic button press. When the asynchronous profile change back to relay mode completes, the driver reads that endpoint's OnOff state so a report received during the transition cannot leave stale UI state. The generic Tuya sub-driver excludes this exact fingerprint to prevent duplicate lifecycle handling. A dedicated driver-switch handler configures the exact parent and provisions relay/scene children. A KES606-specific OnOff DefaultResponse override prevents command acknowledgements from generating unsupported parent switch events or attempting native attribute registration; actual child state still comes from attribute reports/read responses.

The three profile names are retained from the existing manufacturer implementation for continuity. Despite the legacy `ts0601` profile names, this PR adds only the TS0726 fingerprint above. It adds no custom capability and no other device models.

Protocol reference: the exact fingerprint is included in [zigbee-herdsman-converters' TS0726 four-gang definition](https://github.com/Koenkk/zigbee-herdsman-converters/blob/4ed6cf0e261ef9944399bd931dbfd3fdd962f3be/src/devices/tuya.ts#L25040). This implementation is adapted from the manufacturer's existing Edge driver, not a port of the Zigbee2MQTT code.

## Maintainer / certification review requested

This is a **draft** supporting the existing KES-606US-L4 WWST application. Please advise on acceptance of:

- The device's manufacturer-specific `0xE001` mode/power-on attributes, OnOff vendor attributes, and scene command `0xFD`.
- The embedded preference definitions, or the approved shared preferences that should replace them without changing their meaning.
- A refresh-only parent with four independently switchable relay/scene children. Both the existing implementation and candidate tests pass the child automation-presentation checks, while the parent fails the generic requirement for an automation action or condition. We are requesting review of that design; no exception approval is assumed.

# Summary of Completed Tests

- Current commit: **94/94 SDK integration tests passed** with official API v19 (`apiv19_60_beta`) and API v21 (`apiv21_62`) libraries, on genuine Lua 5.3 (Lupa 2.8).
- The 32 new DefaultResponse cases cover all four endpoints, relay/scene modes, On/Off and success/failure responses. Before the fix, all 16 successful-response cases reproduced unsupported parent events and native-registration attempts; after the fix all pass. Subsequent real attribute-report routing is checked in every case.
- Independent source/test review passed. Luacheck 1.2.0 analysis API with repository configuration reports zero warnings/errors on the changed Lua files; the unmodified analyzer is invoked through a local file-discovery/output wrapper.
- Build-only packaging and `git diff --check` passed.
- Current commit `e1c0c3c` passed all six upstream Driver Tests jobs: 57 (4055), 58 (4852), 59 (5231), 60 (5328), 61 (5344), and dev (5344), all with zero failures. CLA status is successful.

## Hardware validation and remaining certification review

Latest candidate package `2026-09-17T04:56:18.385495812` runs this commit on the physical TS0726 sample with one parent and four provisioned children.

- Full Integration Test **6MUUP3G0YH**: **46 passed, 1 failed, 0 skipped/excluded/errors**. Nearby-scan onboarding took 104 seconds. All eight relay commands and parent/four-child removal passed, with operator confirmation of independent operation and all loads off.
- Physical preference verification covered all four countdowns and zero cancellation, all three indicator modes, per-gang Off/On/Memory (including remembered on and off), and global Off/On/Memory. Four per-gang and three global configurations were verified by actual sample power interruptions and protocol readback. Global power-on writes were observed to update all four per-gang registers; saved UI selections are not treated as hardware readback.
- Latest formal scene attempt **XQA0I9YQF8** remains **17 passed, 5 failed, 13 excluded**: parent presentation and four manual button cases. Operator/browser coordination did not complete within the test windows. L1 arrived near the extended deadline and was not confirmed by the suite; the other three were pressed later. This report is not represented as passing.
- A separate continuous-log physical check on the latest package verified all four real `0xFD` presses, correct child `pushed` events and cloud timestamps, with no synthetic initialization press. This is supplemental evidence, not a replacement passing formal report.
- After restoring all four children to Switch, actual mode registers were zero and profiles exposed switch/refresh. A final eight-command regression in L4/L3/L2/L1 order passed without another Hub reboot. Protocol and cloud states agreed; the operator confirmed no cross-routing and all physical loads off. The captured final sequence had eight dedicated acknowledgements, zero native-registration attempts and zero ERROR/FATAL messages.

Earlier candidate `2026-09-17T03:24:02.94756905` reports remain version-specific supplemental evidence: **GH536AKCXW** (29 passed, 1 parent failure, 17 excluded) and **X90YMMGWXX** (21 passed, 1 parent failure, 13 excluded; all four physical button events passed).

During the earlier Scene-to-Switch regression, L3 physically turned off and emitted correct local off reports, but cloud state remained on. The separate DefaultResponse conflict was fixed here, but installing that fix alone did not clear the stale cloud state. An approved test-Hub reboot restored synchronization; the subsequent latest-package full suite and final mode-transition regression did not reproduce it. **The earlier cloud-state stall's root cause remains unproven.**

The full suite's sole failure is the refresh-only parent's automation-presentation requirement; child automation-presentation checks passed. Please review the parent design and supplemental scene evidence. This PR remains a draft. The WWST application was submitted on September 17, 2026 with the full report and these evidence limitations. No exception or certification approval is assumed.


drivers/SmartThings/zigbee-switch/src/zemismart-kes606/can_handle.lua
```diff
@@ -0,0 +1,21 @@
...
+    local gang = tonumber((device.parent_assigned_child_key or ""):match("^gang([1-4])$"))
+    if gang == nil then return false end
+    parent = device:get_parent_device()
+  elseif device.network_type ~= device_lib.NETWORK_TYPE_ZIGBEE then
+    return false
+  end
+  if parent ~= nil and parent:get_manufacturer() == "_TZ3000_hurauima"
+      and parent:get_model() == "TS0726" then
+    return true, require("zemismart-kes606")
+  end
```
cjswedes, line 19: this can_handle is quite confusing:
1. The `get_manufacturer` and `get_model` commands when called on a zigbee child device are overwritten to lookup the mfr/model on the parent device.
2. The checking of if the parent_assigned_child_key parses correctly should not be done in this function. If that fails here, there are no other subdrivers/handlers where it will succeed, so we should just let it fail in this subdriver so that any errors (which shouldnt happen) are actually associated with the code that created the child devices

```suggestion
  if device ~= nil and device:get_manufacturer() == "_TZ3000_hurauima"
      and device:get_model() == "TS0726" then
    return true, require("zemismart-kes606")
  end
  return false
```

drivers/SmartThings/zigbee-switch/src/tuya-multi/can_handle.lua
```diff
@@ -12,6 +12,10 @@ local function is_multi_endpoint(device)
 end
 
 return function(opts, driver, device)
+  -- The composite relay/scene device has its own configure and refresh handlers.
+  if device:get_manufacturer() == "_TZ3000_hurauima" and device:get_model() == "TS0726" then
+    return false
+  end
```
cjswedes, line 18: Rather than excluding like this, we should have the zemismart-kes606 subdriver be a nested subdriver of this Tuya multi subdriver. That will prevent it from using the handlers here that are overridden by the new zemismart subdriver.

drivers/SmartThings/zigbee-switch/src/zemismart-kes606/init.lua
```diff
@@ -0,0 +1,378 @@
...
+        [MODE_ATTRIBUTE] = mode_report,
+        [POWER_ON_ATTRIBUTE] = power_on_report,
+      },
+    },
+    cluster = {
+      [OnOff.ID] = {
+        [OnOff.server.commands.On.ID] = scene_command,
+        [OnOff.server.commands.Off.ID] = scene_command,
+        [OnOff.server.commands.Toggle.ID] = scene_command,
+        [SCENE_PRESS_COMMAND] = scene_command,
```
cjswedes: Is there a real device that reports its scene press as an OnOff server command (On/Off/Toggle) rather than 0xFD? If so please add a test for it. As written, if a device ever sent both an OnOff command and 0xFD for a single physical press, this would emit two pushed events. If it's not actually needed, consider dropping the On/Off/Toggle registrations.

drivers/SmartThings/zigbee-switch/src/zemismart-kes606/init.lua
```diff
@@ -0,0 +1,378 @@
...
+  local parent = parent_from_device(device)
+  if gang == nil or parent == nil or gang_mode(parent, gang) == "scene" then return end
+  local command = on and OnOff.server.commands.On(parent) or OnOff.server.commands.Off(parent)
+  parent:send(command:to_endpoint(gang))
+end
+
+local function onoff_report(_, device, value, zb_rx)
+  local gang = endpoint_from_rx(zb_rx)
+  local parent = parent_from_device(device)
+  if gang == nil or parent == nil or gang_mode(parent, gang) == "scene" then return end
```
cjswedes, line 172: Can you clarify why gang_mode (the persisted field) and child:supports_capability are both consulted in some paths (onoff_report) but only the field is consulted in others (relay_command)?

drivers/SmartThings/zigbee-switch/profiles/ts0601-scene-switch-4-gang.yml
```diff
@@ -0,0 +1,69 @@
+name: ts0601-scene-switch-4-gang
+components:
+  - id: main
+    capabilities:
+      - id: refresh
+        version: 1
+    categories:
+      - name: RemoteController
+preferences:
+  - title: Indicator Light
```
cjswedes, line 10: Regarding your review request:

> The embedded preference definitions, or the approved shared preferences that should replace them without changing their meaning.

We would prefer you use the shared preferences where possible

cjswedes (commented):
> A refresh-only parent with four independently switchable relay/scene children. Both the existing implementation and candidate tests pass the child automation-presentation checks, while the parent fails the generic requirement for an automation action or condition. We are requesting review of that design; no exception approval is assumed.

I am not sure what to do about this; it may need an exception. My only idea is that the parent could have a switch and button capability on its main component: 
- button is pressed whenever a child device button is pressed (this is inline with functionality of other zigbee button devices)
- switch controls all child devices that are in the non-scene mode, and if 1  child is on then the parent is on, otherwise it is off.

PR #3242, opened 2026-09-21
Matter Camera: Exclude `audioVolume` on unusable volume range

samadDotDev:
# Description of Change

In Matter spec, there’s no specific feature-map based gating on microphone/speaker volume settings, outside of the larger Audio (ADO) feature i.e. any camera device supporting audio needs to implement these settings even if the hardware doesn’t support adjusting volume/gain for it. We have encountered one such camera and have confirmed with the partner that this setting cannot be implemented for their hardware. 

Upon following up with Camera TT in CSA, it was noted as a miss that will be followed up separately (to either change the conformance of volume settings from `ADO` to `[ADO]`, or introduce a separate feature map bit for audio gain control) in the future versions of the spec. However to resolve issues in the field and currently available versions of the spec, we found a backward-compatible workaround (which was confirmed to be allowed per DM TT) where min and max can be set to the same value in the constraints defined for these attributes, and that can be keyed off by clients to determine that the volume range is unusable and therefore the capability can be excluded.

# Summary of Completed Tests

- Tested with Matter cameras including usable ranges
- Added unit tests exercising both cases (where these capabilities should be included or excluded)
- [Pending test once the beta firmware is released] To-test with aforementioned Matter camera that will include an unusable range.


ldeora:
The equal min/max convention seems like a reasonable compatibility mechanism for devices whose hardware cannot provide adjustable gain/volume, and handling speaker and microphone independently makes sense.

I noticed two cases that I think should be addressed.

### 1. Range discovery currently depends on `audioVolume` already being present

`should_include_volume_capability()` deliberately preserves the existing capability state while either raw bound is unknown:

```lua
if raw_min_volume == nil or raw_max_volume == nil then
  return device:supports_capability(capabilities.audioVolume, component)
end
```

The problem is that the normal camera subscription currently obtains `SpeakerMinLevel` / `SpeakerMaxLevel` and the microphone equivalents through the `audioVolume` capability mapping itself.

That creates a bootstrap dependency:

```text
audioVolume not currently enabled
    ↓
min/max are not subscribed
    ↓
raw range remains unknown
    ↓
should_include_volume_capability() returns false
    ↓
audioVolume cannot be added
```

The test setup appears to demonstrate this as well: `seed_usable_volume_range()` has to populate the raw ranges before `update_device_profile()` so that the normal camera profile still gets `audioVolume`.

I think range discovery should therefore be independent of whether `audioVolume` is currently exposed - for example, an initial READ/subscription of the relevant min/max attributes, ideally gated by the AVSM metadata/AttributeList. The discovered range can then decide whether the public capability belongs in the profile.

This also matters after `audioVolume` has been removed: if the raw fields are lost on restart, the driver otherwise has no path to discover that a later firmware/device state provides a usable range again.

### 2. The existing range validation still treats the new equal-range convention as invalid

The current handlers use:

```lua
max_volume <= min_volume
min_volume >= max_volume
```

as invalid-range conditions and replace the reported value with the fallback bound.

With the convention introduced by this PR, `min == max` is intentional and should no longer go through the malformed-range path. The raw fields preserve the original values, so the profile decision still works, but the driver's validated internal range then disagrees with the new semantics and an intentional constant range produces an "invalid" warning.

The released [Matter 1.5 camera model](https://github.com/project-chip/connectedhomeip/blob/v1.5.0.0/src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml) requires these range attributes with the corresponding `SPKR`/`ADO` features, and the Matter 1.5 [AVSM conformance test](https://github.com/project-chip/connectedhomeip/blob/v1.5.0.0/src/python_testing/TC_AVSM_2_1.py) explicitly checks:

```text
SpeakerMinLevel <= SpeakerMaxLevel
MicrophoneMinLevel <= MicrophoneMaxLevel
```

so equality is not an invalid range.

I think the driver should distinguish:

```text
min < max  → usable adjustable range
min == max → valid but non-adjustable range; omit audioVolume
min > max  → malformed range
```

and avoid attempting normal volume normalization when the known range is non-adjustable.

A regression test starting with no `audioVolume` and no cached raw range, then discovering a usable range and adding the capability, would cover the first issue nicely.

Other than these two points, the overall approach looks good.

samadDotDev:
Good catch on both. 

For # 1: you're right, it wasn't actually decoupled. `SpeakerMaxLevel`/`MinLevel` and the microphone equivalents were only subscribed as part of `audioVolume`'s own attribute group, so a device without `audioVolume` yet would never receive the values needed to grant it. Fixed by moving those four discovery attributes into the `audioMute` group instead, which is added unconditionally whenever the `AUDIO`/`SPEAKER` feature is present. Therefore, the range discovery no longer depends on `audioVolume` already being active, and it also survives `audioVolume` being dropped and re-evaluated later. Added a regression test that starts from an unseeded device and asserts the subscribe request carries the min/max attributes before `audioVolume` is granted.

For # 2: the capability-inclusion check already treats `min == max` and `min > max` the same (both omit `audioVolume`), so no change needed there. I did tighten the separate logging path in the min/max handlers, which was still using `<=/>=` and would log a false "invalid range" warning + clamp an intentional constant range. This is now changed to strict `<`/`>` per the AVSM conformance check you linked.

ldeora:
Thanks! Both issues I raised are now addressed. Moving the range attributes into the `audioMute` subscription group resolves the discovery dependency, and the strict comparisons correctly accept `min == max`.

One small remaining edge case: `volume_level_handler()` still divides by `max_volume - min_volume`. A guard for a known zero-width range would prevent a late volume report from triggering division by zero while the profile is being updated.

Other than that, the changes look good to me.

samadDotDev:
@Ideora Good catch again, confirmed with a test. Once `min == max` is accepted as valid, `volume_level_handler()`'s `(value - min) * 100 / (max - min)` divides by zero for any report that arrives before the narrowed profile (dropping `audioVolume`) round-trips back down, and the resulting nan gets rejected by the capability layer, crashing the driver. 

Added a guard: when the range is zero-width, it reports a fixed 0 instead of normalizing. Also checked the volume math in `capability_handlers.lua` (set/up/down) - `volume_range` is only ever used as a multiplier there, not a divisor, so it isn't exposed to the same failure, and those commands can't reach the driver once `audioVolume` is hidden anyway. Added a regression test that reproduces the race (narrows the range, then injects a late volume report) and confirmed it crashes without the guard and passes with it.

Kwang-Hui (approved):
Looks good to me

samadDotDev:
This has now been validated with a Matter camera that exposes unusable microphone range; the net effect is that the microphone setting is omitted from UI for that camera. I have updated the testing section of the PR to reflect that.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
```diff
@@ -59,32 +59,45 @@ function CameraAttributeHandlers.volume_level_handler(driver, device, ib, respon
   local component = device:endpoint_to_component(ib)
   local max_volume = device:get_field(camera_fields.MAX_VOLUME_LEVEL .. "_" .. component) or camera_fields.ABS_VOL_MAX
   local min_volume = device:get_field(camera_fields.MIN_VOLUME_LEVEL .. "_" .. component) or camera_fields.ABS_VOL_MIN
-  -- Convert from [min_volume, max_volume] to [0, 100] before emitting capability
   local limited_range = max_volume - min_volume
+  if limited_range <= 0 then
+    -- Non-adjustable (min == max) range: nothing to normalize against. A report can still arrive here
+    -- while the profile update dropping audioVolume is in flight, so report a fixed value instead of
+    -- dividing by zero.
```
hcarter-775: nit: I think this section is self-explanatory

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
```diff
@@ -212,18 +212,22 @@ function CameraUtils.subscribe(device)
       clusters.CameraAvSettingsUserLevelManagement.attributes.TiltMax,
       clusters.CameraAvSettingsUserLevelManagement.attributes.TiltMin
     },
+    -- SpeakerMaxLevel/MinLevel and MicrophoneMaxLevel/MinLevel are subscribed alongside audioMute
+    -- (unconditional on the AUDIO/SPEAKER feature) rather than audioVolume, since audioVolume's own
+    -- presence in the profile depends on first discovering these values -- gating their subscription
+    -- on audioVolume already being present would make that discovery impossible.
```
hcarter-775, line 226: I have a bit of [work](https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/2770) (kinda outdated but not too much I don't think) trying to solve this exact problem a little more clearly (that is, adding the ability to subscribe to attributes/events based on other values besides a capability.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
```diff
@@ -59,32 +59,45 @@ function CameraAttributeHandlers.volume_level_handler(driver, device, ib, respon
...
+  end
+  -- Convert from [min_volume, max_volume] to [0, 100] before emitting capability
   local normalized_volume = utils.round((ib.data.value - min_volume) * 100.0 / limited_range)
   device:emit_event_for_endpoint(ib, capabilities.audioVolume.volume(normalized_volume))
 end
 
 function CameraAttributeHandlers.max_volume_level_handler(driver, device, ib, response)
   local component = device:endpoint_to_component(ib)
   local max_volume = ib.data.value
+  device:set_field(camera_fields.RAW_MAX_VOLUME_LEVEL .. "_" .. component, max_volume)
```
hcarter-775: Maybe we should create a `set_field_for_component` and `get_field_for_component` function, the way we have that helper for endpoints. That would make this (and the other place we do this) a little less arbitrary

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -41,6 +41,17 @@ local function set_hard_privacy_mode_presence(device, hard_privacy_mode_present)
...
 
+-- Decides whether audioVolume belongs on this component. If the min/max levels aren't both known
+-- yet, leaves the capability as-is instead of guessing; otherwise requires max > min to include it.
+local function should_include_volume_capability(device, component)
+  local raw_min_volume = device:get_field(camera_fields.RAW_MIN_VOLUME_LEVEL .. "_" .. component)
+  local raw_max_volume = device:get_field(camera_fields.RAW_MAX_VOLUME_LEVEL .. "_" .. component)
+  if raw_min_volume == nil or raw_max_volume == nil then
+    return device:supports_capability(capabilities.audioVolume, component)
+  end
+  return raw_max_volume > raw_min_volume
```
hcarter-775:
```suggestion
  return device:supports_capability(capabilities.audioVolume, component) or (raw_max_volume > raw_min_volume)
```

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
```diff
@@ -41,6 +41,17 @@ local function set_hard_privacy_mode_presence(device, hard_privacy_mode_present)
   device:set_field(camera_fields.HARD_PRIVACY_MODE_PRESENT, hard_privacy_mode_present == true, { persist = true })
 end
 
+-- Decides whether audioVolume belongs on this component. If the min/max levels aren't both known
+-- yet, leaves the capability as-is instead of guessing; otherwise requires max > min to include it.
```
hcarter-775:
```suggestion
--- @param component string a profile component name
--- @return should_include boolean whether audioVolume belongs on the component argument.
```
if we're gonna have a comment, might as well structure it more neatly

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
hcarter-775: At the same time, the more I look at this, do we even need a "RAW" version of the already existing field? Can't the field we already have, that is, "MAX_VOLUME_LEVEL" do this same job?

hcarter-775: Just looking at it more, if the idea is that we want an un-edited version of the min/max since they may actually be malformed, I'd argue we should just have the "normal" field sets for this value do that, especially given the new logic you're introducing here. There's no need as far as I can see to alter the max and min we get defensively- if the device doesn't report correctly I don't see why we should alter that semi-arbitrarily.

samadDotDev: You're right that we don't need to track the normalized values anymore, and I updated that here: https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3242/commits/d29da26b6a5f66776cb3cc006ce411cfa5b4b377

samadDotDev: created in https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/pull/3242/commits/d29da26b6a5f66776cb3cc006ce411cfa5b4b377

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
samadDotDev: That's short-circuiting when the profile already supports this capability which is not what we want. What we want instead is to retain whatever state of this capability is in the profile when either of the min/max are nil i.e. when we haven't seen the report for either or both attributes then don't change the capability presence/absence prematurely.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/utils.lua
samadDotDev: That looks neat! We can move to those `conditional_subscriptions` post this change (and once you rebase your work later).

ctowns (approved):
LGTM! Thank you for adding the unit tests for the edge cases as well 🚀

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
```diff
@@ -57,34 +57,48 @@ end
...
 
 function CameraAttributeHandlers.max_volume_level_handler(driver, device, ib, response)
   local component = device:endpoint_to_component(ib)
   local max_volume = ib.data.value
-  local min_volume = device:get_field(camera_fields.MIN_VOLUME_LEVEL .. "_" .. component)
-  if max_volume > camera_fields.ABS_VOL_MAX or (min_volume and max_volume <= min_volume) then
+  local min_volume = camera_utils.get_field_for_component(device, camera_fields.MIN_VOLUME_LEVEL, component)
+  -- min == max is a valid (non-adjustable) range per the Matter spec, so only max < min is malformed;
+  -- an out-of-bounds value is clamped, but a malformed relationship is only logged and left as-is,
+  -- since volume_level_handler already guards against a non-positive range.
```
hcarter-775, line 78: nit: remove comment? Seems unnecessary to mention this

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
```diff
@@ -57,34 +57,48 @@ end
...
+  camera_cfg.reconcile_profile_and_capabilities(device)
 end
 
 function CameraAttributeHandlers.min_volume_level_handler(driver, device, ib, response)
   local component = device:endpoint_to_component(ib)
   local min_volume = ib.data.value
-  local max_volume = device:get_field(camera_fields.MAX_VOLUME_LEVEL .. "_" .. component)
-  if min_volume < camera_fields.ABS_VOL_MIN or (max_volume and min_volume >= max_volume) then
+  local max_volume = camera_utils.get_field_for_component(device, camera_fields.MAX_VOLUME_LEVEL, component)
+  -- See max_volume_level_handler: min == max is valid, only min > max is malformed and only logged.
```
hcarter-775, line 93: same thing here about unnecessary comment

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
hcarter-775: Hmm, I guess I was imagining that a capability's prior existence was the primary thing to be checking against? Are you hoping this will un-add the capability for devices that don't properly support this?

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
```diff
@@ -57,34 +57,48 @@ end
...
+  if min_volume < camera_fields.ABS_VOL_MIN then
     device.log.warn(string.format("Device reported invalid minimum (%d) %s volume level range value", ib.data.value, component))
     min_volume = camera_fields.ABS_VOL_MIN
+  elseif max_volume and min_volume > max_volume then
+    device.log.warn(string.format("Device reported invalid minimum (%d) %s volume level range value", ib.data.value, component))
   end
-  device:set_field(camera_fields.MIN_VOLUME_LEVEL .. "_" .. component, min_volume)
+  camera_utils.set_field_for_component(device, camera_fields.MIN_VOLUME_LEVEL, component, min_volume)
```
hcarter-775, line 100: Curious (same thought for max) should this value be persisted? If not, there will always be a period between driver restart and the time these get set when this value is nil

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_utils/device_configuration.lua
samadDotDev: no the prior existence is only checked when we don't have enough data about whether the capability should exist or not. When the device updates its min/max to the unusable ranges (as is the case for aforementioned camera that hasn't been supporting adjustment of microphone volume yet had to expose these values due to conformance, and will update these values to unusable range in the upcoming firmware), we want to remove the capability in existing devices as well.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
samadDotDev: That's fine, we'll retain the existing capability if the value isn't there and will update the inclusion/exclusion once the value arrives, which is functionally equivalent to persisting the last value and making decisions based on that.

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
ctowns: I personally think the comment is useful since it documents atypical behavior (min == max being allowed)

hcarter-775 (approved):

drivers/SmartThings/matter-switch/src/sub_drivers/camera/camera_handlers/attribute_handlers.lua
hcarter-775: Yeah, I think that part is fine. Mostly I was mentioning the extra stuff about only logging the malformed relationship, which is explicitly written in code just below. No strong feelings, but that is what I was primarily pointing out

